Issue #4, opened 2015-01-24
Geometry type system design propersal


sunng87:
I'm not native English speaker, if you have any question, just point it out and I'll try my best to make it clear. 

After we reached agreement to design the type system against some open standard, it's time to go further. 

My idea is to make an enum `Geometry` as the entry point of our type system. It will be like the `Json` enumeration in rustc-serialize, because in Rust, enum is a good way for polymorphism. When we are parsing geometry objects from Geojson or WKT, we might have no idea about the concrete type of them so we can use Geometry. Concretely, we will have a `FromWkt` trait and a `ToWkt` trait for I/O on Geometry without knowing the exact type. We can use pattern matching then to deal with different types. So other operations can also be impled on Geometry, like DE-9IM. It computes relationship between two objects (Point-LineString, Point-Point, LineString-Polygon or whatever). 

The value of enum will be the structs for different types. We still need concrete types because some operations or attributes are unique to some type.

```rust
enum Geometry {
  Point (Point),
  LineString (LineString),
  Polygon (Polygon),

  ...
}
```

We many have some functions like `as_point(&self) -> Option<Point>` to convert it from Geometry to Point.

I will be working on a simple implementation for this. Any question is welcomed.

frewsxcv:
This all sounds good to me. In regards to:

> Concretely, we will have a FromWkt trait and a ToWkt trait for I/O on Geometry without knowing the exact type.

Instead of this, what about having `FromGeo` and `ToGeo` on all the readers/writers (e.g. WKT, GeoJSON, etc). That way `rust-geo` doesn't need to know about all of the readers/writers. Pros/cons/thoughts?

frewsxcv:
I created a couple extremely basic dependency diagrams to illustrate what I'm talking about.

-------------------

What @sunng87 proposed

![copy of untitled drawing](https://cloud.githubusercontent.com/assets/416575/5888477/98770854-a3b4-11e4-885c-4ab5fb246228.png)

-------------------

What I counterproposed

![untitled drawing](https://cloud.githubusercontent.com/assets/416575/5888478/9e3ff3c2-a3b4-11e4-9ee2-b9a1623ef72b.png)

frewsxcv:
@georust/admins This is a pretty significant architecture/design decision. Would be great to get everyone's feedback

skj3gg:
Thanks @sunng87!

I think the To/FromGeo option is better as well, easily allows other markups to interface with `rust-geo`

[...]

Issue #2, opened 2015-01-19
geo structs


calvinmetcalf:
basic geo types need to be defined, some basics
- point, line and poygons
- how to deal with multi versions, separate type or points/line/polygon always store the type in a vector
- minimum number of distinct points in a line (geojson is 2, shapefiles is 1).
- minimum number of distinct points in a polygon (geojson is 3 plus first and last must be the same so 4 total, esri does not have first and last be the same).
- winding order, does it matter (geojson no, esrijson clockwise is a polygon and counterclockwise is a hole).
- if winding order doesn't matter how do you deal with polygons that cross into multiple hemispheres
- how do you deal with holes in polygons (geojson has a polygon object be an outer ring followed by zero or more inner rings, esrijson uses winding order).
- how do you deal with self intersecting polygons.
- how do you deal with polygons which intersect their own holes.
- do we deal with typologies
- how many dimensions
- geometry collections (esri does not touch this)
- projections and that whole can of shit ...

sunng87:
Perhaps we can follow some open standard like [OGC Simple Feature Access](http://www.opengeospatial.org/standards/sfa)

To answer some questions above (Just my opinion)

1. Base types: Point, LineString, Polygon; there are also some other types as defined in SFA 6.1 Geometry Object Model
2. Hole in polygon is represented as InteriorRing (A LineString). A polygon has only one exterorRing and 0 or multiple interiorRing. So wind order doesn't matter.
3. Self intersecting polygons are not supported (? you may check the definition in that document to confirm)
4. We don't need to deal with topology in this lib
5. (x,y) is required and `z`, `m` is optional
6. GeometryCollection is in this spec
7. Projections are out of this library's scope, only numeric coordinates are involved

calvinmetcalf:
not a bad idea, the way point xy is distinct from point xyz fits well with rust fyi ogc has the same types as geojson (Point,LineString,Polygon,MultiPoint,MultiLineString,MultiPolygon,GeometryCollection) plus a few more (MultiCurve, MultiSurface,Curve, Surface, PolyhedralSurface, TIN).

The GeoJSON and some databases like postgres have a somewhat laxer view on self intersections and what not, then tends to be a bit more practical in the real world.

The OGC spec is awful, the page number are romen numerals for gods sack.

Hemisphere spanning polygons are still a problem

sunng87:
[...]

PR #7, opened 2015-01-28
Base types

sunng87:
Just added a few types based on our discussion in #4 and #2. 

Types:

* Point
* MultiPoint
* LineString
* MultiLineString
* Polygon
* MultiPolygon
* GeometryCollection

and a `Coordinate` type wraps (x,y).

I hope we can use these types as common foundation of geojson, wkt and gdal, so we can interop data between these libraries.


mgax:
I kind of like the [newtype](http://doc.rust-lang.org/book/compound-data-types.html#tuple-structs-and-newtypes) approach used in rust-geojson, e.g. [Poly](https://github.com/georust/rust-geojson/blob/master/src/types/poly.rs) and [Ring](https://github.com/georust/rust-geojson/blob/master/src/types/ring.rs).

frewsxcv:
Yeah, for simple structs like these, I'd also suggest newtypes. You can see an example I did in rust-wkt earlier today https://github.com/georust/rust-wkt/commit/8ed3e7fb1a523d7037e73edbdb557275893d630a

sunng87:
Thanks, I will change them to new types.

sunng87:
Changed all type to newtypes. But I left the `Coordinate`, because it will be easier to access its fields by `.`

src/types.rs
```diff
@@ -0,0 +1,57 @@
...
+pub struct MultiPoint(pub Vec<Point>);
+
+#[derive(PartialEq, Clone)]
+pub struct LineString(pub Vec<Point>);
+
+#[derive(PartialEq, Clone)]
+pub struct MultiLineString(pub Vec<LineString>);
+
+#[derive(PartialEq, Clone)]
+pub struct Polygon(pub LineString, pub Vec<LineString>);
```
mgax: Why not simply `pub struct Polygon(pub Vec<LineString>);`?

sunng87: Because the first LineString (LineRing actually)  in the vector is the exterior ring, the rest are holes. They have different meaning. Also this structure implies that Polygon should at least have one exterior ring, zero or multiple interior rings.

frewsxcv: I think both are valid solutions, but for now, I vote to keep what you have just so we could get this merged as soon as possible. We can always change it later

frewsxcv:
Everything else looks good to me

frewsxcv:
For the sake of time, I'm going to merge these now. Thanks for doing this @sunng87 ! These are by no means final and are definitely subject to change

sunng87:
Thanks @frewsxcv , is it possible to publish this to crates.io?

frewsxcv:
I'll do that right now. Do you have an account? I'll add you

frewsxcv:
Published 0.0.2, added a git tag, add you as an owner

sunng87:
Thanks!

PR #13, opened 2015-10-22
Add basic Point methods

Turbo87:
This PR adds some basic methods to the `Point` type. This includes an XY constructor, getters and setters for x, y, lat and lng and implementations for the `Add`, `Sub`, `Neg` traits.

All methods are documented and include examples that are runnable as doctests.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -1,3 +1,7 @@
+use std::ops::Add;
+use std::ops::Neg;
+use std::ops::Sub;
+
 #[derive(PartialEq, Clone, Copy, Debug)]
 pub struct Coordinate {
     pub x: f64,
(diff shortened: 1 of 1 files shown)


frewsxcv:
Thanks, this is great!

Turbo87:
one thing where I'm not entirely sure is: do we prefer immutable methods that create new instances for mutations or is it prefered to mutate the actual object? I guess since we support `clone()` and Rust is pretty good in preventing unintentional mutation we should prefer the in-place mutation, right?

frewsxcv:
That's something that I was undecided about when I thought about it last. I'm still undecided, but there's no reason to hold us back since we changing our decision in the future won't be too bad hopefully.

Turbo87:
okay, agreed. Just FYI I'm not really fluent in Rust yet so it would be great if my pull requests could be thoroughly reviewed.

PR #14, opened 2015-10-22
TravisCI: Deploy to GitHub Pages only for the master branch

Turbo87:


frewsxcv:
Thanks, this looks a lot cleaner :sparkles:

Turbo87:
@frewsxcv this could (and should) be improved further by not using a global GitHub token, but an encrypted private key as part of the repo instead (see https://github.com/Turbo87/sidebar-v2/tree/master/.travis for example). If the token is ever compromised the attacker would have access to **all** repositories, while the encrypted private key could be repository-specific. Please let me know if this is something that I should work on.

frewsxcv:
I think that would be good to move to in the long term, but I'm pretty busy at the moment and probably won't get around to that anytime soon, so feel free to take it up if you want

PR #16, opened 2015-10-22
Merge Coordinate into Point type

### LINKED ISSUE #15: Coordinate vs. Point
what is the purpose of having a `Coordinate` **and** a `Point` type? it seems to me that having `Point` should be sufficient.

Turbo87:
This simplifies the Point API and removes an unnecessary abstraction level.

Resolves #15

### DIFF
--- a/examples/types.rs
+++ b/examples/types.rs
@@ -3,13 +3,6 @@ extern crate geo;
 use geo::*;
 
 fn main() {
-    let c = Coordinate {
-        x: 40.02f64,
-        y: 116.34
-    };
-
-    let p = Point(c);
-
-    let Point(coord) = p;
-    println!("Point at ({}, {})", coord.x, coord.y);
+    let p = Point::new(40.02, 116.34);
+    println!("Point at ({}, {})", p.x(), p.y());
 }
--- a/src/types.rs
+++ b/src/types.rs
@@ -3,10 +3,7 @@ use std::ops::Neg;
 use std::ops::Sub;
 
 #[derive(PartialEq, Clone, Copy, Debug)]
-pub struct Point {
-    x: f64,
-    y: f64,
-}
+pub struct Point(f64, f64);
 
 impl Point {
     /// Creates a new point.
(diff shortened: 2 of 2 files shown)


frewsxcv:
@georust/core Considering this is a breaking change, might be good to get some feedback here

I'm personally fine with this change

frewsxcv:
So after a couple days of thinking, here are my thoughts:

1. If we intend to keep the existing hierarchy (e.g. where `MultiPolygon` contains `Polygons`), I'm still in favor of this change
2. Is there any reason why a `MultiPolygon` should contain multiple `Polygons`? I'm not sure why we designed it that way, but this goes against many other geographic formats (e.g. GeoJSON, WKT) in addition to the [Simple Feature Access standard](http://www.opengeospatial.org/standards/sfa)

I'm slightly in favor of going to down the latter route. What do you think @Turbo87 ?

calvinmetcalf:
@frewsxcv It was designed around geojson which is by far the most pragmatic geospatial format

sunng87:
JTS and libgeos has separated `Coordinate` and `Point`. GeoJSON has a `Position` concept. Personally I'm not quite sure about the benefit and risk to combine Point and Coordinate.

Turbo87:
@frewsxcv I have not thought about MultiPolygons yet and never actually used them, but I also don't see how that is related to Points and Coordinates.

@calvinmetcalf as mentioned in #15 already: modeling after GeoJSON can work, but since JSON doesn't have a proper type system we can actually do much better in Rust. GeoJSON needs both Point and Coordinate because of the `type` attribute for the Point feature, but since Rust has a type system there is basically no difference between the two anymore.

@sunng87 the advantage is a simpler API and I honestly don't see any risks or disadvantages at this point. If I'm missing something then please tell me.

frewsxcv:
> Is there any reason why a MultiPolygon should contain multiple Polygons? I'm not sure why we designed it that way, but this goes against many other geographic formats (e.g. GeoJSON, WKT) in addition to the Simple Feature Access standard

I mentioned this, because if we decide to go this route and not make MultiPolygon *not* consist of Polygons, this pull request should probably *not* merged

Turbo87:
closing due to inactivity

PR #17, opened 2015-10-22
TravisCI: Use encrypted deploy key instead of global token

Turbo87:
see #14


Turbo87:
I've already added the deploy key to https://github.com/georust/rust-geo/settings/keys so this should work  right after merging.

frewsxcv:
Great, thanks! :sparkles:

PR #20, opened 2016-03-04
Fix docstring for impl Neg

timClicks:
This commit changes the docstring for Point.neg() to
be "Returns a point with the x and y compontents negated",
instead of being the clone of Point.add(), e.g.
"Add a point to the given point."


frewsxcv:
Thanks!

PR #23, opened 2016-05-03
fixed two small clippy warnings

llogiq:
Just readability: Arguments with underscores should not be used; this just removes the underscores.

Very clean codebase. Kudos. :+1:


frewsxcv:
Thanks!

llogiq:
You're welcome.

Issue #25, opened 2016-05-12
Generic numeric type for all structs


frewsxcv:
We should allow the user to use whatever numeric type they want (e.g. `f64`, `f32`, `i64`). Just need to parameterize it.

zarch:
Do we really need to use "whatever numeric type" to define spatial coordinates?
To me it looks like f64 is the right type in this case... I don't see much value in supporting general numeric types.

TeXitoi:
Using 32 bits integer is enough for precision and 2 times smaller. But it can be slower and less convenient.

frewsxcv:
Considering both of you expected different types, I think this is the reason why it should be generic :)

frewsxcv:
Completed in #30 by @Kerosene2000 and @zarch

PR #26, opened 2016-05-13
Support generic numeric type for all structs: Issue#25

zarch:
This is a first attempt to fix Issue #25, I'm new to rust... so perhaps there is a more elegant way to fix it, especially I don't like that there is a repetition of the type definition everywhere (`T: Num + Neg<Output = T> + Copy`).

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -186,8 +186,8 @@ impl Add for Point {
     /// assert_eq!(p.x(), 2.75);
     /// assert_eq!(p.y(), 5.0);
     /// ```
-    fn add(self, _rhs: Point) -> Point {
-        Point::new(self.x() + _rhs.x(), self.y() + _rhs.y())
+    fn add(self, rhs: Point) -> Point {
+        Point::new(self.x() + rhs.x(), self.y() + rhs.y())
     }
 }
 
@@ -204,8 +204,8 @@ impl Sub for Point {
     /// assert_eq!(p.x(), -0.25);
     /// assert_eq!(p.y(), 0.5);
     /// ```
-    fn sub(self, _rhs: Point) -> Point {
-        Point::new(self.x() - _rhs.x(), self.y() - _rhs.y())
+    fn sub(self, rhs: Point) -> Point {
+        Point::new(self.x() - rhs.x(), self.y() - rhs.y())
     }
 }
 


frewsxcv:
Any chance you could rebase (or merge) your changes on https://github.com/georust/rust-geo/pull/24? Also, I'm thinking we should just use http://rust-num.github.io/num/num/trait.Num.html instead of specifying all of the operations traits individually. What do you think?

jdroenner:
I think @frewsxcv is right here: using `T: Num ` should be enough for most cases. Also, is it really important to have `Neg<Output = T>` everywhere?

zarch:
My first attempt was to use just the Num trait but the rustc complains about this:
```
src/types.rs:180:20: 180:29 error: cannot apply unary operator `-` to type `T`
src/types.rs:180         Point::new(-self.x(), -self.y())
                                    ^~~~~~~~~
src/types.rs:180:31: 180:40 error: cannot apply unary operator `-` to type `T`
src/types.rs:180         Point::new(-self.x(), -self.y())
                                               ^~~~
```

Add the `Neg<Output = T>` fix this error. Do we need to implement `Neg`? if not we can just remove it and keep the type as `T: Num + Copy`. We still need `Copy` because otherwise rustc in x and y methods:

```
impl<T> Point<T>
    where T: Num + Copy
{
    // [...]
    pub fn x(&self) -> T {
        self.0.x
    }

    pub fn y(&self) -> T {
        self.0.y
    }
```

complains on borrowed content [E0507]:

```
src/types.rs:48:9: 48:13 error: cannot move out of borrowed content [E0507]
src/types.rs:48         self.0.x
                        ^~~~
src/types.rs:48:9: 48:13 help: run `rustc --explain E0507` to see a detailed explanation
src/types.rs:76:9: 76:13 error: cannot move out of borrowed content [E0507]
src/types.rs:76         self.0.y
                        ^~~~
```

Other idea on how we can fix this?

jdroenner:
To implement `Neg` for `Point `it is necessary to have a type which also implements `Neg`:
` impl <T> Neg for Point<T>  where T: Neg<Output = T>`
Only `T: Neg<Output = T>` should be required for this.

To implement `Point ` itself there is no need for `Neg`:
`impl <T> Point<T> where T: Num`

This way `Neg` is available for points (where `T: Neg`) while it is still possible to use other types only limited by `T: Num`. So you can remove the `Neg` bounds from all other implementations and types.



I don't know if we need `T: copy`. In most cases T will be a primitive number so copy should be fine. 
We could also use `T: clone `instead of `T: copy` to allow more types:
```
`impl<T> Point<T> where T: Num + Clone
{
    pub fn x(&self) -> T {
        self.0.x.clone()
    }
}`
```

But it could also be OK to return a reference:
```
`impl<T> Point<T> where T: Num
{
    pub fn x(&self) -> &T {
        self.0.x
    }
}`
```

PR #40, opened 2016-07-17
COORD_PRECISION is now f32

Kerosene2000:
https://github.com/georust/rust-geo/issues/39


frewsxcv:
Version 0.2.0 of geo-rust has been released which includes these changes.

PR #43, opened 2016-07-18
Fix Area trait for Polygon to consider inner isles and implement area tait for MultiPolygon

zarch:


frewsxcv:
Consider this example:

drawing

`A` is the outer ring. `B` and `C` are inner rings. With your `Polygon` `Area` algorithm, we'll calculate `Area(A) - Area(B) - Area(C)`, which is different than calculating the green area in the image.

I wonder if there are other libraries that handle this scenario.

A natural response would be "this is an invalid Polygon since the inner rings fall *outside* the outer ring". I think this is a valid response for the time being unless someone knows of an easy way to do the correct calculation above.

@zarch What do you think? I think your changes are an improvement and I'm 👍 on merging them. Just want to pass this by you. Thanks for the pull request!

waywardmonkeys:
Your polygons are invalid according to the "Simple Features" specification from the OGC.

waywardmonkeys:
Also according to the GeoJSON spec, although that's much less rigorously defined: http://geojson.org/geojson-spec.html#polygon

frewsxcv:
Thanks for your input @waywardmonkeys, much appreciated! 

This pull request looks good to me. @zarch Feel free to merge and deploy a new version (0.2.1). Otherwise I can do it myself sometime tomorrow. :confetti_ball:

zarch:
Thank you @frewsxcv, in these days I was think that would be nice to add a trait like: IsValid (is_valid) that check if a geometry is valid or not.
I'm wondering if this check should be implemented in the rust-geo library or in an external one (rust-geo-validation? where is the right place to raise question on the georust ecosystem and structure?).
I think the validation should use different validation-driver (e.g. geos, grass gis, etc), and would be nice to have also function to fix the input (e.g. converting your example into a multipolygon).

frewsxcv:
>  in these days I was think that would be nice to add a trait like: IsValid (is_valid) that check if a geometry is valid or not. I'm wondering if this check should be implemented in the rust-geo library or in an external one (rust-geo-validation? where is the right place to raise question on the georust ecosystem and structure?).

For now, I think opening a new issue in this repository (rust-geo) should be fine for now.

I also just created https://github.com/georust/meta/ as a place to discuss GeoRust projects and the organization.

PR #44, opened 2016-07-18
Add the Length trait and implement Length for LineString and MultiLineString

zarch:


src/algorithm/length.rs
```diff
@@ -0,0 +1,82 @@
+use num::{Float};
```
waywardmonkeys: You don't need the extra braces here: `use num::Float;`

src/algorithm/length.rs
```diff
@@ -0,0 +1,82 @@
...
+    ///
+    fn length(&self) -> T {
+        let vect = &self.0;
+        if vect.is_empty() || vect.len() == 1 {
+            return T::zero();
+        } else {
+            let ipoints = vect.iter().zip(vect[1..].iter());
+            ipoints.fold(T::zero(), |total_length, (p1, p2)| total_length + p1.distance(&p2))
+        }
+    }
```
frewsxcv: This might be simpler if you used [`slice::windows`](https://doc.rust-lang.org/std/primitive.slice.html#method.windows) here.

src/algorithm/length.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+impl<T> Length<T> for MultiLineString<T>
+    where T: Float
+{
+    ///
+    /// Length of a MultiLineString.
+    ///
```
frewsxcv: You can remove this comment. The implementation signature makes this self explanatory.

src/algorithm/length.rs
```diff
@@ -0,0 +1,82 @@
...
+    fn length(&self) -> T;
+}
+
+impl<T> Length<T> for LineString<T>
+    where T: Float
+{
+    ///
+    /// Length of a LineString.
+    ///
```
frewsxcv: This comment can be removed.

src/algorithm/length.rs
```diff
@@ -26,26 +26,20 @@ pub trait Length<T, RHS = Self> {
 impl<T> Length<T> for LineString<T>
     where T: Float
 {
-    ///
-    /// Length of a LineString.
-    ///
     fn length(&self) -> T {
         let vect = &self.0;
         if vect.is_empty() || vect.len() == 1 {
```
frewsxcv: This `if` can be removed

frewsxcv:
Looks good to me, will wait until tonight to merge in case anyone else has comments.

PR #50, opened 2016-07-22
Iterate over points using windows instead of zip

zarch:
Use windows instead of zip to iterate over points. It seems also slightly faster from my [bench](https://gist.github.com/zarch/3cda1ffda67e156d712f5c1b5a3d3d23).

```
running 2 tests
test bench_win ... bench: 154,231,376 ns/iter (+/- 9,710,675)
test bench_zip ... bench: 173,904,730 ns/iter (+/- 61,305,779)
```

### DIFF
--- a/src/algorithm/area.rs
+++ b/src/algorithm/area.rs
@@ -25,7 +25,7 @@ fn get_linestring_area<T>(linestring: &LineString<T>) -> T where T: Float {
         return T::zero();
     }
     let mut tmp = T::zero();
-    for ps in linestring.0.windows(2).into_iter() {
+    for ps in linestring.0.windows(2) {
         tmp = tmp + (ps[0].x() * ps[1].y() - ps[1].x() * ps[0].y());
     }
     tmp / (T::one() + T::one())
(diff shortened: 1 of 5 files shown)


src/algorithm/area.rs
```diff
@@ -25,8 +25,8 @@ fn get_linestring_area<T>(linestring: &LineString<T>) -> T where T: Float {
         return T::zero();
     }
     let mut tmp = T::zero();
-    for (p1, p2) in linestring.0.iter().zip(linestring.0[1..].iter()) {
-        tmp = tmp + (p1.x() * p2.y() - p2.x() * p1.y());
+    for ps in linestring.0.windows(2).into_iter() {
```
frewsxcv: FYI, you should be able to just do `for ps in linestring.0.windows(2)` here. Using a `for` loop will automatically call [`IntoIterator`](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html) for the item being interated over

https://doc.rust-lang.org/std/iter/index.html#for-loops-and-intoiterator

frewsxcv:
Thanks! It's great that simplifying the logic here results in better performance

PR #53, opened 2016-08-04
Use builtin method for calculation

urschrei:
We can use the built-in `hypot()` function which carries out the multiplication and square root calculation directly.


TeXitoi:
Thanks!

Issue #54, opened 2016-08-04
Perpendicular distance and LineString simplification


urschrei:
Is there any interest in a `simplify` method for LineStrings, using Douglas–Peucker? I've just finished implementing it in a separate crate, and then realised it might be useful here. This would also supply a perpendicular distance (from a LineString to a Point) method…

frewsxcv:
I don't see why not, so +1 from me. I think a `Simplify` trait should be appropriate. Are there any other popular algorithms for doing simplify besides Douglas-Peucker?

urschrei:
Hmm there's Visvalingam-Whyatt, too ([Mike Bostock](https://bost.ocks.org/mike/simplify/) has a nice demo of it).
From a quick look at the source, it looks a bit more complex than RDP to implement. I'd certainly be interested in working *with* someone on it, if people are interested.

urschrei:
Sounds good to me!

PR #55, opened 2016-08-05
Implement LineString simplification

urschrei:
This trait currently uses the [Ramer–Douglas–Peucker](https://en.wikipedia.org/wiki/Ramer–Douglas–Peucker_algorithm) algorithm to
simplify a LineString (see #54), while preserving its overall shape.

This is WIP, for two reasons:
- `rdp()` currently uses `unwrap()`, because the `first()` and `last()` methods on slices return `Option`. It's not immediately clear to me how these could fail in this context, but all input is welcome.
- `point_line_distance()` and `rdp()` are implemented as helper functions, and called by the `Simplify` trait. This is mainly because `rdp()` is recursive and uses sub-slices of its initial input, which AFAICS can't be easily obtained from a `LineString`.

### DIFF
--- a/src/algorithm/simplify.rs
+++ b/src/algorithm/simplify.rs
@@ -4,21 +4,21 @@ use algorithm::distance::Distance;
 
 // perpendicular distance from a point to a line
 fn point_line_distance<T>(point: &Point<T>, start: &Point<T>, end: &Point<T>) -> T
-    where T: Float 
+    where T: Float
 {
     if start == end {
-        return point.distance(&start)
+        point.distance(start)
     } else {
         let numerator = ((end.x() - start.x()) * (start.y() - point.y()) -
-                 (start.x() - point.x()) * (end.y() - start.y()))
+                         (start.x() - point.x()) * (end.y() - start.y()))
             .abs();
-            let denominator = start.distance(&end);
+        let denominator = start.distance(end);
         numerator / denominator
     }
 }
 
 // Ramer–Douglas-Peucker line simplification algorithm
-fn rdp<T>(points: &[Point<T>], epsilon: &T) -> Vec<Point<T>> 
+fn rdp<T>(points: &[Point<T>], epsilon: &T) -> Vec<Point<T>>
     where T: Float
 {
     let mut dmax = T::zero();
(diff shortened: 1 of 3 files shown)


src/algorithm/simplify.rs
```diff
@@ -0,0 +1,111 @@
...
+    #[test]
+    fn perpdistance_test() {
+        let start = Point::new(1.0, 2.0);
+        let end = Point::new(3.0, 4.0);
+        let p = Point::new(1.0, 1.0);
+        let dist = point_line_distance(&p, &start, &end);
+        assert_eq!(dist, 0.7071067811865475);
+    }
+    #[test]
+    fn rdp_test() {
```
frewsxcv: Can you also add a test doing `simplify` on an empty `LineString`?

frewsxcv: Or should we expect all `LineString`s to have at least two points? If so, we should make it explicit in the doc comment and add an `assert`.

src/algorithm/simplify.rs
```diff
@@ -0,0 +1,111 @@
...
+fn rdp<T>(points: &[Point<T>], epsilon: &T) -> Vec<Point<T>> 
+    where T: Float
+{
+    let mut dmax = T::zero();
+    let mut index: usize = 0;
+    let mut distance: T;
+
+    for (i, _) in points.iter().enumerate().take(points.len() - 1).skip(1) {
+        distance = point_line_distance(&points[i],
+                                       &*points.first().unwrap(),
```
frewsxcv: You might as well do `points[0]` here, right?

src/algorithm/simplify.rs
urschrei: Hmm. A 2-point LineString can't be simplified anyway, and both it and an empty `LineString` are valid, so it makes sense to me to just return them both unmodified.

src/algorithm/simplify.rs
urschrei: Instead of `.first()`? Yep, I just instinctively shy away from array indexing, but we know it'll be OK in this case.

frewsxcv:
Everything here looks good to me, is this still a WIP?

urschrei:
No, I'm pretty happy with it. How do you feel about the thoroughness of the tests, though? I feel like some tests for pathological input (tiny epsilon, gigantic LineString) would probably be a good idea at some point, although in the latter case that should probably happen further up the chain.

frewsxcv:
Sounds great, thanks for this! 🎊

urschrei:
\o/

frewsxcv:
:D

One of these I want to get some sort of basic Rust GUI for geo data (maybe written in [conrod](https://github.com/PistonDevelopers/conrod)) so that we can visualize all these rust-geo algorithms and operations.

Issue #56, opened 2016-08-07
Idea: Implement constraints on geometry types


frewsxcv:
Examples:

* LineString: panics if fewer than two points
* Polygon: panics if any rings has fewer than three points
* Polygon should check if first+last points are the same?
* Polygon should ensure interiors are inside exterior? Maybe not this is one as it might be expensive.

TeXitoi:
Checks that the inners don't intersect. 

Checks that multi don't intersect.

TeXitoi:
Maybe that can be an algorithm "check" or something like that. ie optional.

urschrei:
Shapely will allow you to construct invalid [Polygons](http://toblerity.org/shapely/manual.html#polygons) (rings that cross each other, and touch at more than one point), but raises exceptions when they're used, presumably for performance reasons. Maybe a `Valid` trait that returns a bool, which could be checked by any methods implemented on the type.
How would the more straightforward constructors look?  Something like

    impl<T> LineString
    where T: Float {
        pub fn new(v: Vec<T>) -> Result(LineString, _) {
            // Check that the vec contains at least two elements?
            // I have no idea what happens next
        }
    }

?
(I've never written a "validating" constructor before, so I have no idea what I'm doing, really – if anyone's got some examples that would be great)

frewsxcv:
One example in `std` is [`CString::new`](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.new) which will fail if the bytes contain a null character (`\0`)

frewsxcv:
Opened an issue similar to this: https://github.com/georust/rust-geo/issues/127

PR #58, opened 2016-08-10
[wip] Initial pass at LineString constructor

urschrei:
Ref #56 
This checks whether the vector being used has one element, and fails if it does.
The error is currently just a `String`. Is this OK?

### DIFF
--- a/src/algorithm/simplify.rs
+++ b/src/algorithm/simplify.rs
@@ -29,9 +29,7 @@ fn rdp<T>(points: &[Point<T>], epsilon: &T) -> Vec<Point<T>>
     let mut distance: T;
 
     for (i, _) in points.iter().enumerate().take(points.len() - 1).skip(1) {
-        distance = point_line_distance(&points[i],
-                                       &points[0],
-                                       &*points.last().unwrap());
+        distance = point_line_distance(&points[i], &points[0], &*points.last().unwrap());
         if distance > dmax {
             index = i;
             dmax = distance;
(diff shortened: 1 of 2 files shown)


urschrei:
Trying out `from()` and `into()` for constructing LineStrings using common source data structures;
- `Vec<[T; 2]>>`
- `<Vec<(T, T)>>`  
where `T` is `Float`

This works well imo, but where `::new()` returns a `Result()`, and can catch the major stumbling block (1-element sequence), `from()` and into() can't, and will panic if you do something like
`let ls: LineString = vec![[1.0, 2.0]].into()`
I'm not sure to what extent this is in conflict with the guidance that the `From()` and `Into()` traits [must not fail](https://doc.rust-lang.org/std/convert/trait.Into.html) (Is an unrecoverable panic (apart from `catch_unwind`) a failure?) – `TryInto` will be stable soon-ish (I think), so we could wait for that.

frewsxcv:
I like where this is going. I'm wondering, from an API perspective, it makes sense to make the internal field on the `LineString` no longer `pub`. In concrete terms, this would be that construction would be done with `LineString::new` (which could also be called `LineString::from_points`) and if they want to access the underlying points, we could provide methods like:

* `as_slice(&self) -> &[Point]`
* `into_vec(self) -> Vec<Point>`

Though this might be a tangential conversation.

urschrei:
Oh, I like that idea. Will see how it works out…

frewsxcv:
Opened a similar PR: https://github.com/georust/rust-geo/pull/168

PR #61, opened 2016-08-18
Add Point to Polygon and Point to LineString distance methods

### LINKED ISSUE #59: How best to calculate Point to Polygon distance?
I'm working on an algorithm that requires calculating the minimum signed distance from a Point to a Polygon exterior (negative implies that the point lies inside the Polygon / `polygon.contains(point)` is `true`).
I can't see a good way to do this, currently – I've implemented point-line-distance in [simplify](https://github.com/georust/rust-geo/blob/master/src/algorithm/simplify.rs#L6-L18), but that's for segments, and there's currently no way to get Polygon exterior ring segments. Or is there? If anyone can think of an approach here, or point to an existing implementation I'd be very grateful (and I'll implement it ASAP)

urschrei:
This PR adds distance methods from a `Point` to:
- a `LineString`
- a `Polygon`

To note:
- Points **inside a `Polygon`** return `0.0`
- Points on a **`Polygon` boundary** return `0.0`
- Points **on a `LineString`** return `0.0`
- distance to an **empty `LineString`** returns `0.0`
- distance to an **empty `Polygon`** (shouldn't be possible, but we're handling it for now) returns `0.0`
- I'm using a [`BinaryHeap`](https://doc.rust-lang.org/std/collections/struct.BinaryHeap.html) implemented as a minimum priority queue with some floating-point-compatible comparator methods (see the `Ord` and `PartialOrd` `impl`s on `Mindist`) – I just adapted the linked example to use partial ordering. Please chime in if this is Bad, or if you can think of a better design.

Also WIP because:
- <s>no handling of empty geometries yet</s>
- <s>no docs / doctests yet</s>

This will close #59 if/when it merges.

### DIFF
--- a/src/algorithm/distance.rs
+++ b/src/algorithm/distance.rs
@@ -143,8 +143,8 @@ impl<T> Distance<T, Polygon<T>> for Point<T>
         for ring in &polygon.1 {
             dist_queue.push(Mindist { distance: self.distance(ring) })
         }
-        for chunk in ext_ring.chunks(2) {
-            let dist = line_segment_distance(self, &chunk[0], &chunk.last().unwrap_or(&chunk[0]));
+        for chunk in ext_ring.windows(2) {
+            let dist = line_segment_distance(self, &chunk[0], &chunk.last().unwrap());
             dist_queue.push(Mindist { distance: dist });
         }
         dist_queue.pop().unwrap().distance
@@ -164,8 +164,8 @@ impl<T> Distance<T, LineString<T>> for Point<T>
         let mut dist_queue: BinaryHeap<Mindist<T>> = BinaryHeap::new();
         // get points vector
         let points = &linestring.0;
-        for chunk in points.chunks(2) {
-            let dist = line_segment_distance(self, &chunk[0], &chunk.last().unwrap_or(&chunk[0]));
+        for chunk in points.windows(2) {
+            let dist = line_segment_distance(self, &chunk[0], &chunk.last().unwrap());
             dist_queue.push(Mindist { distance: dist });
         }
         dist_queue.pop().unwrap().distance
@@ -333,8 +333,7 @@ mod test {
         let ls = LineString(points.iter().map(|e| Point::new(e.0, e.1)).collect());
         let p = Point::new(3.5, 2.5);
         let dist = p.distance(&ls);
-                      // 0.5 <-- Shapely
-        assert_eq!(dist, 0.5144957554275263);
+        assert_eq!(dist, 0.5);
     }
     #[test]
     // Point to LineString, empty LineString
--- a/src/algorithm/distance.rs
+++ b/src/algorithm/distance.rs
@@ -216,7 +216,7 @@ mod test {
                           (6., 1.), (5., 1.)];
         let ls = LineString(points.iter().map(|e| Point::new(e.0, e.1)).collect());
         let poly = Polygon(ls, vec![]);
-        // A Random point inside the octagon
+        // A point on the octagon
         let p = Point::new(5.0, 1.0);
         let dist = p.distance(&poly);
         assert_eq!(dist, 0.0);
--- a/src/algorithm/distance.rs
+++ b/src/algorithm/distance.rs
@@ -88,7 +88,7 @@ fn line_segment_distance<T>(point: &Point<T>, start: &Point<T>, end: &Point<T>) 
 {
     let dist_squared = pow(start.distance(end), 2);
     // Implies that start == end
-    if dist_squared == T::zero() {
+    if dist_squared.is_zero() {
         return pow(point.distance(start), 2);
     }
     // Consider the line extending the segment, parameterized as start + t (end - start)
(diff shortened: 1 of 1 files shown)


src/algorithm/distance.rs
```diff
@@ -27,10 +67,235 @@ impl<T> Distance<T, Point<T>> for Point<T>
...
+        // exterior ring as a LineString
+        let ext_ring = &exterior.0;
+        for chunk in ext_ring.chunks(2) {
+            let dist = match chunk.len() {
+                2 => line_segment_distance(self, chunk.first().unwrap(), chunk.last().unwrap()),
+                _ => {
+                    // final point in an odd-numbered exterior ring
+                    line_segment_distance(&self, chunk.first().unwrap(), chunk.first().unwrap())
+                }
+            };
```
frewsxcv: Alternatively:

```rust
line_segment_distance(self, chunk[0], chunk.last().or(chunk[0]));
```

The slice returned by the `chunks` iterator guarantees that element `[0]` exists.

urschrei: Had to use `unwrap_or`, but still ✨

urschrei:
Empty geometries are handled gracefully, and everything's documented.

urschrei:
The `Point`-to-`Polygon` method now also accounts for the possibility that the point is inside an interior ring:
screenshot 2016-08-19 12 36 25

src/algorithm/distance.rs
```diff
@@ -131,6 +131,12 @@ impl<T> Distance<T, Polygon<T>> for Point<T>
         }
         // minimum priority queue
         let mut dist_queue: BinaryHeap<Mindist<T>> = BinaryHeap::new();
+        // we've got interior rings
+        if polygon.1.len() > 0 {
```
frewsxcv: This outer `if` isn't necessary. If it's empty, the `for` loop won't iterate, which is fine.

src/algorithm/distance.rs
```diff
@@ -27,13 +72,261 @@ impl<T> Distance<T, Point<T>> for Point<T>
...
 
+// Return minimum distance between a Point and a Line segment
+// This is a helper for Point-to-LineString and Point-to-Polygon distance
+// adapted from http://stackoverflow.com/a/1501725/416626
+fn line_segment_distance<T>(point: &Point<T>, start: &Point<T>, end: &Point<T>) -> T
+    where T: Float + ToPrimitive
+{
+    let dist_squared = pow(start.distance(end), 2);
+    // Implies that start == end
+    if dist_squared == T::zero() {
```
frewsxcv: Alternatively:

```rust
if dist_squared.is_zero() {
```

http://rust-num.github.io/num/num/trait.Zero.html#tymethod.is_zero

src/algorithm/distance.rs
```diff
@@ -27,13 +72,261 @@ impl<T> Distance<T, Point<T>> for Point<T>
...
+    // We find the projection of the point onto the line
+    // This falls where t = [(point - start) . (end - start)] / |end - start|^2, where . is the dot product
+    // We clamp t from [0.0, 1.0] to handle points outside the segment start, end
+    let t = T::zero().max(T::one().min((*point - *start).dot(&(*end - *start)) / dist_squared));
+    let projected = Point::new(start.x() + t * (end.x() - start.x()),
+                               start.y() + t * (end.y() - start.y()));
+    point.distance(&projected)
+}
+
+#[derive(PartialEq, Debug)]
```
frewsxcv: Could this also derive `Ord` and `PartialOrd`?

frewsxcv: And also `Eq`

src/algorithm/distance.rs
```diff
@@ -27,13 +72,261 @@ impl<T> Distance<T, Point<T>> for Point<T>
...
+impl<T> Distance<T, Polygon<T>> for Point<T>
+    where T: Float
+{
+    fn distance(&self, polygon: &Polygon<T>) -> T {
+        // get exterior ring
+        let exterior = &polygon.0;
+        // exterior ring as a LineString
+        let ext_ring = &exterior.0;
+        // No need to continue if the polygon contains the point, or is zero-length
+        if polygon.contains(self) || ext_ring.len() == 0 {
```
frewsxcv: Alternatively:

```rust
ext_ring.len() == 0
```

frewsxcv:
Looks great! Awesome work @urschrei. I left a bunch of nits, but feel free to merge whenever 👍

src/algorithm/distance.rs
urschrei: `Ord`, `PartialOrd`, and `Eq` need custom `impl`s (lines 111, 118, 125) for these, in order for the `BinaryHeap` to work as min-heap, so deriving them causes a conflict…

urschrei:
Now that I've fixed my embarrassing `chunks(2)` mistake, I'm happy for this to merge, the `Ord` etc. stuff notwithstanding.

urschrei:
(I have a working implementation of [Polylabel](https://github.com/mapbox/polylabel) ready to go, when this merges and releases)

frewsxcv:
@aelita-mergebot r+

frewsxcv:
@aelita-mergebot r+

urschrei:
Thanks 🎉!

PR #62, opened 2016-08-21
Implement Haversine distance

tmcw:


tmcw:
Still learning! Anyone know whether there's a better way to avoid this [longwinded series of T:from_i32 statements for an algorithm that works with both variables and constants?](https://github.com/georust/rust-geo/pull/62/files#diff-a6bd5d5a5b68383f3b5a50836952cba2R85)

notriddle:
thread 'main' panicked at 'assertion failed: `(left == right)` (left: `10900.115611845817`, right: `10900.115612674515`)', <anon>:8

Something's going to have to be done about that.

TeXitoi:
Please remove the warnings.

tmcw:
Indeed, I'm very aware of the error, and will work on fixing it, nobody is expecting this PR to be merged with failing tests :)

Anyone know the answer to the T::from_i32 question, though?

urschrei:
I've had the same issue when using "actual numbers" within generic functions, and haven't been able to get around it, but I also haven't taken the time to ask anyone who works on/with `num`. It's probably worth asking on #rust, or users.rust-lang.org.

urschrei:
OK, so the consensus seems to be something like:

    fn f<T: FromPrimitive>(i: i32) -> T { T::from_i32(i).unwrap() }

I'd assumed it would be more complicated than that.

frewsxcv:
Just rebased this off master. @tmcw It looks like there are failing tests. Do you know why they're failing? It'd be great if this could merge before https://github.com/georust/rust-geo/pull/85, but no worries if you're short on time, just let me know :)

Turbo87:
@tmcw @frewsxcv this is most likely related to comparing floats using `assert_eq()`. we should probably use `within_epsilon()` for the assertions (see https://github.com/georust/rust-geo/blob/master/src/algorithm/distance.rs#L347)

Issue #60, opened 2016-08-18
Document what the parameters to the Polygon struct are


frewsxcv:
```rust
pub struct Polygon<T>(pub LineString<T>, pub Vec<LineString<T>>) where T: Float;
```

The first item is the exterior ring. The second item is a vector of interior rings. We should either add a doc comment documenting this, or change the struct to something like:

```rust
pub struct Polygon<T> {
    pub exterior: pub LineString<T>,
    pub interior: Vec<LineString<T>>
} where T: Float;
```

TeXitoi:
I prefere a structure, more explicit in the code. May I propose outer and inners as field names?

frewsxcv:
> May I propose outer and inners as field names?

Sounds good to me!

PR #68, opened 2016-10-09
Document Polygon struct parameters

mbattifarano:
Addresses #60

Replaces the Polygon tuple struct with a regular struct:

```
pub struct Polygon<T>
    where T: Float
{
    pub exterior: LineString<T>,
    pub interiors: Vec<LineString<T>>
}
```

I'm new to rust and eager to learn so comment away!

This breaks the existing Polygon constructor since (correct me if
I'm wrong) structs must be instaniated with curly braces and named
arguments. To mitigate the effect of this change on existing code,
`Polygon` implements a `new` method that preserves the old interface:

`Polygon::new(exterior, interiors)`

I chose `exterior` and `interiors` over `outer` and `inners` as I
believe the former pair is more commonly used. I don't feel strongly
about this choice and will change it upon request.

### DIFF
--- a/src/algorithm/area.rs
+++ b/src/algorithm/area.rs
@@ -14,7 +14,7 @@ pub trait Area<T> where T: Float
     /// let p = |x, y| Point(Coordinate { x: x, y: y });
     /// let v = Vec::new();
     /// let linestring = LineString(vec![p(0., 0.), p(5., 0.), p(5., 6.), p(0., 6.), p(0., 0.)]);
-    /// let poly = Polygon(linestring, v);
+    /// let poly = Polygon::new(linestring, v);
     /// assert_eq!(poly.area(), 30.);
     /// ```
     fn area(&self) -> T;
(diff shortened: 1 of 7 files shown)


notriddle:
As you can see in the little Travis CI section at the bottom, one of the tests failed. I think you got something flipped around...

mbattifarano:
Thanks @notriddle! I left out an import, just pushed the fix.

frewsxcv:
This looks good, but I'm going to leave this open until I/we push a new major version release that includes breaking changes like these. Thanks for doing this @mbattifarano! 🎊

frewsxcv:
@aelita-mergebot r+

frewsxcv:
No reason to hold this up any longer. Thanks again @mbattifarano!

frewsxcv:
Looks like there are a few more places the new `Polygon` needs to be used.

mbattifarano:
Yep, I just needed to update the additions from #70 and #71

frewsxcv:
@aelita-mergebot r+

Issue #70, opened 2016-10-11
Implement `Intersects<Bbox<T>> for Polygon`


frewsxcv:
https://github.com/georust/rust-geo/blob/master/src/algorithm/intersects.rs

https://georust.github.io/rust-geo/geo/algorithm/intersects/trait.Intersects.html

frewsxcv:
Completed in #76

PR #76, opened 2016-10-15
Implement `Intersects<Bbox<T>> for Polygon`

mbattifarano:
#70 

Uses `Intersects<Polygon<T>> for Polygon` to do the work.

Also implements `Intersects<Polygon<T>> for Bbox`


frewsxcv:
Thanks!

PR #82, opened 2016-10-17
Within Epsilon matcher

shterrett:
For https://github.com/georust/rust-geo/issues/75

I couldn't get the `within_epsilon` in the doctest; I think it's because I have it set to compile in test only.

### DIFF
--- a/src/algorithm/area.rs
+++ b/src/algorithm/area.rs
@@ -59,31 +59,33 @@ impl<T> Area<T> for Bbox<T>
 
 #[cfg(test)]
 mod test {
+    use num::Float;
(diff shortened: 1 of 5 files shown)


TeXitoi:
OK for me for the implementation. In practice, the choosen epsilon is quite small. But we can change that later if we have troubles.

shterrett:
I chose `Float::epsilon` for the tests that were "all but integer", partially because they all seemed to be working for everyone. The `1e-15` epsilon was chosen because the relative error between actual and expected in the failing test that I saw was `2e-16`; I made it just a little bit larger.

PR #84, opened 2017-01-10
Add Visvalingam-Whyatt line-simplification algorithm

urschrei:
This is an initial pass at the implementation of the [Visvalingam-Whyatt](http://bost.ocks.org/mike/simplify/) line-simplification algorithm.

This implementation is partially based on a [pre-v1.0 version](https://github.com/huonw/isrustfastyet/blob/gh-pages/mem/line_simplify.rs) written by @huonw, and uses his method of storing retained points in a "linked list", updating it as necessary. I'm not sure if there's a cleaner way of doing this, but I'm open to ideas.

I'd also like to test the output of this algorithm against a more complex LineString using PostGIS – there are several other extant implementations in various languages, but I don't completely trust them.

Another issue is the provision of the `epsilon` parameter; in this case, it refers to the minimum triangle area to retain, and it's somewhat unintuitive compared to `rdp` (although PostGIS provides the same API).

Other questions:
- Should this trait live in the `simplify.rs` module, instead of its own?


frewsxcv:
Awesome, thanks for doing this! I'm a little busy, but I'll try to look at this sometime this weekend unless someone else wants to review this :)

urschrei:
Hmm this is odd; the test passes on my machine (x86_64, OS X):

    stable-x86_64-apple-darwin (default)
    rustc 1.14.0 (e8a012324 2016-12-16)

I don't know where the extra point (`Point(Coordinate { x: 28.770069, y: 54.27718 })`) in position 240 has come from [on Travis](https://travis-ci.org/georust/rust-geo/jobs/191445206#L322)

frewsxcv:
FWIW, I can replicate the error Travis is getting on x86_64 macOS. `rustc 1.16.0-nightly (1a2ed98d3 2017-01-13)`

urschrei:
After `cargo clean`, I can reproduce it on stable and nightly on my local machine. I suspect that it's a precision issue in one of the `num` crates. I'll compare a few more LineStrings, but I'm not too concerned; the algorithm's simple enough that an implementation error would cause differences way beyond 1 point in 401.

frewsxcv:
Thanks!

frewsxcv:
> Should this trait live in the simplify.rs module, instead of its own?

Regarding this question, I think what you have is fine for now. Might change in the future though.

urschrei:
🎊

PR #85, opened 2017-01-15
Trait-based geometry types (no cost abstractions!) WIP.

frewsxcv:
https://github.com/georust/rust-geo/issues/67


src/traits.rs
```diff
@@ -6,3 +6,45 @@ pub trait ToGeo<T: Float>
...
+    type Iter: Iterator<Item=Self::ItemType>;
+
+    fn points(&self) -> Self::Iter;
+}
+
+pub trait PolygonTrait<T: Float> {
+    type ItemType: LineStringTrait<T>;
+    type Iter: Iterator<Item=Self::ItemType>;
+
+    fn rings(&self) -> Self::Iter;
```
jdroenner: shouldn't a polygon have one outer ring and optional inner rings?

frewsxcv: Was going to discuss that eventually, so I guess we could do it now. For now, I'm just going to assume the first item in this iterator is the outer ring and following items are inner rings. We could probably migrate this to:

```rust
    fn outer_ring(&self) -> Self::ItemType;
    fn inner_rings(&self) -> Self::Iter {
        // Default is no inner rings
        std::iter::empty()
    }
```

Unless someone has a better idea for this.

jdroenner: I think this is nice :)

Turbo87:
@frewsxcv any progress on this? if not, any way we can help move this forward?

frewsxcv:
I'm trying to make these changes more incremental with other PRs, for example https://github.com/georust/rust-geo/pull/151

I did hit a roadblock with this and I remember needing access to streaming iterators, which aren't a thing in Rust (yet). This is pretty out-of-date and stale at this point, so I'm going to close it.

PR #87, opened 2017-02-12
Implement a Convex Hull method

urschrei:
The current implementation is slow (the worst case is On<sup>4</sup>) compared to e.g.
[Quickhull](http://www.cse.yorku.ca/~aaw/Hang/quick_hull/Algorithm.html) (I struggled with the recursion + retained points implementation), but it's a reasonable starting point, and a convex hull method is the fundamental building block for Delaunay triangulation and Voronoi tessellation methods.

### DIFF
--- a/src/algorithm/convexhull.rs
+++ b/src/algorithm/convexhull.rs
@@ -1,0 +1,174 @@
+use std::cmp::Ordering;
+use num_traits::Float;
+use types::{Point, Polygon, LineString};
+use std::collections::BTreeSet;
+
+impl<T> Eq for Point<T> where T: Float {}
+
+impl<T> PartialOrd for Point<T>
+    where T: Float
+{
+    fn partial_cmp(&self, other: &Point<T>) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl<T> Ord for Point<T>
+    where T: Float
+{
+    fn cmp(&self, other: &Point<T>) -> Ordering {
+        (self.x(), self.y()).partial_cmp(&(other.x(), other.y())).unwrap()
+    }
+}
+
+impl<T> Point<T>
+    where T: Float
+{
+    // Draw a horizontal line through this point, connect it with other,
+    // and measure the angle between these two lines.
+    fn angle(&self, other: &Point<T>) -> T {
+        if self == other {
+            T::zero()
+        } else {
+            (other.y() - self.y()).atan2((other.x() - self.x()))
+        }
+    }
+}
+
+// Barycentric method for point-in-triangle, see http://blackpawn.com/texts/pointinpoly
+impl<T> Polygon<T>
+    where T: Float
+{
+    fn barycentric_contains(&self, p: &Point<T>) -> bool {
+        let v0 = self.exterior.0[2] - self.exterior.0[0];
+        let v1 = self.exterior.0[1] - self.exterior.0[0];
+        let v2 = *p - self.exterior.0[0];
+        let dot00 = v0.dot(&v0);
+        let dot01 = v0.dot(&v1);
+        let dot02 = v0.dot(&v2);
+        let dot11 = v1.dot(&v1);
+        let dot12 = v1.dot(&v2);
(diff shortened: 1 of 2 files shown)


src/algorithm/convexhull.rs
```diff
@@ -0,0 +1,174 @@
...
+        let u = (dot11 * dot02 - dot01 * dot12) * inv_denom;
+        let v = (dot00 * dot12 - dot01 * dot02) * inv_denom;
+        (u > T::zero()) && (v > T::zero()) && (u + v < T::one())
+    }
+}
+
+// Adapted from http://codereview.stackexchange.com/a/141752/2630
+// The algorithm is from Heineman, G.T., Pollice, G., Selkow, S., 2008.
+// "Algorithms in a Nutshell". O’Reilly Media, Inc., pp261–8
+pub fn convex_hull<T>(points: &BTreeSet<Point<T>>) -> Vec<Point<T>>
```
frewsxcv: This doesn't need to be `pub`, right?

src/algorithm/convexhull.rs
```diff
@@ -0,0 +1,174 @@
+use std::cmp::Ordering;
+use num_traits::Float;
+use types::{Point, Polygon, LineString};
+use std::collections::BTreeSet;
+
+impl<T> Eq for Point<T> where T: Float {}
```
frewsxcv: Why is this needed?

src/algorithm/convexhull.rs
```diff
@@ -0,0 +1,174 @@
...
+    // when that's equal, sort by distance to head
+    hull.sort_by(|a, b| {
+        let angle_a = head.angle(&a);
+        let angle_b = head.angle(&b);
+        angle_a.partial_cmp(&angle_b).unwrap()
+    });
+    // we need to close the Polygon
+    let final_element = hull[0].clone();
+    hull.push(final_element);
+    hull.into_iter().collect::<Vec<Point<T>>>()
```
frewsxcv: Prior to this line, is `hull` not already a `Vec<Point<T>>`?

src/algorithm/convexhull.rs
```diff
@@ -0,0 +1,174 @@
...
+    /// let res = poly.convexhull();
+    /// assert_eq!(res.exterior, correct_hull);
+    /// ```
+    fn convexhull(&self) -> Self where T: Float;
+}
+
+impl<T> Convexhull<T> for Polygon<T>
+    where T: Float
+{
+    fn convexhull(&self) -> Polygon<T> {
```
frewsxcv: Nit: If "convex hull" is two words, I'd personally call this `ConvexHull::convex_hull`

src/algorithm/convexhull.rs
urschrei: `Ord` needs `Eq`, and deriving it in `types.rs` didn't work…

src/algorithm/convexhull.rs
urschrei: Whoops, nope.

src/algorithm/convexhull.rs
urschrei: Fixed!

src/algorithm/convexhull.rs
frewsxcv: Is there some reason you need `Ord`? IIRC, `num::Float` intentionally doesn't implement `Ord` since it can't guarantee total order...I think?

urschrei: This is a problem related to punting on `NaN` values – BTreeSet needs `Ord`, but as you say, there's no total ordering for types which included `NaN`. However, I'm not sure that it should be possible to construct geometries with NaN values in the first place.
I feel like there should be a more elegant solution here, but it'll need input from someone who knows more than I do about floating point…

frewsxcv: > `BTreeSet` needs `Ord`

Ah right, forgot about this. What you have is fine for now. Can you just add a comment next to the `Ord` implementation stating that it's needed for `BTreeSet`?

urschrei: Done!

PR #88, opened 2017-02-16
Replace within_epsilon() function with "approx" dev dependency

Turbo87:
see https://github.com/brendanzab/approx

/cc @frewsxcv


frewsxcv:
Great idea, thanks!

PR #89, opened 2017-02-16
Use QuickHull implementation for convex hull 

urschrei:
This supersedes #87, for a number of reasons:
- It's a lot faster – (O *n* log (*n*)) on average
- It's a far simpler implementation
- it doesn't require those questionable `Ord` and `Eq` traits

There's also a more robust test provided.


src/algorithm/convexhull.rs
```diff
@@ -0,0 +1,199 @@
...
+    // can't build a hull from fewer than four points
+    if points.len() < 4 {
+        return points.to_vec();
+    }
+    let mut hull: Vec<Point<T>> = vec![];
+    let mut min_x_idx = <usize>::min_value();
+    let mut max_x_idx = <usize>::min_value();
+    let mut min_x = Float::max_value();
+    let mut max_x = Float::min_value();
+    let to_retain: BTreeSet<usize> = (0..points.len() - 1).collect();
```
TeXitoi: You can simply write
```rust
let to_retain: BTreeSet<_> = (0..points.len() - 1).collect();'
```

src/algorithm/convexhull.rs
```diff
@@ -0,0 +1,199 @@
...
+    let mut to_remove: BTreeSet<usize> = BTreeSet::new();
```
TeXitoi: useless type annotation

src/algorithm/convexhull.rs
```diff
@@ -0,0 +1,199 @@
...
+    // min x and max x points are always part of the hull
+    hull.push(p_a);
+    hull.push(p_b);
+    let mut left_set: Vec<Point<T>> = vec![];
+    let mut right_set: Vec<Point<T>> = vec![];
+    // divide remaining points into left and right
+    // this is a bit hairy
+    for point in &to_retain.difference(&to_remove)
+        .map(|&idx| points[idx])
+        .collect::<Vec<Point<T>>>() {
```
frewsxcv: Can this `collect` call be removed? Seems like we could just iterate over the `Point`s in the `Iterator`

src/algorithm/convexhull.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+    // Determine points to the left of furthest_point, B
+    for point in set.iter() {
+        if point_location(&furthest_point, p_b, point) {
+            left_pb.push(*point);
+        }
+    }
+    // recur
+    hull_set(p_a, &furthest_point, &mut left_ap, hull);
+    hull_set(&furthest_point, p_b, &mut left_pb, hull);
```
TeXitoi: the hull.insert calls can have a squared effect because you push at the middle of a Vec (you then must shift n/2 elements). You can rewrite it just using `hull.push`:
```rust
// doing the first part
let mut new_set: Vec<_> = set.iter().filter(|point| point_location(p_a, &furthest_point, point)).collect();
hull_set(p_a, &furthest_point, &mut new_set, hull);

// inserting the current point
hull.push(set[furthest_idx]);

// doing the second part, reusing the memory allocated for the first part
new_set.clear();
new_set.extends(set.iter().filter(|point| point_location(&furthest_point, p_b, point));
hull_set(&furthest_point, p_b, &mut new_set, hull);
```

and correct the beginning of the function and the insertion in quick_hull accordingly.

frewsxcv: Seems like these calls could be done concurrently? If so, something we can do later in a future PR. Though we don't really depend on any parallelism libraries like rayon yet, hmm.

src/algorithm/convexhull.rs
```diff
@@ -0,0 +1,199 @@
...
+    fn convex_hull(&self) -> Self where T: Float;
+}
+
+impl<T> ConvexHull<T> for Polygon<T>
+    where T: Float
+{
+    fn convex_hull(&self) -> Polygon<T> {
+        Polygon::new(LineString(quick_hull(&self.exterior.0)), vec![])
+    }
+}
```
frewsxcv: Seems like those can happen in follow up PRs, but would be good to have support for those other types

src/algorithm/convexhull.rs
urschrei: Yep, I've been messing about with Rayon to try to parallelise this. It'll be in a follow-up, if only for discussion, because we're not doing it anywhere yet…

TeXitoi: I think you must have a very big polygon for parallelism being interesting.

src/algorithm/convexhull.rs
urschrei: Removing the `collect()` gives me a trait bound error:

`the trait std::iter::Iterator is not implemented for &std::iter::Map<std::collections::btree_set::Difference<'_, usize>, [closure@convexhull.rs:73:14: 73:32 points:_]>`

TeXitoi: you have to remove the `&` after the `for point in`

src/algorithm/convexhull.rs
urschrei: `let mut new_set: Vec<_> = set.iter().filter(|point| point_location(p_a, &furthest_point, point)).collect();` gives me a reference, but I need the Point…

    = note: expected type `&mut std::vec::Vec<types::Point<T>>`
    = note:    found type `&mut std::vec::Vec<&types::Point<T>>`

TeXitoi: add `.cloned()` after `.iter()`

urschrei: I'm trying to adapt this and https://github.com/georust/rust-geo/pull/89#discussion_r101758526, but it's failing to compile for me in about three different ways – can you send me a PR against https://github.com/urschrei/rust-geo/tree/convex_hull_quick?

urschrei:
I've covered the nits, but:
- https://github.com/georust/rust-geo/pull/89/commits/6aa011532d90c4507c0e0a3cc8a7030f2231b7b5 has exposed what looks like either an off-by-one or a logic error somewhere; very simple Polygons have their initial point duplicated by the algorithm (see the change at https://github.com/georust/rust-geo/pull/89/commits/f31f07e80db878fdb5246d3c0ceedc3819bcb661), but more complex Polygons such as poly1.rs and poly2.rs don't; the result matches Shapely's (and thus GEOS's), though in a different order.
- The vector-insert-avoiding change suggested by @TeXitoi is a crucial enhancement, but I can't seem to get all the way there.

src/algorithm/convexhull.rs
TeXitoi: https://github.com/urschrei/rust-geo/pull/1

urschrei:
Amazing work, thanks @TeXitoi!
I should be able to impl this for MultiPoint, LineString, and MultiLineString, but we're missing a method to get the exterior ring for a MultiPolygon at the moment.

TeXitoi:
For multipolygon, just aggregate all the exterior points of the outer polygons in a Vec. You need to give a mutable slice, so creating the Vec is needed.

src/algorithm/convexhull.rs
```diff
@@ -169,6 +169,44 @@ impl<T> ConvexHull<T> for Polygon<T>
     }
 }
 
+impl<T> ConvexHull<T> for MultiPolygon<T>
+    where T: Float
+{
+    fn convex_hull(&self) -> Polygon<T> {
+        let mut aggregated: Vec<Point<T>> = self.0.iter()
+            .flat_map(|elem| elem.exterior.0.clone())
```
TeXitoi:
```rust
.flat_map(|elem| elem.exterior.0.iter().cloned()
```

src/algorithm/convexhull.rs
```diff
@@ -169,6 +169,44 @@ impl<T> ConvexHull<T> for Polygon<T>
...
+        Polygon::new(LineString(quick_hull(&mut self.0.clone())), vec![])
+    }
+}
+
+impl<T> ConvexHull<T> for MultiLineString<T>
+    where T: Float
+{
+    fn convex_hull(&self) -> Polygon<T> {
+        let mut aggregated: Vec<Point<T>> = self.0.iter()
+            .flat_map(|elem| elem.0.clone())
```
TeXitoi: same

frewsxcv:
@urschrei Anything else here or is this good to merge?

urschrei:
Good to go here. Thanks!

PR #90, opened 2017-02-16
Implement Haversine algorithm

Turbo87:
This PR rebased the Haversine code from #62 and cleaned up the implementation and tests

Closes #62 

/cc @frewsxcv @tmcw

### DIFF
--- a/src/algorithm/haversine_distance.rs
+++ b/src/algorithm/haversine_distance.rs
@@ -8,12 +8,17 @@ pub trait HaversineDistance<T, Rhs = Self>
     /// Returns the distance between two points:
     ///
     /// ```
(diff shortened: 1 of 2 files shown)


Turbo87:
one thing I'm wondering is if we should use an `Angle` trait like http://tomaka.github.io/glium/cgmath/trait.Angle.html to reduce the unnecessary degrees-to-radians conversion if you calculate a lot of distances or other things

frewsxcv:
Awesome, thanks for rebase @Turbo87!

frewsxcv:
Regarding adding Angle, I'm not too familiar with it, but if it reduces and simplifies some of the algebraic logic in the library, I'm open to including it.

PR #95, opened 2017-02-19
Use Angle trait for lat/lng Points

Turbo87:
I've built an initial spike to use the `Angle` trait from the `cgmath` crate to represent lat/lng points.

/cc @tmcw @frewsxcv 

---

**Solved:** I'm flagging this as work-in-progress as I can't seem to be able to make `impl<T> Neg for Point<T> where T: Angle` work. It just fails to compile telling me ``conflicting implementations of trait `std::ops::Neg` for type `types::Point<_>` ``. It's been a long time since I last used Rust and if anyone has ideas on how to solve this I'd be happy to hear them.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ keywords = ["gis", "geo", "geography", "geospatial"]
 
 [dependencies]
 num-traits = "0.1"
+cgmath = "^0.12.0"
 
 [dev-dependencies]
 approx = "0.1.1"
--- a/src/types.rs
+++ b/src/types.rs
@@ -3,6 +3,7 @@ use std::ops::AddAssign;
 use std::ops::Neg;
 use std::ops::Sub;
 
+use cgmath::Angle;
 use num_traits::{Float, ToPrimitive};
 
 pub static COORD_PRECISION: f32 = 1e-1; // 0.1m
@@ -24,9 +25,7 @@ pub struct Bbox<T> {
 #[derive(PartialEq, Clone, Copy, Debug)]
 pub struct Point<T> (pub Coordinate<T>);
 
-impl<T> Point<T>
-    where T: Float + ToPrimitive
-{
+impl<T> Point<T> {
     /// Creates a new point.
     ///
     /// ```
@@ -40,7 +39,11 @@ impl<T> Point<T>
     pub fn new(x: T, y: T) -> Point<T> {
         Point(Coordinate { x: x, y: y })
     }
+}
 
+impl<T> Point<T>
+    where T: Float + ToPrimitive
+{
     /// Returns the x/horizontal component of the point.
     ///
     /// ```
@@ -97,60 +100,6 @@ impl<T> Point<T>
         self
     }
 
-    /// Returns the longitude/horizontal component of the point.
-    ///
-    /// ```
-    /// use geo::Point;
-    ///
-    /// let p = Point::new(1.234, 2.345);
-    ///
-    /// assert_eq!(p.lng(), 1.234);
-    /// ```
-    pub fn lng(&self) -> T {
-        self.x()
-    }
-
-    /// Sets the longitude/horizontal component of the point.
-    ///
-    /// ```
-    /// use geo::Point;
-    ///
-    /// let mut p = Point::new(1.234, 2.345);
-    /// p.set_lng(9.876);
-    ///
-    /// assert_eq!(p.lng(), 9.876);
-    /// ```
-    pub fn set_lng(&mut self, lng: T) -> &mut Point<T> {
-        self.set_x(lng)
-    }
-
-    /// Returns the latitude/vertical component of the point.
-    ///
-    /// ```
-    /// use geo::Point;
-    ///
-    /// let p = Point::new(1.234, 2.345);
-    ///
-    /// assert_eq!(p.lat(), 2.345);
-    /// ```
-    pub fn lat(&self) -> T {
-        self.y()
-    }
-
-    /// Sets the latitude/vertical component of the point.
-    ///
-    /// ```
-    /// use geo::Point;
-    ///
-    /// let mut p = Point::new(1.234, 2.345);
-    /// p.set_lat(9.876);
-    ///
-    /// assert_eq!(p.lat(), 9.876);
-    /// ```
-    pub fn set_lat(&mut self, lat: T) -> &mut Point<T> {
-        self.set_y(lat)
-    }
-
     /// Returns the dot product of the two points:
     /// `dot = x1 * x2 + y1 * y2`
     ///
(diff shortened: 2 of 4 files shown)


Turbo87:
@frewsxcv it looks like I've actually managed to implement Neg, Add and Sub generically now and I think this PR is now ready for review

frewsxcv:
Would it be cleaner if `cgmath::Deg` implemented `Float + ToPrimitive`? Maybe I/we should open a PR to cgmath

Turbo87:
> Would it be cleaner if cgmath::Deg implemented Float + ToPrimitive?

I'm not sure if that would be useful tbh

frewsxcv:
> > Would it be cleaner if cgmath::Deg implemented Float + ToPrimitive?
> 
> I'm not sure if that would be useful tbh

If `Deg` implemented `Float + ToPrimitive`, wouldn't it be able to use rust-geo without any changes?

I'm asking because it seems unfortunate that we have to duplicate some of these definitions. Makes it more confusing to have two ways to do something when we already have a *very* generic way to use rust-geo with `Float`.

Turbo87:
@frewsxcv I'm actually not duplicating any implementations anymore, I've just adjusted what implementations are valid for what kinds of traits

The problem with having `Deg` implement `Float + ToPrimitive` is that you would not gain any advantages from that. The main advantage here is being able to distinguish between radians and degrees, but if everything is handled as floats underneath without caring for their special meaning then we don't gain any advantages.

Turbo87:
@frewsxcv any more thoughts on this? should we merge it?

frewsxcv:
@Turbo87 Sorry for taking so long on this, I've been pretty busy recently. I've already looked through the code for this and it seems fine, I just want to `git checkout` your branch and play around with it. I'll try to review it sometime today, thanks for being patient :)

Turbo87:
cool, thanks!

after merging this we should try to finish up #85, otherwise the conflicts won't get any smaller 😉

src/types.rs
```diff
@@ -44,7 +39,11 @@ impl<T> Point<T>
     pub fn new(x: T, y: T) -> Point<T> {
         Point(Coordinate { x: x, y: y })
     }
+}
 
+impl<T> Point<T>
+    where T: Copy
```
frewsxcv: Not all numeric types implement `Copy`, for example `BigInt`.

src/types.rs
```diff
@@ -100,15 +99,25 @@ impl<T> Point<T>
...
     ///
-    /// let p = Point::new(1.234, 2.345);
+    /// # fn main() {
+    /// let p = Point::new(Deg(1.234), Deg(2.345));
     ///
-    /// assert_eq!(p.lng(), 1.234);
+    /// assert_eq!(p.lng(), Deg(1.234));
+    /// # }
     /// ```
     pub fn lng(&self) -> T {
```
frewsxcv: Why is `lng` only implemented for implementors of `Angle`? What about for other types that implement `Float`?

src/types.rs
```diff
@@ -226,7 +291,25 @@ impl<T> Sub for Point<T>
...
+    /// use cgmath::Deg;
+    ///
+    /// # fn main() {
+    /// let p = Point::new(Deg(1.25), Deg(3.0)) - Point::new(Deg(1.5), Deg(2.5));
+    ///
+    /// assert_eq!(p.lng(), Deg(-0.25));
+    /// assert_eq!(p.lat(), Deg(0.5));
+    /// # }
+    /// ```
+    /// ```
```
frewsxcv: Extra line of backticks here

Cargo.toml
```diff
@@ -11,6 +11,7 @@ keywords = ["gis", "geo", "geography", "geospatial"]
 
 [dependencies]
 num-traits = "0.1"
+cgmath = "^0.12.0"
```
frewsxcv: FYI `^` is redundant here. In a Cargo.toml file, `0.12.0` and `^0.12.0` are the same thing

http://doc.crates.io/specifying-dependencies.html#specifying-dependencies-from-cratesio

frewsxcv:
Just to take a step back here, what's the need for explicitly supporting this specific trait?

Is mixing up radians and degrees a common problem when doing geospatial operations? (This is an actual question, I really don't know the answer)

If so, it seems easy to get around this by one having their own `Degree` and `Radian` type that both implement `Float`. Right now, one cannot do `Point<X> + Point<Y>`, they have to be the same type.

Cargo.toml
Turbo87: didn't know that, thanks!

Turbo87:
okay, so let me start from the beginning. in most serialization formats degrees are used and if angles are presented to the user on the screen degrees are used too. in calculations however (sin, cos, etc.) the computer usually expects radians (see e.g. https://github.com/georust/rust-geo/blob/master/src/algorithm/haversine_distance.rs#L30-L31). if you only do a few calculations then converting from deg to rad for each calculation is fine, but if you do lots of calculations that will add up to a significant overhead.

that is basically the reason why it can make sense for some users to read the angles as degrees, but immediately convert to radians because they will be doing lots of calculations with them and only convert to degree again if the angles are serialized or shown to the user at the end.

> Right now, one cannot do Point<X> + Point<Y>, they have to be the same type.

if `Degree` and `Radian` would implement `Float` and `Float` implemented `Add` then `Degree(90) + Radian(0.5)` would result in `Float(90.5)` which is obviously wrong. The same would be the case for `Point<Degree>` and `Point<Radian>`.

Please let me know if that answers your questions.

frewsxcv:
Regarding your first couple paragraphs, I think you make valid points, but I need to digest and think about it a little bit.

> if `Degree` and `Radian` would implement `Float` and `Float` implemented `Add` then `Degree(90) + Radian(0.5)` would result in `Float(90.5)` which is obviously wrong.

If this were the case, I'd be more in favor of this pull request. Looking at [the docs](https://rust-num.github.io/num/num_traits/float/trait.Float.html), it looks like the `Float` inherits from the `Num` trait. It also looks like `Num` inherits from `Add<Output=Self>` and [since `RHS` isn't specified here](https://doc.rust-lang.org/nightly/core/ops/trait.Add.html), the type being added is the same as self, meaning that `Add` only works in situations where the left-hand-side and the right-hand-side are the same type:

```rust
extern crate num_traits;                                                                               
                                                     
// compiles                                                  
fn add_floats_1<X: num_traits::Float>(x: X, y: X) {                                                    
    let _ = x + y;                                                                                     
}                                                                                                      
         
// does not compile                                                                         
fn add_floats<X: num_traits::Float, Y: num_traits::Float>(x: X, y: Y) {                                
    let _ = x + y;                                                                                     
}                                                                                                      
                                                                                                       
fn main() {}
```

Turbo87:
@frewsxcv any news on this?

frewsxcv:
Sorry for the delay here. This is on my radar, I just haven't had time to look at this but I still plan to :-/

PR #98, opened 2017-03-03
Minor fixes

urschrei:
- Visvalingam algo doesn't need to be public
- Distinguish between Euclidean and Haversine distances
- Typo and minor grammar fixes


frewsxcv:
Thanks!

PR #101, opened 2017-03-14
Haversine fix

urschrei:


frewsxcv:
Thanks for tracking this down @urschrei!

bors r+

PR #104, opened 2017-03-17
Implement Point to multipart geometry distance methods

urschrei:
This adds distance methods and tests from Point to:
- MultiPoint
- MultiLineString
- MultiPolygon

For large MultiPolygons in particular, this could potentially be very slow. One approach would be the use of a spatial index storing the Polygon BoundingBoxes. Retrieving the closest one to the point, then calculating Point-to-Polygon distance would be much faster. There are a number of {quad-, r-, n-}tree implementations around in various states of maintenance, e.g. [acacia](https://github.com/aepsil0n/acacia), or (much heavier, extensively tested) [Spade](https://stoeoef.github.io/spade/spade/struct.RTree.html), so it'd be mostly a case of gluing things together.

(It might also be good to have a spatial index crate backed by one of those implementations in geo-rust, but that's a separate issue)

### DIFF
--- a/src/algorithm/distance.rs
+++ b/src/algorithm/distance.rs
@@ -168,7 +168,7 @@ impl<T> Distance<T, Polygon<T>> for Point<T>
 // Minimum distance from a Point to a MultiPolygon
 impl<T> Distance<T, MultiPolygon<T>> for Point<T>
     where T: Float
-{   
+{
     /// Minimum distance from a Point to a MultiPolygon
     fn distance(&self, mpolygon: &MultiPolygon<T>) -> T {
         let mut dist_queue: BinaryHeap<Mindist<T>> = BinaryHeap::new();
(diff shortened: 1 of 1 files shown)


src/algorithm/distance.rs
```diff
@@ -151,6 +164,32 @@ impl<T> Distance<T, Polygon<T>> for Point<T>
     }
 }
 
+// Minimum distance from a Point to a MultiPolygon
+impl<T> Distance<T, MultiPolygon<T>> for Point<T>
+    where T: Float
+{
+    fn distance(&self, mpolygon: &MultiPolygon<T>) -> T {
```
frewsxcv: I think if you did:

```rust
    /// Minimum distance from a Point to a MultiPolygon
    fn distance(&self, mpolygon: &MultiPolygon<T>) -> T {
```

It would show up in documentation (via rustdoc), just to be explicit about the behavior here.

urschrei: Done!

src/algorithm/distance.rs
```diff
@@ -167,7 +168,8 @@ impl<T> Distance<T, Polygon<T>> for Point<T>
 // Minimum distance from a Point to a MultiPolygon
 impl<T> Distance<T, MultiPolygon<T>> for Point<T>
     where T: Float
-{
+{
```
frewsxcv: Nit: whitespace

src/algorithm/distance.rs
```diff
@@ -180,7 +182,8 @@ impl<T> Distance<T, MultiPolygon<T>> for Point<T>
 // Minimum distance from a Point to a MultiLineString
 impl<T> Distance<T, MultiLineString<T>> for Point<T>
     where T: Float
-{
+{
```
frewsxcv: Nit: whitespace

frewsxcv:
LGTM after whitespace is addressed

PR #107, opened 2017-03-23
Add geometry rotation

urschrei:
This trait allows a geometry to be rotated about
a given origin by the given number of degrees,
using a transformation matrix

### DIFF
--- a/src/algorithm/rotate.rs
+++ b/src/algorithm/rotate.rs
@@ -23,6 +23,7 @@ fn rotation_matrix<T>(angle: T, origin: &Point<T>, points: &[Point<T>]) -> Vec<P
 
 pub trait Rotate<T> {
     /// Rotate a Geometry around a given origin by an angle, in degrees
+    /// Positive angles are counter-clockwise, and negative angles are clockwise rotations.
     ///
     /// ```
     /// use geo::{Point, LineString};
--- a/src/algorithm/rotate.rs
+++ b/src/algorithm/rotate.rs
@@ -48,7 +48,7 @@ impl<T> Rotate<T> for Point<T>
     where T: Float
 {
     /// Rotate the Point about the origin by the given number of degrees
-    fn rotate(&self, angle: T, origin: &Point<T>) -> Point<T> {
+    fn rotate(&self, angle: T, origin: &Point<T>) -> Self {
         rotation_matrix(angle, origin, &[*self])[0]
     }
 }
@@ -57,7 +57,7 @@ impl<T> Rotate<T> for LineString<T>
     where T: Float
 {
     /// Rotate the LineString about the origin by the given number of degrees
-    fn rotate(&self, angle: T, origin: &Point<T>) -> LineString<T> {
+    fn rotate(&self, angle: T, origin: &Point<T>) -> Self {
         LineString(rotation_matrix(angle, origin, &self.0))
     }
 }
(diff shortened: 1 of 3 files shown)


urschrei:
Thinking more about this, I feel like the default is rotation around the geometry's centroid. I think that means two traits…

urschrei:
Added a trait to rotate by an arbitrary Point, while `rotate` now always uses the centroid. One of the tests is failing, until the changes to the `centroid` trait merge.

frewsxcv:
FYI, looks like Travis found some test failures.

urschrei:
That failure is expected, because complex polygon centroids aren't currently calculated correctly. Fix is in #112.

urschrei:
FINALLY. Tests passing, and conforming with Shapely / GEOS results.
I'm gonna write the impls for the `Multi` geometries next. How do people feel about the traits themselves?

src/algorithm/rotate.rs
```diff
@@ -0,0 +1,251 @@
...
+    /// let rotated = linestring.rotate(-45.0);
+    /// let mut correct = Vec::new();
+    /// correct.push(Point::new(-2.0710678118654755, 5.0));
+    /// correct.push(Point::new(5.0, 5.0));
+    /// correct.push(Point::new(12.071067811865476, 5.0));
+    /// let correct_ls = LineString(correct);
+    /// assert_eq!(rotated, correct_ls);
+    /// ```
+    fn rotate(&self, angle: T) -> Self where T: Float;
+}
```
frewsxcv: I think you should be able to make a generic implementation for this, right? Something like:

```rust
impl<T> Rotate<T> for G<T>
    where T: Float,
          G: Centroid, RotatePoint
{
        self.rotate_point(self.centroid())
    }
}
```

I haven't actually tested this though

frewsxcv:
This all looks good to me! I might have called the method `rotate_around_point`, but not that big of a deal.

src/algorithm/rotate.rs
urschrei: Ah! Yep,

    impl<T, G> Rotate<T> for G
        where T: Float,
              G: Centroid<T> + RotatePoint<T>
    {
           self.rotate_point(angle, &self.centroid().unwrap())
        }
    }


Works. The only problem is the generic implementation doesn't work for Polygon, because rotation of a complex polygon uses the centroid of its *simple* representation. I already asked on IRC, but I don't think we can specify a more specific impl for Polygon at the moment?

urschrei:
Hmm you're right, it could be a bit confusing…

src/algorithm/rotate.rs
frewsxcv: I think what you have is fine, we can always make it more generic in the future if we find a better way.

frewsxcv:
Looks good! r=me, feel free to change the name of the method

urschrei:
bors r=frewsxcv

src/algorithm/rotate.rs
frewsxcv: @urschrei I'm revisiting this. Regarding polygon rotation, what's wrong with using the centroid of a complex polygon (using the weighted center) instead of how it works right now (just using the centroid of the exterior)?

urschrei: IIRC it's because that's what GEOS does; even if the polygon is complex, it's treated as simple when calculating the centroid. I don't know _why_ this is, though.

PR #108, opened 2017-03-26
Add "Orient" trait for Polygon

urschrei:
This returns a correctly-oriented copy of the polygon; `Default` returns counter-clockwise exterior rings and clockwise interior rings. `Reversed` returns rings with the opposite orientation.

### DIFF
--- a/src/algorithm/orient.rs
+++ b/src/algorithm/orient.rs
@@ -24,18 +24,29 @@ pub trait Orient<T> {
     /// let oriented_int = vec![(1.0, 0.5), (0.5, 1.0), (1.0, 1.5), (1.5, 1.0), (1.0, 0.5)];
     /// let oriented_int_ls = LineString(oriented_int.iter().map(|e| Point::new(e.0, e.1)).collect::<Vec<_>>());
     /// // build corrected Polygon
-    /// let oriented = poly.orient(Direction::Default);
+    /// let oriented = poly.orient(&Direction::Default);
     /// assert_eq!(oriented.exterior.0, oriented_ext_ls.0);
     /// assert_eq!(oriented.interiors[0].0, oriented_int_ls.0);
     /// ```
-    fn orient(&self, orientation: Direction) -> Self;
+    fn orient(&self, orientation: &Direction) -> Self;
 }
 
 impl<T> Orient<T> for Polygon<T>
     where T: Float
 {
-    fn orient(&self, direction: Direction) -> Polygon<T> {
+    fn orient(&self, direction: &Direction) -> Polygon<T> {
         orient(self, direction)
+    }
+}
+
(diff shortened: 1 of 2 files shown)


urschrei:
Could I get a review of this, if someone's got time?
Landing this would enable the `Rotation` trait to merge, but more interestingly, I've got a working implementation of the [Rotating Calipers](http://web.archive.org/web/20150330010154/http://cgm.cs.mcgill.ca/%7Eorm/rotcal.html) algorithm for finding the minimum distance between two Polygons – It requires both the `Orient` trait and an `Extremes` trait for finding minimum and maximum vertices of a convex polygon, which also depends on `Orient`.

src/algorithm/orient.rs
```diff
@@ -0,0 +1,139 @@
...
+                         .map(|poly| poly.orient(direction))
+                         .collect())
+    }
+}
+
+/// By default, a properly-oriented Polygon has its outer ring oriented counter-clockwise,
+/// and its inner ring(s) oriented clockwise. Selecting `Reversed` will result in a Polygon
+/// with a clockwise-oriented exterior ring, and counter-clockwise interior ring(s)
+#[derive(Debug)]
+pub enum Direction {
```
frewsxcv: If you added `#[derive(Copy, Clone, Debug)]` here, you wouldn't need to pass `&Direction` references everywhere since it's just as cheap to pass the enum value than it is to pass a reference to the enum value.

src/algorithm/orient.rs
```diff
@@ -0,0 +1,139 @@
...
+    Default,
+    Reversed,
+}
```
frewsxcv: Might be good to explicitly mention:

```rust
enum Direction {
    /// exterior ring is oriented counter-clockwise, interior rings are oriented clockwise
    Default,
    /// exterior ring is oriented clockwise, interior rings are oriented counter-clockwise
    Reversed,
}
```

frewsxcv:
> Could I get a review of this, if someone's got time?

Sorry about that, this PR slipped by me. Feel free to ping me directly if you need a review for GeoRust things. I also just realized you aren't part of the GeoRust GitHub organization, sorry about that! Just added you.

r=me after my couple minor comments (you should be able to add a comment `bors r=frewsxcv` to trigger [the mergebot I have enabled for this repo](https://bors-ng.github.io/)0

frewsxcv:
bors r-

no bors, stop, not yet

frewsxcv:
Unrelated to this PR, I've been working on a [geospatial data viewer](https://github.com/frewsxcv/rgis/). My next steps are to add a basic UI and add in some of the operations/algorithms you've implemented in this repository, so thanks for all your work here :)

urschrei:
bors r=frewsxcv

PR #110, opened 2017-03-27
Quickhull orientation fix

### LINKED ISSUE #109: Convex Hull is oriented clockwise
The convex hull implementation outputs a clockwise-oriented exterior ring. This isn't strictly wrong, but many other algorithms assume a counter-clockwise exterior ring. The simplest fix here is to [reverse `hull`](https://github.com/georust/rust-geo/blob/master/src/algorithm/convexhull.rs#L103) with `into_iter().rev().collect()`, but I'm not sure what its overhead is – it'd be better to fix the implementation to order the edge segments correctly.

urschrei:
This closes #109, but is WIP until I write a few more tests and get some more eyes on it.


urschrei:
I'm happy that this is generating counter-clockwise hulls now – the tests are pretty exhaustive, match Shapely / GEOS output, and resulting polygons are verified as valid, convex, and counter-clockwise. If anyone can think of any other ways to exercise the algorithm I'm all ears, but otherwise I'm happy for this to merge.

PR #112, opened 2017-03-29
Correctly calculate centroids for complex polygons.

### LINKED ISSUE #111: Polygon centroid calculation should account for interior holes
It doesn't at the moment, meaning that rotation around a polygon centroid (see #107) is only possible for simple polygons. If anyone has suggestions for existing clean algorithm implementations, I'm happy to rewrite the existing implementation.

urschrei:
This PR closes #111. I've added helper functions for calculation of a simple polygon centroid, and for simple polygon area. The latter is functionally identical to `get_linestring_area()`, but we can't use that since it's a private function in a public module.

### DIFF
--- a/src/algorithm/centroid.rs
+++ b/src/algorithm/centroid.rs
@@ -23,21 +23,47 @@ pub trait Centroid<T: Float> {
     fn centroid(&self) -> Option<Point<T>>;
 }
 
+// Calculation of simple (no interior holes) Polygon area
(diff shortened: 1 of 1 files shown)


frewsxcv:
Is this ready to go? Looks good to me if so!

urschrei:
Yep!

frewsxcv:
bors=frewsxcv,texitoi

frewsxcv:
bors r=frewsxcv,texitoi

PR #113, opened 2017-04-01
Implement distance traits for more geometries

urschrei:
This PR adds distance traits for all geometries for which the trait was already implemented in one direction. E.g. if `Point`-to-`Polygon` distance existed, `Polygon`-to-`Point` distance now also exists. This adds a few boilerplate functions, but I can't think of a more ergonomic solution.

Still missing:
- `LineString`-to-`LineString` minimum distance (is there a good algorithm for this?)
- `Polygon`-to-`Polygon` minimum distance (I'm (slowly) working on a [rotating calipers](http://web.archive.org/web/20150407022621/http://cgm.cs.mcgill.ca/~orm/mind2p.html) implementation)

### DIFF
--- a/src/algorithm/distance.rs
+++ b/src/algorithm/distance.rs
@@ -10,10 +10,10 @@ use num_traits::pow::pow;
 pub trait Distance<T, Rhs = Self> {
     /// Returns the distance between two geometries
     ///
-    /// If a `Point` is contained by a `Polygon`, the distance is `0.0`  
-    /// If a `Point` lies on a `Polygon`'s exterior or interior rings, the distance is `0.0`  
-    /// If a `Point` lies on a `LineString`, the distance is `0.0`  
-    /// The distance between a `Point` and an empty `LineString` is `0.0`  
+    /// If a `Point` is contained by a `Polygon`, the distance is `0.0`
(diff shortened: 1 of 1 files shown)


src/algorithm/distance.rs
```diff
@@ -10,10 +10,10 @@ use num_traits::pow::pow;
 pub trait Distance<T, Rhs = Self> {
     /// Returns the distance between two geometries
     ///
-    /// If a `Point` is contained by a `Polygon`, the distance is `0.0`
```
frewsxcv: I think these spaces might have been intentional because markdown treats them as newlines:

http://spec.commonmark.org/0.26/#hard-line-breaks

I think with your changes, these sentences get put on the same line

urschrei: Oops. Fixed.

frewsxcv:
Thanks for this! It's unfortunate that we can't do something like https://github.com/georust/rust-geo/issues/73

bors r+

urschrei:
That would be great, and I did wonder whether something like that would be possible given the symmetry of `From<T> for U` and `Into<U> for T` impls, but I don't know nearly enough about how it would work under the hood. Maybe a post for Internals…

PR #114, opened 2017-04-10
Add extreme point-finding

urschrei:
This PR allows the calculation of the indices containing the extreme (min and max) `x` and `y` values of a geometry (any geometry for which we can build a convex hull). There are some subtleties:

The algorithm is simple, and runs in `O(n)` time, but only works on convex geometries, which it expects to be oriented counter-clockwise. Since convex-hull processing using QuickHull is `O(n log n)` on average, that's the lower bound if the geometry needs to be processed first. However, if you know your geometries are convex and oriented, you can specify this, and get `O(n)` processing.

There's also a binary-search algorithm available, which cuts processing time to `O(log n)`, but I haven't been able to make it work. It's [available in a previous commit](https://github.com/urschrei/rust-geo/commit/4f71b24030a888cd95506747a0f1feaa3cce4201#diff-6128c8bf9f8def540989e4f8abf862c7L85) along with a link to the source, if anyone wants to have a look at it and figure out where the problem is – it incorrectly returns minimum `x`. The current algorithm is implemented using a wrapper function that allows the algorithms to be easily swapped out.

I've also defaulted to returning indices, not points, because this functionality is primarily intended as a helper for other algorithms.

### DIFF
--- a/src/algorithm/extremes.rs
+++ b/src/algorithm/extremes.rs
@@ -109,19 +109,19 @@ pub trait ExtremePoints<T: Float> {
     /// let points = points_raw.iter().map(|e| Point::new(e.0, e.1)).collect::<Vec<_>>();
     /// let poly = Polygon::new(LineString(points), vec![]);
     /// // Polygon is both convex and oriented counter-clockwise
-    /// let extremes = poly.extreme_points(true, true);
+    /// let extremes = poly.extreme_indices(true, true);
     /// assert_eq!(extremes.ymin, 0);
     /// assert_eq!(extremes.xmax, 1);
     /// assert_eq!(extremes.ymax, 2);
     /// assert_eq!(extremes.xmin, 3);
     /// ```
-    fn extreme_points(&self, convex: bool, oriented: bool) -> Extremes;
+    fn extreme_indices(&self, convex: bool, oriented: bool) -> Extremes;
 }
 
 impl<T> ExtremePoints<T> for Polygon<T>
     where T: Float
 {
-    fn extreme_points(&self, convex: bool, oriented: bool) -> Extremes {
+    fn extreme_indices(&self, convex: bool, oriented: bool) -> Extremes {
         find_extremes(polymax_naive, self, convex, oriented)
     }
 }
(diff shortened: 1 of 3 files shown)


urschrei:
Argh, I've just realised that the indices are only valid for input geometries that are convex. Because the convex-hull algorithm alters vertex order **and** may remove vertices, the returned struct contains indices that are wrong if the input geometry wasn't convex to begin with.

It's trivial to use this work as the basis for a trait that returns the extreme points (which is useful functionality imo), and it's necessary to have the extreme indices of a convex geometry as part of the polygon-to-polygon distance functionality, but it shouldn't be a public API.

frewsxcv:
I think it's fine leaving it public, as long as the limitations of the algorithm are documented, I think it's fine. r=me if this is good to go

urschrei:
OK, the `ExtremeIndices` trait now checks the input Polygon for convexity, and returns a `Result`, which is the safest way to handle this, I think.

I'm using a [wrapping index lookup](https://github.com/georust/rust-geo/pull/114/files#diff-6128c8bf9f8def540989e4f8abf862c7R53) to get the two previous points. I don't think I can over / underflow it, and it's guaranteed to always receive a valid index, since it's only called by `.iter().enumerate()`. Still, if someone could have a look at it, since I'm always nervous about doing manual index arithmetic.

I've also added an `ExtremePoints` trait, which I've made available to anything implementing `ConvexHull + ExtremeIndices`. I noticed that when you implement it like this, it doesn't explicitly show up in the docs as a trait available to the relevant structs – do I have to do anything else to get it to show up, or is it a doc limitation?

src/algorithm/extremes.rs
```diff
@@ -37,21 +37,79 @@ fn below<T>(u: &Point<T>, vi: &Point<T>, vj: &Point<T>) -> bool
...
+{
+    fn previous_vertex(&self, current_vertex: &usize) -> usize
+        where T: Float
+    {
+        (current_vertex + (self.exterior.0.len() - 1) - 1) % (self.exterior.0.len() - 1)
+    }
+}
+
+// positive implies a -> b -> c is counter-clockwise, negative implies clockwise
+pub fn cross_prod<T>(p_a: &Point<T>, p_b: &Point<T>, p_c: &Point<T>) -> T
```
frewsxcv: Did you mean for this to be public?

frewsxcv:
LGTM. r=me if it's ready

urschrei:
r=frewsxcv

PR #115, opened 2017-04-17
WIP: implement Polygon-Polygon, LineString-Polygon, Line-Polygon distance

urschrei:
This PR is about 80% of the work towards implementing Polygon-to-Polygon distance. I'm opening it now in order to commit myself to finishing it.

The problem has four variations:
1. Disjoint polygons whose convex hulls are disjoint. This is the most common case
2. Disjoint polygons whose convex hulls intersect
3. Intersecting Polygons
4. Complex polygons whose internal holes contain Polygons.

---
screenshot 2017-04-12 00 41 56
**In the case of 4**, I have no idea how to deal with it yet (it may be solved by the same solution as 2) 

**In the case of 3**, we don't need to worry about it. The distance is 0

screen shot 2017-04-17 at 17 18 37
<s>**In the case of 2**, we can use a variation of the algorithm described by [Toussaint and Bhattacharya](http://www-cgrl.cs.mcgill.ca/~godfried/publications/mindist.pdf), which is fairly straightforward:
1. Form a Delaunay triangulation of the union of the two sets (polygons) of points
2. (Optionally, it cuts down the processing time in step 3) compute the [Gabriel Graph](https://en.wikipedia.org/wiki/Gabriel_graph)
3. Find all the edges from 1 (2) whose pairs of points contain one point from Polygon P and one point from Polygon Q
4. Compute the length of all these edges
5. The minimum length is also the minimum set distance
6. Since the minimum distance could alternatively involve an edge, check the distance between each of the two candidate points in step 4, and the edge formed by `[idx -1, other point idx, idx + 1]`
7. The minimum of 5 and 6 is the minimum distance.

I've prototyped this, but need to do more work to find a Delaunay data structure that can keep track of input indices and easily enumerate edges. [Spade](https://docs.rs/spade/1.1.0/spade/delaunay/index.html) is looking like the likely candidate here.</s>
(This didn't work out, so I'm going to fall back to what GEOS does here, with an R*-Tree and a nearest-neighbour search)

screenshot 2017-04-17 17 21 53
**In the case of 1**, we can use the [Rotating Calipers](http://digitool.library.mcgill.ca/R/?func=dbin-jump-full&object_id=21623&local_base=GEN01-MCG02) method to find the minimum distance. This work is complete, tested, and needs review. Some notes:
- It's fast; the algorithm runs in `O(n + m)`, though we have to compute the convex hulls, which is `O(n log n)`.
- It doesn't allocate much; one of the distance calculations creates a new `LineString`, which I suspect I'll be able to get rid of.
- It's conceptually simple; spin two parallel lines around the convex hulls of the polygons, and after each step, do some distance calculations. When you're back at your start points, the smallest distance you found is the minimum distance.
- In practice it's really complicated; [`unitvector`](https://github.com/urschrei/rust-geo/blob/372a301ddc9cc0eb7dd4e2b1477b2bfa7fed5e0a/src/algorithm/distance.rs#L406-L552) is basically a black box; I ported it from Java, simplified the branching using clippy, and then left it as is. It has to handle a lot of special cases related to slopes and orientation, and I'm not sure whether they can be simplified. It may be possible to break them up and remove some redundancy, though.

The algorithm itself runs as follows:
1. We compute the convex hulls of the polygons P and Q, and find the minimum P y point, and the maximum Q y point. These are our start positions.
2. We initialise our `Polydist` struct with the polygons, extreme y points, and some placeholder values.
3. We enter the loop. At each iteration, we pass `Polydist` to [`nextpoints`](https://github.com/urschrei/rust-geo/blob/372a301ddc9cc0eb7dd4e2b1477b2bfa7fed5e0a/src/algorithm/distance.rs#L740-L830), to calculate the next set of caliper points, then to [`computemin`](https://github.com/urschrei/rust-geo/blob/372a301ddc9cc0eb7dd4e2b1477b2bfa7fed5e0a/src/algorithm/distance.rs#L837-L953), to compute the minimum distance for this iteration
4. At each iteration, one of three states is possible:
    1. One caliper is aligned with an edge on `P`, and a vertex on `Q`
    2. One caliper is aligned with an edge on `Q`, and a vertex on `P`
    3. Both calipers are aligned with edges, which implies they're parallel.
5. Depending on the state of 4, a series of distance calculations and intersection checks are carried out, and the minimum distance updated as necessary. See p31 of the paper for details of how to check for orthogonal intersections.
6. When we're back at our start positions, we exit the loop.

Three measures of distance are possible:
1. Between a vertex from P and a vertex from Q
2. Between a vertex from P and an edge from Q, or vice versa
3. Between an edge from P and an edge from Q.

I've written tests for each of these possibilities, and the results agree with GEOS.
Which brings me to my next point. I initially thought I could copy GEOS or JTS, but it turns out they punted on all this and just brute-force the problem, which is much more simple, and takes ✨ quadratic ✨ time.

OK, this is already the longest PR of all time. The code is carefully commented, and in good shape. I've used enums and matches where it makes sense, including replacing simple if / else clauses. The method of passing around a struct of mutable (and in some cases, borrowed) values feel a little bit un-Rust-like, but again, I think it makes sense in the context of how the algorithm works; there's no way of using `iter()` for this.


urschrei:
Now that we have an R*-Tree implementation available, I'm going to get on with this. Contrary to my initial enthusiasm, Case 2 above can't easily be solved, so I'm going to do what GEOS does, and use a nearest neighbour-search to calculate the point-to-line-segment distances between two polygons, keeping the smallest distance.
That leaves us with Case 4, which I still haven't looked at.

src/algorithm/distance.rs
```diff
@@ -250,35 +273,783 @@ impl<T> Distance<T, LineString<T>> for Point<T>
...
+
+// Angle between a vertex and an edge
+fn vertex_line_angle<T>(poly: &Polygon<T>, p: &Point<T>, m: &T, vertical: bool, idx: &usize) -> T
+where
+    T: Float + FloatConst,
+{
+    let hundred = T::from(100).unwrap();
+    let pnext = poly.exterior.0[poly.next_vertex(idx)];
+    let pprev = poly.exterior.0[poly.prev_vertex(idx)];
+    let clockwise = if cross_prod(&pprev, p, &pnext) < T::zero() {
```
frewsxcv:
```rust
let clockwise = cross_prod(&pprev, p, &pnext) < T::zero();
```

urschrei:
Some updates:
- Polygon to Polygon distance is now working, and has good test coverage
- I've moved the "fast path" option out of the `Distance` module, because it's 700 loc, and it's currently in a crate-only sibling module called `polygon_distance_fast_path`

I'm actually in two minds about keeping it in rust-geo:
- it's a lot of code
- some of it is a black box, in the sense that I know what it does, but I don't know _why_ it works. Another maintainer would have a hill to climb if they needed to debug it (I've read the original paper about a dozen times, and spent a couple of weeks with the original code)
- it only works for convex input polygons. This is a fairly common case, but I've got no statistics to back it up
- that means that all Polygon-Polygon distance operations are paying an additional `O(M+N)`cost for the convexity check
    - in practice, the R* tree method is slow, so I'm pretty sure it bounds `O(M+N)` anyway, but it's an additional slow-down in real-world usage.

On the other hand:
- It's very fast, and doesn't allocate.
- No other open-source geo library has it

What do people think? @frewsxcv @mbattifarano @rory 

Some TODOs:

- [ ] All Polygon distance checks have a corner case in which the other geometry is contained inside an inner ring. When this is the case it's disjoint, so you can't detect it using `intersects` or `contains`. The best I've come up with is checking whether it's contained by a simple polygon built from the outer shell. But that needs a `clone()` operation, and I'd love to get rid of it. Anyone got ideas that don't involve `unsafe`?
- [ ] Line to Polygon distance needs a couple of special cases covered, but it's mostly done

urschrei:
From beyond the mists of time, an improvement I suggested to the R* crate over the summer has landed on master (https://github.com/Stoeoef/spade/commit/fcff8b1489f93a470407f89e708ee3a347d6c82e), which halves the time taken by the naïve polygon-polygon distance calculation. Distance from a 9k-vertex non-convex polygon to another polygon runs in around 40ms on my 3.4 GHz Core i7 \o/

(This case is still around 5x slower than GEOS, but it's a lot of progress)

PR #116, opened 2017-04-23
Address some compiler warnings

michaelkirk:
Just some code cleanup

- unused imports
- dead code
- unneccesary mut

### DIFF
--- a/src/algorithm/extremes.rs
+++ b/src/algorithm/extremes.rs
@@ -11,6 +11,14 @@ use types::{Extremes, ExtremePoint};
 
 // various tests for vector orientation relative to a direction vector u
 
+// Not currently used, but maybe useful in the future
+#[allow(dead_code)]
+fn up<T>(u: &Point<T>, v: &Point<T>) -> bool
+    where T: Float
+{
+    u.dot(v) > T::zero()
+}
+
 fn direction_sign<T>(u: &Point<T>, vi: &Point<T>, vj: &Point<T>) -> T
     where T: Float
 {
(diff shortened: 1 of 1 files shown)


src/algorithm/extremes.rs
```diff
@@ -282,7 +279,7 @@ mod test {
     #[test]
     fn test_polygon_extreme_wrapper_convex() {
         // convex, with a bump on the top-right edge
-        let mut points_raw =
+        let points_raw =
```
urschrei: the `reverse()` call on line 288 requires this to be `mut`

src/algorithm/extremes.rs
```diff
@@ -11,11 +10,6 @@ use types::{Extremes, ExtremePoint};
 // -1, 0. = smallest x
 
 // various tests for vector orientation relative to a direction vector u
-fn up<T>(u: &Point<T>, v: &Point<T>) -> bool
```
urschrei: I'd prefer this to be kept too, in case binary search lands, so just tagging it with `#[allow(dead_code)]` is fine.

michaelkirk: Can you recommend a useful comment? I'm not sure what the method does.

src/algorithm/extremes.rs
michaelkirk: @urschrei  Sorry, what reverse call on line 288? I don't see `reverse` anywhere in this file.

src/algorithm/extremes.rs
michaelkirk: On second thought, it's pretty straight forward.

src/algorithm/extremes.rs
urschrei: My bad, I was looking at the wrong branch. Sorry!

src/algorithm/extremes.rs
urschrei: Yep, but something like "is vector v pointing up, relative to vector u" (as per the [original](http://geomalgorithms.com/a14-_extreme_pts.html#polyMax_2D())) is fine, just in case.

PR #117, opened 2017-05-05
Add contains point impl for bbox

etrombly:
This makes it quicker to parse shapefiles, there's usually a bbox associated with a polygon. The contains check is a lot faster for the bbox.

### DIFF
--- a/src/algorithm/contains.rs
+++ b/src/algorithm/contains.rs
@@ -152,6 +152,15 @@ impl<T> Contains<LineString<T>> for Polygon<T>
     }
 }
 
+
+impl<T> Contains<Point<T>> for Bbox<T>
+    where T: Float
+{
+    fn contains(&self, p: &Point<T>) -> bool {
+        p.x() >= self.xmin && p.x() <= self.xmax && p.y() >= self.ymin && p.y() <= self.ymax
+    }
+}
+
 impl<T> Contains<Bbox<T>> for Bbox<T>
     where T: Float
 {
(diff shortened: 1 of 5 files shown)


src/algorithm/contains.rs
```diff
@@ -166,6 +175,7 @@ impl<T> Contains<Bbox<T>> for Bbox<T>
 mod test {
     use types::{Coordinate, Point, LineString, Polygon, MultiPolygon, Bbox};
     use algorithm::contains::Contains;
+        #[test]
```
frewsxcv: Any reason for this addition?

etrombly: That was unintentional. I had temporarily made a test to verify it worked, then deleted it.

frewsxcv:
I'm not opposed to adding serialization support, but out of curiosity, what's your use case? Just trying to learn how people are using this library

etrombly:
Oh, sorry. Didn't mean for these to go in. I was wanting to add map data using include bytes so I didn't have to package the maps with the professor I'm working on. Was just seeing if it would work.

etrombly:
Program, not professor. On mobile

frewsxcv:
No worries! I'll cherry pick your first commit right now

src/algorithm/intersects.rs
```diff
@@ -73,7 +73,7 @@ impl<T> Intersects<Bbox<T>> for Bbox<T>
 {
     fn intersects(&self, bbox: &Bbox<T>) -> bool {
         // line intersects inner or outer polygon edge
-        if bbox.contains(&self) {
+        if bbox.contains(self) {
```
frewsxcv: Opened an issue for this: https://github.com/rust-lang/rust/issues/41812

frewsxcv:
Merged in https://github.com/georust/rust-geo/commit/3e00ef94c3d69e6d0b1caab86224469ced9444e6 . Will publish a new release right now

Issue #99, opened 2017-03-13
Add a line/line-segment type?


kestred:
Does it make sense to have a separate type for a Line, like https://github.com/paulmach/go.geo/blob/master/line.go (in that it is treated as a LineString for GeoJSON/WKT)?

In particular, I would like this feature for strictly typed conversions using Postgres's builtin Line Segment type (using https://github.com/sfackler/rust-postgres). I had heard that it might be the plan to represent `geo::types` as traits; in which case this could either be its own trait or an implementation of `LineString`.  (Aside: if you've written up the current direction of georust somewhere, I'd love a link!).

frewsxcv:
Is the distinction here that a `Line` is a series of `Coordinate`s instead of a series of `Point`s?

TeXitoi:
I think here a line is a segment, composed of 2 coordinates.

frewsxcv:
> (Aside: if you've written up the current direction of georust somewhere, I'd love a link!)

I don't have anything in writing really. I haven't done a great job at establishing a direction for rust-geo, other than saying it's a place for geospatial types and algorithms. I don't do much geospatial work anymore, and have been pretty busy with some other Rust projects, so if anyone has good ideas for the direction of this crate, let me know.

> I had heard that it might be the plan to represent geo::types as traits; in which case this could either be its own trait or an implementation of LineString.

Traits would be cool, and I have a pull request open experimenting with them, but with the direction I was heading, it would have made the ergonomics worse than they currently are, and I wasn't a fan of that. That's not to say traits aren't the way forward, they just might require more thinking.

> Does it make sense to have a separate type for a Line, like https://github.com/paulmach/go.geo/blob/master/line.go (in that it is treated as a LineString for GeoJSON/WKT)?

I'm not really sure about this. For now, I've been very loosely following version 1.1.0 of the Simple Features spec

* https://en.wikipedia.org/wiki/Simple_Features
* http://www.opengeospatial.org/standards/sfa

which has a data model that looks like:

<img width="737" alt="screen shot 2017-03-18 at 4 11 44 pm" src="https://cloud.githubusercontent.com/assets/416575/24075778/36e8ba98-0bf8-11e7-891f-fb760deac042.png">

the latest version of the spec (v 1.2.1) has a data model that looks like:

[...]

Issue #81, opened 2016-10-17
Create `Union` algorithm trait


frewsxcv:
![](https://www.mathworks.com/help/map/polybool.png)

Similar to https://github.com/georust/rust-geo/issues/80

acfoltzer:
Would this be a good issue for a new contributor to work on? I could really use polygon union for one of my projects, though I'd have to study up on potential algorithms.

urschrei:
@acfoltzer Absolutely! Some ideas have already been kicked around in #80. In short, it requires a lot of other things to be in place – the implementation of the binary search tree (Bentley-Ottmann needs a tree that can update its sweepline value, see https://github.com/ideasman42/isect_segments-bentley_ottmann/issues/10) and priority queue that @shterrett refers to remain open questions, though as I mention there's now an excellent DCEL implementation available in Spade. Progress on any aspect of this would be great. It's also worth looking at the clipping crate, depending on what kind of polygons you're working with. It's less general, but may be suitable for your use case.

urschrei:
Closed by #835!

Issue #80, opened 2016-10-17
Create `Intersection` algorithm trait 


frewsxcv:
![](https://www.mathworks.com/help/map/polybool.png)

It might be the case (in the case of intersecting two polygons) that the operation will return a polygon _or_ a linestring _or_ a point _or_ nothing, the return type will need to reflect this. Maybe something like this?

``` rust
trait Intersection<...> {
    fn intersection(&self, other) -> Option<IntersectionResult> {
        ...
    }
}

// not a fan of "Result" here since it sounds similar to std::result::Result
enum IntersectionResult {   
    Polygon(Polygon),
    LineString(LineString),
    Point(Point),
}
```

Note that this is different from the [`Intersects`](https://georust.github.io/rust-geo/geo/algorithm/intersects/trait.Intersects.html) trait which simply returns a `bool` indicating if two geometries intersect at any point

ambaxter:
Yup. That would be perfect.

mbattifarano:
I think Vatti's Algorithm would work well here. My understanding is that it can be parameterized to produce union, intersection, xor and difference.

https://en.wikipedia.org/wiki/Vatti_clipping_algorithm
extras.springer.com/2005/978-1-84628-108-2/VattiClip.pdf

shterrett:
@mbattifarano and I were looking into this the other day. I am planning on implementing the algorithm that is outlined in [de Berg, Cheong, van Kreveld, Overmars Computational Geometry](http://www.cs.uu.nl/geobook/). Briefly

+ Represent the polygons in a doubly-linked edge list so that each edge has a reference to its successor and predecessor, and each face has a reference to an incident edge
+ Combine the two polygons into a single list, ignoring edge intersections
+ Use a scanning line algorithm to detect all of the intersections, and add a node at each point, aligning the references as necessary, including a reference to the originating polygon for each face (which may be both)
+ Use boolean operations on the faces + originating polygons to keep only those desired (intersection/union/etc...)

This calls for a few new data structures:
+ Doubly linked polygon edge list
+ Line segment (only two points; not a `LineString`)
+ Priority Queue that allows testing for containment and maintains order of the "event points" by upper-left-ness
+ Binary search tree that returns the adjacent segments to given segment at any point

[...]

PR #118, opened 2017-05-24
Add `Line` type

mbattifarano:
Adds a `Line` type to represent a line segment by its two endpoints.

This PR is the first step to implementing Union/Intersection traits as outlined in #80 (See also #81). Additionally, it addresses #99.

Implements the following (with converse operations where applicable) for `Line`:
- `Length`
- `Area`
- `BoundingBox`
- `Centroid`
- `Rotate`
- `Contains` (`Point`, `Line`, `LineString`)
- `Intersects` (`Point`, `Line`, `LineString`, `Polygon`)
- `Distance` (`Point`)

I wanted to get this up sooner rather than later, and it's already on the large side so there are a few things left unimplemented (operations on `Multi*` geometries for example). Additionally, several operations on `LineString` (`Length` for example) could be nicely expressed as operations on an `Line` iterator. As a result, there is some redundancy that could be resolved in subsequent PRs.

### DIFF
--- a/src/algorithm/area.rs
+++ b/src/algorithm/area.rs
@@ -121,7 +121,7 @@ mod test {
     #[test]
     fn area_line_test() {
         let p = |x, y| Point(Coordinate { x: x, y: y });
-        let line1 = Line((p(0.0, 0.0), p(1.0, 1.0)));
+        let line1 = Line::new(p(0.0, 0.0), p(1.0, 1.0));
         assert_eq!(line1.area(), 0.);
     }
 }
--- a/src/algorithm/area.rs
+++ b/src/algorithm/area.rs
@@ -1,5 +1,5 @@
 use num_traits::Float;
-use types::{LineString, Polygon, MultiPolygon, Bbox};
+use types::{Line, LineString, Polygon, MultiPolygon, Bbox};
 
 /// Calculation of the area.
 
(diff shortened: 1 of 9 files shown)


frewsxcv:
sorry for the delay! this seems alright with me. my main hesitation with this is https://github.com/georust/rust-geo/pull/118/commits/7a4e3980253ac2dc9b8d1d60846dc1598ac69ddd, but i think i have hesitations about the `Geometry` `enum` in general. anyone else have thoughts? @urschrei?

mbattifarano:
@frewsxcv I added `Line` to `Geometry` in this PR purely for the sake of consistency. I'm happy to revert the commit, or leave it as is and deal with the fate of the enum separately.

frewsxcv:
let's wait a couple and see if anyone else has anything to say, otherwise I'll ditch the `Geometry` `enum` change and merge this in

urschrei:
Sorry for the delay I agree that having it in the enum could lead to confusion. Related: Are any of the public / private visibility refinements that are landing in 1.19 useful here?

frewsxcv:
> Related: Are any of the public / private visibility refinements that are landing in 1.19 useful here?

Not that I know of, if you find anything, please please open an issue :)

frewsxcv:
fyi, I rebased and merged without the `Geometry` enum change. thanks again for the contribution!

PR #120, opened 2017-06-20
Incorrect usage of `abs_sub`

mbattifarano:
`x.abs_sub(y)` is defined ([confusingly](https://github.com/rust-lang/rust/issues/30315)) as `max(x-y, 0)`, not `abs(x-y)`. This wasn't caught in tests since `t_x - t_y` in [src/algorithm/intersects.rs:48](https://github.com/georust/rust-geo/blob/master/src/algorithm/intersects.rs#L48) happens to be positive in all of the given examples so I've added a new example where `t_x - t_y` is negative.

### DIFF
--- a/src/algorithm/intersects.rs
+++ b/src/algorithm/intersects.rs
@@ -45,7 +45,7 @@ impl<T> Intersects<Point<T>> for Line<T>
                     t <= T::one()
             },
             (Some(t_x), Some(t_y)) => { // All other lines
-                t_x.abs_sub(t_y) <= T::epsilon()  &&
+                (t_x - t_y).abs() <= T::epsilon()  &&
                     T::zero() <= t_x &&
                     t_x <= T::one()
             }
@@ -442,6 +442,7 @@ mod test {
     #[test]
     fn point_intersects_line_test() {
         let p0 = Point::new(2., 4.);
+        let p1 = Point::new(0., 0.);
         // vertical line
         let line1 = Line::new(Point::new(2., 0.), Point::new(2., 5.));
         // point on line, but outside line segment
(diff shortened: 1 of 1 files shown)


src/algorithm/intersects.rs
```diff
@@ -442,6 +442,7 @@ mod test {
     #[test]
     fn point_intersects_line_test() {
         let p0 = Point::new(2., 4.);
+        let p1 = Point::new(0., 0.);
```
michaelkirk: No big deal, but rather than adding another point, it seems like the existing pattern of this test is to have one point, and then a line for every test case with a comment explaining why that particular line is interesting. 

As written, it's not clear why that point is an interesting test case outside of the context of this PR.

Maybe this would be a little clearer:

```
// Line below point
let line4 = Line::new(Point::new(0.,0.), Point::new(5., 0.)
assert!(!line4.intersects(&p0));

// Line above point
let line5 = Line::new(Point::new(0.,5.), Point::new(5., 5.)
assert!(!line5.intersects(&p0));
```

src/algorithm/intersects.rs
```diff
@@ -448,12 +448,28 @@ mod test {
...
         assert!(line3.intersects(&p0));
         assert!(p0.intersects(&line3));
+        assert!(!line4.intersects(&p0));
+        assert!(!p0.intersects(&line4));
+        assert!(!line5.intersects(&p0));
+        assert!(!p0.intersects(&line5));
+        assert!(!line6.intersects(&p0));
+        assert!(!p0.intersects(&line6));
+        assert!(!line7.intersects(&p0));
+        assert!(!p0.intersects(&line7));
```
mbattifarano: @michaelkirk: to follow up on your previous comment, the failure occurs when the point is above a line with positive slope or below a line with negative slope. So I included examples above and below lines with positive and negative slopes. On master, lines 465-466 (`line4`) and lines 471-472 (`line7`) will fail.

frewsxcv:
Wow, good catch! Thanks! 🙇‍♂️

frewsxcv:
bors: r+

Issue #119, opened 2017-06-04
Determine fate of `Geometry` enum


frewsxcv:
What is the point of this `enum`? Doesn't seem like it serves much of a purpose. What are qualifications for a `struct`/variant to be added to it?

urschrei:
The only practical reason I can think of is for FFI; having an enum of valid types to match against means you don't have to maintain them on the calling side.

rory:
Personally I like this enum. For example, I've used it in my [`wkb` crate](https://crates.io/crates/wkb). If you have some WKB bytes, and you want to parse that into a geometry, you don't really know what sort of geometry you'll get out. The dataformat itself defines what type the geometry is.

Michael-F-Bryan:
What about replacing the `Geometry` enum with a `Geometry` marker trait? 

So we could say that anything which implements `Geometry` must implement a set of commonly used algorithm traits. As a tentative list, I'd say that any "geometric object" should implement:

```rust
trait Geometry<F>: Distance<F> + BoundingBox<F> + Intersects<F> + MapCoords<F>
        + Rotate<F> + Translate<F> {}
```
- `Distance`
- `BoundingBox`
- `Intersects`
- `MapCoords`
- `Rotate`
- `Translate`

Would the ability to create trait objects for a geometric thing be useful? i.e. if we made `Geometry` a trait, would it be object safe? We could probably tweak a couple algorithms to mutate inplace v, then add a default method to return a mutated copy when `Self: Sized + Clone`... That's not exactly a small API change though, and would break a lot of existing code...

rory:
> What about replacing the Geometry enum with a Geometry marker trait?

One question: I wrote a [Well Known Binary reader/writer](https://github.com/rory/rust-wkb), and i have a method which takes a Geometry, and then write it/encodes it. The [function](https://github.com/rory/rust-wkb/blob/16eae8ed22b2f8e4dd8d3ea62c8683866f6a5210/src/lib.rs#L100) is a big switch statement. "If it's a point do X, if a linestring do Y..." etc. Would this be possible with your suggestion?

> That's not exactly a small API change though, and would break a lot of existing code...

The crate is 0.x.y which semver says *" Anything may change at any time. The public API should not be considered stable"* so no-one should rely on it.

frewsxcv:
[...]

PR #122, opened 2017-06-22
WIP line-line intersection

mbattifarano:
This PR continues the work towards intersection operators for basic types (#80).

I'm marking this as WIP since I'm not wholly satisfied by a few of the details and would appreciate some feedback.

## The return type of `Intersection`

Two line segments can share nothing, a single point, or a line segment between them. Here, I've chosen to implement intersection to return `Option<LineIntersection>` where `LineIntersection` is an enum over the set {`Point`, `Line`}. 

go.geo side-steps this by returning [`NewPoint(math.Inf(1), math.Inf(1))`](https://github.com/paulmach/go.geo/blob/7c91620e5d753c84d6149b42d39ba6556e1a5f2f/line.go#L166) when the two lines are co-linear. I don't like this approach because I think this concept is much more easily expressed by the type system than by what amounts to the `NaN` of points. That said, I'm not wholly satisfied with `Option<LineIntersection>`: it's cumbersome to write and read, and we'd need one for almost every type. 

I considered defining `LineIntersection` like `{Empty, Point(Point), Line(Line)}`, but that doesn't change much.

I also considered using the `Geometry` enum instead of the specialized `LineIntersection`, this is basically how [PostGIS handles intersection](http://postgis.net/docs/ST_Intersection.html), but I decided against that first because of #119, and second because it over-specifies the types that can actually be returned. 

ArcGIS solves this problem by having separate [intersect and overlap](https://gis.stackexchange.com/questions/130034/what-is-the-difference-between-intersect-overlap-in-arcgis-server) methods.

## Handling degenerate lines

A `Valid` trait? Validating constructors? Should intersection return a `Result` or panic instead of handling various degeneracies or invalidities? Should we have a separate `Degeneracy` trait that repairs degeneracies (e.g. converts a degenerate `Line` to `Point`)?

## Project structure

For now, I added the point ordering as module directly under `algorithm`, but as I add more helper functions to support the line sweep it might make more sense to gather all of that code (trait, impls, and support) in a subdirectory just for intersection.

### DIFF
--- a/src/algorithm/intersection.rs
+++ b/src/algorithm/intersection.rs
@@ -1,0 +1,157 @@
+use num_traits::{Float};
+use algorithm::intersects::Intersects;
+use types::{Point, Line};
+use std::fmt::Debug;
+
+pub trait Intersection<Rhs = Self> {
+    type OutputGeometry;
+
+    /// Constructs the intersection of two geometries
+    fn intersection(&self, rhs: &Rhs) -> Option<Self::OutputGeometry>;
+}
+
+#[derive(PartialEq, Debug)]
+pub enum LineIntersection<T>
+    where T: Float
+{
+    Point(Point<T>),
+    Line(Line<T>)
+}
+
+impl<T> Intersection<Line<T>> for Line<T>
+    where T: Float + Debug
+{
+    type OutputGeometry = LineIntersection<T>;
+    fn intersection(&self, line: &Line<T>) -> Option<LineIntersection<T>> {
+        match (degenerate(&self), degenerate(&line)) {
+            (true, true) => {
+                if self.start == line.start {
+                    Some(LineIntersection::Point(self.start))
+                } else {
+                    None
+                }
+            }
+            (true, false) => {
+                if self.start.intersects(&line) {
+                    Some(LineIntersection::Point(self.start))
+                } else {
+                    None
+                }
+            }
+            (false, true) => {
+                if line.start.intersects(&self) {
+                    Some(LineIntersection::Point(line.start))
+                } else {
+                    None
+                }
+            }
+            (false, false) => {
+                // Using Cramer's Rule:
+                // https://en.wikipedia.org/wiki/Intersection_%28Euclidean_geometry%29#Two_line_segments
+                let (x1, y1, x2, y2) = (self.start.x(), self.start.y(),
+                                        self.end.x(), self.end.y());
+                let (x3, y3, x4, y4) = (line.start.x(), line.start.y(),
+                                        line.end.x(), line.end.y());
+                // self: (x(s), y(s)) = (x1 + s * a1, y1 + s * a2)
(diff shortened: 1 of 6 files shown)


urschrei:
I've opened #123 to hive off discussion about validity.

src/algorithm/intersection.rs
```diff
@@ -0,0 +1,187 @@
...
+use std::fmt::Debug;
+
+pub trait Intersection<Rhs = Self> {
+    type OutputGeometry;
+
+    /// Constructs the intersection of two geometries
+    fn intersection(&self, rhs: &Rhs) -> Option<Self::OutputGeometry>;
+}
+
+#[derive(PartialEq, Debug)]
```
frewsxcv: this can also derive `Copy`

src/types.rs
```diff
@@ -344,6 +344,15 @@ impl<T> Line<T>
     }
 }
 
+impl<T> PartialEq for Line<T>
+    where T: Float
+{
+    fn eq(&self, other: &Self) -> bool {
+        (self.start == other.start && self.end == other.end) ||
+            (self.start == other.end && self.end == other.start)
```
frewsxcv: hmmm. thinking out loud, i'm wondering if we should avoid `start` and `end` as terms for `Line`. 'start' and 'end' seem to imply that a `Line` is directed, even though it's not.

src/algorithm/mod.rs
```diff
@@ -4,6 +4,10 @@ pub mod centroid;
 pub mod contains;
 /// Checks if the geometry A intersects the geometry B.
 pub mod intersects;
+/// Construct the intersections of geometries.
+pub mod intersection;
+/// Implements total ordering on points
+pub mod point_ord;
```
frewsxcv: where does `point_ord` get used in the `intersection` `impl`? i'm a little confused how both of these modules relate to each other

src/types.rs
mbattifarano: I agree. `a` and `b`?

src/algorithm/mod.rs
mbattifarano: It currently is used [here](https://github.com/georust/rust-geo/pull/122/files#diff-aff11658469bb534220beba341bdef9bR72) to make computing the intersection of two parallel, but potentially overlapping line segments easier. The basic idea is that if you sort the endpoints while keeping track of the line segment to which each endpoint belongs, you can figure out the intersection by looking at their order. For example if the sorted endpoints look like `[l1a, l2b, l1b, l2a]` then you know (provided none of them are equal) that there is a line segment overlap from l2b to l2a.

mbattifarano: It will also appear again in the more general intersection algorithm which requires the endpoints to be stored in a binary search tree.

frewsxcv:
opened a tracking issue for this https://github.com/georust/rust-geo/issues/288

PR #124, opened 2017-06-27
Added destination trait.

mkulke:
The code is inspired from the turf.js equivalent fn in https://github.com/Turfjs/turf/tree/master/packages/turf-destination

It basically shoots a ray on the sphere using a distance and an angle from an existing point to create a new point.

This is useful when creating pseudo shapes like circles (split circle in _n_ angle angle slices, create _n_ new Points via `destination` from a circle center Point, then combine the new Points to a Polygon.

### DIFF
--- a/src/algorithm/destination.rs
+++ b/src/algorithm/destination.rs
@@ -1,0 +1,62 @@
+use num_traits::{Float, FromPrimitive};
+use types::Point;
+
+/// Returns a new Point using the distance to the existing Point and a bearing for the direction
+
+pub trait Destination<T: Float> {
+    /// Returns a new Point using distance to the existing Point and a bearing for the direction
+    ///
+    /// ```
+    /// # extern crate geo;
+    /// # #[macro_use] extern crate approx;
+    /// #
+    /// use geo::Point;
+    /// use geo::algorithm::destination::Destination;
+    ///
+    /// # fn main() {
+    /// let p_1 = Point::<f64>::new(9.177789688110352, 48.776781529534965);
+    /// let p_2 = p_1.destination(45., 10000.);
+    /// assert_eq!(p_2, Point::<f64>::new(9.274410083250379, 48.84033282787534))
+    /// # }
+    /// ```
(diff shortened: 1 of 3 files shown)


urschrei:
I'm uneasy about the generality implied by the name of the trait as is. At the very least, the description should make it clear that this is distance projected using the WGS84 spheroid (if I understand correctly).

mkulke:
@urschrei: yup true, what about `HaversineDestination`?

src/algorithm/destination.rs
```diff
@@ -0,0 +1,62 @@
...
+
+/// Returns a new Point using the distance to the existing Point and a bearing for the direction
+
+pub trait Destination<T: Float> {
+    /// Returns a new Point using distance to the existing Point and a bearing for the direction
+    ///
+    /// ```
+    /// # extern crate geo;
+    /// # #[macro_use] extern crate approx;
+    /// #
```
frewsxcv: you shouldn't need these three lines for this doc example. `extern crate geo` is already implicitly added and you aren't using the 'approx' provided `assert_relative_eq!` macro

src/algorithm/destination.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// use geo::Point;
+    /// use geo::algorithm::destination::Destination;
+    ///
+    /// # fn main() {
```
frewsxcv: once you remove the three lines above, you can remove this `fn main() {` and closing `}`

src/algorithm/destination.rs
```diff
@@ -0,0 +1,62 @@
...
+                bearing_rad.sin() * rad.sin() * center_lat.cos()
+            }
+            .atan2(rad.cos() - center_lat.sin() * lat.sin()) + center_lng;
+
+        Point::new(lng.to_degrees(), lat.to_degrees())
+    }
+}
+
+#[cfg(test)]
+mod destination {
```
frewsxcv: not a big deal, but usually tests live in a module named `test`

mkulke:
@frewsxcv thx for your remarks, I think i covered those.

PR #125, opened 2017-07-10
add auto derive for serde

etrombly:
I'm using rust-geo to parse a shapefile into country borders. I do this in the build.rs then include_bytes the resulting struct into the binary. Thought other people might be able to use a similar tactic. Example at https://github.com/etrombly/country_parser/blob/master/src/build.rs


frewsxcv:
Seems alright to me! This could live behind a feature flag (maybe enabled by default) but not a big deal

PR #128, opened 2017-07-27
Add translation trait

urschrei:


src/algorithm/translate.rs
```diff
@@ -0,0 +1,196 @@
...
+// translate a slice of points by the given offsets
+fn translation_matrix<T>(x_offset: T, y_offset: T, points: &[Point<T>]) -> Vec<Point<T>>
+    where T: Float
+{
+    points
+        .iter()
+        .map(|point| {
+                 let new_x = point.x() + x_offset;
+                 let new_y = point.y() + y_offset;
+                 Point::new(new_x, new_y)
```
frewsxcv: wonder if we should add a `Points::map` method to map coordinates to new values

rory: I am working on something like this which would do that.

urschrei: I'm not sure I follow, but if @rory does, I'll just wait on the PR!
IMO `rotate` / `translate` / `skew` / `scale` should all be implemented on top of a more general `affine` impl that populates an affine transformation matrix with the relevant values depending on the transformation type, but that would require us to pull in a linear algebra crate…

frewsxcv: > I'm not sure I follow, but if @rory does, I'll just wait on the PR!

for example:

```rust
fn translation_matrix<T>(x_offset: T, y_offset: T, points: &[Point<T>]) -> Vec<Point<T>>
    where T: Float
{
    points
        .iter()
        .map(|point| point.map(|x, y| (x + x_offset, y + y_offset))
        .collect::<Vec<_>>()
}
```

rory: My WIP is here: https://github.com/rory/rust-geo/tree/coord_change. Specifically you can see the file here https://github.com/rory/rust-geo/commit/b7a01f5a79eca2de7b581aa192b023778b122787 . My idea is to allow you to pass in a function which will take a `Coordinate` and return a new `Coordinate`, which gives you maximum flexibility. I don't like the name (`coord_change`), but can't think of much better.

frewsxcv: Maybe `map_coords`? That's what python-geojson calls it https://pypi.python.org/pypi/geojson#map-coords

PR #129, opened 2017-07-29
Add some Into's

rory:
In this patch, I've added some Into<_> for things, to make it easy to convert from one type to another where appropriate, such as convert a `Point` into a `Geometry` (a `Geometry::Point(p)` obviously), likewise 

This should make the types easier to use.


frewsxcv:
Thanks!

bors r+

Issue #123, opened 2017-06-23
Dealing with Validity


urschrei:
See previous discussions in #56 and #122 
First, what do we mean when we talk about validity? The OGC OpenGIS spec [discussed in the PostGIS docs](https://postgis.net/docs/using_postgis_dbmanagement.html#OGC_Validity) gives a good overview.

- It's probably not desirable to have validating constructors by default, as validation can incur a significant performance overhead, but they should probably be _available_, either as separate constructors, or with a `validate=True` argument, although the latter could cause a lot of headaches in terms of revising calls to the existing constructors, and dealing with invalid geometries; e.g. both of the simplification algorithms can return invalid geometries, so those traits would always have to return a `Result` (they should arguably already be doing this)
- Should we check for validity before carrying out certain operations? This is what Shapely does, raising an error if invalid geometries are detected. I'm not sure what JTS / GEOS do.
- I think the two points above directly imply that every geometry operation must return a `Result`.

frewsxcv:
Regardless of _when_ we check validity of geometry, we'll need to implement the validity checks before that happens. Unless anyone else has a better suggestion, I think we can use the same validity checks [described by OGC](https://postgis.net/docs/using_postgis_dbmanagement.html#OGC_Validity), mentioned in the original post above. Opening a new issue for this: https://github.com/georust/rust-geo/issues/127

Michael-F-Bryan:
I like the concept of "making illegal states unrepresentable". That way you can do things knowing that a particular assumption you make (e.g. all `Polygon`s are closed) will always be valid. 

These checks can often introduce `O(n)` or `O(n^2)` slow downs though, so what about adding validity checks which are only enabled in debug builds so all issues are found during development without hurting the performance of production code?

For example, you might implement the `Polygon::new()` function like this:

```rust
impl<F: Float> Polygon<F> {
  fn new(exterior: LineString<F>, interior: Vec<LineString<F>>) -> Polygon<F> {
    let poly = Polygon { exterior, interior };
    debug_assert!(poly.is_valid(), "All rings and lines in a polygon must be closed");
    poly
  }
  
  fn is_valid(&self) -> bool {
    exterior.is_closed() && interior.iter().all(|ring| ring.is_closed())
  }
}
```

Another option would be to add `new_checked()` constructors which create a new shape and then call `is_valid()` to make sure it's valid. 

---

[...]

Issue #32, opened 2016-05-30
Spatial Reference System of geometries


jdroenner:
Spatial data is always represented in a spatial / coordinate reference system. While most data, e.g. from GPS sensors, is represented in WGS84 (LatLon) this is not always the case.

Currently geometries in rust-geo have no information about a coordinate reference system. This allows to compare a LatLon- `Point ` with an UTM- `Point`.

In [JTS ](http://www.vividsolutions.com/jts/JTSHome.htm) and [GEOS ](https://trac.osgeo.org/geos) each `Geometry` has a CRS-id field. Algorithms have to check if the CRS of two geometries are the same. This could also be done for geometries in rust-geo.

Maybe it would also be possible to use the advanced type-system of Rust to add a generic CRS-type to geometries....

tmcw:
My 2c is that this should be low on the list of priorities. While projected data is an intense focal point of traditional GIS, its modern relevance is iffy. Here's the author of PostGIS [saying that projected data is mainly a case of legacy technology and expectations](http://gis.stackexchange.com/a/16252/895). Libraries like [Shapely](https://pypi.python.org/pypi/Shapely) and [Turf](http://turfjs.org/) successfully avoid the question of projections, doing calculations in WGS84 only and expecting users to project on the way in & out if they must. And for that, we have the very robust proj4 library, which [one fellow has already started porting into rust](https://github.com/JamesFysh/rust-proj4).

Relative to other priorities at this early stage, my personal instinct is that dealing with projected data shouldn't be one of them. We _should_ make a distinction between x/y points and lat/lon points, but after that distinction is made, a distance algorithm should _only_ deal with lat/lon points.

frewsxcv:
I agree with @tmcw. At least for now, I think it makes sense to leave projections out of rust-geo for now. If anyone is particularly interested in projections though, feel free to create a new repo+crate under the GeoRust organization and experiment.

frewsxcv:
> And for that, we have the very robust proj4 library, which one fellow has already started porting into rust.

BTW https://github.com/georust/rust-proj

jdroenner:
Even if the most data created today (locations / GPS) is in WGS84 there are a lot of datasets with different coordinate systems. So a geometry should have a type or field representing its CRS. My experience is that no matter what you do there is always one dataset with an esoteric coordinate system ;)

[...]

PR #131, opened 2017-08-04
Added {From,To}Iterator for appropriate types. Also some doccomments

rory:
I added a few more `Into<X>` and `FromIterator` and `IntoIterator`, also some comments.

The doctest comments show how to use some of the new converts which should make construction of `geo` structs much easier and ergnomic.

e.g.:

    let line: LineString<f32> = vec![(0., 0.), (10., 0.)].into();

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -59,6 +59,12 @@ pub struct Point<T> (pub Coordinate<T>) where T: Float;
 
 impl<T: Float> From<Coordinate<T>> for Point<T> { fn from(x: Coordinate<T>) -> Point<T> { Point(x) } }
 
+impl<T: Float> From<(T, T)> for Point<T> {
+    fn from(coords: (T, T)) -> Point<T> {
+        Point::new(coords.0, coords.1)
+    }
+}
+
 impl<T> Point<T>
     where T: Float + ToPrimitive
 {
@@ -433,6 +439,11 @@ mod test {
         assert_eq!(c, c2);
         assert_eq!(c.x, c2.x);
         assert_eq!(c.y, c2.y);
+
+        let p: Point<f32> = (0f32, 1f32).into();
+        assert_eq!(p.x(), 0.);
+        assert_eq!(p.y(), 1.);
+
     }
 
     #[test]
--- a/src/types.rs
+++ b/src/types.rs
@@ -3,6 +3,8 @@ use std::ops::AddAssign;
 use std::ops::Neg;
 use std::ops::Sub;
 
+use std::iter::{Iterator, FromIterator};
+
 use num_traits::{Float, ToPrimitive};
 
 pub static COORD_PRECISION: f32 = 1e-1; // 0.1m
@@ -63,7 +65,11 @@ pub struct ExtremePoint<T>
 #[derive(PartialEq, Clone, Copy, Debug, Serialize, Deserialize)]
 pub struct Point<T> (pub Coordinate<T>) where T: Float;
 
-impl<T: Float> From<Coordinate<T>> for Point<T> { fn from(x: Coordinate<T>) -> Point<T> { Point(x) } }
+impl<T: Float> From<Coordinate<T>> for Point<T> {
+    fn from(x: Coordinate<T>) -> Point<T> {
+        Point(x)
+    }
+}
 
 impl<T: Float> From<(T, T)> for Point<T> {
     fn from(coords: (T, T)) -> Point<T> {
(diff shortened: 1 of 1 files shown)


src/types.rs
```diff
@@ -15,6 +17,12 @@ pub struct Coordinate<T>
     pub y: T,
 }
 
+impl<T: Float> From<(T, T)> for Coordinate<T> {
```
frewsxcv: I'm _very_ hesitant about adding this implementation. Projects are not consistent about `(x, y)` vs. `(y, x)`: https://macwright.org/lonlat/

rory: I was just trying to mimic the `Point::new` constructor, which is (x, y). I'm not 100% wedded to any one way.

Perhaps I am unsure about a fundamental point of this crate. I thought it represented geometric objects in any coordinate reference system, and hence the X/Y order makes sense, since another CRS/SRS would not be lat/lon. However if everything is aimed to be lat/lon, that changes things.

I am willing to split this part off (perhaps into another PR).

src/types.rs
```diff
@@ -348,14 +418,99 @@ impl<T> Line<T>
...
+/// You can iterate over the points in the `LineString`
+///
+/// ```
+/// use geo::{LineString, Point};
+/// let line = LineString(vec![Point::new(0., 0.), Point::new(10., 0.)]);
+/// for point in line {
+///     println!("Point x = {}, y = {}", point.x(), point.y());
+/// }
+/// ```
+///
```
urschrei: This won't fail if you try to convert a `Vec` containing a single point, but the resulting LineString will be invalid. This is currently the case using the manual method, but there are open questions around how constructors should deal with it: #123 #58

src/types.rs
urschrei: Related: #32
I'm not sure where I stand on this. On the one hand, I'm personally in favour of coming down firmly on one side of the consensus, clearly stating that it's `x, y`, and letting libraries that want to use georust deal with that (which is what the ecosystem in general has to do now). On the other hand, I realise that's just my personal preference.

frewsxcv: After sleeping on this, I think you make a good point (heh) about `Point::new`. As long as we have that constructor looking like `Point::new(3, 7)` which implicitly is `(x, y)`, it doesn't seem much worse to make a `From` impl for `(T, T)`. For the time being, I think this new `impl` is fine and maybe in the future we should state in the module level documentation that all coordinate pairs in rust-geo will be `x, y`.

rory: Remember, rust-geo is still 0.x.y, so according to semver you can change things however you want. :)

src/types.rs
rory: With these PR I just wanted to mimic the standard constructors, so that it would easier to create those objects. Hence I didn't do anything about validity. (But the project would have decide *what* to do w.r.t. validity anyway ;) )

urschrei: Yes, true, and I don't think the lack of a check here is a problem at all. I'm just waiting for some intrepid soul to write the Polygon validity algorithm(s), since we can't do much checking without them anyway.

frewsxcv: All good points. `From` makes sense for now, but whenever we cross the bridge into validation-land, it should become `TryFrom`.

src/types.rs
```diff
@@ -54,10 +62,30 @@ pub struct ExtremePoint<T>
     pub xmin: Point<T>,
 }
 
+/// A single Point in 2D space.
+///
+/// Points can be created using the `new(x, y)` constructor, or from a `Coordinate` or pair of points.
+///
+/// ```
```
frewsxcv: Examples sections [should have an `# Examples` heading](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#examples-in-api-docs) before the code blocks, though it's not a big deal such that we need to hold up this PR

src/types.rs
rory: w.r.t. validity, I think it's better to allow people to create invalid geometries and then have some sort of `is_valid` check, so that geometry creation is fast, and validity checks is opt-in. This is sorta like how PostGIS does it I think. (I wonder if you could implemented `From` (w/o validity check), and `TryFrom` which does do valitory check 🤔🤔 )

urschrei: A `From` operation definitionally "cannot fail", so this all depends on whether we consider the creation of invalid geometries to be a failure, in which case we'll have to split conversions between `From` and `TryFrom`.

src/types.rs
rory: w.r.t. `# Examples`, most of the existing doc comments don't have it as a separate heading, so for consistency reasons I didn't put it in here. :)

Issue #127, opened 2017-07-26
Implement validity checks for each geometry type


frewsxcv:
See discussion in https://github.com/georust/rust-geo/issues/123

Validity checks are described here: https://postgis.net/docs/using_postgis_dbmanagement.html#OGC_Validity

frewsxcv:
Relevant: https://github.com/georust/rust-geo/issues/56

rory:
Something I'm working on appears to require valid geometries, so I'm interested in this. I'd be half tempted to implement what I want as a starting point sometime.

frewsxcv:
i have a WIP PR for ensuring valid linestrings: https://github.com/georust/rust-geo/pull/155/files?w=1 if you have any feedback, let me know

rory:
Looks like it solves one case, but there are many ways for a linestring to be invalid, and I hit an odd one (a linestring of 2 identical points!).

However that [link from PostGIS](https://postgis.net/docs/using_postgis_dbmanagement.html#OGC_Validity) splits things 2 ways. Simple, and Valid. In it, there are no invalid linestrings (so long as there's >=2 points). So maybe It's worth adding simplicity check? Or call do a simple & valid check but call that "validity" (my preferred option)?

frewsxcv:
for what it's worth, i'm in favor of making validity opt-out. if i have a `LineString` instance, i'd rather not have to check everytime i work with it if it's valid or not. if we can guarantee validity with low overhead (see my branch), i don't see the downside (besides being a little less ergonomic)

yeah, i knew about the 'no duplicate points' simplicity rule for linestrings, but wasn't sure the best way to address it. logically, it's a pretty straightforward check, but i'm not sure how best to implement it since it could affect performance. for example, if we kept a `Vec` for `LineString`, concatenating two large `LineString`s could be pretty expensive since we'd have to check the intersection of two `Vec`s. another option is to use a `BTreeSet` (an ordered set) which would allow us to use set operations, but i haven't thought too hard about it.

rory:
I also haven't thought about it too much. I think validity should be opt-out too. Sometimes you don't care, or more importantly, you can assume (or know) that the thing is valid, in which case there shouldn't be a performance hit.

urschrei:
I agree about optional validity. I really like the idea of validity checks being easily chainable, so you could e.g. carry out a series of "cheap" geo operations, only checking validity before a potentially expensive one, and bailing out with an Err if need be.
[...]

PR #132, opened 2017-08-05
Add winding order for linestrings

rory:
For some data formats (e.g. ESRI Shapefiles), rings in a polygon are stored in a specific winding order (clockwise for exterior ring, anti-clockwise for interior rings IIRC). So I added support to detect this, to make a linestring clockwise/counterclockwise, and to iterate over the points in that order.

The code is already there for the Area calculation. I split the shoelace formula code into it's own file, since it felt weird importing something from the `area` module into `types`. But I'm open to suggestions on this, because I'm not 100% sure it's the best approach.

### DIFF
--- a/src/algorithm/shoelace_formula.rs
+++ b/src/algorithm/shoelace_formula.rs
@@ -1,0 +1,14 @@
+use num_traits::Float;
+use types::LineString;
+
+pub fn twice_area<T>(linestring: &LineString<T>) -> T where T: Float {
+    if linestring.0.is_empty() || linestring.0.len() == 1 {
+        return T::zero();
+    }
+    let mut tmp = T::zero();
+    for ps in linestring.0.windows(2) {
+        tmp = tmp + (ps[0].x() * ps[1].y() - ps[1].x() * ps[0].y());
+    }
+
+    tmp
+}
--- a/src/types.rs
+++ b/src/types.rs
@@ -5,6 +5,8 @@ use std::ops::Sub;
 
 use num_traits::{Float, ToPrimitive};
 
+use algorithm::shoelace_formula::twice_area;
+
 pub static COORD_PRECISION: f32 = 1e-1; // 0.1m
 
 #[derive(PartialEq, Clone, Copy, Debug, Serialize, Deserialize)]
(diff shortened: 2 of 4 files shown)


src/types.rs
```diff
@@ -348,9 +350,86 @@ impl<T> Line<T>
...
+
+    /// Returns the winding order of this line
+    pub fn winding_order(&self) -> WindingOrder {
+        let shoelace = twice_area(self);
+        if shoelace < T::zero() {
+            WindingOrder::Clockwise
+        } else if shoelace > T::zero() {
+            WindingOrder::CounterClockwise
+        } else if shoelace == T::zero() {
+            // what should be done here?
```
mbattifarano: It looks like `shoelace == 0` when the line is degenerate (empty, single point, multiple coincident points) which brings us back to the validity conversation ( #123 #127 ). 

It will also be zero for linestrings co-incident with the line `y = x` or either of the axes -- (and perhaps some other cases?).

I think changing the return type to `Option<WindingOrder>` would be a reasonable choice here.

src/types.rs
```diff
@@ -348,9 +350,86 @@ impl<T> Line<T>
...
+
 #[derive(PartialEq, Clone, Debug, Serialize, Deserialize)]
 pub struct LineString<T>(pub Vec<Point<T>>) where T: Float;
 
+impl<T: Float> LineString<T> {
+
+    /// Returns the winding order of this line
+    pub fn winding_order(&self) -> WindingOrder {
+        let shoelace = twice_area(self);
+        if shoelace < T::zero() {
```
mbattifarano: imo I think a `match` here instead would be a little easier to read.

src/algorithm/shoelace_formula.rs
```diff
@@ -0,0 +1,14 @@
+use num_traits::Float;
+use types::LineString;
+
+pub fn twice_area<T>(linestring: &LineString<T>) -> T where T: Float {
+    if linestring.0.is_empty() || linestring.0.len() == 1 {
+        return T::zero();
+    }
+    let mut tmp = T::zero();
+    for ps in linestring.0.windows(2) {
+        tmp = tmp + (ps[0].x() * ps[1].y() - ps[1].x() * ps[0].y());
```
mbattifarano: Does it make sense to extract the cross product into a trait?

src/types.rs
```diff
@@ -445,5 +524,64 @@ mod test {
...
+        assert_eq!(new_line1, cw_line);
+        assert_ne!(new_line1, ccw_line);
+
+        // test make_counterclockwise_winding
+        let mut new_line2 = cw_line.clone();
+        new_line2.make_counterclockwise_winding();
+        assert_eq!(new_line2.winding_order(), WindingOrder::CounterClockwise);
+        assert_ne!(new_line2, cw_line);
+        assert_eq!(new_line2, ccw_line);
+
```
mbattifarano: I think it would be worth testing a few cases where we should expect a panic.

src/algorithm/shoelace_formula.rs
urschrei: @mbattifarano we're also already computing it in [convexhull.rs](https://github.com/georust/rust-geo/blob/master/src/algorithm/convexhull.rs#L51) although in a slightly different way, and it's used in a couple of places in the [Polygon Distance PR](https://github.com/georust/rust-geo/pull/115/files#diff-af7ca56bf0fcfc5aced02a9a325eecc6R713), so it's probably worth thinking about how to unify them.

urschrei:
This seems relevant to the [Orient](https://github.com/georust/rust-geo/blob/master/src/algorithm/orient.rs) trait, although I'm not sure how best to combine them…

src/algorithm/shoelace_formula.rs
mbattifarano: @urschrei I agree. It also crops up in #122. I've opened up an issue to that effect https://github.com/georust/rust-geo/issues/133

src/types.rs
rory: I was considering making `WindingOrder` have 3 enum values: `Clockwise`, `CounterClockwise" and `Open` (as in 'this linestring is open, so there is no winding order'), since in theory this method can be called on every linestring, including linestrings which are not a closed ring.

However I didn't know about the other failure cases. So maybe a `None` would be better then.

rory:
> This seems relevant to the Orient trait, although I'm not sure how best to combine them…

🤦 Gah! I didn't know about that!  I saw the file and just thought it was something to do with rotations. That'll teach me for redoing work.

Yes, there shouldn't be both, and hence it makes sense to "merge" them somehow.

urschrei:
I'm not tied to `Orient` (I stole the name from [Shapely](http://toblerity.org/shapely/manual.html#shapely.geometry.polygon.orient)'s method of the same name), its functionality could easily be re-implemented on top of this PR, and if we iterate on this one, we only have to worry about the edge cases @mbattifarano pointed out in https://github.com/georust/rust-geo/pull/132#discussion_r131529586 – which I also failed to consider – in one place.

rory:
I've redid this as #134, so I'm closing this.

PR #134, opened 2017-08-06
Winding/Orientation for LineStrings

rory:
After feedback on #132, I've redid this, using the `Orient` trait.

The `Orient` trait is very useful for encoding (Multi)Polygons, since you can ensure the data is correctly ordered, but it's not helpful for reading bytes. In that case, you need a way to look at a linestring and see if it's clockwise or counter-clockwise, to see if it's an interior or exterior ring. That's what I've added in this PR. Since this functionality only makes sense for LineStrings, I implemented it as methods on the LineString struct, rather than create a trait.

Taking advice from #132, I've made `.winding_order()` return `Option<WindingOrder>`.

I also added some shorthand methods to make it easier to call (e.g. `.orient_default()`)

### DIFF
--- a/src/algorithm/orient.rs
+++ b/src/algorithm/orient.rs
@@ -29,6 +29,20 @@ pub trait Orient<T> {
     /// assert_eq!(oriented.interiors[0].0, oriented_int_ls.0);
     /// ```
     fn orient(&self, orientation: Direction) -> Self;
+
+    /// Orient a Polygon according to default orientation
+    fn orient_default(&self) -> Self
+        where Self: Sized
+    {
+        self.orient(Direction::Default)
+    }
+
+    /// Orient a Polygon, but reversed from the default
+    fn orient_reversed(&self) -> Self
+        where Self: Sized
+    {
+        self.orient(Direction::Reversed)
+    }
 }
 
 impl<T> Orient<T> for Polygon<T>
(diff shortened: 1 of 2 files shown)


rory:
Any feedback on this?

src/algorithm/orient.rs
```diff
@@ -29,6 +29,20 @@ pub trait Orient<T> {
     /// assert_eq!(oriented.interiors[0].0, oriented_int_ls.0);
     /// ```
     fn orient(&self, orientation: Direction) -> Self;
+
+    /// Orient a Polygon according to default orientation
+    fn orient_default(&self) -> Self
```
frewsxcv: I personally think it's a little redundant having these new methods in addition to `orient`, but if people find it helpful then I'm cool with it

src/algorithm/orient.rs
```diff
@@ -61,8 +75,14 @@ pub enum Direction {
     Reversed,
 }
 
+#[derive(PartialEq, Copy, Clone, Debug)]
```
frewsxcv: This could also derive `Eq`

src/algorithm/orient.rs
urschrei: Yeah, I think it should be one or the other – I'm not tied to using an enum, though.

rory:
OK, those are good points. I'll reimplent this feature this way shortly and update this pull request. Stay tuned!

rory:
continued in #169

PR #135, opened 2017-08-07
Implement Simplification traits for more things

rory:
If you can simplify a `LineString`, then you can simplify a `MultiLineString`.

I also added implementations for the simplification traits for`Polygon` (and `MultiPolygon`), which just perform the simplification on the constituant linestrings that make up polygon. PostGIS's `ST_Simplfiy` does this AFAIK. The link for VM algorithm shows a polygon (the USA) being simplified, so the concept of "simplifying a polygon" is well understood.

### DIFF
--- a/src/algorithm/simplify.rs
+++ b/src/algorithm/simplify.rs
@@ -47,8 +47,16 @@ fn rdp<T>(points: &[Point<T>], epsilon: &T) -> Vec<Point<T>>
     }
 }
 
+/// Simplifies a geometry.
+///
+/// The [Ramer–Douglas–Peucker
+/// algorithm](https://en.wikipedia.org/wiki/Ramer–Douglas–Peucker_algorithm) simplifes a
+/// linestring. Polygons are simplified by running the RDP algorithm on all their constituent
+/// rings. This may result in invalid Polygons, and has no guarantee of preserving topology.
+///
+/// Multi* objects are simplified by simplifing all their constituent geometries individually.
 pub trait Simplify<T, Epsilon = T> {
-    /// Returns the simplified representation of a LineString, using the [Ramer–Douglas–Peucker](https://en.wikipedia.org/wiki/Ramer–Douglas–Peucker_algorithm) algorithm
+    /// Returns the simplified representation of a geometry, using the [Ramer–Douglas–Peucker](https://en.wikipedia.org/wiki/Ramer–Douglas–Peucker_algorithm) algorithm
     ///
     /// ```
     /// use geo::{Point, LineString};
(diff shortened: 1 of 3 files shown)


rory:
I was tempted to implement the simplification for `Point` (& `MultiPoint`), which would do nothing and just return a clone of the point. PostGIS's `ST_Simplify` does this. This allows you to implement simplication traits on `Geometry`, which can make it easier to mass simplify a lot of geometry, because you don't have to do any logic yourself to split out point/multipoints. 

But I didn't do it this time, I'm in two minds about whether it sensible.

urschrei:
We should note that in the case of Polygons and MultiPolygons, the simplification operation is not guaranteed to return a valid geometry (and we're back to validity!). [A well-performing fix](https://www.jasondavies.com/simplify/) is dependent upon a usable R* implementation (I have a look occasionally, but results have been mixed).

urschrei:
(I've had a look at the JTS topology-preserving implementation – I don't do much Java, but it looks to be the naïve version described above. I'm not sure about GEOS, but I assume it's the same)

rory:
I have expanded the comments w.r.t. validity and topology

src/algorithm/simplify.rs
```diff
@@ -52,7 +52,7 @@ fn rdp<T>(points: &[Point<T>], epsilon: &T) -> Vec<Point<T>>
 /// The [Ramer–Douglas–Peucker
 /// algorithm](https://en.wikipedia.org/wiki/Ramer–Douglas–Peucker_algorithm) simplifes a
 /// linestring. Polygons are simplified by running the RDP algorithm on all their constituant
-/// rings.
+/// rings. This may result in invalid Polygons, and has no guarantee of perserving topology.
```
urschrei: Should be "constituent" and "preserving".

src/algorithm/simplifyvw.rs
```diff
@@ -146,8 +146,9 @@ fn area<T>(p1: &Point<T>, p2: &Point<T>, p3: &Point<T>) -> T
 
 /// Simplifies a geometry.
 ///
-/// Polygons are simplified by running the algorithm on all their constituant rings. Multi* objects
-/// are simplified by simplifing all their constituant geometries individually.
+/// Polygons are simplified by running the algorithm on all their constituant rings.  This may
+/// result in invalid Polygons, and has no guarantee of perserving topology.  Multi* objects are
+/// simplified by simplifing all their constituant geometries individually.
```
urschrei: Should be "simplifying" and "constituent".

rory:
I have rebased & fixed up those typo fixes now.

urschrei:
Thanks for making the changes. PR looks good to me. 👍

frewsxcv:
this is great, thanks! opened a new issue for reminding us about this potentially creating invalid geometries (https://github.com/georust/rust-geo/issues/142)

bors r+

PR #136, opened 2017-08-09
add .map_coords() function to convert all coordinates in a shape

rory:
I added the ability to apply a function to all coordinates. This is a general function which will change all the coordinates in a geometry. You can even change the type of the coordinates, e.g. if you want to change from `f32` to `f64`. This gives ultimate flexibility.

Some of the other traits (e.g. Translation) could probably be rewritten to take advantage of this change, but my first simple attempt failed.

I'm not 100% happy with the syntax, `p.map_coords(&|&(x, y)| ...` has too many `&` and stuff, so I'm open to suggestions on if this can be improved.

### DIFF
--- a/src/algorithm/map_coords.rs
+++ b/src/algorithm/map_coords.rs
@@ -1,0 +1,251 @@
+use num_traits::Float;
+use types::{Point, Polygon, LineString, Line, MultiPoint, MultiPolygon, MultiLineString, GeometryCollection, Geometry};
+
+/// Map all the coordinates in an object, returning a new one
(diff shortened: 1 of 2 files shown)


frewsxcv:
In the future, it might be nice if these methods returned structures that implement `Iterator` instead of having some of the implementations call `collect`. The syntax is a little clunky with the &'s but I can't think of anything better right now and now point on waiting. Thanks for doing this!

bors r+

frewsxcv:
Will cut a new release later when I'm back at my computer

PR #139, opened 2017-08-10
Fix Floating-point instability in distance test

### LINKED ISSUE #138: Possible bug in distance between point and polygon
Hi everybody, I'm starting to use this library, and while experimenting I found a possible bug.

I tried this both on the pinned version on crate.io and the master branch here.

Basically in some cases asking for the distance between a polygon and a point that lies on the polygon's exterior ring will not return `0.` 

Here is a basic `main.rs` to reproduce the problem. 
```rust
extern crate geo;

use geo::{Point, Polygon, LineString};
use geo::distance::Distance;


fn main() {
    let exterior = LineString(vec![
                              Point::new(0., 0.),
                              Point::new(0., 0.0004),
                              Point::new(0.0004, 0.0004),
                              Point::new(0.0004, 0.),
                              Point::new(0., 0.)]);

    let poly = Polygon::new(exterior.clone(), vec![]);

    let good_point_1 = Point::new(0.0001, 0.0001);
    let good_point_2 = Point::new(0.0002, 0.);
    let bugged_point = Point::new(0.0001, 0.);

    assert_eq!(poly.distance(&good_point_1), 0.);
    assert_eq!(poly.distance(&good_point_2), 0.);
    // This fails!
    assert_eq!(poly.distance(&bugged_point), 0.);
}


```

As you see the point `(0.0001, 0.)` should be inside the polygon (which goes from `x=0` to `x=0.0004`), but the distance result is instead `0.00000000000000000006776263578034403`

Could you confirm this is not the expected behaviour?

urschrei:
This is a more robust way to calculate distance from a point to a
line-segment. It's a direct port from GEOS:
https://github.com/OSGeo/geos/blob/master/src/algorithm/CGAlgorithms.cpp#L191

Note the small difference between the test distances in  `point_linestring_distance_test`, which match Shapely (GEOS) for the same polygon and point.

Old: 1.1313708498984758
New 1.1313708498984762

Fixes #138

### DIFF
--- a/src/algorithm/distance.rs
+++ b/src/algorithm/distance.rs
@@ -67,15 +67,7 @@ pub trait Distance<T, Rhs = Self> {
 
 // Return minimum distance between a Point and a Line segment
 // This is a helper for Point-to-LineString and Point-to-Polygon distance
-// adapted from http://stackoverflow.com/a/1501725/416626. Quoting the author:
-//
-// The projection of point p onto a line is the point on the line closest to p.
-// (and a perpendicular to the line at the projection will pass through p).
-// The number t is how far along the line segment from start to end that the projection falls:
-// If t is 0, the projection falls right on start; if it's 1, it falls on end; if it's 0.5,
-// then it's halfway between. If t is less than 0 or greater than 1, it
-// falls on the line past one end or the other of the segment. In that case the
-// distance to the segment will be the distance to the nearer end
+// adapted from https://github.com/OSGeo/geos/blob/master/src/algorithm/CGAlgorithms.cpp#L191
 fn line_segment_distance<T>(point: &Point<T>, start: &Point<T>, end: &Point<T>) -> T
 where
     T: Float + ToPrimitive,
--- a/src/algorithm/distance.rs
+++ b/src/algorithm/distance.rs
@@ -3,7 +3,6 @@ use std::collections::BinaryHeap;
 use num_traits::{Float, ToPrimitive};
 use types::{Point, Line, MultiPoint, LineString, MultiLineString, Polygon, MultiPolygon};
 use algorithm::contains::Contains;
-use num_traits::pow::pow;
 
 /// Returns the distance between two geometries.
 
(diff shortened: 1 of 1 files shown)


urschrei:
PR also includes a test with the values that caused the bug found in #138

src/algorithm/distance.rs
```diff
@@ -61,38 +60,32 @@ pub trait Distance<T, Rhs = Self> {
...
-    let t = T::zero().max(T::one().min((*point - *start).dot(&(*end - *start)) / dist_squared));
-    let projected = Point::new(start.x() + t * (end.x() - start.x()),
-                               start.y() + t * (end.y() - start.y()));
-    point.distance(&projected)
+    if start == end {
+        return point.distance(start);
+    }
+    let r = ((point.x() - start.x()) * (end.x() - start.x()) + (point.y() - start.y()) * (end.y() - start.y())) /
+        ((end.x() - start.x()) * (end.x() - start.x()) + (end.y() - start.y()) * (end.y() - start.y()));
+    if r <= T::zero() {
```
mbattifarano: I think this would be easier to read as a `match` statement on `r.partial_cmp(T::zero())`, but that's just my personal preference so feel free to ignore.

src/algorithm/distance.rs
```diff
@@ -61,38 +60,32 @@ pub trait Distance<T, Rhs = Self> {
...
+        return point.distance(start);
+    }
+    if r >= T::one() {
+        return point.distance(end);
+    }
+    let s = ((start.y() - point.y()) * (end.x() - start.x()) - (start.x() - point.x()) * (end.y() - start.y())) /
```
mbattifarano: defining `dx = end.x() - start.x()` and `dy = end.y() - start.y()` (or similar) would make these three lines easier to read IMO and I don't think would change the computation.

urschrei: Good idea. Done!

src/algorithm/distance.rs
urschrei: Not sure I follow. Can you sketch it out?

mbattifarano: Sorry nevermind. I misread L83 my first time through. This is good the way it is.

umurgdk:
@urschrei do you think this PR could help with possible intersection (polygon to polygon) bug?

urschrei:
frewsxcv r+

frewsxcv:
thanks for fixing this @urschrei!

urschrei:
np! (I'm the one who caused it!)

PR #143, opened 2017-08-17
Topology-preserving VW

urschrei:
This is a first attempt at topology-preserving VW, based on https://www.jasondavies.com/simplify/.
**Update**: see subsequent comments for a refinement of the approach.
<s>It's WIP for two reasons:
- I need a longer test input which produces self-intersections for a given epsilon. This is surprisingly difficult to cook up.</s>
- I'm not sure the approach in this fork is correct:
    1. If an intersection is found, move the triangle (`a`) into a vec, and exit the loop early
    2. pop the next triangle (`b`) and start processing it
    3. `peek_mut()` at the _next_ triangle (`c`) in the queue, and set its area to `a`s area. This guarantees its removal
   4. push `a` back onto the heap. If we modified the correct `c` triangle, `a` will be subsequently skipped in the main loop, as one of its support points was removed
   5. if it wasn't the triangle we needed to remove, `a` will be re-detected as causing an intersection.

<s>I've got an alternate implementation [here](https://github.com/urschrei/rust-geo/blob/vw_topology_preserve/src/algorithm/simplifyvw.rs#L168). The difference:
1. If an intersection is found, move the triangle (`a`) into a vec, and exit the loop early
2. pop the next triangle (`b`) and start processing it
3. `peek` at the _next_ triangle (`c`) in the queue, **and increase the epsilon to its area. This guarantees its removal**
4. push `a` back onto the heap. If the `c` triangle we remove was the correct one, `a` will be subsequently skipped in the main loop, as one of its support points was removed
5. If it wasn't, `a` will be re-detected as causing an intersection, and the epsilon will be increased again.

The drawback of the second approach is that it permanently increases the user-defined epsilon, leading to the removal of more points overall than the first approach.
Also, both of these approaches could be wrong. More complex test inputs will _really_ help here.</s>


urschrei:
And now for the bad news:
`cargo bench` for the naïve version (reduction of `norway_main`) runs in around 6.5 ms
For the topo-preserving version, it's around 370 ms

urschrei:
I need to dig into this more, but it doesn't seem to be thrashing anywhere obvious; the naïve version runs the main loop 12.9k times, whereas the new version runs it 13.5k times (Only a 4.5% increase). This is somewhat complicated by not being able to run perf (bc macOS).

mbattifarano:
Cool! I can take a look at this over the weekend

urschrei:
I hooked up the same test to Instruments. 427 ms total with `--release`. Of which:
271 ms (**64%**) inserting into the R* tree
117 ms (**27.5%**) removing stale segments from the R* tree
Oof. Insertion is apparently `O(log(n))` on average, but I'm struggling to reconcile that with the numbers. The JS implementation (using RBush) is beating this by 200ms.

urschrei:
OK, this is beginning to feel somewhat more coherent. I've refactored the intersection-removal logic, which has made it a lot simpler:  
Whereas before, we were relying on a heuristic of "we need to remove a previous point in order to affect this point's geometry, so we modify the next entry in the heap and hope for the best. If it's the wrong point, the intersection persists until we've removed the right one". This worked OK in practice, but it was imprecise, and resulted in the removal of more points than necessary.
The new approach is conceptually the same, but more precise: the VScore struct has a new boolean field initialised to `false`. If a point's removal causes a self-intersection, the boolean is flipped, and the loop continues until it reaches the adjacent-triangle re-calculation logic.
At that point, the smaller of the adjacent point _indices_ is identified, and its associated VScore area is set to negative epsilon before being pushed onto the heap, ensuring it's popped off again as soon as the loop resets.
This is producing a non-intersecting Norway polygon with the same number of vertices as the self-intersecting version (I'm not sure why yet, I was expecting n-1 vertices).

urschrei:
Progress:
- Everything is faster. The  8k -> 3.2k simplification is running in around 300 ms for me. That's still slow, but it's a pretty large real-world example. For `n` in the 100's-of-points range, performance is perfectly acceptable IMO.
- I spent a few days banging my head against a solution to a common problem: what to do if an additional point needs to be removed to prevent self-intersection, but removal wouldn't leave enough points to form a valid geometry (see illustration below). This is now handled by bailing out of the simplification if removal would cause an intersection, but we only have `n + 2` points left, where `n` is the minimum number of points needed for a given geometry (2 for a `LineString`, 4 for a `Polygon`). The rationale is that if removal of the second point _still_ didn't remove the intersection, we'd be out of points to remove, and be left with an invalid geometry. This strikes me as a pretty good compromise.

screenshot 2017-08-23 13 46 17
screenshot 2017-08-23 13 47 22

Removing the point denoted by the dot causes a self-intersection. Ordinarily, that could be fixed by removing the previous point too, but that's impossible because that wouldn't leave enough points to form a ring, so we keep the exterior as is, and end the simplification. This doesn't affect simplification of the inner ring(s), and there's a test for that.

urschrei:
OK! I think this death march is over for now. I'm pretty happy with the tests (although ideally I'd like some more weird adversarial examples) and the docs. Could some people have a look at it / try it out / try to break it / comment on the code? @frewsxcv @pka @mbattifarano @jdroenner?

I'm still not delighted about the Spade R* tree performance – someone has suggested using SQLite's R* functionality, but AFAIK it's not included in the standard SQLite distribution, and SQLite isn't available everywhere anyway, and I have no experience using it for spatial queries (Spatialite uses it, of course, so it certainly works). Other points in Spade's favour are the fact that we need an R or R* tree for a Polygon distance variant, so we're not pulling it in as a dependency just for this.

src/algorithm/simplifyvw.rs
```diff
@@ -173,36 +437,165 @@ pub trait SimplifyVW<T, Epsilon = T> {
...
+            initial_min: 4,
+            min_points: 6,
+            geomtype: GeomType::Ring,
+        };
+        let mut simplified = vwp_wrapper(&gt, &self.exterior.0, Some(&self.interiors), epsilon);
+        let exterior = LineString(simplified.remove(0));
+        let interiors = match simplified.is_empty() {
+            true => vec![],
+            false => simplified.into_iter().map(LineString).collect(),
+        };
```
frewsxcv: seems fine to just do `let interiors = simplified.into_iter().map(LineString).collect();` here. an empty iterator will result in an empty `Vec`

src/algorithm/simplifyvw.rs
```diff
@@ -110,45 +156,263 @@ fn visvalingam<T>(orig: &[Point<T>], epsilon: &T) -> Vec<Point<T>>
...
+        orig[triangle.current],
+        orig[triangle.right],
+    ]).bbox()
+        .unwrap();
+    let br = Point::new(bbox.xmin, bbox.ymin);
+    let tl = Point::new(bbox.xmax, bbox.ymax);
+    let candidates = tree.lookup_in_rectangle(&BoundingRect::from_corners(&br, &tl));
+    candidates
+        .iter()
+        .map(|c| {
```
frewsxcv: you can get rid of the `map` here i think

```
    .iter()
    .any(|c| {
        ...
    })
```

src/algorithm/simplifyvw.rs
```diff
@@ -2,37 +2,81 @@ use std::cmp::Ordering;
...
+where
+    T: Float,
 {
     fn partial_cmp(&self, other: &VScore<T>) -> Option<Ordering> {
         Some(self.cmp(other))
     }
 }
 
-impl<T> Eq for VScore<T> where T: Float {}
+impl<T> Eq for VScore<T>
```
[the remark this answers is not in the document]
urschrei: The compiler complains about this for some reason, which is (IIRC) why I had to implement the blank `Eq`.

frewsxcv:
I just read through this again one more time, and everything seems good enough to land. I'm still not very knowledgeable about the algorithm (yet), but I don't think it makes sense to hold up merging this. If I learn more about it in the future, I can always revisit the code.

I noticed the title of the PR has 'WIP'. Is this ready to go or do you still have more changes?

urschrei:
I don't think there's anything left for me to do – edited. I'm happy to label this as "experimental" in the docs for a while, maybe? I'm just going to squash my commits, and then it's ready to go, apart from the possible doc change.

frewsxcv:
> I'm happy to label this as "experimental" in the docs for a while, maybe?

up to you if you want to do this, but i don't think it's necessary. i don't see any reason to consider this much more experimental than other algorithms in rust-geo.

thanks for doing this! your comments you added to this PR along the way were very helpful!

bors r+

PR #148, opened 2017-09-01
Translation bounds

### LINKED ISSUE #147: Seemingly unnecessary bounds on `Translate`
Is the `FromPrimitive` bound still necessary for the `Translate` trait's blanket impl? It doesn't look like `MapCoords` requires `FromPrimitive`, and I can't see anything which explicitly requires it.

The code in question:

```rust
impl<T, G> Translate<T> for G
    where T: Float + FromPrimitive,
        G: MapCoords<T, T, Output=G>
{
    fn translate(&self, xoff: T, yoff: T) -> Self {
        self.map_coords(&|&(x, y)| (x + xoff, y + yoff))
    }
}
```

This isn't really an issue in practice because all types implementing `num_traits::Float` are also `FromPrimitive`, but it feels a bit odd to add the extra constraint.

---

Also, I don't think the `&` in front of the closure is necessary. `MapCoords::map_coords()` should be generic over anything which is `Fn(&(T, T)) -> (NT, NT)` so you can skip the extra syntax and dynamic dispatch, although fixing that now would be a breaking change...

Michael-F-Bryan:
As we discussed in #147 I've removed the unnecessary `FromPrimitive` bound on the `Translate` trait. 

I also made `MapCoords::map_coords()` generic over any `Fn(&(T, T)) -> (NT, NT)` so it uses static dispatch instead of dynamic dispatch and you don't need to write funky `&|&(x, y)| { ... }` borrowed closures any more. 

(fixes #147)

### DIFF
--- a/src/algorithm/translate.rs
+++ b/src/algorithm/translate.rs
@@ -1,4 +1,4 @@
-use num_traits::{Float, FromPrimitive};
+use num_traits::Float;
 use algorithm::map_coords::MapCoords;
 
 pub trait Translate<T> {
@@ -26,7 +26,7 @@ pub trait Translate<T> {
 }
 
 impl<T, G> Translate<T> for G
-    where T: Float + FromPrimitive,
+    where T: Float,
         G: MapCoords<T, T, Output=G>
 {
     fn translate(&self, xoff: T, yoff: T) -> Self {


Michael-F-Bryan:
I wasn't able to remove all `&`'s. Apparently the `iterator.map()` function can't capture a closure, so on a couple impls I had to pass in a pointer to the closure instead. I'm assuming the compiler/LLVM will be able to dereference these and inline them anyway, so hopefully it turns into a no-op for optimised builds.

Michael-F-Bryan:
Looks like it's not possible to remove the dynamic dispatch in `MapCoords::map_coords()` because of the way implementations iterate over their components and recursively call `map_coords()`. This means during the monomorphising stage you end up with an infinitely nested type because the closure is defined in terms of itself.

```
error[E0275]: overflow evaluating the requirement `[closure@src/algorithm/map_coords.rs:245:34: 245:59]: std::ops::Fn<(&(f64, f64),)>`
  |
  = help: consider adding a `#![recursion_limit="128"]` attribute to your crate
  = note: required because of the requirements on the impl of `std::ops::Fn<(&(f64, f64),)>` for `&[closure@src/algorithm/map_coords.rs:245:34: 245:59]`
  = note: required because of the requirements on the impl of `std::ops::Fn<(&(f64, f64),)>` for `&&[closure@src/algorithm/map_coords.rs:245:34: 245:59]`
...
  = note: required because of the requirements on the impl of `std::ops::Fn<(&(f64, f64),)>` for `&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&[closure@src/algorithm/map_coords.rs:245:34: 245:59]`
```

I'm going to roll back to 9b4a16b so you can merge that. I'll see if I can figure out a clean way of implementing a `MapCoords::map_coords()` with static dispatch on a different branch.

rory:
Getting rid of all the `&`s would be cool. I tried that, but also ran into weird compiler problems. I was out of my depth, so I just backed off and did the thing that works. If you can figure out a solution that would be cool. 👍

On 1 September 2017 05:07:30 GMT+02:00, Michael Bryan <notifications@github.com> wrote:
>As we discussed in #147 I've removed the unnecessary `FromPrimitive`
>bound on the `Translate` trait. 
>
>I also made `MapCoords::map_coords()` generic over any `Fn(&(T, T)) ->
>(NT, NT)` so it uses static dispatch instead of dynamic dispatch and
>you don't need to write funky `&|&(x, y)| { ... }` borrowed closures
>any more. 
>
>(fixes #147)
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/georust/rust-geo/pull/148
>
>-- Commit Summary --
>
>  * Removed the unnecessary trait bound on Translate
>  * Made `MapCoords::map_coords()` generic over any Fn()
>
>-- File Changes --
>
>    M src/algorithm/map_coords.rs (47)
>    M src/algorithm/translate.rs (6)
>
>-- Patch Links --
>
>https://github.com/georust/rust-geo/pull/148.patch
>https://github.com/georust/rust-geo/pull/148.diff
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/georust/rust-geo/pull/148

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #151, opened 2017-09-03
Add generic impls for Rotate, RotatePoint. [WIP]

frewsxcv:
Depends on this conversation: https://github.com/georust/rust-geo/pull/107#discussion_r136720014

PR #153, opened 2017-09-04
Add generic trait impl for RotatePoint.

frewsxcv:
Originally added in https://github.com/georust/rust-geo/pull/151, but
extracted out due to indecision about how to move forward with Rotate.


urschrei:
I can't think of a good way to unify this, short of splitting the Centroid trait into `CentroidWeighted` and `CentroidUnweighted` or similar. I don't think it would cause confusion, but it's certainly less simple than the current approach.

frewsxcv:
@urschrei You're talking about `Rotate` (not `RotatePoint`) right? This PR doesn't change the centroid or rotation behavior

urschrei:
@frewsxcv Yep, sorry, should probably just have left it as a comment on #151.

PR #154, opened 2017-09-04
Add associated return type for Centroid algorithm trait.

frewsxcv:


frewsxcv:
In the future, we can remove the `Option` from the polygon and linestring implementations once we implement validation (I'm working on linestring validation right now in a branch).

src/algorithm/centroid.rs
```diff
@@ -6,6 +6,8 @@ use algorithm::distance::Distance;
 
 /// Calculation of the centroid.
 pub trait Centroid<T: Float> {
+    type Centroid;
```
frewsxcv: i originally titled this `Result`, but that's confusing since there's `std::result::Result` which is conceptually different

frewsxcv: actually, maybe `Output` would be a better name here. i'm doing the same thing with bounding box right now and it's confusing that the trait is `BoundingBox`, the associated return type name is `BoundingBox` and the actual return type is `Bbox`.

PR #158, opened 2017-09-06
Correctly check for LineString containment in Polygon

### LINKED ISSUE #157: Polygon-LineString containment check seems to be wrong
Given an approximately circular Polygon, and an approximately circular concentric LineString: 

circle

The [containment check](https://github.com/georust/rust-geo/blob/master/src/algorithm/contains.rs#L214-L225) should return `true` - it doesn't.

[Polygon](https://github.com/urschrei/rust-geo/blob/polygon_distance/src/algorithm/test_fixtures/shell.rs)
[LineString](https://github.com/urschrei/rust-geo/blob/polygon_distance/src/algorithm/test_fixtures/poly_in_ring.rs)

urschrei:
The DE-9IM specification for [`containment`](http://docs.safe.com/fme/html/FME_Desktop_Documentation/FME_Transformers/Transformers/spatialrelations.htm#DE9IM_Matrix) states:

> The **interiors intersect** and no part of the candidate's interior
or boundary intersects the base's exterior. It is possible for the boundaries to intersect.

For `intersection`

> The two features are not disjoint

For `disjoint`

> The boundaries and interiors do not intersect.

The current containment check deviates from this, in that it forbids
intersection (`!self.intersects(linestring)`), when in fact it should
be forbidding intersection with any _interior rings_. This PR fixes
that, and adds a test.

closes #157

### DIFF
--- a/src/algorithm/contains.rs
+++ b/src/algorithm/contains.rs
@@ -268,6 +268,8 @@ mod test {
         let v = Polygon::new(vec![(150., 350.), (100., 350.), (210., 160.), (290., 350.), (250., 350.), (200., 250.), (150., 350.)].into(), vec![]);
         let rect = Polygon::new(vec![(185., 237.), (220., 237.), (220., 220.), (185., 220.), (185., 237.)].into(), vec![]);
         assert_eq!(v.contains(&rect), true);
+    }
+    #[test]
     // LineString is fully contained
     fn linestring_fully_contained_in_polygon() {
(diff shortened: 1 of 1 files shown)


frewsxcv:
lgtm! r=me once the merge conflict is addressed

urschrei:
r=frewsxcv

frewsxcv:
bors r=frewsxcv

PR #162, opened 2017-09-14
Added a prelude

### LINKED ISSUE #146: Prelude
A common Rust pattern is to expose some top-level `prelude` module which re-exports all the commonly-used traits in a module. 

Seeing as a lot of the functionality for various operations is separated into their various submodules in `geo::algorithms`, and you have to manually import each trait, would it be possible to expose a `geo::prelude` module which people can glob import (e.g. `use geo::prelude::*`)?

Michael-F-Bryan:
Closes #146

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,3 +15,27 @@ pub mod algorithm;
 #[cfg(test)]
 #[macro_use]
 extern crate approx;
+
+/// A prelude which re-exports the traits for manipulating objects in this
+/// crate. Typically imported with `use geo::prelude::*`.
+pub mod prelude {
+    pub use algorithm::area::Area;
+    pub use algorithm::boundingbox::BoundingBox;
+    pub use algorithm::centroid::Centroid;
+    pub use algorithm::contains::Contains;
+    pub use algorithm::convexhull::ConvexHull;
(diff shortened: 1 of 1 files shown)


Michael-F-Bryan:
For the next major version, would it be a good idea to "flatten" rust-geo's public API? The way I see it, you only want access to the datatypes and the traits used to manipulate them. Therefore you could alter things so that the only things exported are the `prelude` module and `types::*`. 

We'd still leave the algorithms module as it is on disk because the way it is currently structured allows nice separation between components. This would just rearrange how things are viewed by the user.

I'm imagining a top level `lib.rs` which looks like this:

```rust
mod traits;
mod types;
mod algorithm;

// re-exports to provide a more flattened API
pub use types::*;
pub mod prelude {
  ...
}
```

---

What are people's thoughts? This would technically be a breaking change (therefore you'd need to bump the major version) because people will have code which does `use geo::algorithm::distance::Distance`, but I feel like restructuring the public API like this will make it much flatter and simpler to deal with.

frewsxcv:
bors r+

thanks! opened a new issue for flattening the API: https://github.com/georust/rust-geo/issues/163

PR #164, opened 2017-09-19
Add `CoordsIter` trait for iterating over coordinates in geometries.

frewsxcv:
A new trait to iterate over all coordinates in a geometry. In the future, we should add a new method to this trait called `external_coords_iter` which iterates over the external coordinates of a geometry, excluding coordinates of interior holes. Then we can make algorithms like `ConvexHull` have a generic implementation where `G: CoordsIter`


frewsxcv:
working on refactoring some of the rust-geo traits right now. this is similar to `MapCoords` but unlike `MapCoords` the methods on this trait don't result in any new memory allocations.

rory:
I have a `map_coords_inplace` branch lying around at home that I should make a PR for. It allows you to change things without memory allocations But maybe this would suffice.

frewsxcv:
> It allows you to change things without memory allocations But maybe this would suffice.

yep, the main reason i opened this is so that i can modify `Coordinate` values without allocations:

https://github.com/georust/rust-geo/compare/frewsxcv-rotate-refactgor?expand=1#diff-64481126a8d2c515ad92dc33cc71a856R18

rory:
:) I also wanted to modify objects without allocations.

I should clean up my code and start opening PRs

frewsxcv:
@rory don't want to pressure you into reviewing, but did you happen to already look at the code here? if so, any thoughts?

rory:
I only skimmed it earlier. I might be able to look at it properly within the next few days

src/algorithm/coords_iter.rs
```diff
@@ -0,0 +1,89 @@
+use num_traits::Float;
+use std::iter;
+use types::{Coordinate, Point, Polygon, LineString, Line, MultiPoint, MultiPolygon,
+            MultiLineString};
+
+pub trait CoordsIter<T: Float> {
+    fn coords_iter<'a>(&'a self) -> Box<Iterator<Item = &'a Coordinate<T>> + 'a>;
```
frewsxcv: just like the other boxed iterators that get returned in this crate, we can remove the `Box` once return `impl`s stabilize

frewsxcv:
just finished this! i'll try to find some time in the next few days to add tests. most of the time i spent on this was trying to remove `Box<Iterator>`, which is why we have to specify the specific types of the iterator for each implementation

frewsxcv:
also will add docs to the trait. just realized they're missing

michaelkirk (approved):
nice!

frewsxcv:
bors r=michaelkirk

PR #165, opened 2017-09-19
Use folds instead of min-heaps for Polygon distances

urschrei:
This results in fewer allocations, fewer structs we need to worry about, and cuts Point-Polygon-distance time by ~30%


urschrei:
@frewsxcv: bors r+

frewsxcv:
🎉

bors r+

PR #167, opened 2017-09-21
Closest point

### LINKED ISSUE #161: Closest Point
Would it be possible to add a closest Points algorithm? I'm guessing that it's be a trait which returns either a single point (intersection), a pair of points (closest Points on object A and B), or some sort of "infinite points" marker for when two lines are exactly parallel.

Michael-F-Bryan:
The initial implementation of a `ClosestPoint` algorithm. The trait itself looks something like this:

```rust
pub trait ClosestPoint<F: Float, Rhs = Point<F>> {
    fn closest_point(&self, other: &Rhs) -> Closest<F>;
}

pub enum Closest<F: Float> {
    Intersection(Point<F>),
    SinglePoint(Point<F>),
    Indeterminate,
}
```

If you can think of a better name for `Closest` please let me know (names are hard!).

(fixes #161)

### DIFF
--- a/src/algorithm/closest_point.rs
+++ b/src/algorithm/closest_point.rs
@@ -1,0 +1,12 @@
+use num_traits::Float;
+use Point;
+
+pub enum Closest<F: Float> {
+    SinglePoint,
+    Intersect(Point<F>),
+    Indeterminate,
+}
+
+pub trait ClosestPoint<F: Float, Other = Point<F>> {
+    fn closest_point(&self, other: &Other) -> Closest<F>;
+}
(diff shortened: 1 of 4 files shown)


Michael-F-Bryan:
I've finished off most of the `ClosestPoint` impls where the right hand side is a `Point`. Luckily I was able to find [a generic way][1] to do things early on, so that eliminated loads of boilerplate.

[1]: https://github.com/Michael-F-Bryan/rust-geo/blob/6faf09021b642c6415b0d96ec712f3e5770eaf8e/src/algorithm/closest_point.rs#L89-L139

frewsxcv:
looks great!

bors r+

PR #168, opened 2017-09-23
Add validation constructors for LineString, and some other methods.

frewsxcv:
Add new constructor functions for `LineString`, and a few other misc.
functions:

* `LineString`
  * `LineString::new`
  * `LineString::new_unchecked`
  * `LineString::points`
  * `LineString::points_mut`
  * `LineString::is_closed`
* `LineStringError`
  * `LineStringError::into_vec`

This is backwards compatible such that it doesn't un-`pub`licize the
`LineString` internals (yet). The intention is for this to land, migrate
everything to the new functions, then close off access to the internals.

Partially extracted from https://github.com/georust/rust-geo/pull/155.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -1,3 +1,5 @@
+use std::error;
+use std::fmt;
 use std::ops::Add;
 use std::ops::AddAssign;
 use std::ops::Neg;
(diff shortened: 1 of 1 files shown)


src/types.rs
```diff
@@ -488,6 +490,146 @@ impl<T> Line<T>
...
+    /// let points = vec![
+    ///     Point::new(1., 5.),
+    ///     Point::new(2., -10.),
+    ///     Point::new(-4., 5.),
+    /// ];
+    /// let line_string = LineString::new(points).unwrap();
+    /// assert!(!line_string.is_closed());
+    /// ```
+    pub fn is_closed(&self) -> bool {
+        assert!(self.0.len() > 1);
```
frewsxcv: this assertion can be removed once we remove the `pub` in `LineString<T>(pub Vec<Point<T>>)`

src/types.rs
```diff
@@ -488,6 +490,146 @@ impl<T> Line<T>
...
+    /// Attempting to create a `LineString` with fewer than two `Point`s
+    /// results in an error:
+    ///
+    /// ```
+    /// use geo::{LineString, Point};
+    ///
+    /// let points = vec![Point::new(6., 14.)];
+    /// assert!(LineString::new(points).is_err());
+    /// ```
+    pub fn new(points: Vec<Point<T>>) -> Result<Self, LineStringError<T>> {
```
Michael-F-Bryan: I feel like this might make the library painful to use. Returning a `Result` means you now need to add loads of `unwrap()` statements to your code seeing as 99% of the time you already *know* that the line you created was valid.

What about an `assert!()` so you get a runtime panic instead? I'd argue creating an invalid `LineString` is a programming error, so it should blow up loudly during development time. In production code it should be impossible to create an invalid line, and if you do then chances are your program is FUBAR anyway...

src/types.rs
```diff
@@ -488,6 +490,146 @@ impl<T> Line<T>
...
+    /// Creating a `LineString`:
+    ///
+    /// ```
+    /// use geo::{LineString, Point};
+    ///
+    /// let points = vec![Point::new(6., 14.), Point::new(2., -10.)];
+    /// let line_string = unsafe { LineString::new_unchecked(points) };
+    /// ```
+    pub unsafe fn new_unchecked(points: Vec<Point<T>>) -> Self {
+        debug_assert!(points.len() > 1);
```
Michael-F-Bryan: Hmm... I'm not sure if this `debug_assert!()` would be valid here. If a developer is already using `unsafe` then that's a promise by him/her that the thing they're passing in is valid and doesn't need to be checked. 

Also, doesn't naming the function `new_unchecked()` and then checking it seem like a bit of an oxymoron? :stuck_out_tongue_winking_eye:

src/types.rs
Michael-F-Bryan: It might be useful to keep the function around and use it for validation in the constructors, but we can probably remove the `pub` from `pub fn is_closed()`. Users shouldn't need to worry about checking if a line is closed seeing as it's an explicit assumption upheld by the constructor.

src/types.rs
frewsxcv: A small detail, but since this returns a `Result`, one should be able to do `LineString::new(points)?` in many contexts

src/types.rs
frewsxcv: Yeah, you're right. For comparison, the `unsafe` function [`CString::from_vec_unchecked`](https://github.com/rust-lang/rust/blob/23aaeb573b626a51af9ecc97680663153e4ab2b0/src/libstd/ffi/c_str.rs#L272-L276) could very easily do a `debug_assert` to make sure that the `Vec` that is passed in does _not_ have a trailing null byte, but it doesn't.

src/types.rs
frewsxcv: > Users shouldn't need to worry about checking if a line is closed seeing as it's an explicit assumption upheld by the constructor.

To clarify, this is an `is_closed` method on a _line string_ and not a _polygon_. I anticipate the polygon constructor in the future will ensure the interior data is a closed line string, but it could use this method.

frewsxcv:
tracking issue https://github.com/georust/rust-geo/issues/289

PR #169, opened 2017-09-23
Winding/Orientation for LineStrings

rory:
This is a rework of #134 with the suggested changes.

The `Orient` trait is very useful for encoding `(Multi)Polygons`, since you can ensure the data is correctly ordered, but it's not helpful for reading bytes. In that case, you need a way to look at a linestring and see if it's clockwise or counter-clockwise, to see if it's an interior or exterior ring. That's what I've added in this PR.

### DIFF
--- a/src/algorithm/area.rs
+++ b/src/algorithm/area.rs
@@ -1,6 +1,8 @@
 use num_traits::Float;
 use types::{Bbox, Line, LineString, MultiPolygon, Polygon};
 
+use algorithm::winding_order::twice_signed_ring_area;
+
 /// Calculation of the area.
 
 pub trait Area<T>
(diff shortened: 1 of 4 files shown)


rory:
I've rebased this branch on current master. Is there any feedback? Is this a good feature?

urschrei:
I don't think we should extract the shoelace formula into its own public module; can we change `Orient::signed_ring_area` to be `pub(crate)` instead? That would make it importable in other modules that need it (`area`)?

rory:
> I don't think we should extract the shoelace formula into its own public module

Changed (with a history rewrite)

> can we change Orient::signed_ring_area to be pub(crate) instead? That would make it importable in other modules that need it (area)?

I'm not sure what you mean here, there's (now) `algorithm::winding_order::twice_signed_ring_area`, not `Orient::signed_ring_area`, and I've changed it to `pub(crate)`

urschrei:
> I'm not sure what you mean here

Yeah, I misread that change – sorry about that. This looks good to me now!

rory:
Don't merge this now. I've done an explicit `panic!()` for a scenario that I thought was impossible, and I'm using this code more myself and hitting this scenario.

rory:
I have fixed this `panic` issue, and rebased, and pushed to the branch. Is this all OK and merge-able?

frewsxcv:
bors r=urschrei

Issue #171, opened 2017-09-26
Primitive operations on Points


Michael-F-Bryan:
I just noticed there isn't an impl for multiplying/dividing a `Point<F>` by `F`. Has this been left out on purpose?

My specific use case is creating a unit vector. My current code looks like this

```rust
let magnitude = other.dot(&other);
let unit_v = Point::new(other.x() / magnitude, other.y() / magnitude);
```

However it'd be nice if I could skip the `Point::new()` call and just do something like this:

```rust
let magnitude = other.dot(&other);
let unit_v = other / magnitude;
```

urschrei:
I've implemented the same functionality you describe in a couple of places, and used `Point` interchangeably as a vector, although I don't feel great about it. We should probably just have a vector type – I'm already getting confused when I look at function signatures that are doing vector ops.

Michael-F-Bryan:
That's a good point, although what's the difference between a `Point` and a `Coordinate`? My thoughts were that `Point` is actually a generic 2D vector, whereas `Coordinate` is a thing which has a `x` and `y` and denotes positions. 

That said, `Coordinate` is essentially useless at the moment...

urschrei:
>Point is actually a generic 2D vector

Hmm. I think this depends on the model, but I'd consider points to represent an absolute position in space, without any other dimensions. Which also describes `Coordinate`, I realise. The `Point` / `Coordinate` distinction precedes my existence in rust-geo, but I know this question's come up before.

frewsxcv:
> That said, `Coordinate` is essentially useless at the moment...

For the record, there was some discussion here: https://github.com/georust/rust-geo/issues/15

michaelkirk:
This was addressed as part of https://github.com/georust/geo/pull/505/files#diff-c7745eb5b54b1f8ebb4ed8f0b180cf8ee4fe4f2a771539abd51167fae3a194bdR383

PR #172, opened 2017-10-04
Update geometry descriptions

urschrei:
Add missing geometry and container struct descriptions, and regularise
existing descriptions to more closely match the SimpleGeo descriptions.

The description of `Coordinate` may be controversial, but I've kept it as general as possible (see discussions in #171 and #15).


frewsxcv:
looks great, thanks for improving the descriptions!

PR #178, opened 2018-01-07
Use new bulk-load method for initial R* Tree population

urschrei:
This gets us a big speedup, and a higher-quality tree for nearest-neighbour searches.


frewsxcv:
fancy! looks good to me 👌 

bors r+

PR #179, opened 2018-01-10
Clippy pass, rustfmt pass, typo fixes, enum addition

urschrei:
99% of these changes are from `rustfmt` – the rest are changes suggested by Clippy, which now only finds a single issue.
I've also tidied up some doc typos, and added `Line` to the `Geometry` enum.

### DIFF
--- a/src/algorithm/area.rs
+++ b/src/algorithm/area.rs
@@ -1,9 +1,11 @@
 use num_traits::Float;
-use types::{Line, LineString, Polygon, MultiPolygon, Bbox};
+use types::{Bbox, Line, LineString, MultiPolygon, Polygon};
 
 /// Calculation of the area.
 
-pub trait Area<T> where T: Float
+pub trait Area<T>
+where
+    T: Float,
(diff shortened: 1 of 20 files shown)


frewsxcv:
everything looks good here, nice work! the only thing to note here is that https://github.com/georust/rust-geo/pull/179/commits/816e2e8ef5a10d2d9fdcb11db995153bc61e5294 should be considered a breaking change (which is fine!) since someone could have been `match`ing on the `Geometry` `enum` prior to that commit and now it will error since they're missing a `Geometry::Line` branch. if that's cool, r=me

urschrei:
Ah, good catch. I think it'll have to be OK, since we're pre-1.0, and it was going have to go in at some point. We should probably cut a release in any case - lmk if you want me to write up the notes etc.

urschrei:
r=frewsxcv

frewsxcv:
bors r+

i don't think you can edit a comment and have bors pick it up

PR #180, opened 2018-01-21
Add PostGIS and GeoJSON integration/conversions

eeeeeta:
This PR adds conversions from PostGIS types to `geo` primitives, and conversions from `geo` primitives to GeoJSON and PostGIS types.

This integration is gated behind two new features, `postgis-integration` and `geojson-integration`.


urschrei:
The PostGIS part looks great, but GeoJSON already provides [fallible conversions for Geo items](https://github.com/georust/rust-geojson/blob/master/src/conversion.rs), however, the documentation was hidden for some reason when the original PR merged. [I unhid it](https://github.com/georust/rust-geojson/commit/a3fad98dd166d6364493f9fe6094f1a793527681) recently, but it won't hit docs.rs until we cut a new release, so nobody knows about it – sorry about that :/

eeeeeta:
@urschrei No problem - I got rid of the GeoJSON parts of this PR. Nice to know that a (better) implementation of it exists already! (Perhaps you *should* cut a new release? I don't know...)

Anyway, let me know if there's anything more you want me to do w.r.t. the PostGIS integration stuff.

urschrei:
>Perhaps you should cut a new release?

Yes, absolutely. I thought we had after I updated the docs last time, but it seems we didn't. Anyway, the PostGIS conversions lgtm.

frewsxcv:
this looks good to me! when i originally was thinking about this crate, i had envisioned 'geo' not having many dependencies and all of the associated crates (e.g. geojson, postgis, gpx) would have all of the conversion logic. but after seeing this pull request, it seems appropriate to have it live in this crate by an opt-in feature flag, good idea!

thanks for the contribution!

bors r+

PR #181, opened 2018-01-22
Implement SpatialObject for Line type

urschrei:
This doesn't get us any speedup, but it _does_ let us use the `Line` type directly in an `RTree`, which reduces complexity quite a bit.

### DIFF
--- a/src/algorithm/simplifyvw.rs
+++ b/src/algorithm/simplifyvw.rs
@@ -8,7 +8,7 @@ use spade::SpadeFloat;
 use spade::BoundingRect;
 use spade::rtree::RTree;
 
-// Store triangle information
+/// Store triangle information
 // current is the candidate point for removal
 #[derive(Debug)]
 struct VScore<T>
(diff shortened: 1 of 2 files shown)


floatdrop:
Not exactly question about this PR, but generally – is implementing `SpatialObject` for other types (like Polygon) is planned (or how do I implement them in my code)?

urschrei:
@floatdrop I haven't really thought about implementing it for any other types, because we haven't yet had to use `RTree` to store them. It should be easy to do for `LineString`, but I'm less sure whether it would be useful for `Polygon`. In terms of implementing it for your own types, you first need to Implement `PointN` and `TwoDimensional` for your point type (see [here](https://github.com/georust/rust-geo/blob/523ac0393744d5216ae525188678691abfb7c9a4/src/types.rs#L318-L351) for ours), and then implement `SpatialObject` as in this PR.

floatdrop:
@urschrei thanks for quick response! My question was more about – how do I implement `SpatialObject` for `Polygon`? Because these traits are not in scope of my library – rust refuses to accept implementation (or I just [do it](https://gist.github.com/anonymous/ecb8850bc00627aa36601a1c417f5466) wrong).

urschrei:
@floatdrop As you say, you won't be able to do it yourself, because `Polygon` is a foreign type. So you have two options: use a [wrapper type](https://stackoverflow.com/a/25415289/416626), or store the `Polygon`'s exterior as`Line`s (when this PR lands), and then use that for querying.

urschrei:
bors r=frewsxcv

PR #182, opened 2018-01-24
Fix Github language detection

urschrei:
Github is currently mis-classifying Rust-Geo as a RenderScript project because it can't recognise some of the test fixtures in use as Rust files. This override fixes that, so that we can be found / seen by users again.


frewsxcv:
oooh i haven't seen this before, looks good!

bors r+

PR #183, opened 2018-01-27
Add Haversine length algorithm

eeeeeta:
This PR adds a new algorithm, `haversine_length`, that computes the "Haversine length" of a line or linestring - i.e. the sum of the Haversine distances between every two points in the line.

It also reexports this algorithm, and the PostGIS algorithms introduced in #180, in the prelude (because I realised I forgot to reexport them in #180...)


frewsxcv:
looks good, thanks!

bors r+

PR #184, opened 2018-02-10
Easy methods to convert a Geometry to the underlying type

rory:
Added a few `.as_point()` etc. methods to "unwrap" `Geometry`s. It helps with ergonomics

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -833,6 +833,79 @@ impl<T: Float> From<MultiPolygon<T>> for Geometry<T> {
     }
 }
 
+impl<T: Float> Geometry<T> {
+
+    /// If this Geometry is a Point, then return that, else None.
+    ///
+    /// ```
+    /// use geo::*;
+    /// let g = Geometry::Point(Point::new(0., 0.));
+    /// let p2: Point<f32> = g.as_point().unwrap();
+    /// assert_eq!(p2, Point::new(0., 0.,));
+    /// ```
+    pub fn as_point(self) -> Option<Point<T>> {
+        if let Geometry::Point(x) = self {
+            Some(x)
+        } else {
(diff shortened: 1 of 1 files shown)


src/types.rs
```diff
@@ -833,6 +833,79 @@ impl<T: Float> From<MultiPolygon<T>> for Geometry<T> {
...
+    pub fn as_point(self) -> Option<Point<T>> {
+        if let Geometry::Point(x) = self {
+            Some(x)
+        } else {
+            None
+        }
+    }
+
+    /// If this Geometry is a LineString, then return that LineString, else None.
+    pub fn as_linestring(self) -> Option<LineString<T>> {
```
frewsxcv: i don't feel strongly about this, but should this be `as_line_string` since the struct is called a `LineString` (and not a `Linestring`). if we end up doing this, it would make `as_multi_line_string` a little verbose 😅

rory:
So if this is good, can it be merged? Should I do it?

frewsxcv:
@rory i think it's mergeable, though i didn't merge it earlier in case you had any thoughts about [this comment here](https://github.com/georust/rust-geo/pull/184#discussion_r167751901)

rory:
Ah yes, sorry. I'm not fussed about the wording change. Obviously I have mentally been using `linestring` as the lowercase of `LineString` (and `multipolygon` instead of `multi_polygon` etc.). But I don't feel strongly about it too much. If you'd rather, I can change it.

frewsxcv:
I don't feel strongly either, so I'm fine with it as is. Can you try approving this pull request with bors? I'm curious why the other one you approved didn't work

rory:
bors r=frewsxcv

Issue #133, opened 2017-08-05
Extract cross product operation


mbattifarano:
The cross product of two points is implemented inline in several places. We should implement a cross product method directly on `Point`.

See also: https://github.com/georust/rust-geo/pull/132#discussion_r131529789

frewsxcv:
here are the two implementations which could be put on `Point`:

* https://github.com/georust/rust-geo/blob/523ac0393744d5216ae525188678691abfb7c9a4/src/algorithm/extremes.rs#L70-L76
* https://github.com/georust/rust-geo/blob/523ac0393744d5216ae525188678691abfb7c9a4/src/algorithm/convexhull.rs#L38-L56

Jay9596:
I would like to work on this issue.
Should I change the occurrences or leave them be and change them after submitting the PR with the cross fn as part of Point.

frewsxcv:
@Jay9596 thanks for taking this on! you can go ahead and change the existing occurrences to use the new method

PR #189, opened 2018-03-11
cross_prod method added to Point

Jay9596:
Solves #133 

`Point` struct  now implements a cross product fuction `cross_prod` that use previously used inline in 2 different parts of the codebase.
- algorithm/extremes.rs
- algorithm/convexhull.rs

It uses the same implementation the other inline function were using.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -254,7 +254,12 @@ where
         self.x() * point.x() + self.y() * point.y()
     }
 
-    /// Returns the cross product of 3 points
+    /// Returns the cross product of 3 points. A positive value implies
+    /// `self` → `point_b` → `point_c` is counter-clockwise, negative implies
+    /// clockwise.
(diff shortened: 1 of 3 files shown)


src/types.rs
```diff
@@ -254,6 +254,25 @@ where
         self.x() * point.x() + self.y() * point.y()
     }
 
+    /// Returns the cross product of 3 points
```
frewsxcv: can you add another sentence here saying something like:

> A positive value implies `self` → `point_b` → `point_c` is counter-clockwise, negative implies clockwise.

frewsxcv:
thanks for the pull request @Jay9596! ✨ 

i added one small commit fbfbac2e4a047b4c083684219f141f1f2e32ec11 adjusting the formatting

bors r+

PR #190, opened 2018-03-11
[WIP] implement coordinate reprojection

urschrei:
This should implement coordinate reprojection on all Geometries which implement MapCoords. <s>However, I'm not at all sure I've got the type bounds correct (It _looks_ like `Reproject` shouldn't need any, but I think I need it for the impl on MapCoords).</s> The actual problem is that I'm getting a type mismatch error in the `project` function:

      --> src/algorithm/reproject.rs:22:62
       |
    22 |             let reprojected = source_proj.project(dest_proj, Point::new(x, y));
       |                                                              ^^^^^^^^^^^^^^^^ expected struct `geo::types::Point`, found struct `types::Point`
       |
       = note: expected type `geo::types::Point<_>`
                  found type `types::Point<T>`

And I have no idea why it's occurring, so if anyone has any ideas…

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ serde = "1.0"
 serde_derive = "1.0"
 spade = "1.3.0"
 postgis = { version = "0.5", optional = true }
+proj = { version = "0.3", optional = true }
 
 [features]
 default = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ proj = { version = "0.3", optional = true }
 [features]
 default = []
 postgis-integration = ["postgis"]
+reprojection = ["proj"]
 
 [dev-dependencies]
 approx = "0.1.1"
--- a/src/algorithm/mod.rs
+++ b/src/algorithm/mod.rs
@@ -46,6 +46,6 @@ pub mod to_postgis;
 pub mod haversine_length;
 /// Calculate, and work with, the winding order of Linestrings
 pub mod winding_order;
-/// Apply a projection transformation to a Geometry
+/// Apply a coordinate projection transformation to a Geometry using `rust-proj`
 #[cfg(feature = "reprojection")]
 pub mod reproject;
--- a/src/algorithm/mod.rs
+++ b/src/algorithm/mod.rs
@@ -46,3 +46,6 @@ pub mod to_postgis;
 pub mod haversine_length;
 /// Calculate, and work with, the winding order of Linestrings
 pub mod winding_order;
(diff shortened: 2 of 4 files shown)


rory:
I don't really know a lot of the theory behind spatial reference projections. But I have a gut feeling that this isn't totally right. You're doing a simple MapCoords on each coordinate to convert from one SRS to another. But does that work? If you have a simple linestring, with 2 points (a start & an end), this code will project that to a simple straight line in the other SRS. But that's not always the case, right? Sometimes a straight line in one coordinate system would be a curved line in another, right?

urschrei:
> Sometimes a straight line in one coordinate system would be a curved line in another, right?

I don't *think* this is a problem. Two lines with the same geodetic coordinates would appear curved _relative_ to each other when projected into different spatial reference systems (or if one were cartesian coordinates, and the other were in e.g. Robinson). AFAIK (I need to double-check this using GEOS) it's not a problem since the curvature would only be applied when representing the line visually; within the projected SRS, and for the purposes of operations, lines are straight.

urschrei:
The type mismatch error is probably due to [this](https://github.com/rust-lang/rust/issues/48953).

 Having projection / transformation functionality in a separate crate also seems redundant, since it depends on `Geo` anyway, so I think there's a case for just integrating it as an optional feature, then implementing the trait(s) on `MapCoords` (assuming the GEOS comparison works out as expected) as per this PR. Now that [`proj-sys`](https://github.com/georust/proj-sys) exists, people can do their own thing with non-`Geo` types if they'd prefer to stay outside the ecosystem in any case. There would have to be some housekeeping on the `rust-proj` crate to ensure that potential users can easily find it, too. @frewsxcv, what do you think?

urschrei:
(Closing in favour of a new PR)

rory:
FYI, since you seem to be aware of projections and stuff, my objection is addressed. :)

frewsxcv:
@urschrei i'm okay with your other PR that uses `proj-sys`, but it seems like it could be cleaner just to remove the usage of `geo::Point` from `proj`, add a `proj::Point`, and remove the `geo` dependency altogether, then use `proj` here. what do you think?

frewsxcv:
if you'd prefer, we can merge your other `proj-sys` pull request for now and open an issue for my idea in my previous comment

urschrei:
My thinking with the all-inclusive PR was that anyone wanting to reproject coordinates would probably also be using Geo, so why have separate crates. Also, a `Proj`-native `Point` would be missing the nice `from` things we have for primitive `Point`-like things here, although it's not a ton of work to implement those.

PR #191, opened 2018-03-14
Implement a MapCoordsFallible trait

urschrei:
This trait is identical to MapCoords, but is intended for use when mapping fallible functions over geometries.

It also adds Failure as a dependency, but I think that's OK.

### DIFF
--- a/src/algorithm/map_coords.rs
+++ b/src/algorithm/map_coords.rs
@@ -727,6 +727,13 @@ mod test {
 
     #[test]
     fn test_fallible_inplace() {
+        let f = |x: f64, y: f64| {
+            if x != 2.0 {
+                Ok((x * 2., y + 100.))
+            } else {
+                Err(format_err!("Ugh"))
+            }
+        };
         // this should produce an error
         let mut bad_ls: LineString<_> = vec![
             Point::new(1.0, 1.0),
@@ -741,25 +748,24 @@ mod test {
         ].into();
         assert!(
             bad_ls
-                .map_coords_inplace_fallible(&|&(x, y)| if x != 2.0 {
-                    Ok((x * 2., y + 100.))
-                } else {
-                    Err(format_err!("Ugh"))
-                })
+                .map_coords_inplace_fallible(&|&(x, y)| f(x, y))
                 .is_err()
         );
         assert!(
             good_ls
-                .map_coords_inplace_fallible(&|&(x, y)| if x != 2.0 {
-                    Ok((x * 2., y + 100.))
-                } else {
-                    Err(format_err!("Ugh"))
-                })
+                .map_coords_inplace_fallible(&|&(x, y)| f(x, y))
                 .is_ok()
         );
     }
     #[test]
     fn test_fallible() {
+        let f = |x: f64, y: f64| {
+            if x != 2.0 {
+                Ok((x * 2., y + 100.))
+            } else {
+                Err(format_err!("Ugh"))
+            }
+        };
         // this should produce an error
         let bad_ls: LineString<_> = vec![
             Point::new(1.0, 1.0),
@@ -772,21 +778,9 @@ mod test {
             Point::new(2.1, 2.0),
             Point::new(3.0, 3.0),
         ].into();
-        let bad = bad_ls.map_coords_fallible(&|&(x, y)| {
-            if x != 2.0 {
-                Ok((x * 2., y + 100.))
-            } else {
-                Err(format_err!("Ugh"))
-            }
-        });
+        let bad = bad_ls.map_coords_fallible(&|&(x, y)| f(x, y));
         assert!(bad.is_err());
-        let good = good_ls.map_coords_fallible(&|&(x, y)| {
-            if x != 2.0 {
-                Ok((x * 2., y + 100.))
-            } else {
-                Err(format_err!("Ugh"))
-            }
-        });
+        let good = good_ls.map_coords_fallible(&|&(x, y)| f(x, y));
         assert!(good.is_ok());
         assert_eq!(
             good.unwrap(),
--- a/src/algorithm/map_coords.rs
+++ b/src/algorithm/map_coords.rs
@@ -77,29 +77,6 @@ pub trait MapCoordsInplace<T> {
         T: CoordinateType;
 }
 
-/// Map a fallible function over the coordinates in a geometry, mutating them
-/// **NOTE**: If the result is `Err`, at least one of the coordinate transformation
-/// operations was unsuccessful, and the geometry is now in an inconsistent state.
-pub trait MapCoordsInplaceFallible<T> {
-    /// Map a fallible function over the coordinates in a geometry, mutating them
-    ///
-    /// ```
-    /// use geo::Point;
-    /// use geo::algorithm::map_coords::MapCoordsInplaceFallible;
-    ///
-    /// let mut p = Point::new(10., 20.);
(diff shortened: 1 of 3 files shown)


urschrei:
> If there's a 100 point LineString, and you call map_coords_inplace_failible, which is OK for first 10 points, and then fails, won't this make the LineString wrong because only the first 10 will have changed? The only way to work around this is to create a new LineString and std::mem::replace it in, right?

I should probably write a test that simulates this, but my intent is that each map call `Result` is checked. If it's an `Error`, that bubbles up to the main function via early return.

urschrei:
> What's the point of using a new dependency (failure)? I've seen that crate around in Rust news, but I've never used it nor looked into it. Rust has an Err option, which is what I'd use if I was implementing this.

`Failure` is a lot more flexible than the built-in `Err`. Using it means we can choose between several different error-handling strategies, and even mix them as necessary. For instance, it's easy to return a generic `Error` with a `String` as the cause (which is what I'm doing in #190) if it's coming from an external C lib. But we can also specify `Geo` specific errors, and because they can be converted into `Error`, and the function signature wouldn't change. `Failure`'s errors can also contain backtrace information, and nest, and it's possible to iterate over nested causes to see exactly what went wrong, and where.
Geo's error-handling story is (imo, and I'm certainly guilty of exacerbating it with `unwrap()` everywhere) under-specified at the moment, and this PR (and hopefully #190) both do things that are assumed to break – it seemed like a good time to get the ball rolling on the use of a library which will more than likely become the de-facto error-handling library in Rust.

rory:
> my intent is that each map call Result is checked. If it's an Error, that bubbles up to the main function via early return.

Sure, but since you're modifying the object in place, won't the modifications to the first 10 points still be there? Those can't be undone.

You could change the docs to say "Check the returned error of this, and if it's an error *the object is now in an inconsistant state* and shouldn't be used"

urschrei:
>You could change the docs to say "Check the returned error of this, and if it's an error the object is now in an inconsistant state and shouldn't be used"

Ah, I see what you mean. I had assumed that an `Err` return value would signal that the operation was unsuccessful, and the result should be discarded, but it's better to be explicit about what the `Err` means in this context.

urschrei:
Hmm the more I think about the fallible in-place version, the less I like what I've done. @rory is correct here I think – mem::replace is a better idea. If any of the operations *do* fail, it seems much better to return the unchanged version as an `Err`. But there's now very little difference to `MapCoordsFallible`.

src/algorithm/map_coords.rs
```diff
@@ -59,16 +109,37 @@ impl<T: CoordinateType, NT: CoordinateType> MapCoords<T, NT> for Point<T> {
     }
 }
 
-impl<T: CoordinateType> MapCoordsInplace<T> for Point<T> {
+impl<T: CoordinateType, NT: CoordinateType> MapCoordsFallible<T, NT> for Point<T> {
```
frewsxcv: instead of having all these explicit `MapCoordsFallible` `impl`s, what if we had a blanket `impl` like this (i haven't tried compiling it so it might be wrong):

```rust
impl<T: CoordinateType, NT: CoordinateType, G: MapCoords<T, NT>> MapCoordsFallible<T, NT> for G {
    type Output = G::Output;

    fn map_coords_fallible(
        &self,
        func: &Fn(&(T, T)) -> Result<(NT, NT), Error>,
    ) -> Result<Self::Output, Error> {
        Ok(self.map_coords(func))
    }
}
```

urschrei: I don't *think* we can implement it in terms of `map_coords`, because of the different input function signatures:
`map_coords` is `func: &Fn(&(T, T)) -> (NT, NT)`  
`map_coords_fallible` is `func: &Fn(&(T, T)) -> Result<(NT, NT), Error>`

And (I think) because we're also missing a `try!` or `?` to check for an early return, somewhere.

rory:
I haven't throught about this, but you could probably replace all the MapCoords with calls to MapCoordsInplaceFalliable and appropriate clones and wrapping functions.

MapcoordsFaillible is just a clone & MapCoordsInplaceFalliable. Mapcoordsinplace is just a wrapping the function to always return Ok() and then call MapCoordsInplaceFalliable etc etc.

On 18 March 2018 15:55:21 CET, "Stephan Hügel" <notifications@github.com> wrote:
>urschrei commented on this pull request.
>
>
>
>> @@ -59,16 +109,37 @@ impl<T: CoordinateType, NT: CoordinateType>
>MapCoords<T, NT> for Point<T> {
>     }
> }
> 
>-impl<T: CoordinateType> MapCoordsInplace<T> for Point<T> {
>+impl<T: CoordinateType, NT: CoordinateType> MapCoordsFallible<T, NT>
>for Point<T> {
>
>I don't *think* we can implement it in terms of `map_coords`, because
>of the different input function signatures:
>`map_coords` is `func: &Fn(&(T, T)) -> (NT, NT)`  
>`map_coords_fallible` is `func: &Fn(&(T, T)) -> Result<(NT, NT),
>Error>`
>
>And (I think) because we're also missing a `try!` or `?` to check for
>an early return, somewhere.

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

urschrei:
> MapcoordsFaillible is just a clone & MapCoordsInplaceFalliable. Mapcoordsinplace is just a wrapping the function to always return Ok() and then call MapCoordsInplaceFalliable etc etc.

Can you sketch this out (or send a PR to my branch)?

As for `MapCoordsinplaceFallible`, I've been tying myself in knots trying to get it right, and I don't know whether it's worth it – the best case scenario is that it's hiding a clone operation which means that someone (probably one of us) will end up complaining about memory usage and/or slowness at some point.

rory:
TBH, I'm not convinced that in place failible is a good idea.

Leaving an object in a strange/invalid state if one transform fails sounds like a really bad idea, because programmers won't read the docs and it will cause a bug. 🙂

I made the MapCoordsinplace because I didn't want clones, and inplacefailible having "hidden" clones seems bad. So I don't know the solution.

We could replace mapcoords with mapcoordsinplace though. I'll do a patch later 

On 18 March 2018 16:44:48 CET, "Stephan Hügel" <notifications@github.com> wrote:
>> MapcoordsFaillible is just a clone & MapCoordsInplaceFalliable.
>Mapcoordsinplace is just a wrapping the function to always return Ok()
>and then call MapCoordsInplaceFalliable etc etc.
>
>Can you sketch this out (or send a PR to my branch)?
>
>As for `MapCoordsinplaceFallible`, I've been tying myself in knots
>trying to get it right, and I don't know whether it's worth it – the
>best case scenario is that it's hiding a clone operation which means
>that someone (probably one of us) will end up complaining about memory
>usage and/or slowness at some point.

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

urschrei:
I agree, I'll get rid of it.

src/algorithm/map_coords.rs
frewsxcv: yep, you're right 👍 

i don't think there's a way to make this work since we can't return from the outer function within the closure if `func` returs an error

frewsxcv:
looks good to me. a little skeptical about incorporating `failure`, but it seems fine for now and we can always restart that discussion in the future

bors r+

frewsxcv:
i guess i should wait for @rory to dismiss their review (or approve the pr). any other concerns?

PR #192, opened 2018-03-14
Implement rust-proj as an optional feature within geo

urschrei:
Continuing the discussion from #190 and https://github.com/georust/rust-geo/issues/32.

## Introduction
`rust-proj` is currently (as of `0.3.0`) using PROJ v4.9.x, and depends upon `Geo` for its `Point` type. However, we now have a `proj-sys` crate, with low-level bindings to PROJ v5.0.0. Unfortunately, `rust-proj's` dependency on `Geo` means that we can't directly use its functions within `Geo`, as this introduces a circular dependency (see #190).
### This PR:
- Re-implements `rust-proj`'s functionality as an optional feature ("`proj`") within `Geo` using `proj-sys`
    - it has to be optional, as it requires PROJ v5.0.0 to be present
- Makes two new traits –`Project` and `Convert` – available to any geometries implementing the `MapCoordsFallible` trait (see #191), allowing them to be (inverse) projected to and from Geodetic coordinates, and converted between projections. This functionality is a slightly more granular version (in keeping with the new PROJ API, which makes the same distinction) of what's possible using `pyproj` and `Shapely` geometries (see [here](http://shapely.readthedocs.io/en/stable/manual.html#other-transformations) for the official example):
```python
from shapely.geometry import Point, LineString
import pyproj
# LineString in projected coordinates (OSGB36)
ls = LineString([(548295.39, 182498.46), (548296.39, 182499.46)])
osgb36=pyproj.Proj("+init=EPSG:27700")
eastings, northings = list(zip(*ls.coords))
# inverse-project to geodetic lon, lat coordinates (WGS 84)
conv_ls = zip(*osgb36(eastings, northings, inverse=True))
# compare to command-line proj output
assert conv_ls == [(0.13774816464635564, 51.52142699952049), (0.1377629902608001, 51.5214357231341)]
```

I propose that we leave `rust-proj` at 0.3.0, clearly marking it as deprecated, directing users to install `Geo` with the `proj` feature if they want v5.0.x features, and linking to `proj-sys` for implementers who don't wish to use `Geo`, for whatever reason.

We can also (as per my and @turboladen's discussion in #32) highlight the availability of this feature more clearly, if need be.


src/algorithm/proj.rs
```diff
@@ -0,0 +1,423 @@
...
+    /// are assumed to be geodetic in radians, unless an inverse projection is intended.
+    ///
+    /// For conversion operations, `definition` defines input, output, and
+    /// any intermediate steps that are required. See the `convert` example for more details.
+
+    // In contrast to PROJ v4.x, the type of transformation
+    // is signalled by the choice of enum used as input to the PJ_COORD union
+    // PJ_LP signals projection of geodetic coordinates, with output being PJ_XY
+    // and vice versa, or using PJ_XY for conversion operations
+    pub fn new(definition: &str) -> Option<Proj> {
```
turboladen: Having never used PROJ directly, what's an example `str` that one would pass in here? WKT of the projection?

turboladen: Nevermind--I just realized that's a dumb question. 😬

turboladen:
Sweet!

src/lib.rs
```diff
@@ -45,4 +51,6 @@ pub mod prelude {
     pub use algorithm::from_postgis::FromPostgis;
     #[cfg(feature = "postgis-integration")]
     pub use algorithm::to_postgis::ToPostgis;
+    #[cfg(feature = "projection")]
```
turboladen: Maybe I missed it, but I don't see `projection` as a defined feature--should this be `feature = "proj"`?

urschrei: It absolutely should. Fixed!

PR #193, opened 2018-03-18
Impl iter() and iter_mut() for some geometries

urschrei:
Implementing [`Deref`](https://doc.rust-lang.org/std/ops/trait.Deref.html) and [`DerefMut`](https://doc.rust-lang.org/std/ops/trait.DerefMut.html) for `LineString`, `Multi`-geometries, and
`GeometryCollection`  gives us `iter()` and `iter_mut()` for
them, allowing the removal of `self.0` in _lots_ of places.

### DIFF
--- a/src/algorithm/area.rs
+++ b/src/algorithm/area.rs
@@ -24,7 +24,10 @@ where
     fn area(&self) -> T;
 }
 
-fn get_linestring_area<T>(linestring: &LineString<T>) -> T where T: Float {
+fn get_linestring_area<T>(linestring: &LineString<T>) -> T
+where
+    T: Float,
+{
     twice_signed_ring_area(linestring) / (T::one() + T::one())
 }
 
@@ -55,11 +58,8 @@ where
     T: Float,
 {
     fn area(&self) -> T {
-        self.0
-            .iter()
-            .fold(T::zero(), |total, next| {
-                total + next.area()
-            })
+        self.iter()
+            .fold(T::zero(), |total, next| total + next.area())
     }
 }
 
--- a/src/algorithm/boundingbox.rs
+++ b/src/algorithm/boundingbox.rs
@@ -130,7 +130,7 @@ where
     /// Return the BoundingBox for a MultiLineString
     ///
     fn bbox(&self) -> Self::Output {
-        get_bbox(self.0.iter().flat_map(|line| line.0.iter()))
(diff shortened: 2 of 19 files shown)


urschrei:
I've also added [`Index`](https://doc.rust-lang.org/std/ops/trait.Index.html) and [`IndexMut`](https://doc.rust-lang.org/std/ops/trait.IndexMut.html) for the same traits, so we can e.g. index into a LineString's backing `Vec` directly:

    let line: LineString<f64> = vec![(0., 0.), (10., 0.)].into();
    assert_eq!(ls[1], Point::new(10., 0.));

frewsxcv:
i've seen this for a week or so now and am conflicted. as much as i'd love to r+ this since it makes some of our geometry structures more ergonomic (from an api standpoint), it _seems_ to be bad practice to use `Deref` in this manner:

* https://stackoverflow.com/a/45086999
* https://botbot.me/mozilla/rust/2018-03-29/?msg=98368531&page=8

i'm still thinking about this though. at the end of the day, i don't think implementing `Deref` as you did is causing any harm, but wondering if we should try try to stick to community undocumented conventionos here. ¯\\\_(ツ)\_/¯

urschrei:
Hmm. I resorted to `Deref` because I banged my head against `typeck` for a couple of hours trying to implement `iter` and `iter_mut` directly to no avail, so I can revisit that…

frewsxcv:
> trying to implement `iter` and `iter_mut` directly to no avail, so I can revisit that…

this compiles:

```diff
diff --git a/src/types.rs b/src/types.rs
index 68e8d8cc..e3e6d9e7 100644
--- a/src/types.rs
+++ b/src/types.rs
@@ -649,6 +649,14 @@ impl<T: CoordinateType> LineString<T> {
             Line::new(*w.get_unchecked(0), *w.get_unchecked(1))
         }))
     }
+
+    pub fn points(&self) -> ::std::slice::Iter<Point<T>> {
+        self.0.iter()
+    }
+
+    pub fn points_mut(&mut self) -> ::std::slice::IterMut<Point<T>> {
+        self.0.iter_mut()
+    }
 }
 
 /// Turn a `Vec` of `Point`-ish objects into a `LineString`.
```

it'll be nicer though once `impl Trait` hits stable, then we can just write:

```rust
pub fn points_mut(&mut self) -> impl Iterator<Item = &mut Point<T>> {
    self.0.iter_mut()
}
```

PR #195, opened 2018-03-21
Impl Polygon convexity function on the type

urschrei:
The convexity function was previously only available within the
Extremes module, but it's generally useful and will be required by
other algorithms. This moves the function and implements it directly
on the type.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -771,7 +771,7 @@ where
     // The polygon is convex if the z-components of the cross products are either
     // all positive or all negative. Otherwise, the polygon is non-convex.
     // see: http://stackoverflow.com/a/1881201/416626
-    pub fn convex(&self) -> bool {
+    pub fn is_convex(&self) -> bool {
         let convex = self
             .exterior
             .0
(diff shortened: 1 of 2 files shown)


src/types.rs
```diff
@@ -741,6 +741,66 @@ where
...
+                    (ListSign::Empty, true) | (ListSign::Positive, true) => ListSign::Positive,
+                    (ListSign::Empty, false) | (ListSign::Negative, false) => ListSign::Negative,
+                    _ => ListSign::Mixed
+                }
+            });
+        if convex == ListSign::Mixed {
+            false
+        } else {
+            true
+        }
```
frewsxcv: these last five lines can just be `convex != ListSign::Mixed`

src/types.rs
```diff
@@ -741,6 +741,66 @@ where
...
+{
+    /// Determine whether a Polygon is convex
+    // For each consecutive pair of edges of the polygon (each triplet of points),
+    // compute the z-component of the cross product of the vectors defined by the
+    // edges pointing towards the points in increasing order.
+    // Take the cross product of these vectors
+    // The polygon is convex if the z-components of the cross products are either
+    // all positive or all negative. Otherwise, the polygon is non-convex.
+    // see: http://stackoverflow.com/a/1881201/416626
+    pub fn convex(&self) -> bool {
```
frewsxcv: most things that return `bool` usually start with `is_` or `has_`. what do you think about calling this `is_convex`?

urschrei:
Addressed!

PR #200, opened 2018-04-11
Prefix Euclidean distance/length traits with 'Euclidean'.

frewsxcv:
I think we should make it explicit what algorithm the `Distance`/`Length` traits use. The `Distance`/`Length` naming suggests they should be used for most scenarios, even though the user probably wants Haversine, right?


urschrei:
I would have thought most people wanted Euclidean, but this removes any ambiguity, so 👍.

frewsxcv:
bors r=urschrei

PR #201, opened 2018-04-14
Experiment with a new geo-types crate. [feedback requested]

frewsxcv:
The idea is to move the core data types into a new, minimal crate called `geo-types`. The idea here is to have all third party crates (e.g. `rust-gdal`, `rust-geojson`) rely on this new crate instead pulling in all the baggage of the `geo` crate.


urschrei:
Need to have a closer look, but very much in favour of the idea. This keeps tripping us up in other ecosystem crates.

jdroenner:
I like this idea. `rust-gdal` currently only uses the core types. (Could we also have a geo-traits later?)

geo-types/src/algorithms.rs
```diff
@@ -0,0 +1,94 @@
+// FIXME: everything in this file is copy/paste from the 'geo' crate. ideally we
+//        wouldn't have this duplication
```
frewsxcv: the (non-public) traits below are unfortunately necessary to keep the `spade` trait `impl`s on the geo types in `geo-types/src/lib.rs`. for now it's a little messy, but i think we can reevaluate this in the future. i.e. do we really need `spade`?

frewsxcv:
@urschrei @jdroenner Alright, finished working on this. If either of you have some time, would love a pair of eyes 👀 for a review!

PR #203, opened 2018-04-18
fix polygon centroid for flat polygon

antoine-de:
if the polygon is flat (area = 0), it is considered as a linestring for the centroid computation.

without this for a flat polygon eg:

A ----> B -----> C
^-------------------|

the centroid was `Some(NaN, NaN)`

without this, the test `empty_interior_polygon_test` was also causing a `panic!` because of the unwrap in the centroid method.

### DIFF
--- a/src/algorithm/centroid.rs
+++ b/src/algorithm/centroid.rs
@@ -46,6 +46,10 @@ where
     T: Float + FromPrimitive,
 {
     let area = simple_polygon_area(poly_ext);
+    if area == T::zero() {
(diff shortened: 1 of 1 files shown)


urschrei:
Good catch. Thank you!

urschrei:
bors r=urschrei

antoine-de:
great, thanks!

when do you think this will be released?

urschrei:
@antoine-de released to crates.io as `0.8.1`

antoine-de:
thanks 👍

PR #213, opened 2018-05-11
Implement VincentyDistance and VincentyLength

frewsxcv:
Fixes https://github.com/georust/rust-geo/issues/205

Based off this implementation: https://github.com/janantala/GPS-distance/blob/master/java/Distance.java

### DIFF
--- a/geo/benches/vincenty_distance.rs
+++ b/geo/benches/vincenty_distance.rs
@@ -1,0 +1,19 @@
+#![feature(test)]
+
+extern crate geo;
+extern crate test;
+
+use geo::algorithm::vincenty_distance::VincentyDistance;
+
+const NITER: usize = 10_000;
+
+#[bench]
+fn bench_vincenty(bencher: &mut test::Bencher) {
+    let a = geo::Point::<f64>::new(17.107558, 48.148636);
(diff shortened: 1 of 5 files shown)


urschrei:
Vincenty can fail to converge for certain antipodal points (see here: https://github.com/geopy/geopy/issues/30), and [Charles Karney](https://link.springer.com/article/10.1007%2Fs00190-012-0578-z) has actually provided a method that's [guaranteed to converge](https://github.com/geopy/geopy/pull/144) in Geopy. I think we should return a `Result` if we keep Vincenty, or switch to Karney's method (it doesn't look particularly more complex, but I understand if you don't want to re-implement now).

frewsxcv:
> I think we should return a Result if we keep Vincenty

Done in https://github.com/georust/rust-geo/pull/213/commits/bcbbbd9b3c0dfb6c042376532a8aea657b5fa8b4

> or switch to Karney's method (it doesn't look particularly more complex, but I understand if you don't want to re-implement now)

I think we should add this in too, so I opened https://github.com/georust/rust-geo/issues/234

urschrei:
lgtm!

urschrei:
bors r=urschrei

Issue #198, opened 2018-04-09
Replace '-> Box<Iterator>' with '-> impl Iterator' once return impls stabilize


frewsxcv:
https://github.com/georust/rust-geo/blob/f082fe8656bf9a0a1639d5b4d28f2e0f6df369ee/src/types.rs#L642

lnicola:
`lines` can be easily replaced. `points_cw` and `points_ccw` however, depend on https://github.com/rust-lang/rust/issues/34511.

frewsxcv:
gonna close this. we don't have any instances of `Box<Iterator>` anymore

PR #218, opened 2018-05-12
Avoid Box in methods returning iterators

lnicola:
Part of #198.

This only uses `impl Iterator` for `lines`. The other methods can't use it because they're part of a trait.

The code is a little ugly, and replacing `Points` with `impl Trait` in the future will be a breaking change. Feel free to close if you think it's not worth it.

For what it's worth, if we do switch to `impl Iterator` in the future, writing down the ugly `EitherIter<&'a Point<T>, Iter<'a, Point<T>>, Rev<Iter<'a, Point<T>>>>` type will no longer be needed.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@
   * <https://github.com/georust/rust-geo/pull/209>
 * Use the `proj` crate, rename feature to `use-proj`
   * <https://github.com/georust/rust-geo/pull/214>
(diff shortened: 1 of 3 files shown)


frewsxcv:
Thanks!

bors r+

lnicola:
Maybe it would be worth noting the minimum compiler version in the release notes?

PR #220, opened 2018-05-13
Clarify wording for Contains algorithm.

frewsxcv:


frewsxcv:
Agreed, I think we could have a couple edge-case examples for this. Going to do that in a follow-up PR at some point if that's cool, instead of blocking this one

bors r+

frewsxcv:
bors r=antoine-de

PR #222, opened 2018-05-15
Extract geo types into separate files.

frewsxcv:


urschrei:
You are a hero for doing this.

PR #226, opened 2018-05-15
Polygon-Line euclidean distance fix

urschrei:
See here: https://github.com/georust/rust-geo/pull/219#pullrequestreview-120296643

WIP until I add a tests for a Line contained in a Polygon interior ring and intersection


urschrei:
This implementation is incorrect for `Lines` contained by interior `Polygon` rings, because it does `Point`-`Line` distance for each `Point` in the containing interior `Polygon` ring. This is wrong, though – it should be doing `Line`-`LineString` distance (which is just the minimum of a series of `Line`-`Line` distances). Alas, this requires a `Line`-`Line` distance implementation, which I now have to write.

PR #227, opened 2018-05-15
implement EuclideanDistance for MultiPolygon to Line and Line to MultiPolygon

JivanRoquet:
Implements the `EuclideanDistance` trait for the two following cases:

```rust
impl<T> EuclideanDistance<T, MultiPolygon<T>> for Line<T> { .. }
impl<T> EuclideanDistance<T, Line<T>> for MultiPolygon<T> { .. }
```

The test is not opinionated about the actual value being computed. It's only checking that the distance between the `Line` and the `MultiPolygon` is equal to the distance between the `Line` and the actual closest `Polygon` from this line.

More opinionated tests (aka, testing for real values) will be added once https://github.com/urschrei/rust-geo/commit/90dee9da71a55230b3c78fe80fb308d8fa933a1d is merged, as this PR depends on this commit for its results.

### DIFF
--- a/geo/src/algorithm/euclidean_distance.rs
+++ b/geo/src/algorithm/euclidean_distance.rs
@@ -324,6 +324,30 @@ where
     }
 }
 
+/// Line to MultiPolygon distance
+impl<T> EuclideanDistance<T, MultiPolygon<T>> for Line<T>
+where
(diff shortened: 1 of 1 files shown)


urschrei:
Looks good to me!

urschrei:
bors r=urschrei

PR #230, opened 2018-05-16
HaversineIntermediate trait to find intermediate points along a great circle route

JivanRoquet:
Given two existing points, implements the ability to find intermediate points along the great circle route joining the initial points.

The `haversine_intermediate` method takes three parameters:

- `&self`
- `other: &Point<T>` which represents the other side of the great circle route
- `f: T` which represents the fraction of the route at which to put the intermediate point

`f` must be included between `T::zero()` and `T::one()`.

Return value is `Point<T>`.

### DIFF
--- a/geo/src/algorithm/haversine_intermediate.rs
+++ b/geo/src/algorithm/haversine_intermediate.rs
@@ -32,7 +32,7 @@ pub trait HaversineIntermediate<T: Float> {
     /// ```
 
     fn haversine_intermediate(&self, other: &Point<T>, f: T) -> Point<T>;
-    fn haversine_intermediate_fill(&self, other: &Point<T>, max_dist: T) -> Vec<Point<T>>;
+    fn haversine_intermediate_fill(&self, other: &Point<T>, max_dist: T, include_ends: bool) -> Vec<Point<T>>;
 }
 
 impl<T> HaversineIntermediate<T> for Point<T>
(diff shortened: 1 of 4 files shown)


geo/src/algorithm/haversine_intermediate.rs
```diff
@@ -0,0 +1,118 @@
...
+        let d = two * k.asin();
+
+        let a = ((one - f) * d).sin() / d.sin();
+        let b = (f * d).sin() / d.sin();
+
+        let x = a * lat1.cos() * lon1.cos() + b * lat2.cos() * lon2.cos();
+        let y = a * lat1.cos() * lon1.sin() + b * lat2.cos() * lon2.sin();
+        let z = a * lat1.sin() + b * lat2.sin();
+
+        let dxy = (x.powi(2) + y.powi(2)).sqrt();
```
frewsxcv: I think this should be the same as `x.hypot(y)`

https://docs.rs/num/0.1.42/num/trait.Float.html#tymethod.hypot

JivanRoquet: agreed!

JivanRoquet:
Added the ability to "fill" a great circle route with intermediate points at regular interval, as described in issue https://github.com/georust/rust-geo/issues/228.

JivanRoquet:
up? :)

frewsxcv:
lgtm, thanks!

bors r+

PR #239, opened 2018-05-20
Add contribution guidelines and code of conduct

urschrei:
See [here](https://github.com/orgs/georust/teams/core/discussions/2) for prior discussion. This is a first effort, adapted from the originals at https://github.com/rust-lang/rust. I haven't specified any humans as part of the mod team, as this doesn't really make sense for us  – I've just said "contact the  the core team". Maybe we'd prefer specifically contactable people?

### DIFF
--- a/geo/src/algorithm/CODE_OF_CONDUCT.md
+++ b/geo/src/algorithm/CODE_OF_CONDUCT.md
@@ -1,0 +1,36 @@
+# The Rust-Geo Code of Conduct
+
+This document is based on, and aims to track the [Rust Code of Conduct](https://www.rust-lang.org/conduct.html)
+
+## Conduct
+
+* We are committed to providing a friendly, safe and welcoming environment for all, regardless of level of experience, gender identity and expression, sexual orientation, disability, personal appearance, body size, race, ethnicity, age, religion, nationality, or other similar characteristic.
+* On IRC, please avoid using overtly sexual nicknames or other nicknames that might detract from a friendly, safe and welcoming environment for all.
+* Please be kind and courteous. There's no need to be mean or rude.
+* Respect that people have differences of opinion and that every design or implementation choice carries a trade-off and numerous costs. There is seldom a right answer.
+* Please keep unstructured critique to a minimum. If you have solid ideas you want to experiment with, make a fork and see how it works.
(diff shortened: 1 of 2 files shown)


geo/src/algorithm/CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,36 @@
+# The Rust-Geo Code of Conduct
```
frewsxcv: Is our CoC going to cover all GeoRust projects? If so, should this be the GeoRust CoC?

geo/src/algorithm/CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,36 @@
...
+
+And if someone takes issue with something you said or did, resist the urge to be defensive. Just stop doing what it was they complained about and apologize. Even if you feel you were misinterpreted or unfairly accused, chances are good there was something you could've communicated better — remember that it's your responsibility to make your fellow Rustaceans comfortable. Everyone wants to get along and we are all here first and foremost because we want to talk about cool technology. You will find that people will be eager to assume good intent and forgive as long as you earn their trust.
+
+*Adapted from the [Node.js Policy on Trolling](http://blog.izs.me/post/30036893703/policy-on-trolling) as well as the [Contributor Covenant v1.3.0](https://www.contributor-covenant.org/version/1/3/0/).*
+
+[mod_team]: https://github.com/orgs/georust/teams/core
```
frewsxcv: I'm not entirely sure about this, so I can check later, but I don't think non-organization-members can access this page. I was also considering registering georust.org and setting up an alias like code-of-conduct@georust.org that would get sent to some of us, but haven't had the time.

geo/src/algorithm/CODE_OF_CONDUCT.md
urschrei: I think that's a good idea – should we ask individual project owners?

geo/src/algorithm/CODE_OF_CONDUCT.md
urschrei: I just checked, and you're right – you can't see teams. A `.org` and dedicated email is a good idea for this. I'm happy to do it during the week and send over the creds if you're too busy.

frewsxcv:
I registered georust.org and setup mods@georust.org to forward to a private Google Group

urschrei:
Amazing! Shall we merge?

Issue #176, opened 2017-10-24
Wrong earth radius in haversine_destination?


Boscop:
WGS84 says earth radius is 6378137.0 meters (like the comment says) but the code uses 6371000.0 meters:
https://georust.github.io/rust-geo/src/geo/algorithm/haversine_destination.rs.html#29

urschrei:
IIRC, it was [because](https://rosettacode.org/wiki/Haversine_formula)
>the derivation of this ellipsoidal quadratic mean radius is wrong (the averaging over azimuth is biased). When applying these examples in real applications, it is better to use the mean earth radius, 6371 km. *This value is recommended by the International Union of Geodesy and Geophysics and it minimizes the RMS relative error between the great circle and geodesic distance*.
(emphasis mine)

See also: https://en.wikipedia.org/wiki/Earth_radius#Mean_radius, which references ftp://athena.fsv.cvut.cz/ZFG/grs80-Moritz.pdf section 4a.

Boscop:
Ah ok. And is it correct that bearing should be in degrees, clockwise with 0° at north for [`haversine_destination()`](https://georust.github.io/rust-geo/geo/algorithm/haversine_destination/trait.HaversineDestination.html)?

I'm trying to convert lat/lng coords into flattened cartesian coords for the purpose of vehicle routing within a confined area (less than 50km in diameter) using the first point as origin of the cartesian coord system, but somehow I get a huge error when converting back. This is my code:
```rust
pub type F = f64;
pub type Vec2 = Vector2<F>; // km

fn to_cartesian(origin: &Location, loc: Location) -> Vec2 {
	use geo::Point;
	use geo::algorithm::haversine_distance::HaversineDistance;

	let orig = Point::new(origin.longitude, origin.latitude);
	let delta_x_m = orig.haversine_distance(&Point::new(loc.longitude, origin.latitude));
	let delta_y_m = orig.haversine_distance(&Point::new(origin.longitude, loc.latitude));
	Vec2::new(delta_x_m as F * 0.001, delta_y_m as F * 0.001)
}

fn to_lat_lng(origin: &Location, loc: &Vec2) -> Location {
	use geo::Point;
	use geo::algorithm::haversine_destination::HaversineDestination;

	let origin = Point::new(origin.longitude, origin.latitude);
	let ang_deg = loc.x.atan2(loc.y).to_degrees() as f64;
	let dist_km = loc.norm() as f64;
[...]

PR #247, opened 2018-06-04
Organize all Earth radius/constants.

frewsxcv:


geo/src/lib.rs
```diff
@@ -28,6 +28,19 @@ pub mod algorithm;
 #[macro_use]
 extern crate approx;
 
+// Mean radius of Earth in meters
```
urschrei: I'd add a note (or maybe just link to the explanations in #176), or something that makes the distinction between uses of mean spherical and ellipsoidal earth radii here somewhere, just so we don't have to remember which is used for what application when this comes up again.

frewsxcv:
bors r=urschrei

PR #255, opened 2018-06-04
Remove redundant distance calculation

### LINKED ISSUE #252: Point/Line distance function defined twice
On paper, these sound like they do the same thing, but it looks like the implementation might be different?

https://github.com/georust/rust-geo/blob/fbcabaae3dc5f2f0dfe73ec97ae2b22475d4ae82/geo/src/algorithm/euclidean_distance.rs#L74-L96

https://github.com/georust/rust-geo/blob/fbcabaae3dc5f2f0dfe73ec97ae2b22475d4ae82/geo/src/algorithm/simplify.rs#L5-L19

urschrei:
closes #252


urschrei:
Uhhh also I'm now getting a failing test. Hrm.

Issue #254, opened 2018-06-04
Trait implementations that calculate sums could utilize Sum trait


frewsxcv:
https://doc.rust-lang.org/std/iter/trait.Sum.html

```diff
diff --git a/geo/src/algorithm/euclidean_length.rs b/geo/src/algorithm/euclidean_length.rs
index 1c07b754..49c8da7e 100644
--- a/geo/src/algorithm/euclidean_length.rs
+++ b/geo/src/algorithm/euclidean_length.rs
@@ -1,5 +1,6 @@
 use num_traits::Float;
 
+use std::iter::Sum;
 use ::{Line, LineString, MultiLineString};
 use algorithm::euclidean_distance::EuclideanDistance;
 
@@ -36,12 +37,12 @@ where
 
 impl<T> EuclideanLength<T> for LineString<T>
 where
-    T: Float,
+    T: Float + Sum,
 {
     fn euclidean_length(&self) -> T {
         self.lines()
             .map(|line| line.euclidean_length())
-            .fold(T::zero(), |total_length, length| total_length + length)
+            .sum()
     }
 }
```

frewsxcv:
Basically just search for `fold.*zero` or just `fold` in the code and you'll find a bunch of places to cleanup

urschrei:
I had so much fun writing those folds, too.

PR #260, opened 2018-06-06
Replace folds with sum()

urschrei:
Removes some of the unnecessary `fold`s (see #254). WIP til I've tracked them all down and checked to see whether the lifetime issues that crop up with the remaining uses can be resolved.

### DIFF
--- a/geo/src/algorithm/centroid.rs
+++ b/geo/src/algorithm/centroid.rs
@@ -1,8 +1,9 @@
 use num_traits::{Float, FromPrimitive};
+use std::iter::Sum;
 
-use ::{Bbox, Line, LineString, MultiPolygon, Point, Polygon};
 use algorithm::area::Area;
 use algorithm::euclidean_length::EuclideanLength;
+use {Bbox, Line, LineString, MultiPolygon, Point, Polygon};
 
 /// Calculation of the centroid.
 pub trait Centroid<T: Float> {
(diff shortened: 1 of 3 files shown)


urschrei:
I think this actually covers all the cases in which it's just a simple accumulating fold.

geo/src/algorithm/euclidean_length.rs
```diff
@@ -35,18 +36,18 @@ where
...
         self.lines()
             .map(|line| line.euclidean_length())
-            .fold(T::zero(), |total_length, length| total_length + length)
+            .sum()
     }
 }
 
 impl<T> EuclideanLength<T> for MultiLineString<T>
 where
-    T: Float,
```
frewsxcv: Did you mean to convert the `fold` usage below into `sum`?

geo/src/algorithm/centroid.rs
```diff
@@ -157,7 +161,7 @@ where
 
 impl<T> Centroid<T> for MultiPolygon<T>
 where
-    T: Float + FromPrimitive,
+    T: Float + FromPrimitive + Sum,
```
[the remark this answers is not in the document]
urschrei: Oops, removing…

geo/src/algorithm/euclidean_length.rs
urschrei: It's needed because of the `Sum` bound on `EuclideanLength` for `LineString`.

Issue #263, opened 2018-06-08
Spade 1.5.0 causes tests to fail


urschrei:
I can repro the failures, but I currently have no idea why they're occurring (I can't repro them using Spade's native types, which makes me wonder whether I've messed up a trait implementation somewhere.)

urschrei:
Welp, doesn't look like it's the `SpatialObject` impl on `Line`…

urschrei:
Issue open here: https://github.com/Stoeoef/spade/issues/32

urschrei:
(We should just pin Spade 1.4.0 'til this is fixed)

frewsxcv:
https://github.com/georust/rust-geo/pull/266

PR #264, opened 2018-06-08
Add test for Spade failure

urschrei:
This test currently fails using Spade 1.5.0

See also: #263


urschrei:
Closing – it's a Spade bug and they can repro it, so we don't need to incorporate a specific test for it.

PR #265, opened 2018-06-09
Remove unnecessary borrows in function params for `Copy` types.

frewsxcv:
Breaking changes for `geo` and `geo-types`.

### DIFF
--- a/geo-types/src/algorithms.rs
+++ b/geo-types/src/algorithms.rs
@@ -8,22 +8,22 @@ pub trait EuclideanDistance<T, Rhs = Self> {
     fn euclidean_distance(&self, rhs: &Rhs) -> T;
 }
 
-fn line_segment_distance<T>(point: &Point<T>, start: &Point<T>, end: &Point<T>) -> T
+fn line_segment_distance<T>(point: Point<T>, start: Point<T>, end: Point<T>) -> T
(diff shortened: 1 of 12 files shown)


frewsxcv:
I need to benchmark this...

frewsxcv:
Okay, just benchmarked this with our current set of benchmarks. With these changes, `convex hull f32` and `convex hull f64` speed up ~17% and `Polygon Euclidean distance RTree f64` speeds up ~4%. Everything else that we have benchmarks for there's not much of a performance change. I'm gonna approve...

bors r+

urschrei:
How are we getting 17% improvement for this?! (I mean, that's amazing, but I thought it was largely a semantic change, why is there such a perf difference?)

frewsxcv:
> How are we getting 17% improvement for this?! (I mean, that's amazing, but I thought it was largely a semantic change, why is there such a perf difference?)

Not sure! It looks like convex hull is a recursive algorithm and maybe by in-lining the points/coordinates instead of references, it's able to do something like [tail call optimization](http://wiki.c2.com/?TailCallOptimization)? 🤷‍♂️ If I had some more time, it could be interesting to dig into the compiled LLVM to see what's going on behind the scenes.

PR #269, opened 2018-06-13
Add more benchmarks

urschrei:
For euclidean distance (both types of Polygon-Polygon), simplification, containment, convex hull


frewsxcv:
I was _just_ about to add a bunch of benchmarks like these. Glad I didn't otherwise we could have done duplicate work 😅

PR #271, opened 2018-06-14
Migrate a couple benchmarks to use Criterion.

frewsxcv:


frewsxcv:
criterion is just like the default `cargo bench` backend, but has some extra feature like:

- reporting performance changes across `cargo bench` runs
- better UI in the console
- reports outliers

try it out! just check out this branch and run `cargo bench`

urschrei:
Ah, amazing.

PR #276, opened 2018-06-17
Refactor rotate algorithm. Add rotate benchmark.

frewsxcv:


geo/src/algorithm/rotate.rs
```diff
@@ -111,8 +118,8 @@ where
 {
     /// Rotate the Point about itself by the given number of degrees
     /// This operation leaves the point coordinates unchanged
-    fn rotate(&self, angle: T) -> Self {
-        rotation_matrix(angle, self.centroid(), &[*self])[0]
+    fn rotate(&self, _angle: T) -> Self {
+        *self
```
frewsxcv: Rotating a single point around itself is a no-op

geo/src/algorithm/rotate.rs
```diff
@@ -121,9 +128,11 @@ where
...
 {
     fn rotate(&self, angle: T) -> Self {
-        let pts = vec![self.start, self.end];
-        let rotated = rotation_matrix(angle, self.centroid(), &pts);
-        Line::new(rotated[0], rotated[1])
+        let centroid = self.centroid();
+        Line::new(
+            rotate_one(angle, centroid, self.start),
+            rotate_one(angle, centroid, self.end),
+        )
```
frewsxcv: Previous implementation resulted in heap allocations with the `vec![...]` creation and also the `Vec` returned from `rotation_matrix`. New implementation shouldn't have any heap allocations

frewsxcv:
Also, no performance difference with these changes based off the new benchmark

PR #285, opened 2018-07-01
Introduce Triangle geometry type.

frewsxcv:
Fixes https://github.com/georust/rust-geo/issues/261.

GEOS also offers a Triangle type:

- http://geos.osgeo.org/doxygen/classgeos_1_1geom_1_1Triangle.html

I updated the `polygon_distance_fast_path` and `simplifyvw` algorithms
to utilize this new geometry type. As a result of this change and
tangential optimizations, the `simplify vw simple f32` benchmark is now
10% faster.

### DIFF
--- a/geo-types/src/lib.rs
+++ b/geo-types/src/lib.rs
@@ -50,6 +50,9 @@ pub use geometry::Geometry;
 mod geometry_collection;
 pub use geometry_collection::GeometryCollection;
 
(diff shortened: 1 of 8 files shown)


frewsxcv:
Opened a follow-up issue for that: https://github.com/georust/rust-geo/issues/286

bors r=urschrei

PR #295, opened 2018-07-04
Rename bounding ‘box’ to ‘rect’; move structure to geo-types.

frewsxcv:
- Rename `Bbox` structure to `Rect`, move to `geo_types`, transition
  from four `T` values to two `Coordinate<T>` values
- Rename `BoundingBox` to `BoundingRect`

Fixes https://github.com/georust/rust-geo/issues/292.


geo/src/algorithm/bounding_rect.rs
```diff
@@ -0,0 +1,290 @@
...
+
+    ///
+    /// Return the BoundingRect for a MultiPoint
+    ///
+    fn bounding_rect(&self) -> Self::Output {
+        get_bounding_rect(self.0.iter().map(|p| p.0))
+    }
+}
+
+impl<T> BoundingRect<T> for Line<T>
```
michaelkirk: Are the Line and LineString implementations intentionally duplicated in geo-types/src/algorithms.rs?

geo-types/src/line_string.rs
```diff
@@ -174,7 +174,7 @@ where
     type Point = Point<T>;
 
     fn mbr(&self) -> ::spade::BoundingRect<Self::Point> {
-        let bbox = self.bbox();
+        let bbox = self.bounding_rect();
```
michaelkirk: nit: seems like you're preferring `let bounding_rect =` elsewhere

frewsxcv: yep, good catch, thanks

geo/src/algorithm/bounding_rect.rs
frewsxcv: yeah, intentional. to have `spade` trait implementations for the types in `geo-types`, we need to have some basic algorithms included in the `geo-types` crate. but i'm planning to consolidate them in a follow-up pull request to this one.

PR #296, opened 2018-07-04
Distance and Contains for Triangle, Point

michaelkirk:


michaelkirk:
Rebased and conflicts addressed.

Issue #250, opened 2018-06-04
Simple polygon area defined twice


frewsxcv:
https://github.com/georust/rust-geo/blob/fbcabaae3dc5f2f0dfe73ec97ae2b22475d4ae82/geo/src/algorithm/winding_order.rs#L5-L15

https://github.com/georust/rust-geo/blob/fbcabaae3dc5f2f0dfe73ec97ae2b22475d4ae82/geo/src/algorithm/centroid.rs#L30-L43

Can merge these two functions. Would be good to hold off until https://github.com/georust/rust-geo/pull/246 merges to avoid conflicts.

PR #301, opened 2018-08-05
algorithm: Merge polygon area calculation into a single function

ketotek:
Fix for https://github.com/georust/geo/issues/250.
I considered using the map/sum iterator, but that would have required changing many type restrictions. Let me know if that would be better.

Signed-off-by: Alexandru Copot <alex.mihai.c@gmail.com>

### DIFF
--- a/geo/src/algorithm/area.rs
+++ b/geo/src/algorithm/area.rs
@@ -34,7 +34,8 @@ where
     fn area(&self) -> T;
 }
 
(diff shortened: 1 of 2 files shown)


ketotek:
> Thanks for the contribution! I hope you don't mind that I added one small commit to make pub into pub(crate) for get_linestring_area. We don't want to expose this function to users outside of the crate, so we'll restrict the scope to callers within the crate.

Sure, thanks, I missed that aspect.

PR #307, opened 2018-10-25
Create feature flags for optional dependencies

fschutt:
If a crate depends on geo-types this makes it easier to turn off serde and spade. Since these features are enabled by default, this is not a breaking change.

Right now there is no way to actually turn off `serde` and `spade` - `geo-types = { version = "0.2.0", default-features = false }` *still* includes those dependencies. If this gets merged, it would be great if a new minor version of geo-types could be published.

### DIFF
--- a/geo-types/Cargo.toml
+++ b/geo-types/Cargo.toml
@@ -9,6 +9,11 @@ readme = "../README.md"
 keywords = ["gis", "geo", "geography", "geospatial"]
 description = "Geospatial primitive data types"
 
+[features]
+serde_derive = ["serde"]
+spade_ext = ["spade"]
+default = ["serde_derive", "spade_ext"]
+
 [dependencies]
 num-traits = "0.2"
 serde = { version = "1", optional = true, features = ["derive"] }


KodrAus:
Hi @fschutt :wave:

Hmm, the `serde` and `spade` dependencies are already optional in the `geo-types` crate, and don't appear to be enabled by default, so unless another dependency in your graph has explicitly opted in to either of those features they shouldn't be getting pulled in.

fschutt:
Yeah, that's what should happen, but that's not what happens in reality. Try it yourself - make a new crate, put `geo-types = { version = "0.2.0", default-features = false }` and run `cargo tree` on top of it to see the dependencies. The only way to disable these features right now is to make feature flags for them. Otherwise they still get pulled in, although that technically shouldn't happen.

KodrAus:
There must be something else going on here, when I make a new crate with the following `Cargo.toml`:

```toml
[dependencies]
geo-types = { version = "0.2.0", default-features = false }
```

And run `cargo build` I don't compile `serde` or `space`, and they don't find their way into my `Cargo.lock`:

```toml
[[package]]
name = "geo-type-test"
version = "0.0.0"
dependencies = [
 "geo-types 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
]

[[package]]
name = "geo-types"
version = "0.2.0"
source = "registry+https://github.com/rust-lang/crates.io-index"
dependencies = [
 "num-traits 0.2.6 (registry+https://github.com/rust-lang/crates.io-index)",
]

[[package]]
name = "num-traits"
version = "0.2.6"
source = "registry+https://github.com/rust-lang/crates.io-index"

[metadata]
"checksum geo-types 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)" = "7260ac8abfc04cc21b914254b69875c2548faf6f791f88ad9ec1a131d4251729"
"checksum num-traits 0.2.6 (registry+https://github.com/rust-lang/crates.io-index)" = "0b3a5d7cc97d6d30d8b9bc8fa19bf45349ffe46241e8816f50f62f6d6aaabee1"
```

frewsxcv:
It's possible `cargo tree` shows all dependencies, even if they're not enabled in the current feature set

PR #314, opened 2018-11-22
Replaced spade with rstar

Stoeoef:
Hi there!

I'm in the process of extracting spade's rtree and was finally able to get the geo tests running with the planned replacement crate: `rstar`.
Since the results so far look really promising, I'd like to share them with this PR to get some first reactions.
The goodies:
- Reduced compilation times for `geo` workspace: `cargo clean && cargo build` was reduced to 11s (from 40s) on my local machine
- Benchmark improvements:

> simplify vwp f32        time:   [17.300 ms 17.413 ms 17.551 ms]                          
>                        change: [-70.837% **-70.698%** -70.558%] (p = 0.00 < 0.05)
>                        Performance has improved.

The smaller build times may actually justify to use `rstar` without a feature flag.

Some TODO's before I'm willing to remove the WIP status:
- [x] I'll perform a review of the changes myself again
- [x] The simplify vwp algorithm was adapted a bit, I'll need to investigate if it is still correct. I'll add a comment to the appropriate section and ask for a special review.
- [x] I need to check the test coverage of `rstar` for the methods used by `geo`, Should not be too dramatic, but I want those tests anyway...

Also, `rstars` public interface is likely to change in the near future - however, I would keep `geo` updated should any breaking change make this necessary. That is, after all, in my own interest, as failing geo test cases will likely mean that something is wrong with my changes.

Any thoughts or opinions are welcome. Please keep in mind that, as I just published the `rstar` crate, it will take a while until it is in a polished state. I hope does not pose too much of a problem.

### DIFF
--- a/geo/Cargo.toml
+++ b/geo/Cargo.toml
@@ -16,11 +16,11 @@ travis-ci = { repository = "georust/geo" }
 [dependencies]
 num-traits = "0.2"
 serde = { version = "1.0", optional = true, features = ["derive"] }
-spade = "1.5.1"
+rstar = { version = "0.1" }
 failure = "0.1.2"
 postgis = { version = "0.6", optional = true }
 proj = { version = "0.5", optional = true }
-geo-types = { version = "0.2", features = ["spade"] }
+geo-types = { path = "../geo-types", features = ["rstar"] }
 
 [features]
 default = []
--- a/geo/src/algorithm/simplifyvw.rs
+++ b/geo/src/algorithm/simplifyvw.rs
@@ -4,8 +4,7 @@ use std::cmp::Ordering;
 use std::collections::BinaryHeap;
 use {Coordinate, Line, LineString, MultiLineString, MultiPolygon, Point, Polygon, Triangle};
 
-use spade::rtree::RTree;
-use spade::{self, SpadeFloat};
+use rstar::{RTree, RTreeNum};
 
 /// Store triangle information
 // current is the candidate point for removal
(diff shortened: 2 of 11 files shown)


urschrei:
Oh, amazing. Thank you! Feel free to ping me about the vwp changes – if the tests are still passing, I'm pretty confident that it'll be OK, but I can certainly cook up some more to exercise it even more thoroughly.

frewsxcv:
Wow, this looks awesome! Extracting out the r*-tree was a great idea 👍 

I just skimmed through the code and everything is look good, ping me or @urschrei if you need any help! 👋

geo/Cargo.toml
```diff
@@ -16,11 +16,11 @@ travis-ci = { repository = "georust/geo" }
 [dependencies]
 num-traits = "0.2"
 serde = { version = "1.0", optional = true, features = ["derive"] }
-spade = "1.5.1"
+rstar = { version = "0.1" }
 failure = "0.1.2"
 postgis = { version = "0.6", optional = true }
 proj = { version = "0.5", optional = true }
-geo-types = { version = "0.2", features = ["spade"] }
+geo-types = { path = "../geo-types", features = ["rstar"] }
```
Stoeoef: This will need to be updated in order to publish the crate... feel free to change it to any version you see fit.

geo/src/algorithm/simplifyvw.rs
```diff
@@ -290,11 +289,20 @@ where
...
-        tree.lookup_and_remove(&Point(orig.0[smallest.left]));
+        // Remove stale segments from R* tree
+        let left_point = Point(orig.0[left as usize]);
+        let middle_point = Point(orig.0[smallest.current]);
+        let right_point = Point(orig.0[right as usize]);
+
+        let line_1 = Line::new(left_point, middle_point);
+        let line_2 = Line::new(middle_point, right_point);
+        assert!(tree.remove(&line_1).is_some());
+        assert!(tree.remove(&line_2).is_some());
```
Stoeoef: Note that I originally had a different assertion:
`assert!(tree.remove(&line_1).is_some() || tree.remove(&line1.reverse().is_some())`
as I was not sure in which direction a line would have been loaded.

I then realized that, as long as all lines are going "in the same direction", these assumptions should hold: If edges A->B and B->C exist, then they should be loaded as `(A,B), (B,C)` in contrast to `(A,B),(C,B))`.

Since all incoming lines are either the lines of a line string (which is directed) or the inner rings of a polygon (which are also directed), these assertions should hold. However, you may want to check this yourselves. Please note that the spade algorithm was not susceptible to this, that's why I'm so cautious. It was though, at the same time, much more inefficient.

urschrei: As you say, this could only be a problem for us if someone constructs an invalid geometry. @frewsxcv, I know we aren't currently checking for that, but I don't know if this is the right place to enforce it – I think it's reasonable for individual algorithms to assume that incoming geometries are valid.

geo/Cargo.toml
frewsxcv: yep, this makes sense 👍

geo/src/algorithm/simplifyvw.rs
frewsxcv: agreed. we still need to work on our validation story in this crate, but for the time being, i think it's okay to assume geometry coming in is invalid so this looks good 👍

PR #315, opened 2018-11-22
Grab bag of changes ✨

frewsxcv:


urschrei:
How does the `VScore` struct change work? It looks like the `I` type is generic over…everything?

frewsxcv:
> How does the `VScore` struct change work? It looks like the `I` type is generic over…everything?

I noticed that `intersector` is effectively unused in the `visvalingam` function, it's always `false`. We're using extra memory space for that field, so to decrease memory usage slightly, we could remove that field.

By making `intersector` generic, we can choose what the type of that field is on a per-usage basis. [Here's a playground I threw together](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2015&gist=ccbe35b33cfa02a03d2830031a85cd90) that hopefully helps demonstrate this.

It's true that `I` is generic over everything, _until we make the `I` type concrete_.

For example, we can't do:

```rust
impl VScore<I> {
    fn add_one(&mut self) {
        self.intersector += 1;
    }
}
```

...since `intersector` is `I`, which can be _anything_.

_But_, once we make `intersector` a concrete type, then we can perform actions on it:

```rust
impl VScore<u32> {
    fn add_one(&mut self) {
        self.intersector += 1;
    }
}

fn main() {
    let mut v = VScore { intersector: 1u32 };
    v.add_one();

    let mut v = VScore { intersector: 1u64 };
    // This won't work since `add_one` isn't implemented for `VScore<u64>`
    // v.add_one();
}
```

Let me know if I can clarify further, always happy to help!

urschrei:
Ahh, it never occurred to me that you could do that, but it makes perfect sense. Thanks for the explanation!

PR #321, opened 2018-12-03
Allow output type of Area to differ from coordinate type.

frewsxcv:
I feel like many of our algorithms should do this...

Thoughts?


urschrei:
So IIUC, `NumCast` allows for conversions between machine scalar types, returning an `Option<T>`? How does conversion failure handling look?

PR #322, opened 2018-12-17
add centroid implementation for MultiPoint

nberard:
Adds missing implementation of centroid on MultiPoint struct.

### DIFF
--- a/geo/src/algorithm/centroid.rs
+++ b/geo/src/algorithm/centroid.rs
@@ -3,7 +3,7 @@ use std::iter::Sum;
 
 use algorithm::area::{get_linestring_area, Area};
 use algorithm::euclidean_length::EuclideanLength;
-use {Line, LineString, MultiPolygon, Point, Polygon, Rect};
+use {Line, LineString, MultiPoint, MultiPolygon, Point, Polygon, Rect};
 
 /// Calculation of the centroid.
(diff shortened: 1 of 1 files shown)


geo/src/algorithm/centroid.rs
```diff
@@ -244,6 +244,40 @@ where
...
+{
+    type Output = Option<Point<T>>;
+
+    fn centroid(&self) -> Self::Output {
+        if self.0.is_empty() {
+            return None;
+        }
+        let sum = self.0.iter().fold(
+            Point::new(T::zero(), T::zero()),
+            |a: Point<T>, b: &Point<T>| Point::new(a.x() + b.x(), a.y() + b.y()),
```
urschrei: do we need `b` to be **`&`**`Point<T>`?

TeXitoi: `.iter()` returns a `Iterator<Item=&Point<T>>` thus you can't really do any other thing (type inference might work here).

nberard:
the test is in the comment ;)

TeXitoi:
That's OK for me, @urschrei I let you merge if everything is good on your side.

frewsxcv:
https://github.com/Stoeoef/rstar/commit/d4ca14e29a9d4b567e1f80a9a9296bd8b0a7adc7

PR #325, opened 2018-12-23
Rewrite `partition` util function, move to new `utils` module.

frewsxcv:
I read through this code yesterday and got stumped by the `swap_remove_to_first` and `swap_remove_to_last` functions. I understand them now, but I found this other iterative implementation easier to follow personally, so I swapped it out 🙈. I benchmarked this PR, no performance changes.


urschrei:
Those have been confusing me for _years_, and I never followed up on the offer of an explanation of how they worked. ✨

Issue #166, opened 2017-09-21
Do polygons need to be closed?


Michael-F-Bryan:
I don't think it specifies in the documentation whether a polygon should be closed or not, although I'm assuming that's an implicit assumption. 

Would it be worth adding this in `Polygon`'s docs (and possibly adding a `debug_assert!()` in the constructor) if this constraint must be upheld?

rory:
I also wondered about this, I have been closing my polygons (ie last point = first point). At a minimum the docs should clarify this.

urschrei:
According to the OpenGeo spec, [the outer boundary of a Polygon must be closed and simple](https://connect.boundlessgeo.com/docs/suite/4.8/dataadmin/pgBasics/geometries.html#polygons). A debug assert is a good idea, but I think we should be implicitly closing Polygons when the `new` method is called, and making that clear in the docs (it's what Shapely does, not totally sure about JTS / GEOS, but TestBuilder does it too) – there will be lots of cases where they're constructed at runtime. This is part of the larger "should it be possible to construct invalid geometries" discussion, but in this case, it's a fundamental problem if the shell is left open.

rory:
Idea: Automatically close the linestrings if they are not closed. Just look if the first & last are the same, and if not `.clone()` and `.push(...)`. You know what the user is trying to do.

*But* (and this goes back to the validity argument) is it worth having some sort of `.new_unchecked()` method which just takes what you give it (potentially resulting in invalid geometries).

Michael-F-Bryan:
It's probably not. A `.clone()` in this case is just copying a pair of floats and appending them to your line. It might be expensive if the line's inner `Vec` is at full capacity and the push forces it to resize, but I'm going to assume that's a fairly rare edge case.

urschrei:
Another (probably rare in practice) downside is if you keep the original Vec around and construct the Polygon by cloning it. If you depend on their equality (or even the length being the same), that's going to cause problems.

rory:
My suggestion for `.new_unchecked()` is not for performance reasons, but because I think it's good to have a way for the user (of the library) to say "No, I *really* know what I'm doing, Yes I'm willing to deal with the consequences". Sort of a unix philosophy for software libraries. :)

urschrei:
I agree with that in principle, but in the interest of striking a balance between "no surprises", and API bloat, I think it'd be good to come up with at least one reason why a user might want to do that.

Michael-F-Bryan:
[...]

PR #337, opened 2019-01-21
Rewrite Polygon structure to enforce closed LineString rings.

frewsxcv:
Prior to this pull request, the `exterior` and `interior` fields on a `Polygon` were public, allowing users to modify their `Coordinate`s as they saw fit.

To ensure all rings in a `Polygon`s remain closed, this pull request:

- privatizes the `exterior` and `interior` fields
- adds getters/mutators
- closes all `LineString` rings upon construction or after mutation

Fixes https://github.com/georust/geo/issues/166.

### DIFF
--- a/geo/src/algorithm/contains.rs
+++ b/geo/src/algorithm/contains.rs
@@ -164,10 +164,10 @@ where
     T: Float,
 {
     fn contains(&self, p: &Point<T>) -> bool {
-        match get_position(*p, &self.exterior) {
+        match get_position(*p, &self.exterior()) {
             PositionPoint::OnBoundary | PositionPoint::Outside => false,
             _ => self
-                .interiors
+                .interiors()
                 .iter()
                 .all(|ls| get_position(*p, ls) == PositionPoint::Outside),
         }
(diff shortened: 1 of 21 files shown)


geo/src/algorithm/contains.rs
```diff
@@ -404,18 +404,19 @@ mod test {
...
-            (0.0, 0.0),
+            [0.5, 0.5],
+            [1.5, 0.5],
+            [1.5, 1.5],
+            [0.0, 1.5],
+            [0.0, 0.0],
         ]);
         let poly = Polygon::new(linestring, vec![inner_linestring]);
-        assert!(poly.contains(&Point::new(0.25, 0.25)));
+        assert!(!poly.contains(&Point::new(0.25, 0.25)));
```
frewsxcv: Confusingly, as a result of closing the ring, this went from `true` to `false` 🤔

frewsxcv:
> Won't we need an (optional) validation step as part of this? It's relatively straightforward to inadvertently create an invalid geometry (e.g. using https://gist.github.com/urschrei/aa95f3681a09ff1a9d82935ff82ae9c4 as the exterior ring). Not validating isn't strictly worse than what we do now, but auto-closing could be seen as implying validity…

Oh for sure, we still need to address the validity issue. It's certainly still possible to create invalid geometries with this crate. For instance, in addition to your example, it's still possible to create a `LineString` with zero coordinates, which we shouldn't allow and would be easy to check.

To address your concerns, do you think it'd be sufficient to add a disclaimer in the docs warning that there are currently no geometric validity checks in-place?

urschrei:
> do you think it'd be sufficient to add a disclaimer in the docs warning that there are currently no geometric validity checks in-place?

Absolutely – being clear in the docs is all we can do anyway. It's a great feature, I didn't mean to imply we shouldn't land it!

PR #339, opened 2019-02-15
Geometrycollection ergonomics

rory:
just some minor ergonomic improvements

### DIFF
--- a/geo-types/src/geometry_collection.rs
+++ b/geo-types/src/geometry_collection.rs
@@ -29,12 +29,15 @@ impl<T: CoordinateType> GeometryCollection<T> {
 }
 
 
+/// Convert any Geometry (or anything that can be converted to a Geometry) into a
+/// GeometryCollection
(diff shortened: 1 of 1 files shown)


geo-types/src/geometry_collection.rs
```diff
@@ -11,12 +11,33 @@ pub struct GeometryCollection<T>(pub Vec<Geometry<T>>)
 where
     T: CoordinateType;
 
+impl<T: CoordinateType> GeometryCollection<T> {
+    /// Return an empty GeometryCollection
+    pub fn empty() -> GeometryCollection<T> {
```
frewsxcv:
```suggestion
    pub fn new() -> GeometryCollection<T> {
```

thoughts on calling this `new`? this matches other empty collection constructors in rust, e.g. `Vec::new`, `HashMap::new`

antoine-de: or implement `Default` ? even if I'm not sure it makes sense for a GeometryCollection

PR #347, opened 2019-02-20
Polygon::interiors_push - Add an interior to the current list

rory:
For a project, I am slowly creating polygons, and I needed a way to add an interior ring to a polygon, so I added `Polygon::interiors_push`.

I couldn't see how to do it with `Polygon::interiors_mut`.

### DIFF
--- a/geo-types/src/polygon.rs
+++ b/geo-types/src/polygon.rs
@@ -358,8 +358,9 @@ where
         }
     }
 
-    /// Add to the interiors
-    /// The new `LineString` [will be closed]:
+    /// Add an interior ring to the `Polygon`.
+    ///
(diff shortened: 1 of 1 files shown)


rory:
The rust-geo-booleanop library probably needs this functionality too:

https://github.com/21re/rust-geo-booleanop/blob/0776c7f4e27d3d60adad92cbdeaf6340b82e3360/lib/src/boolean/connect_edges.rs#L117

antoine-de:
Great catch, we [need](https://github.com/QwantResearch/osm_boundaries_utils_rs/blob/master/src/boundaries.rs#L125) this functionality too.

antoine-de:
great, thanks @frewsxcv !

PR #348, opened 2019-02-28
Implement discrete Frechet distance

BaptisteB-A:
This is a proposition of discrete Frechet distance implementation, based on Eiter and Mannila's [Computing Discrete Fréchet Distance](http://www.kr.tuwien.ac.at/staff/eiter/et-archive/cdtr9464.pdf)

### DIFF
--- a/geo-types/src/line_string.rs
+++ b/geo-types/src/line_string.rs
@@ -159,9 +159,9 @@ impl<T: CoordinateType> LineString<T> {
     ///
     /// let mut coords = vec![(0., 0.), (5., 0.), (7., 9.)];
     /// let line_string: LineString<f32> = coords.into_iter().collect();
-    /// assert_eq!(3, line_string.nb_coords());
+    /// assert_eq!(3, line_string.num_coords());
     /// ```
-    pub fn nb_coords(&self) -> usize {
+    pub fn num_coords(&self) -> usize {
         self.0.len()
     }
 }
--- a/geo-types/src/line_string.rs
+++ b/geo-types/src/line_string.rs
@@ -159,9 +159,9 @@ impl<T: CoordinateType> LineString<T> {
     ///
     /// let mut coords = vec![(0., 0.), (5., 0.), (7., 9.)];
     /// let line_string: LineString<f32> = coords.into_iter().collect();
-    /// assert_eq!(3, line_string.len());
+    /// assert_eq!(3, line_string.nb_coords());
     /// ```
-    pub fn len(&self) -> usize {
+    pub fn nb_coords(&self) -> usize {
         self.0.len()
     }
 }
--- a/geo-types/src/line_string.rs
+++ b/geo-types/src/line_string.rs
@@ -149,6 +149,21 @@ impl<T: CoordinateType> LineString<T> {
             }
         }
     }
+
+    /// Return the number of coordinates in the `LineString`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use geo_types::LineString;
+    ///
+    /// let mut coords = vec![(0., 0.), (5., 0.), (7., 9.)];
+    /// let line_string: LineString<f32> = coords.into_iter().collect();
+    /// assert_eq!(3, line_string.len());
+    /// ```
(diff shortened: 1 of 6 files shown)


geo/src/algorithm/frechet_distance.rs
```diff
@@ -0,0 +1,125 @@
...
+    ///
+    /// let ls_a = LineString::from(vec![(1., 1.), (2., 1.), (2., 2.), (3., 3.)]);
+    /// let ls_b = LineString::from(vec![(2., 2.), (0., 1.), (2., 4.), (3., 4.)]);
+    ///
+    /// let distance = ls_a.frechet_distance(&ls_b);
+    ///
+    /// assert_eq!(2., distance);
+    /// ```
+    ///
+    /// [frechet distance]: https://en.wikipedia.org/wiki/Fr%C3%A9chet_distance
```
antoine-de: can you put in the comment the research paper used for the implementation too ?

geo/src/algorithm/frechet_distance.rs
```diff
@@ -0,0 +1,125 @@
...
+use crate::{LineString, Point};
+use num_traits::{Float, FromPrimitive};
+
+/// Determine the similarity between two `LineStrings` using the [frechet distance].
+///
+/// [frechet distance]: https://en.wikipedia.org/wiki/Fr%C3%A9chet_distance
+pub trait FrechetDistance<T, Rhs = Self> {
+    /// Determine the similarity between two `LineStrings` using the [frechet distance].
+    ///
+    /// # Units
```
antoine-de: I'm not sure to understand this `# Units` section.

what do you mean ?

geo/src/algorithm/frechet_distance.rs
```diff
@@ -0,0 +1,125 @@
...
+        }
+    }
+}
+
+struct Data<T>
+where
+    T: Float + FromPrimitive,
+{
+    cache: Vec<Vec<T>>,
+    points_a: Vec<Point<T>>,
```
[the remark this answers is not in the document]
Tristramg: The problem is with linestring’s accessors: https://docs.rs/geo/0.12.0/geo/struct.LineString.html#methods

We need a random access to the points (in order to be able to use the cache), hence we can’t use `points_iter`, and `into_points` returns anyway a new Vec.

Do you have a better idea?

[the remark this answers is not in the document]
antoine-de: yes, I think we should add a random access to the linestring's point if it is needed.

geo-types/src/line_string.rs
```diff
@@ -149,6 +149,21 @@ impl<T: CoordinateType> LineString<T> {
...
+    /// # Examples
+    ///
+    /// ```
+    /// use geo_types::LineString;
+    ///
+    /// let mut coords = vec![(0., 0.), (5., 0.), (7., 9.)];
+    /// let line_string: LineString<f32> = coords.into_iter().collect();
+    /// assert_eq!(3, line_string.len());
+    /// ```
+    pub fn len(&self) -> usize {
```
antoine-de: hum I'm not sure about this name, wouldn't `nb_points` be more explicit ? The length of a linestring might mean different things.

BaptisteB-A: True, I will rename it to `nb_coords`. Thanks!

geo-types/src/line_string.rs
```diff
@@ -149,6 +149,21 @@ impl<T: CoordinateType> LineString<T> {
...
+    /// # Examples
+    ///
+    /// ```
+    /// use geo_types::LineString;
+    ///
+    /// let mut coords = vec![(0., 0.), (5., 0.), (7., 9.)];
+    /// let line_string: LineString<f32> = coords.into_iter().collect();
+    /// assert_eq!(3, line_string.nb_coords());
+    /// ```
+    pub fn nb_coords(&self) -> usize {
```
frewsxcv: thoughts on calling this `num_coords`? i personally haven't seen the `nb_` naming convention in the ecosystem before

frewsxcv:
bors retry

PR #351, opened 2019-04-01
Implement rstar::RTreeObject for Polygon

FliegendeWurst:


geo-types/src/polygon.rs
```diff
@@ -468,3 +468,43 @@ impl<T: CoordinateType> From<Rect<T>> for Polygon<T> {
...
+            }
+            if p.y < miny {
+                miny = p.y;
+            }
+            if p.y > maxy {
+                maxy = p.y;
+            }
+        }
+        ::rstar::AABB::from_corners(Point::new(minx, miny), Point::new(maxx, maxy))
+    }
```
frewsxcv: is this implementation different than doing this?

```rust
#[cfg(feature = "rstar")]
impl<T> ::rstar::RTreeObject for Polygon<T>
where
    T: ::num_traits::Float + ::rstar::RTreeNum,
{
    type Envelope = ::rstar::AABB<Point<T>>;

    fn envelope(&self) -> Self::Envelope {
        self.exterior.envelope()
    }
}
```

which would [reuse this implementation](https://github.com/georust/geo/blob/be288752844bfa96220cf030a985c0b6487c7c44/geo-types/src/line_string.rs#L192-L213)

FliegendeWurst: It would create the same envelope. Didn't know that it was implemented for LineString already.

urschrei:
bors try

urschrei:
bors retry

urschrei:
bors try

michaelkirk:
Hi @FliegendeWurst  - sorry this sat here for so long. Is it still something you're interesting in getting mergd? If so could you rebase and address any conflicts?

urschrei:
bors try

lnicola (approved):

lnicola:
Somewhat less related, but do we want the equivalent of #682 for `geo`?

urschrei:
> Somewhat less related, but do we want the equivalent of #682 for `geo`?

Yes I _think_ we do. @rmanoka WDYT?

geo-types/src/polygon.rs
```diff
@@ -549,3 +549,25 @@ impl<T: AbsDiffEq<Epsilon = T> + CoordNum> AbsDiffEq for Polygon<T> {
...
+}
+
+#[cfg(feature = "rstar")]
+impl<T> ::rstar::PointDistance for Polygon<T>
+where
+    T: ::num_traits::Float + ::rstar::RTreeNum,
+{
+    fn distance_2(&self, point: &Point<T>) -> T {
+        self.exterior.distance_2(point)
```
michaelkirk, line 571: I'm not sure how rstar uses these methods, but this seems incorrect.

A little perusing of the rstar codebase uncovered this usage:
```
    /// Returns `true` if a point is contained within this object.
    ///
    /// By default, any point returning a `distance_2` less than or equal to zero is considered to be
    /// contained within `self`. Changing this default behavior is advised if calculating the squared distance
    /// is more computationally expensive than a point containment check.
    fn contains_point(&self, point: &<Self::Envelope as Envelope>::Point) -> bool {
        self.distance_2(point) <= num_traits::zero()
    }
```

It seems surprising that a point within a polygon would return a >0 distance with the implementation in this PR.

It seems like it'd be nice to use the `EuclideanDistance` trait... but it's in geo.

urschrei: Urgh thanks for catching this @michaelkirk. Time for a EuclideanDistance in private_utils, AKA the module that dare not speak its name?

rmanoka:
> > Somewhat less related, but do we want the equivalent of #682 for `geo`?
> 
> Yes I _think_ we do. @rmanoka WDYT?

Yeah, that would be the most technically correct way to handle this.  If it's too messy, we could opt to just supporting the latest as `geo` is anyway a reasonably fast-moving crate at least as of now.

As it stands, the PR currently is fully within geo-types, so we should probably support both `rtree_0_9` and `rtree` features.  I don't mind moving it into geo to re-use an existing algo impl (as suggested by @urschrei @michaelkirk ).   It may be a bit confusing that some types have the traits in geo-types and the rest in `geo` but we can explain that it is natural due to the complexity of the impl.

geo-types/src/polygon.rs
urschrei: Doing this correctly involves duplicating:

- `Point`-`Polygon` euclidean distance
    - The `CoordinatePosition` logic for `Polygon,` and `CoordPos`

into `private_utils`, which is a **lot** of code. There has to be a better way of doing this.

michaelkirk: One alternative would be to move the integration onto rstar - adding a a new optional `geo` dependency.

Pros: 
- keeps geo-types small

Cons:
- one time breaking change for the geo-types rstar users.
- maybe annoying to update rstar with latest geo all the time since I think geo breaks more than rstar (but not really that often)

rmanoka: How about only implementing `RTreeObject` and not `PointDistance` ?   I guess the nearest-neighbor queries won't be available on `RTree<Polygon>` but that seems fair as it feels quite expensive to do the point-in-polygon computation for this query. 

We could provide the full impl. for simpler cases like `Rect`, so that users can get back nearest-neighbor functionality on polygon via `RTree<GeomWithData<Rect, Polygon>>`  and the `RTree::nearest_neighbor_iter` along with additional logic to then invoke geo algos to find the actual nearest neighbor.

frewsxcv: Instead of adding to `private_utils`, another option is to create a small `euclidean_distance` crate that `geo` depends on and that `geo-types` conditionally depends on if the `rstar` feature is enabled.

michaelkirk: That seems like a good compromise that doesn't break anything for anyone. Good idea @rmanoka.

urschrei: Those are both sensible ideas. In terms of long-term maintenance, I'm somewhat in favour of @frewsxcv's (to be clear I'm happy to start off with @rmanoka's, though)

rmanoka: Yeah what @frewsxcv suggests is the ideal way for handling private_utils in general, and the current issue.  Just to be clear, that new crate cannot use types from `geo-types` right?  Currently `private_utils` freely uses a lot of geo-types.

urschrei: > Just to be clear, that new crate cannot use types from geo-types right? Currently private_utils freely uses a lot of geo-types.

Hmm true. Euclidean distance also depends on a lot of _other_ algorithms (`Contains`, `Intersects`) and we can't import those without introducing a circular dependency, right?

michaelkirk:
This is a really old PR! 

It seems like the controversial part was removed in https://github.com/georust/geo/pull/351/commits/a61a2d4f2fbf97cb89ed62c4f11a732a061009dd, so it should be good to merge now.

Right? Anyone opposed? Otherwise I'll merge in a few days.

michaelkirk:
This is a really old PR! 

It seems like the controversial part was removed in https://github.com/georust/geo/pull/351/commits/a61a2d4f2fbf97cb89ed62c4f11a732a061009dd, so it should be good to merge now.

Right? Anyone opposed? Otherwise I'll merge in a few days.

frewsxcv (approved):
I'm surprised there's no merge conflicts with the changelog, because we've added a lot to the changelog since this was last modified

urschrei:
> This is a really old PR! 
> 
> 
> 
> It seems like the controversial part was removed in https://github.com/georust/geo/pull/351/commits/a61a2d4f2fbf97cb89ed62c4f11a732a061009dd, so it should be good to merge now.
> 
> 
> 
> Right? Anyone opposed? Otherwise I'll merge in a few days.

Yeah, the distance2 impl (and its implications) was the problem.

michaelkirk:
Ok, let's do it!

bors r+

michaelkirk:
Note that this had to be substantially reworked because we now support multiple versions of rstart. I've made those changes here: https://github.com/georust/geo/pull/827

I'm going to close this PR since it's superseded by the rework in #827.

PR #358, opened 2019-05-03
Remove macro usage from README

Mec-iS:
the old example is not working after interface changes


frewsxcv:
In what way does the example not work with the macros?

frewsxcv:
Ah I never published a new version of `geo` or `geo-types` that includes these macros. I'll fix this now.

frewsxcv:
Just published new versions of `geo` and `geo-types`. Reopen if something still doesn't work here!

Issue #359, opened 2019-05-06
Rect constructor should enforce (min.x < max.x) and (min.y < max.y)


frewsxcv:
See `Use case 1` in https://github.com/georust/geo/issues/354

rmanoka:
Please consider swapping the coords to enforce the constraint, rather than `panic`-ing if it is not true.

Use case: for instance, GDAL libraries use an inverted coord. system for the pixel vs world.  That is, pixel (0,0) is top-left of raster in accordance with usual CV notion, while the min in true world coords (such as long, lat or other projected coords) occurs at bottom-left of the raster.

It is intuitive to provide a point-wise transform function, which is expected to work on all shapes.  Unfortunately, it breaks down with `Rect` due to the aforementioned panic.

Can provide a PR if this makes sense to you.

michaelkirk:
Looks like this was resolved with #360

PR #360, opened 2019-05-07
Adding Rect constructor to address issue #359: 

Chopinsky:
To address the issue #359 of lacking a Rect geo-type constructor that can perform coordinates validations, I've made the following changes: 
1. Add a `new` constructor to the Rect geo-type.
2. Update all internal use of the Rect struct literals to the new constructor.

However, existing use of the Rect struct literals in unit tests are left alone.

### DIFF
--- a/geo-types/src/rect.rs
+++ b/geo-types/src/rect.rs
@@ -12,7 +12,13 @@ where
 }
 
 impl<T: CoordinateType> Rect<T> {
-    /// Creates a new rectangle.
+    /// Constructor to creates a new rectangle from coordinates, where `min` denotes to the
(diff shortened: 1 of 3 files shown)


geo-types/src/rect.rs
```diff
@@ -12,6 +12,35 @@ where
 }
 
 impl<T: CoordinateType> Rect<T> {
+    /// Creates a new rectangle.
+    ///
+    /// # Examples
```
frewsxcv: Would you mind adding a `# Panics` section above this documenting how this function can panic? Thanks!

Chopinsky: Just update the doc to mention the panic conditions.

Issue #353, opened 2019-04-12
Shortest path calculation algorithm


Bartoshko:
New `geo::algorithm` module to calculate shortest path between two given points based on given `MultiLineString` struct.
Proposal of implementation is to build it on two functions: 
1. First function is going to take `MultiLineString` and calculate (return) `VertexString` that will hold information about 'Line' connection between each Coordinate if present and children tree containing `Coordinate` it is connected to by `Line`  and calculated travel cost to this `Coordinate`. 

- This struct will be called Vertex:
`
struct Vertex {
    coordinates: Coordinate,
    graphs: Vec<GraphRelation>,
}
` 

- GraphRelation:
`
struct GraphRelation {
    vertex_index: usize,
    cost: f64,
}
`

- VertexString:
`
pub struct VertexString {
    cost_calc_type: Cost,
    vector: Vec<Vertex>,
}
`
- Cost enum 
`
enum Cost {
     EUCLIDEAN,
     HAVERSINE,
     ...,
}
`
- I have been implementing this here : [vertex idea](https://github.com/Bartoshko/path_navigator/blob/vertex/src/vertex.rs) I have had implemented only haversine formula.
- But I think your lib has mostly everything I would do and more. Maybe there will be possibility to add proposed feature to your lib.

- Travel cost can be implemented using any of available algorithm as there is an variety of them.

- distance calculation is than going to need to have passed struct `MultiLineString` that implements trait CostCalculation that is going to return f64 and. CostCalculation will contain all functions to calculate cost, passed enum variable determinate which function is going to be used.
- nice to have: recalculate distance cost, that will simply apply given function to each `Vertex` graph_relation recalculating all graph costs.

2. Second function is going to be Dijkstra shortest path algorithm that is taking `VertexString` and start `Coordinate`, finish `Coordinate` than returns `LineString` that represents shortest path.
Shortest Path:
- can be applied to any VertexString,
- do not know anything about travel cost,
- can be implemented for `VertexString` as `pub fn djikstra_shortest_path(&self, start: &Coordinate, stop: &Coordinate) -> LineString`

I would be glad to do pull request with proposed feature if You agree on the proposition.

frewsxcv:
Sounds good to me! I think this would be a great addition to the library! Sorry this response took so long 😬

Bartoshko:
Great. What do You think to implement Vertex first? And to write some test for this then, We will discuss implementation of shortest path algorithm ?

frewsxcv:
[...]

PR #366, opened 2019-06-04
Create VertexString data structure for shortest-path feature as discussed in issue #353

Chopinsky:
This PR creates the `VertexString`, `Vertex`, `GraphRelation` structs, and `Cost` enum types as the backbone data structure for the addition of the shortest-path algorithm to the algorithm set. APIs that facilitate the access of the internal data of these struct have also been included. 

The implementation is a close replica of the proposed structs with the same names as discussed in issue #353, with small improvement for various use cases. I've also added preliminary docs, examples, as well as unit tests.


frewsxcv:
Thanks for your patience! I'll look over this now...

geo-types/src/vertex_string.rs
```diff
@@ -0,0 +1,674 @@
...
+        Vertex {
+            coordinate,
+            id,
+            vertices: Vec::with_capacity(DEFAULT_SIZE),
+            closest_neighbor: None,
+        }
+    }
+
+    /// Get the vertex id, which also serves as the vector index in the VertexString
+    pub fn get_id(&self) -> usize {
```
frewsxcv: getters in rust shouldn't have a `get_` prefix

https://rust-lang-nursery.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

geo-types/src/vertex_string.rs
```diff
@@ -0,0 +1,674 @@
...
+    where
+        C: Iterator<Item = Line<T>>
+    {
+        VertexString::build(src, None)
+    }
+
+    /// Create a `VertexString` from the iterator of the edges connecting the vertices.
+    /// If the size of the vertex coordinates container is known, using this API will
+    /// help improve the overall performance.
+    pub fn new_with_size<C>(src: C, size: usize) -> Self
```
frewsxcv: this should be called `with_size`. see the 'General constructors' section [here](https://rust-lang-nursery.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case)

.gitignore
```diff
@@ -1,2 +1,3 @@
 **/target
 **/Cargo.lock
+.idea
```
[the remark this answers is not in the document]
Chopinsky: Sorry, I've overlooked this when committing. I will revert this change and restore the gitignore file.

geo-types/src/coordinate.rs
```diff
@@ -68,3 +69,12 @@ where
         (self.x, self.y)
     }
 }
+
+impl<T> fmt::Display for Coordinate<T>
```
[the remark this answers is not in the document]
Chopinsky: This was added mainly for hashing the coordinates (to a string, i.e. enabling the underlying `to_string` method) such that we can build up an index for the vertices, and retrieve them in O(1) time when queried. I'm not a super fan of this approach either, so I'm open to other workarounds. 

Alternatively, we can just loop over all the vertices to find an arbitrary vertex, though this may become a costly operation when dealing with a large graph.

Another approach would be hashing the [f64; 2] array, which is essentially what's behind the `Coordinate<T>` type. Yet we need to convert the inner coordinates to be `Float` or `f64` compatible.

geo-types/src/vertex_string.rs
```diff
@@ -0,0 +1,674 @@
...
+
+    // the cost of the edge connecting the vertex and this neighbor
+    cost: f64,
+}
+
+impl GraphRelation {
+    /// Build a graph relationship between two neighboring vertices and connect them. The source
+    /// vertex is where the `GraphRelation` is added to, and the target vertex's id is saved into
+    /// the struct. The cost is calculated and passed in as the known value.
+    pub fn new_with_cost<T: Float>(index: usize, cost: T) -> Self {
```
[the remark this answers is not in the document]
Chopinsky: I was trying to make the constructor more flexible, and I agree that we should probably ask developers to convert the coordinates such that they can handle exceptions better than a default `unwrap_or` method. I will change the generic type to `f64`.

geo-types/src/vertex_string.rs
Chopinsky: Yep, I will remove the `get_` prefix

geo-types/src/vertex_string.rs
Chopinsky: I'll update the constructor name to match the linked instructions.

Chopinsky:
I'm terribly sorry for the messed up commits ... just did a rebase and that seemed to bring up all previous commits as well ... Please discard any irrelevant old commits (anything between `f0968a1` and `4f68a7a`).

Chopinsky:
> Sorry that this has sat for so long. Is this still something you're interested in working on?

I'm not sure if this PR is still relevant, or if the code can compile anymore, I haven't worked on this for about a year. Please feel free to the close the PR if nobody else would be interested in taking it over.

michaelkirk:
Alright, I'm going to close this for now.

Issue #361, opened 2019-05-13
Bump rstar in geo-types to 0.4.0


frewsxcv:
Gonna hold off on this until we're ready to do a geo-types bump.

Stoeoef:
I've recently published 0.4.0. Only new features were added, hopefully no bug this time...
You should not see any breaking changes in regards to rstar's API surface.

urschrei:
Just trying to fix this, and getting a confusing error related to `RTreeObject`:

    the trait `rstar::Point` is not implemented for geo_types::Line<T>
    note: required because of the requirements on the impl of rstar::RTreeObject for geo_types::Line<T>
    note: required by rstar::RTree

I'm not sure why this is arising, since our `Point<T>` has an `rstar::Point` impl…

Stoeoef:
Just bumping the version information works for me ([comparison](https://github.com/georust/geo/compare/master...Stoeoef:master)), all tests pass and the benchmarks run on nightly. What did you do differently?

urschrei:
@Stoeoef I think I bumped the `geo_types` version without bumping the `geo` version 🤦‍♂️. Works fine here too.

PR #373, opened 2019-07-24
Bump rstar to 0.4.

frewsxcv:
Fixes https://github.com/georust/geo/issues/361


urschrei:
I think this is failing because `geo_types` needs the `rstar` version bump too

Issue #387, opened 2019-10-14
Replacing Failure with std::Error and a helper crate.


urschrei:
Using `Failure` was my (with the benefit of hindsight, premature) call, and it obviously hasn't worked out as intended. With the proliferation of helper crates that don't appear in the public API (snafu, thiserror), I think it would be best to go with that model, and present a `GeoError` enum that implements `std::Error`. This will also require a change to Geojson and one or two other crates, but not to worry…

frewsxcv:
https://github.com/georust/geo/pull/388

PR #388, opened 2019-10-26
Remove usages of 'failure' crate.

frewsxcv:
https://github.com/georust/geo/issues/387


frewsxcv:
> FWIW `Box<dyn Error>` is much worse in a library API than `failure::Error` because callers can't assume it is Send + Sync. If you are planning to stick with dyn Error, you at least want `Box<dyn Error + Send + Sync>` everywhere.

Great point! I'll adjust this when I'm back at my computer

PR #402, opened 2019-12-18
Simplify indices

urschrei:
This PR adds a new trait to the `simplify` and `simplifyVW` modules: `SimplifyIdx` and `SimplifyVwIdx`.
Both traits do the same thing: they run the respective simplification algorithm, but return the _indices_ of the simplified `LineString`, instead of the simplified geometry. This is mostly in response to ongoing requests from FFI consumers: because their use-cases don't necessarily map onto the Simple Features geometry model, returning geometries is less useful than returning indices, which can be used as mask values / input into native filter functions.

The only current drawback to these new traits is that because they return `Vec<usize>`, they're only implemented for `LineString`. If there's a more flexible way to do it, I'd love to extend it to some other geometries.

This change introduces some additional indirection for the RDP implementation, but `cargo bench` hasn't shown a perf hit…

### DIFF
--- a/geo-types/src/lib.rs
+++ b/geo-types/src/lib.rs
@@ -63,7 +63,8 @@ pub mod private_utils;
 #[macro_use]
 extern crate approx;
 
-mod test {
+#[cfg(test)]
+mod tests {
     use super::*;
 
     #[test]
--- a/geo/src/algorithm/simplify.rs
+++ b/geo/src/algorithm/simplify.rs
@@ -6,12 +6,12 @@ use num_traits::Float;
 // instead, we wrap a simple struct around index and point in a wrapper function,
 // passing that around instead, extracting either points or indices on the way back out
 #[derive(Copy, Clone)]
-struct RdpIndex<'a, T>
+struct RdpIndex<T>
 where
     T: Float,
 {
     index: usize,
-    point: &'a Point<T>,
+    point: Point<T>,
 }
 
 // Wrapper for the RDP algorithm, returning simplified points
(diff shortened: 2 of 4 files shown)


geo-types/src/lib.rs
```diff
@@ -63,7 +63,8 @@ pub mod private_utils;
 #[macro_use]
 extern crate approx;
 
-mod test {
+#[cfg(test)]
```
frewsxcv: 👏

geo/src/algorithm/simplify.rs
```diff
@@ -69,20 +117,74 @@ pub trait Simplify<T, Epsilon = T> {
         T: Float;
 }
 
+/// Simplifies a geometry, returning the retained _indices_ of the input.
+///
+/// This operation uses the [Ramer–Douglas–Peucker algorithm](https://en.wikipedia.org/wiki/Ramer–Douglas–Peucker_algorithm)
+/// and does not guarantee that the returned geometry is valid.
```
frewsxcv: thanks for adding docs!

geo/src/algorithm/simplify.rs
```diff
@@ -69,20 +117,74 @@ pub trait Simplify<T, Epsilon = T> {
...
+impl<T> SimplifyIdx<T> for LineString<T>
+where
+    T: Float,
+{
+    fn simplify_idx(&self, epsilon: &T) -> Vec<usize> {
+        rdp_indices(
+            &self
+                .clone()
+                .into_points()
+                .iter()
```
frewsxcv: instead of `self.clone().into_points().iter()`, you can avoid the memory allocation by doing `self.points_iter()`

urschrei: ✨

PR #403, opened 2019-12-21
Doc updates

urschrei:
(To be merged after #402)

This PR adds top-level docs (mostly the same as what's currently in the README), and the following changes:

- Revised `LineString` docs
- Revised `Point` docs
- Revised module docs to better distinguish `closest_point` and `euclidean_distance`
- Revised `chamberlain_duquette` docs
- Revised `haversine_destination` module doc
- Note existence of some non-euclidean distance algorithms
- Detail optional features
- Make GeoJSON interop more obvious. This is somewhat redundant, since the GeoJSON crate says the same thing, but we don't know what crate people will see first, so I think it's OK
- Hidden the `PointsIter` struct from docs.rs

### DIFF
--- a/geo-types/src/lib.rs
+++ b/geo-types/src/lib.rs
@@ -1,3 +1,11 @@
+#![doc(html_logo_url = "https://raw.githubusercontent.com/georust/meta/master/logo/logo.png")]
+//! The `geo-types` library provides geospatial primitive types and traits to the [`GeoRust`](https://github.com/georust)
+//! crate ecosystem.
+//!
+//! In most cases, you will only need to use this crate if you're a crate author and want compatibility
+//! with other `GeoRust` crates. Otherwise, the [`geo`](https://crates.io/crates/geo) crate re-exports these types and
+//! provides geospatial algorithms, while the [`geojson`](https://crates.io/crates/geojson) crate allows serialising
+//! and de-serialising `geo-types` primitives to GeoJSON.
 extern crate num_traits;
 
 #[cfg(feature = "serde")]
--- a/geo-types/src/line_string.rs
+++ b/geo-types/src/line_string.rs
@@ -66,12 +66,28 @@ use std::ops::{Index, IndexMut};
 /// }
 /// ```
 ///
+/// You can also iterate over the coordinates in the `LineString` as `Point`s:
+///
+/// ```
+/// use geo_types::{LineString, Coordinate};
+///
+/// let line_string = LineString(vec![
+///     Coordinate { x: 0., y: 0. },
+///     Coordinate { x: 10., y: 0. },
+/// ]);
+///
+/// for point in line_string.points_iter() {
+///     println!("Point x = {}, y = {}", point.x(), point.y());
+/// }
+/// ```
+
 #[derive(PartialEq, Clone, Debug, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct LineString<T>(pub Vec<Coordinate<T>>)
 where
     T: CoordinateType;
 
+/// A `Point` iterator returned by the `points_iter` method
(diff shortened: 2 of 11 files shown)


geo-types/src/lib.rs
```diff
@@ -1,3 +1,11 @@
+#![doc(html_logo_url = "https://avatars1.githubusercontent.com/u/10320338?s=1000&v=4")]
+//! The `geo-types` library provides geospatial primitive types and traits to the [`Georust`](https://github.com/georust)
```
frewsxcv:
```suggestion
//! The `geo-types` library provides geospatial primitive types and traits to the [`GeoRust`](https://github.com/georust)
```

geo-types/src/lib.rs
```diff
@@ -1,3 +1,11 @@
...
+//! crate ecosystem.
+//!
+//! In most cases, you will only need to use this crate if you're a crate author and want compatibility
+//! with other `Georust` crates. Otherwise, the [`geo`](https://crates.io/crates/geo) crate re-exports these types and
```
frewsxcv:
```suggestion
//! with other `GeoRust` crates. Otherwise, the [`geo`](https://crates.io/crates/geo) crate re-exports these types and
```

geo-types/src/lib.rs
```diff
@@ -1,3 +1,11 @@
+#![doc(html_logo_url = "https://avatars1.githubusercontent.com/u/10320338?s=1000&v=4")]
```
frewsxcv: Would it also work to use this url? https://raw.githubusercontent.com/georust/meta/master/logo/logo.png This might be better since we can swap it out at any point without breaking old docs

geo-types/src/line_string.rs
```diff
@@ -66,12 +66,28 @@ use std::ops::{Index, IndexMut};
...
+/// }
+/// ```
+
 #[derive(PartialEq, Clone, Debug, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct LineString<T>(pub Vec<Coordinate<T>>)
 where
     T: CoordinateType;
 
+#[doc(hidden)]
```
frewsxcv: Any reason for hiding this?

geo/src/lib.rs
```diff
@@ -1,3 +1,42 @@
...
+//! - `use-proj`: enable coordinate conversion and transformation of `Point` geometries using the [`proj`](https://docs.rs/proj) crate
+//! - `use-serde`: enable serialisation of geometries using `serde`.
+//!
+//! ## GeoJSON
+//! If you wish to read or write `GeoJSON`, use the [`geojson`](https://docs.rs/geojson) crate, with the `geo-types` feature activated.
+//! This provides fallible conversions **to** `geo-types` primitives such as `Point` and `Polygon` from `geojson` `Value`
+//! structs using the standard [`TryFrom`](https://doc.rust-lang.org/stable/std/convert/trait.TryFrom.html) 
+//! and [`TryInto`](https://doc.rust-lang.org/stable/std/convert/trait.TryInto.html) traits,
+//! and conversion **from** `geo-types` primitives to `geojson`
+//! `Value` structs using the [`From`](https://doc.rust-lang.org/stable/std/convert/trait.TryFrom.html) trait.
```
frewsxcv: 👏

geo-types/src/lib.rs
urschrei: 👍

geo-types/src/line_string.rs
urschrei: No, just couldn't figure out a good way to explain what it was in the presence of `points_iter`, but got it now, I think.

urschrei:
Changes addressed, commit squashed!

PR #407, opened 2019-12-28
Remove trait object arguments for MapCoord traits.

frewsxcv:
Now callers don't need to pass a reference to a closure, they can just
pass the closure. No performance difference.


frewsxcv:
yay for ci catching a bug!

PR #410, opened 2020-01-03
Update Geometry enum and GeometryCollection

urschrei:
This PR:

- Adds non-consuming iteration functionality (`iter()`, `iter_mut()`) and indexing to `GeometryCollection`
- Adds `TryFrom` impls from `Geometry` enums to `Geo` primitives (deprecating but not removing the previous functions)
- Doctest examples of working with `GeometryCollection`

### DIFF
--- a/geo-types/src/geometry.rs
+++ b/geo-types/src/geometry.rs
@@ -86,7 +86,7 @@ impl<T: CoordinateType> Geometry<T> {
     /// assert_eq!(p2, Point::new(0., 0.,));
     /// ```
     #[deprecated(
-        note = "Will be removed in an upcoming version. Switch to std::convert::TryFrom<Point>"
+        note = "Will be removed in an upcoming version. Switch to std::convert::TryInto<Point>"
     )]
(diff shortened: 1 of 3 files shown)


geo-types/src/geometry.rs
```diff
@@ -69,6 +85,9 @@ impl<T: CoordinateType> Geometry<T> {
     /// let p2: Point<f32> = g.into_point().unwrap();
     /// assert_eq!(p2, Point::new(0., 0.,));
     /// ```
+    #[deprecated(
+        note = "Will be removed in an upcoming version. Switch to std::convert::TryFrom<Point>"
```
frewsxcv: shouldn't this, and all the following doc comments, be `std::convert::TryInto<...>`?

```suggestion
        note = "Will be removed in an upcoming version. Switch to std::convert::TryInto<Point>"
```

urschrei: Oops. Yep!

urschrei:
bors retry

Issue #390, opened 2019-10-31
Triangle and Rect are missing from enum Geometry<T>


guidorice:
I noticed that Triangle and Rect are missing from enum Geometry<T> in geo-types/src/geometry.rs. I could add them? This might make a good PR for me, being new to Rust. 

Relates to #119 
Relates to #4

guidorice:
I think I could get it to work by adding Geometry::Rect and Geometry::Triangle to the match statement in src/algorthm/to_postgis.rs:

```rust
impl ToPostgis<ewkb::Geometry> for Geometry<f64> 
```

and convert Triangles and Rects to ewkb::GeometryT::Polygon() on the fly. 

However, that seems like a bit of a hack, especially for Triangle which is actually part of the OGC Geometry class heirarchy, and it looks like PostGIS recently added it.  The postgis crate doesn't support Triangle yet, it appears.

Seems like better to hold off, considering #119 .

guidorice:
@frewsxcv cool I am looking at your proof-of-concept PR. I see now that

In [to_postgis.rs](https://github.com/georust/geo/blob/8bf248e37d85e3a2f544ddc20e9cc1ccd0f6c2b5/geo/src/algorithm/to_postgis.rs#L91-L96) I see you convert `to_polygon` instead of `Rect` or `Triangle`. 

I understand now that converting to/from Polygon is commonly how things are done when converting types to/from PostGIS. Please ignore what I wrote above and close this whenever you finish your PR #421 . Thanks much!

frewsxcv:
Opened a PR in https://github.com/georust/geo/pull/432

PR #421, opened 2020-03-03
Proof of concept – Improved ergonomics by allowing algorithms to be called on non-geo types

frewsxcv:
This pull request does a few things:

- Adds `Triangle` and `Rect` to the `Geometry` enum (https://github.com/georust/geo/issues/390)
- Introduces a new `GeometryCow` enum, that's like `Geometry`, but each variant has a `Cow` internally 🐮 
- Redefines the `Area` algorithm to be implemented for any type that implements `Into<GeometryCow>`

The benefit of all of this, is it means any user can implement `Into<GeometryCow>` for their types, and the types will have access to the `Area` algorithm (and other algorithms in the future). e.g.

```rust
use geo::algorithm::Area;

let geojson = geojson_str.parse::<GeoJson>().unwrap();
let area = geojson.area();
```

This improves the ergonomics since the user doesn't need to explicitly convert via geo-types. Though if a user does `geojson.area()` twice, it means they're unnecessarily converting to the intermediary geo-type `GeometryCow` twice. It would be more performant in that scenario for the user to explicitly convert to a geo-type, and _then_ call `area` on that.

Thoughts? If we think this is an improvement, I can break each bullet point above into a separate pull request.

### DIFF
--- a/geo/src/algorithm/map_coords.rs
+++ b/geo/src/algorithm/map_coords.rs
@@ -539,8 +539,8 @@ impl<T: CoordinateType, NT: CoordinateType> TryMapCoords<T, NT> for Triangle<T> 
         func: impl Fn(&(T, T)) -> Result<(NT, NT), Box<dyn Error + Send + Sync>>,
     ) -> Result<Self::Output, Box<dyn Error + Send + Sync>> {
         let p1 = func(&self.0.x_y())?;
-        let p2 = func(&self.0.x_y())?;
-        let p3 = func(&self.0.x_y())?;
+        let p2 = func(&self.1.x_y())?;
+        let p3 = func(&self.2.x_y())?;
 
         Ok(Triangle(
             Coordinate { x: p1.0, y: p1.1 },
--- a/geo/src/algorithm/map_coords.rs
+++ b/geo/src/algorithm/map_coords.rs
@@ -520,8 +520,8 @@ impl<T: CoordinateType, NT: CoordinateType> MapCoords<T, NT> for Triangle<T> {
 
     fn map_coords(&self, func: impl Fn(&(T, T)) -> (NT, NT) + Copy) -> Self::Output {
         let p1 = func(&self.0.x_y());
-        let p2 = func(&self.0.x_y());
-        let p3 = func(&self.0.x_y());
+        let p2 = func(&self.1.x_y());
+        let p3 = func(&self.2.x_y());
 
         Triangle(
             Coordinate { x: p1.0, y: p1.1 },
--- a/geo/src/algorithm/map_coords.rs
+++ b/geo/src/algorithm/map_coords.rs
@@ -1,6 +1,6 @@
 use crate::{
     Coordinate, CoordinateType, Geometry, GeometryCollection, Line, LineString, MultiLineString,
-    MultiPoint, MultiPolygon, Point, Polygon, Rect, Triangle
+    MultiPoint, MultiPolygon, Point, Polygon, Rect, Triangle,
 };
 use std::error::Error;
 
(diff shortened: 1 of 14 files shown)


urschrei:
I really like this idea – we'll need to make sure that it's both well-documented and advertised for people who want to integrate it, though.

guidorice:
@frewsxcv @urschrei  this looks awesome! I am going to look into the new changes, and then I will update and/or close #390 accordingly.

geo/src/algorithm/map_coords.rs
```diff
@@ -509,6 +515,57 @@ impl<T: CoordinateType> MapCoordsInplace<T> for Rect<T> {
...
+impl<T: CoordinateType, NT: CoordinateType> TryMapCoords<T, NT> for Triangle<T> {
+    type Output = Triangle<NT>;
+
+    fn try_map_coords(
+        &self,
+        func: impl Fn(&(T, T)) -> Result<(NT, NT), Box<dyn Error + Send + Sync>>,
+    ) -> Result<Self::Output, Box<dyn Error + Send + Sync>> {
+        let p1 = func(&self.0.x_y())?;
+        let p2 = func(&self.0.x_y())?;
+        let p3 = func(&self.0.x_y())?;
```
michaelkirk:
```suggestion
        let p1 = func(&self.0.x_y())?;
        let p2 = func(&self.1.x_y())?;
        let p3 = func(&self.2.x_y())?;
```

geo/src/algorithm/map_coords.rs
```diff
@@ -509,6 +515,57 @@ impl<T: CoordinateType> MapCoordsInplace<T> for Rect<T> {
...
+            Coordinate { x: p3.0, y: p3.1 },
+        ))
+    }
+}
+
+impl<T: CoordinateType> MapCoordsInplace<T> for Triangle<T> {
+    fn map_coords_inplace(&mut self, func: impl Fn(&(T, T)) -> (T, T)) {
+        let p1 = func(&self.0.x_y());
+        let p2 = func(&self.0.x_y());
+        let p3 = func(&self.0.x_y());
```
michaelkirk:
```suggestion
        let p1 = func(&self.0.x_y());
        let p2 = func(&self.1.x_y());
        let p3 = func(&self.2.x_y());
```

geo/src/algorithm/map_coords.rs
```diff
@@ -509,6 +515,57 @@ impl<T: CoordinateType> MapCoordsInplace<T> for Rect<T> {
     }
 }
 
+impl<T: CoordinateType, NT: CoordinateType> MapCoords<T, NT> for Triangle<T> {
+    type Output = Triangle<NT>;
+
+    fn map_coords(&self, func: impl Fn(&(T, T)) -> (NT, NT) + Copy) -> Self::Output {
+        let p1 = func(&self.0.x_y());
+        let p2 = func(&self.0.x_y());
+        let p3 = func(&self.0.x_y());
```
michaelkirk:
```suggestion
        let p1 = func(&self.0.x_y());
        let p2 = func(&self.1.x_y());
        let p3 = func(&self.2.x_y());
```

frewsxcv: tyty

michaelkirk:
I'm considering reviving this functionality for geo in pursuit of some of the geometry graph work.

Any lessons learned or reasons I shouldn't @frewsxcv?

frewsxcv:
The only thing I remember is that it felt like too much was happening behind the scenes. If someone were to implement `Into<GeometryCow>` for their geometry type, everytime they'd do a geo algorithm call there would most likely be a non-trivial conversion happening behind the scenes. Compared to if a user implement `CoordsIter` for their type, no conversion happens when they call a geo algorithm.

michaelkirk:
FYI: this is what I'm doing with the COW stuff for my geometry graph work - it's actually pretty different in motivation from this PR:

https://github.com/georust/geo/commit/c7de1ffa05b87b9148917085779e6f57d6fe15c4

For now I've not made Cow public. I haven't found a use for it outside of the geometry graph, so I didn't want to surface it. If people like the idea, I'd be happy to make it public.

frewsxcv:
Just took a look and a couple thoughts came to mind:

- If the inner type is no longer a `Cow`, maybe `GeometryRef` is a more appropriate name? 🤷🏻 
- I wonder if having `Geometry` _and_ `GeometryCow`/`GeometryRef` is confusing in the context of `GeometryCollection` which would contain the former but not the latter

Only spent a few minutes looking at your commit above, but will try to look again with some more thorough thoughts

PR #436, opened 2020-04-01
fix rect intersection

michaelkirk:
Addendum to https://github.com/georust/geo/pull/420

This fixes the intersection test for Rect when `self` is contained
wholly within `arg` - it was already working for when `arg` was within
`self` as of #420

### DIFF
--- a/geo/src/algorithm/intersects.rs
+++ b/geo/src/algorithm/intersects.rs
@@ -211,32 +211,27 @@ where
     T: Float,
 {
     fn intersects(&self, bounding_rect: &Rect<T>) -> bool {
-        // line intersects inner or outer polygon edge
-        if bounding_rect.contains(self) {
-            false
-        } else {
-            let x_overlap = value_in_range(
-                self.min().x,
-                bounding_rect.min().x,
-                bounding_rect.min().x + bounding_rect.width(),
-            ) || value_in_range(
-                bounding_rect.min().x,
-                self.min().x,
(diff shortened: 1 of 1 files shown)


geo/src/algorithm/intersects.rs
```diff
@@ -211,32 +211,27 @@ where
...
-                self.min().y,
-                self.min().y + self.height(),
-            );
-
-            x_overlap && y_overlap
-        }
+        let x_overlap = value_in_range(
+            self.min().x,
+            bounding_rect.min().x,
+            bounding_rect.max().x
```
michaelkirk, line 217: Simplified this math a bit.

width is computed as  `self.max().x - self.min().x`

So this: 
`self.min().x + self.width()`

previously expanded as:
`self.min().x + self.max().x - self.min().x`

Which is more simply `self.max().x`

geo/src/algorithm/intersects.rs
```diff
@@ -211,32 +211,27 @@ where
     T: Float,
 {
     fn intersects(&self, bounding_rect: &Rect<T>) -> bool {
-        // line intersects inner or outer polygon edge
-        if bounding_rect.contains(self) {
-            false
```
michaelkirk, line 216: This contains check is backwards. If `a` contains `b` then we should always consider the rects to intersect.

We could fix it by changing this to return true, but it's wholly redundant with the subsequent check for overlap, so I removed it altogether.

PR #437, opened 2020-04-01
Change COORD_PRECISION from 1e-1 to std::f32::EPSILON.

### LINKED ISSUE #392: question about COORD_PRECISION vs std::f32::EPSILON
```rust
pub static COORD_PRECISION: f32 = 1e-1; // 0.1m
```

Consider changing to `std::f32::EPSILON` instead? For example, if the coordinates are decimal degrees then 0.1 difference in longitude at the equator is 11 kilometers (Haversine distance). That's an arbitrary distance, and actually it is being used by at least one algorithm:

```rust
pub fn point_contains_point<T>(p1: Point<T>, p2: Point<T>) -> bool
where
    T: Float,
{
    line_euclidean_length(Line::new(p1, p2)).to_f32().unwrap() < COORD_PRECISION
}
```

guidorice:
Since #392 has been open a while, and it's a 1 line change, I thought I would open this PR. Please let me know if I am misunderstanding why the `0.1m` value was hard-coded.

Fixes #392

### DIFF
--- a/geo-types/src/private_utils.rs
+++ b/geo-types/src/private_utils.rs
@@ -6,7 +6,7 @@
 use crate::{Coordinate, CoordinateType, Line, LineString, Point, Rect};
 use num_traits::Float;
 
-pub static COORD_PRECISION: f32 = 1e-1; // 0.1m
+pub static COORD_PRECISION: f32 = std::f32::EPSILON;
 
 pub fn line_string_bounding_rect<T>(line_string: &LineString<T>) -> Option<Rect<T>>
 where
--- a/geo/src/algorithm/euclidean_distance.rs
+++ b/geo/src/algorithm/euclidean_distance.rs
@@ -28,12 +28,12 @@ pub trait EuclideanDistance<T, Rhs = Self> {
     ///
     /// ```
     /// use geo::algorithm::euclidean_distance::EuclideanDistance;
-    /// use geo::{LineString, Point, Polygon, COORD_PRECISION};
+    /// use geo::{LineString, Point, Polygon};
     ///
     /// // Point to Point example
     /// let p = Point::new(-72.1235, 42.3521);
     /// let dist = p.euclidean_distance(&Point::new(-72.1260, 42.45));
-    /// assert!(dist < COORD_PRECISION);
+    /// assert!(dist < 0.1);
     ///
     /// // Point to Polygon example
     /// let points = vec![


guidorice:
@frewsxcv yes that is a good suggestion! Let's hold off merging this, and I will see if I can implement that in an alternate branch which would

- remove COORD_PRECISION
- replace it's usages with Float::epsilon and using T: Float

michaelkirk:
>  i wonder if we should be using Float::epsilon instead 

That would be a bit tricky to do generically since COOD_PRECISION is a static.

Stepping back a bit, what is `COORD_PRECISION`? I'm having a hard time making sense of a what a universal "precision" parameter could mean (beyond the size of float the float we're using: e.g. f32 vs f64). 

Different algorithms have inherently different precision behavior, and they can't all be controlled by a single var. 

Currently `COORD_PRECISION` is used in exactly one place in production code:

```
pub fn point_contains_point<T>(p1: Point<T>, p2: Point<T>) -> bool
where
    T: Float,
{
    line_euclidean_length(Line::new(p1, p2)).to_f32().unwrap() < COORD_PRECISION
}
```

The only other place it is used is in a test.

I think we should:
- remove the static
- inline the usage of T::epsilon in the two places it's currently used
- individual methods should document their own unique precision characteristics, since a universal "precision" parameter is not possible.

Does that make sense? @guidorice do you want to make that change? Otherwise I can follow up with a PR.

guidorice:
@michaelkirk  I agree with the sentiment to step back and ask what is this COORD_PRECISION to begin with! The steps you wrote sound good to me, and I will be glad to make these changes.

> remove the static
> inline the usage of T::epsilon in the two places it's currently used

I am not sure about this last one:

> individual methods should document their own unique precision characteristics, since a universal "precision" parameter is not possible

This could be an ongoing effort or a specific enhancement issue?

cc @frewsxcv

michaelkirk:
> I am not sure about this last one:
>> individual methods should document their own unique precision characteristics, since a universal "precision" parameter is not possible
> This could be an ongoing effort or a specific enhancement issue?

Sorry to be unclear, I meant that "in general" algorithms should be documenting their own precision, since a global notion of "precision" is not possible.

I did not mean to imply that you should go around and document every existing method as part of this PR. =)

frewsxcv:
Completely agree with everything both of you just said, and agree we should remove `COORD_PRECISION` ✅ . I honestly forgot how it came to be, and the historical context, if any

guidorice:
Closing because of decision to remove COORD_PRECISION. I will open a separate PR for that. 👍

PR #438, opened 2020-04-02
fix vincenty for equatorial and coincident points

### LINKED ISSUE #418: Fail on vincenty distance
I am not expert in vincenty distance at all. And I was just testing how this functionality works, So I got a bug already on my second test. Code:

    let point1: Point<f64> = Point::new(0.0, 0.0);
    let point2: Point<f64> = Point::new(0.001, 0.0);
    let distance = point1.vincenty_distance(&point2).unwrap();

giving equal points will also give same error

    let point1: Point<f64> = Point::new(0.51, 88.1);
    let point2: Point<f64> = Point::new(0.51, 88.1);

The error:

    thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: FailedToConvergeError'

Looks like it panics when distance is 0?
I am not sure, but looks like it is wrong behavior.
Forward thank you!

    geo = "0.12.2"
    rustc 1.43.0-nightly

michaelkirk:
fixes #418 

From Algorithms for geodesics, Charles F. F. Karney∗
available https://arxiv.org/pdf/1109.4448.pdf

> For equatorial geodesics (φ1 = 0 and α1 = 1 2π), Eq. (11) is
> indeterminate; in this case, take σ1 = 0

Also fixes vincenty for coincident points.

### DIFF
--- a/geo/src/algorithm/vincenty_distance.rs
+++ b/geo/src/algorithm/vincenty_distance.rs
@@ -92,9 +92,17 @@ where
                 + (cosU1 * sinU2 - sinU1 * cosU2 * cosLambda)
                     * (cosU1 * sinU2 - sinU1 * cosU2 * cosLambda))
                 .sqrt();
+
             if sinSigma.is_zero() {
-                return Err(FailedToConvergeError);
+                if self == rhs {
+                    // coincident points
+                    return Ok(T::zero());
+                } else {
+                    // anitpodal points, for which vincenty does not converge
+                    return Err(FailedToConvergeError);
+                }
             }
+
             cosSigma = sinU1 * sinU2 + cosU1 * cosU2 * cosLambda;
             sigma = sinSigma.atan2(cosSigma);
             let sinAlpha = cosU1 * cosU2 * sinLambda / sinSigma;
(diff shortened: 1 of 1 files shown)


geo/src/algorithm/vincenty_distance.rs
```diff
@@ -196,4 +211,33 @@ mod test {
...
         );
     }
+
+    #[test]
+    fn test_vincenty_distance_equatorial() {
+        let a = Point::<f64>::new(0.0, 0.0);
+        let b = Point::<f64>::new(100.0, 0.0);
+        assert_relative_eq!(
+            a.vincenty_distance(&b).unwrap(),
+            11131949.079,
```
michaelkirk, line 221: This value matches what comes out of geopy.vincenty 1.21.0

urschrei:
lgtm! Could you leave a link to the PDF with the updated algorithm in the source, in case someone needs to go looking for it in the future?

michaelkirk:
> lgtm! Could you leave a link to the PDF with the updated algorithm in the source, in case someone needs to go looking for it in the future?

Added! 

Please take another look @urschrei

urschrei:
Thanks!

Issue #425, opened 2020-03-10
FailedToConvergeError for vincenty_distance({0,0},{180,0})


brenzi:
Vincenty fails ungracefully when asked to compute the distance between opposite aequatorial points:
{0,0},{180,0}

What I get is `FailedToConvergeError`. 
I think the function shouldn't panic in this case but just return a predefined value for that case.

urschrei:
So if I understand correctly, Vincenty can fail for antipodal points (anything over ~19,936 km). The issue here is that we could provide the correct distance for your particular case, but not for all nearly-antipodal points, since that would require an additional distance calculation, and the whole point of Vincenty is that it's very cheap to calculate. I wonder whether it would be enough to check for _strictly_ antipodal points, which is [fast](https://mathworld.wolfram.com/AntipodalPoints.html) and then simply return the maximum possible Vincenty value.

brenzi:
In that case I guess I can just apply error handling in my crate ;-) For such large distances I don't care about accuracy.

frewsxcv:
Can anyone else confirm that other Vincenty implementations also error with these coordinates?

michaelkirk:
I think there are two issues... we weren't correctly special casing equatorial points, so any equatorial points would fail to converge.  I opened a PR for that here: https://github.com/georust/geo/pull/438

But the antipodal thing, as I understand it, is strictly a limitation of vincenty. caveat is that I only just started studying vincenty a few hours ago, so i could be off base. 

In my short study, it seems like there are other options that don't have these shortcomings. In fact geopy has deprecated their vincenty implementation in favor of [geographiclib](https://geographiclib.sourceforge.io/html/python/index.html), which I believe calculates distance based on an implementation of the methods described in https://arxiv.org/pdf/1109.4448.pdf

michaelkirk:
> In my short study, it seems like there are other options that don't have these shortcomings

related: https://github.com/georust/geo/issues/234

michaelkirk:
> I think there are two issues... we weren't correctly special casing equatorial points, so any equatorial points would fail to converge. I opened a PR for that here: #438

Fixes here were merged in https://github.com/georust/geo/pull/438

> But the antipodal thing, as I understand it, is strictly a limitation of vincenty. 

In #440 we merged an implementation of the geodesic computation for distance, which, unlike vincenty, always converges.

From master, you can use it like this:
```rust
use geo::prelude::*;
use geo::Point;

// New York City
[...]

Issue #234, opened 2018-05-19
Add geodesic distance/length algorithm


frewsxcv:
It seems to be more accurate than Vincenty and always converges: https://github.com/geopy/geopy/pull/144

Psykopear:
I'm working on this.
I have the Inverse algorithm implemented, and it passes all the test cases found on the python implementation of geographiclib (I used that as a reference since it looked more readable to me).
At the moment it is a 1:1 implementation, so lots of not idiomatic code, but it should be easier to debug the port.

I'll publish it on github and link it here once a I have a working implementation of at least the "Direct" algorithm too, so you guys can take a look at it.

frewsxcv:
Awesome! Looking forward to it!

Psykopear:
Here is the [repo](https://github.com/Psykopear/geodesic-rs).

Considerations:
- I don't know what half of the code does (all the calculations), I just ported it and made it work
- The repo contains the reference python implementation, and also depends on an external crate that binds a C implementation of the algorithm,  so I can run benchmarks against it

This port has two problems:
1) As I said, the code isn't idiomatic, it's mostly a 1:1 port.
2) The implementation seems to be on average 3 times slower than bindings to a C implementation I found [here](https://github.com/savage13/geographiclib) (you can try it by following the README)


About 1) the code could easily be made more idiomatic, but losing the naming might make porting the rest of the features, or future updates, harder.

As for 2) performance wise, there are a lot of low hanging fruit that would be solved by just refactoring the code for the first point, using right data structures, avoiding unnecessary clones ecc.

Anyway it's a starting point, a big part of the dirty work is done.

That said, I did this port as an exercise, and I'm not going to spend much more time over it (at least in the near future), so if anyone is willing to take it from here, the code is there for everyone to use, and I can answer questions about what I did.

michaelkirk:
Hi @Psykopear - thanks for publishing that transcription!

I've forked it here: 
You can see my WIP here: https://github.com/Psykopear/geodesic-rs/compare/master...michaelkirk:mkirk/correctness?expand=1

Beyond the unit tests you've implemented, I wanted to further validate the behavior using the test dataset published here: https://geographiclib.sourceforge.io/html/geodesic.html#testgeod

I built a test harness to do that here: https://github.com/michaelkirk/validate_geographiclib
Which compares results against the reference cpp and python implementations.

This surfaced some correctness issues in the rust transcription, addressed in my fork linked above.

[...]

PR #440, opened 2020-04-23
geodesic distance

michaelkirk:
Integrates with the geographiclib rust port started by @Psykopear and touched up by me. 

FIXES https://github.com/georust/geo/issues/234
FIXES https://github.com/georust/geo/issues/425

### DIFF
--- a/geo/Cargo.toml
+++ b/geo/Cargo.toml
@@ -18,7 +18,7 @@ travis-ci = { repository = "georust/geo" }
 num-traits = "0.2"
 serde = { version = "1.0", optional = true, features = ["derive"] }
 rstar = { version = "0.7" }
-geographiclib-rs = { version = "0.1", git = "https://github.com/michaelkirk/geographiclib-rs" }
+geographiclib-rs = { version = "0.2" }
 
 postgis = { version = "0.6", optional = true }
 proj = { version = "0.15.1", optional = true }
--- a/geo/Cargo.toml
+++ b/geo/Cargo.toml
@@ -18,6 +18,8 @@ travis-ci = { repository = "georust/geo" }
 num-traits = "0.2"
 serde = { version = "1.0", optional = true, features = ["derive"] }
 rstar = { version = "0.7" }
+geographiclib-rs = { version = "0.1", git = "https://github.com/michaelkirk/geographiclib-rs" }
+
 postgis = { version = "0.6", optional = true }
 proj = { version = "0.15.1", optional = true }
 
@@ -45,6 +47,10 @@ harness = false
 
 [[bench]]
 name = "vincenty_distance"
+harness = false
+
+[[bench]]
+name = "geodesic_distance"
 harness = false
 
 [[bench]]
--- a/geo/benches/geodesic_distance.rs
+++ b/geo/benches/geodesic_distance.rs
@@ -1,0 +1,19 @@
+#[macro_use]
+extern crate criterion;
+extern crate geo;
+
+use geo::prelude::*;
+
+fn criterion_benchmark(c: &mut criterion::Criterion) {
+    c.bench_function("geodesic distance f64", |bencher| {
+        let a = geo::Point::<f64>::new(17.107558, 48.148636);
+        let b = geo::Point::<f64>::new(16.372477, 48.208810);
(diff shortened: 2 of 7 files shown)


geo/Cargo.toml
```diff
@@ -18,6 +18,8 @@ travis-ci = { repository = "georust/geo" }
 num-traits = "0.2"
 serde = { version = "1.0", optional = true, features = ["derive"] }
 rstar = { version = "0.7" }
+geographiclib-rs = { version = "0.1", git = "https://github.com/michaelkirk/geographiclib-rs" }
```
michaelkirk, line 21: I don't want to merge this repo reference as is. 

My preference would be moving the geographiclib-rs repo into the georust org and publishing from there. Thoughts?

geo/src/algorithm/geodesic_distance.rs
```diff
@@ -0,0 +1,46 @@
...
+    ///     5_585_234., // meters
+    ///     distance.round()
+    /// );
+    /// ```
+    /// [Karney (2013)]:  https://arxiv.org/pdf/1109.4448.pdf
+    fn geodesic_distance(&self, rhs: &Rhs) -> T;
+}
+
+impl GeodesicDistance<f64> for Point<f64> {
+    fn geodesic_distance(&self, rhs: &Point<f64>) -> f64 {
```
michaelkirk, line 43: Note that, currently anyway, geographiclib only works with f64.

geo/benches/geodesic_distance.rs
```diff
@@ -0,0 +1,19 @@
...
+
+use geo::prelude::*;
+
+fn criterion_benchmark(c: &mut criterion::Criterion) {
+    c.bench_function("geodesic distance f64", |bencher| {
+        let a = geo::Point::<f64>::new(17.107558, 48.148636);
+        let b = geo::Point::<f64>::new(16.372477, 48.208810);
+
+        bencher.iter(|| {
+            let _ = a.geodesic_distance(&b);
```
michaelkirk, line 13: The geodesic computation is 2-3x slower than vincenty

```
geodesic distance f64   time:   [873.59 ns 896.28 ns 922.24 ns]                                   
vincenty distance f64   time:   [357.93 ns 359.94 ns 361.92 ns]                                  
```

geo/Cargo.toml
michaelkirk: /cc @Psykopear

geo/benches/geodesic_distance.rs
michaelkirk: FYI we can get this down 20% if we're willing to make a hard dependency on the `lazy_static` crate (we already have several transitive dependencies on it). Shown here:  https://github.com/michaelkirk/rust-geo/commit/7e2c22c91b019f032afc9d23b9858f83b6b01b1e

geodesic distance f64   time:   [720.97 ns 725.92 ns 732.69 ns]

geo/Cargo.toml
frewsxcv: if you get @Psykopear's approval, +1 to move into the georust org. let me know if you need permissions for that

geo/benches/geodesic_distance.rs
frewsxcv: would `lazy_static` be added to this crate or the `geodesiclib` crate? + 1 from me either way

michaelkirk: I was originally thinking here in geo, but now that you mention it, it makes more sense in the geographiclib crate so any other potential users could benefit.

michaelkirk: Done in https://github.com/michaelkirk/geographiclib-rs/pull/4

geo/Cargo.toml
Psykopear: Hi @frewsxcv and @michaelkirk 
The code is released under MIT License on purpose, so please take whatever you need and use it how you please.
I just had an email exchange with Michael, with the work he did you can consider him an author as much as me. As I said to him I probably won't be working on this for a while.

michaelkirk:
bors r=frewsxcv

PR #442, opened 2020-04-25
Introduce Rect constructor that doesn't panic.

frewsxcv:
`Rect::new` panics if the user supplies invalid rectangle coordinates. This pull request introduces `Rect::try_new` that returns an `Option<Rect>`.

### DIFF
--- a/geo-types/src/rect.rs
+++ b/geo-types/src/rect.rs
@@ -34,10 +34,10 @@ impl<T: CoordinateType> Rect<T> {
     where
         C: Into<Coordinate<T>>,
     {
-        Rect::try_new(min, max).expect(RECT_INVALID_BOUNDS_ERROR)
+        Rect::try_new(min, max).unwrap()
     }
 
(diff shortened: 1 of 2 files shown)


geo-types/src/rect.rs
```diff
@@ -34,11 +34,23 @@ impl<T: CoordinateType> Rect<T> {
...
         C: Into<Coordinate<T>>,
     {
-        let (min, max) = (min.into(), max.into());
-
-        Self::assert_valid_bounds(min, max);
+        Rect::try_new(min, max).expect(RECT_INVALID_BOUNDS_ERROR)
+    }
 
-        Rect { min, max }
+    pub fn try_new<C>(min: C, max: C) -> Option<Rect<T>>
```
michaelkirk, line 40: Take it or leave it, but would this make more sense as a Result?

frewsxcv: https://github.com/georust/geo/pull/442/commits/da09297b9ab787423f9649063a223359fc73c413

PR #443, opened 2020-04-25
Implement BoundingRect for Point, Geometry, GeometryCollection.

frewsxcv:
This is dependent on https://github.com/georust/geo/pull/442

### DIFF
--- a/geo/src/algorithm/bounding_rect.rs
+++ b/geo/src/algorithm/bounding_rect.rs
@@ -1,10 +1,10 @@
 use crate::utils::{partial_max, partial_min};
 use crate::{
     Coordinate, CoordinateType, Geometry, GeometryCollection, Line, LineString, MultiLineString,
-    MultiPoint, MultiPolygon, Polygon, Rect, Triangle
+    MultiPoint, MultiPolygon, Polygon, Rect, Triangle,
 };
-use geo_types::InvalidRectCoordinatesError;
 use geo_types::private_utils::{get_bounding_rect, line_string_bounding_rect};
+use geo_types::InvalidRectCoordinatesError;
 
 /// Calculation of the bounding rectangle of a geometry.
 pub trait BoundingRect<T: CoordinateType> {
@@ -185,7 +185,10 @@ where
 }
 
 // Return a new rectangle that encompasses the provided rectangles
-fn bounding_rect_merge<T: CoordinateType>(a: Rect<T>, b: Rect<T>) -> Result<Rect<T>, InvalidRectCoordinatesError> {
+fn bounding_rect_merge<T: CoordinateType>(
+    a: Rect<T>,
+    b: Rect<T>,
+) -> Result<Rect<T>, InvalidRectCoordinatesError> {
     Rect::try_new(
         Coordinate {
             x: partial_min(a.min().x, b.min().x),
@@ -200,13 +203,13 @@ fn bounding_rect_merge<T: CoordinateType>(a: Rect<T>, b: Rect<T>) -> Result<Rect
 
 #[cfg(test)]
 mod test {
+    use super::bounding_rect_merge;
     use crate::algorithm::bounding_rect::BoundingRect;
     use crate::line_string;
     use crate::{
         polygon, Coordinate, Line, LineString, MultiLineString, MultiPoint, MultiPolygon, Polygon,
         Rect,
     };
-    use super::bounding_rect_merge;
 
     #[test]
     fn empty_linestring_test() {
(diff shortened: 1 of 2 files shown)


geo/src/algorithm/bounding_rect.rs
```diff
@@ -141,6 +142,61 @@ where
...
 
+impl<T> BoundingRect<T> for Geometry<T>
+where
+    T: CoordinateType,
+{
+    type Output = Option<Rect<T>>;
+
+    fn bounding_rect(&self) -> Self::Output {
+        match self {
+            Geometry::Point(_) => None,
```
michaelkirk, line 153: Won't this give the wrong answer for some collection that include points?

e.g.
```
Collection = [ 
  Point (0, 0), 
  Rect { (1, 1), (2, 2) }
  Rect { (2, 2), (3, 3) }
] 
```

I would expect the bounding_rect to be `Rect { (0, 0), (3, 3) }`

But the current impl seems like it would return `Rect { (1, 1), (3, 3) }`

It seems like we could solve this by having point.bounding_rect() return a zero area rect (`Rect { point, point }` )

geo/src/utils.rs
```diff
@@ -53,3 +53,21 @@ where
         }
     }
 }
+
+// The Rust standard library has `max` for `Ord`, but not for `PartialOrd`
+pub fn partial_max<T: PartialOrd>(a: T, b: T) -> T {
```
michaelkirk, line 58: hm... can we use T::min from float traits?

michaelkirk: oh, I forgot CoordinateType isn't necessarily a Float. 😅

geo/src/algorithm/bounding_rect.rs
```diff
@@ -141,6 +142,61 @@ where
...
+            }
+        })
+    }
+}
+
+// Return a new rectangle that encompasses the provided rectangles
+fn bounding_rect_merge<T: CoordinateType>(a: Rect<T>, b: Rect<T>) -> Option<Rect<T>> {
+    Rect::try_new(
+        Coordinate {
+            x: partial_min(a.min().y, b.min().y),
```
michaelkirk, line 190: Looks like a typo - x: partial_min(a.min().**x**, b.min().**x**),

geo/src/utils.rs
```diff
@@ -53,3 +53,21 @@ where
...
+    if a > b {
+        a
+    } else {
+        b
+    }
+}
+
+// The Rust standard library has `min` for `Ord`, but not for `PartialOrd`
+pub fn partial_min<T: PartialOrd>(a: T, b: T) -> T {
+    if a > b {
```
michaelkirk, line 68: shouldn't this be `<` since this method wants the min?

frewsxcv:
why yes @michaelkirk, i should write tests 😉

frewsxcv:
also thanks so much for reviewing!

geo/src/utils.rs
frewsxcv: yep totally, good catch! addressed in https://github.com/georust/geo/pull/443/commits/ff08c8aa044fb8f9c9b1e545977d396539a9875e

geo/src/algorithm/bounding_rect.rs
frewsxcv: addressed in https://github.com/georust/geo/pull/443/commits/0db539603f733e6a2e82d892650a7d0a70584a22

geo/src/utils.rs
frewsxcv: yeah...our usage of both `Float` and `CoordinateType` is confusing at times

geo/src/algorithm/bounding_rect.rs
frewsxcv: addressed in https://github.com/georust/geo/pull/443/commits/83c90abfcef1a072181d10ce731fe0a48891c800

geo/src/algorithm/bounding_rect.rs
```diff
@@ -141,13 +156,72 @@ where
     }
 }
 
+impl<T> BoundingRect<T> for Geometry<T>
+where
+    T: CoordinateType,
+{
+    type Output = Option<Rect<T>>;
+
+    fn bounding_rect(&self) -> Self::Output {
```
michaelkirk, line 165: Take it or leave it, or possibly for another PR, but can this can return a Rect instead of an Option<Rect>?

Or is there still something returning a `None` bounding rect?

frewsxcv: completely agree with you that it should return a `Rect`, but there are existing implementations (like `impl BoundingRect for LineString`) that will return `None` if the `LineString` has zero coordinates. we should change this in the future though!

frewsxcv:
bors r=michaelkirk

frewsxcv:
bors r=michaelkirk

fyi, i just gave you bors review access @michaelkirk . let's see if this works now

PR #456, opened 2020-05-09
speed up ci

michaelkirk:
Several CI speedups in separate commits. Probably best to review the commits individually.

**Before**
Screen Shot 2020-05-09 at 6 15 26 PM

**After**

Overall things are much faster.

First run still builds tarpaulin, which takes an extra ~8m, but still at least only one job does it now, as opposed to before when all jobs did it.

Screen Shot 2020-05-09 at 6 10 33 PM

Subsequent builds use the cached tarpaulin so we shave off that ~8min.

Screen Shot 2020-05-09 at 6 10 17 PM

The next biggest thing in the CI build is the proj build. One quick fix might be to use the apt package, but I'm assuming we're doing it this way to peg to a specific version so we're not at the whim of ubuntu's proj packaging - is that right @urschrei? 

If I get the inspiration - maybe I'll try to knock out a script to cache the proj build based on version in a followup PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,12 +9,12 @@ addons:
 
 matrix:
   include:
-    - env: GEO_TYPES_FEATURES="" GEO_FEATURES="" RUN_TARPAULIN=1
+    - env: GEO_FEATURES="" GEO_TYPES_FEATURES="" RUN_TARPAULIN=1
+    - env: GEO_FEATURES="--features use-proj"
+    - env: GEO_FEATURES="--features postgis-integration"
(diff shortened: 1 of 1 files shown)


michaelkirk:
bors try

michaelkirk:
bors retry

michaelkirk:
bors retry

😓

michaelkirk:
bors try

michaelkirk:
Ok! Sorry for all the bors noise - this is ready for review!

michaelkirk:
Lol you inspired me.

PR #458, opened 2020-05-10
Replace source-built proj with libproj-dev from Debian Sid

urschrei:


urschrei:
Ref @michaelkirk in https://github.com/georust/geo/pull/456#issue-415628941.

Looks like this isn't currently an improvement – ~44 mins cumulative vs ~41 mins when building proj from source, even though that specific build is shorter. Bah.

urschrei:
It looks like switching to Sid adds ~60s to build times across the board from current master, but reduces the proj build time.
I don't use apt enough to try to fine-tune the builds to use stable for `libssl-dev`, but Sid for `proj-dev`, and I don't know whether that would actually help. In conclusion: ¯\\_(ツ)_/¯.

What shall we do? The source build isn't onerous or fiddly, so I'm happy with the status quo.

frewsxcv:
yeah i'm fine to leave as is if it's faster 👍

michaelkirk:
Thanks for looking at this @urschrei!

PR #461, opened 2020-05-11
Use macro constructors in doc examples.

frewsxcv:


geo/src/algorithm/euclidean_distance.rs
```diff
@@ -27,50 +27,67 @@ pub trait EuclideanDistance<T, Rhs = Self> {
...
     /// use geo::algorithm::euclidean_distance::EuclideanDistance;
-    /// use geo::{LineString, Point, Polygon, COORD_PRECISION};
+    /// use geo::point;
+    ///
+    /// let p1 = point!(x: -72.1235, y: 42.3521);
+    /// let p2 = point!(x: -72.1260, y: 42.45);
+    ///
+    /// let distance = p1.euclidean_distance(&p2);
+    ///
+    /// assert_eq!(distance, 0.09793191512474639);
```
michaelkirk, line 41: Maybe better to use assert_relative_eq for the comparisons, but no biggie

frewsxcv: hmm i forgot we had access to that function. i just changed this line to `assert_relative_eq!(distance, 0.09793);` locally, and now the test fails:

```
thread 'main' panicked at 'assert_relative_eq!(distance, 0.09793)

    left  = 0.09793191512474639
    right = 0.09793

', src/algorithm/euclidean_distance.rs:13:1
```

i maybe need to specify an `epsilon`? and if i do that, seems like that'd be just as verbose as writing out the entire floating point number. thoughts?

michaelkirk: Right - I think the default epsilon is [f64::EPSILON](https://github.com/brendanzab/approx/blob/master/src/abs_diff_eq.rs#L87).

> if i do that, seems like that'd be just as verbose as writing out the entire floating point number

To clarify, my concern wasn't that typing out the entire floating point would be too verbose - it was about avoiding a future false alarm failing tests if some inconsequential change causes floating point math to round *slightly* differently. I assumed that's why we were using `assert_relative_eq!` at all.

So, I was expecting something like:
```
assert_relative_eq!(distance, 0.09793191512474639);
```

Or

```
assert_relative_eq!(distance, 0.09793191512474639, epsilon = 1e-10);
```

Take it or leave it though - I'm not hung up on it.

frewsxcv: ohhh that totally makes sense. thanks for the info!

frewsxcv:
bors r=michaelkirk

PR #462, opened 2020-05-14
Remove COORD_PRECISION and replace usages with approx macros.

### LINKED ISSUE #392: question about COORD_PRECISION vs std::f32::EPSILON

guidorice:
## Context

The `COORD_PRECISION` constant is an arbitrary value of questionable utility. Additional background is in PR https://github.com/georust/geo/pull/437

## Action

This pull requests removes `COORD_PRECISION` and instead `approx` crate macros are used for floating point comparisons.

Closes #392

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Changes
 
+## Unreleased
+
+* geo-types: Remove COORD_PRECISION which was an arbitrary constant of 0.1m
+
 ## geo 0.13.0
 
 * Bump geo-types dependency to 0.5.0
--- a/geo-types/Cargo.toml
+++ b/geo-types/Cargo.toml
@@ -11,6 +11,7 @@ description = "Geospatial primitive data types"
 edition = "2018"
 
 [dependencies]
+approx = "0.3"
 num-traits = "0.2"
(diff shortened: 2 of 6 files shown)


guidorice:
bors: r+

michaelkirk:
I’m still pretty much a tourist in this repo, so I’ll wait for another review before merging, especially since we’re introducing a new release dependency. 

Personally, I think it makes sense to leverage the kind of comparisons in approx here and anywhere else we’d do “equalish” checks with floating points.

guidorice:
@michaelkirk sounds good! thanks 👍

Issue #460, opened 2020-05-10
Idea: Deprecate Area#area, and add Area#signed_area and Area#unsigned_area


frewsxcv:
Thoughts? More verbose, but it makes it much clearer what the return value is.

Related to:

- https://github.com/georust/geo/issues/93
- https://github.com/georust/geo/issues/194#issuecomment-525304959
- https://github.com/georust/geo/issues/194#issuecomment-621477598

urschrei:
Yes, absolutely!

PR #463, opened 2020-05-16
Separate `area` method into `{signed,unsigned}_area`.

frewsxcv:
Fixes https://github.com/georust/geo/issues/460


geo/src/algorithm/chamberlain_duquette_area.rs
```diff
@@ -32,31 +33,41 @@ use num_traits::Float;
...
+
+    fn unsigned_area(&self) -> T;
 }
 
 impl<T> ChamberlainDuquetteArea<T> for Polygon<T>
 where
     T: Float + CoordinateType,
 {
-    fn chamberlain_duquette_area(&self) -> T {
+    fn signed_area(&self) -> T {
```
michaelkirk, line 60: I guess `chamberlain_duquette_signed_area`/`chamberlain_duquette_unsigned_area` would be too verbose?

Just weighing that against the ambiguity of using the same method names here - seems like a reasonable tradeoff.

geo/src/algorithm/area.rs
```diff
@@ -107,18 +135,27 @@ impl<T> Area<T> for MultiPolygon<T>
...
-    fn area(&self) -> T {
+    fn signed_area(&self) -> T {
         self.0
             .iter()
-            .fold(T::zero(), |total, next| total + next.area())
+            .fold(T::zero(), |total, next| total + next.signed_area())
+    }
+
+    fn unsigned_area(&self) -> T {
+        self.signed_area().abs()
```
michaelkirk, line 145: hm... is this correct?

e.g. if I had 
```
polygon_1.signed_area == 1 
polygon_2.signed_area == -1
```

and
```
m = MultiPolygon::new(polygon_1, polygon_2)
```

then shouldn't `m.unsigned_area == 2`?

But the current code would give `m.unsigned_area == 0`

Should we instead sum each polygon's unsigned area like you did with GeometryCollection?

geo/src/algorithm/centroid.rs
```diff
@@ -205,8 +205,7 @@ where
             return None;
         }
         for poly in &self.0 {
-            // the area is signed
-            let area = poly.area().abs();
+            let area = poly.unsigned_area();
```
michaelkirk, line 208: yay for readable code over copious comments.

geo/src/algorithm/chamberlain_duquette_area.rs
urschrei: Just to note that the ([Chamberlain-Duquette algorithm](https://trs.jpl.nasa.gov/bitstream/handle/2014/41271/07-0286.pdf?sequence=1&isAllowed=y), p4) explicitly calculates the signed area – I think an unsigned variant would be confusing / no longer strictly C-D, so this is OK. We point out that it's signed in the docs, too.

urschrei:
> One concern about the implementation for MultiPolygon.
> 
> Overall, I think it reads much better.
> 
> Looking back, I think there were kind of two issues - the first being that `area` is inherently ambiguous, and secondly, even though it's ambiguous, it seems like some other popular libs (geos/shapely) have decided that `area` means "unsigned area".
> 
> So I think at this point, being super explicit makes sense for us.
> 
> Maybe we should add some migration help to the release notes, like:
> 
> > Introduces `Area#unsigned_area`. For clarity, the existing `Area#area`, which can return a negative value depending on winding order, has been renamed to `Area#signed_area`.  Most likely, if you aren't sure which one to use, you'll want `unsigned_area` which is always positive.
> 
> (feel free to take, leave, or modify the suggested release note)

I agree, this is a good idea for inclusion.

geo/src/algorithm/chamberlain_duquette_area.rs
frewsxcv: @urschrei to clarify, how do you feel about having signed/unsigned variants for C-D? if i didn't read the docs, and i saw an algorithm that returned meters, i would expect it to be unsigned, which is one reason i like having the distinction here in the method name. but i definitely don't feel strongly if this is a weird convention!

urschrei: > if i didn't read the docs, and i saw an algorithm that returned meters, i would expect it to be unsigned

That’s an interesting point. I’d prefer to be practical at the expense of strict adherence to the paper, and we do note that the original is signed.

geo/src/algorithm/area.rs
frewsxcv: @michaelkirk great catch 🏆 

fixed in https://github.com/georust/geo/pull/463/commits/a030e0b9cf67a0b9b87e3fcc55e23766511d5a02

geo/src/algorithm/chamberlain_duquette_area.rs
frewsxcv: > I guess `chamberlain_duquette_signed_area`/`chamberlain_duquette_unsigned_area` would be too verbose?
> 
> Just weighing that against the ambiguity of using the same method names here - seems like a reasonable tradeoff as you have it.

yeah i meant to mention this when i opened the pr. i was undecided what to go with. for now, i'm going with the verbose since that's consistent with what we do for other traits (e.g. `EuclideanDistance#euclidean_distance`). https://github.com/georust/geo/pull/463/commits/8393b73e4950f03cd457f4ba8b2925d8397d2eed

we can always change this in the future if it's annoying though

PR #466, opened 2020-05-23
Extract PostGIS integration from 'geo' to a separate 'geo-postgis' crate.

frewsxcv:
- Reduces `geo`'s full dependency tree quite a bit
- Our PostGIS compatibility feature in `geo` lived in `geo::algorithms`, but is it really an algorithm? It's really more just glue between `geo-types` and `postgis`
- We can move this out to a separate repo if we want

Thoughts?

### DIFF
--- a/geo-postgis/Cargo.toml
+++ b/geo-postgis/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "geo-postgis"
+version = "0.1.0"
+authors = ["Corey Farwell <coreyf@rwell.org>"]
+license = "MIT/Apache-2.0"
+repository = "https://github.com/georust/geo" # TODO: change this?
+documentation = "https://docs.rs/geo-postgis/"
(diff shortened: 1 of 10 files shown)


geo-postgis/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+name = "geo-postgis"
+version = "0.1.0"
+authors = ["Corey Farwell <coreyf@rwell.org>"]
+license = "MIT/Apache-2.0"
+repository = "https://github.com/georust/geo" # TODO: change this?
+documentation = "https://docs.rs/geo-postgis/"
+readme = "../README.md" # TODO
+keywords = ["gis", "geo", "geography", "geospatial"] # TODO
+description = "" # TODO
+edition = "2018"
```
frewsxcv, line 11: I'll fill in the `TODO`s above if this PR gets approved

PR #469, opened 2020-05-28
Migrate from Travis CI to GitHub Actions

frewsxcv:
Down from 15 minutes to ~10~ 3 minutes :)


frewsxcv:
Working on creating a Docker image so we don't have to compile proj and tarpaulin each time we run CI https://github.com/georust/docker-images

Issue #473, opened 2020-07-02
Triangle.contains(Point) fails for degenerate triangles


msalib:
Given a degenerate triangle  `t` where all coordinates are the same, `t.contains(p)` will return `true` for any Point `p`. 

I tried running this example:

```rust
use geo::prelude::*;
use geo::{Coordinate, Point, Triangle};

fn main() {
    let origin: Coordinate<f64> = (0., 0.).into();
    let tri = Triangle(origin, origin, origin);
    let pt: Point<f64> = (0., 1.23456).into();
    assert!(!tri.contains(&pt));
    println!("Hello, world!");
}
```

It fails with an assertion error, but it should pass.

One fix would be to include a check for degenerate self in the `contains` call; in that case, we should just return an equality check between any of the triangle's points (they're all the same) and `other`. Alternatively, perhaps the documentation could be updated to explain that degenerate Triangles are forbidden?

My rustc info is:
```
rustc 1.44.1 (c7087fe00 2020-06-17)
binary: rustc
commit-hash: c7087fe00d2ba919df1d813c040a5d47e43b0fe7
commit-date: 2020-06-17
host: x86_64-unknown-linux-gnu
release: 1.44.1
LLVM version: 9.0
```

Thanks so much for an excellent package! I only stumbled onto this because I've been building some spatial-join code using it and my property tests constructed this weird case.

michaelkirk:
Fixed in #474

PR #474, opened 2020-07-03
Fix contains impl of triangle for collinear points

mbrobbel:
A fix for #473. It replaces the `Contains` implementation for `Triangle` with the method described [here](https://stackoverflow.com/a/2049712/3751841) and [here](https://stackoverflow.com/a/20861130/3751841).

### DIFF
--- a/geo/src/algorithm/contains.rs
+++ b/geo/src/algorithm/contains.rs
@@ -340,11 +340,29 @@ where
     T: CoordinateType,
 {
     fn contains(&self, coord: &Coordinate<T>) -> bool {
-        let sign_1 = utils::sign(coord, &self.0, &self.1);
-        let sign_2 = utils::sign(coord, &self.1, &self.2);
-        let sign_3 = utils::sign(coord, &self.2, &self.0);
+        let s = self.0.y * self.2.x - self.0.x * self.2.y
+            + (self.2.y - self.0.y) * coord.x
+            + (self.0.x - self.2.x) * coord.y;
+        let t = self.0.x * self.1.y - self.0.y * self.1.x
+            + (self.0.y - self.1.y) * coord.x
+            + (self.1.x - self.0.x) * coord.y;
 
-        (sign_1 == sign_2) && (sign_2 == sign_3)
+        if (s < T::zero()) != (t < T::zero()) {
+            return false;
+        }
+
+        let a = (T::zero() - self.1.y) * self.2.x
+            + self.0.y * (self.2.x - self.1.x)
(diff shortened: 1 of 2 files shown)


urschrei:
This the correct solution, but I think we have some existing implementations of cross-product and left / right querying that we can re-use for it. I'll have a look over the weekend.

urschrei:
> I think we have some existing implementations of cross-product and left / right querying that we can re-use for it.
Turns out that we dont'!

geo/src/algorithm/contains.rs
```diff
@@ -340,11 +340,29 @@ where
     T: CoordinateType,
 {
     fn contains(&self, coord: &Coordinate<T>) -> bool {
-        let sign_1 = utils::sign(coord, &self.0, &self.1);
-        let sign_2 = utils::sign(coord, &self.1, &self.2);
-        let sign_3 = utils::sign(coord, &self.2, &self.0);
+        let s = self.0.y * self.2.x - self.0.x * self.2.y
```
urschrei, line 343: Could you link to the SO solution and / or that this function uses barycentric coordinates to obtain the answer? Just so readers know what's happening and where to look for more information.

geo/src/algorithm/contains.rs
```diff
@@ -340,11 +340,29 @@ where
...
+            return false;
+        }
+
+        let a = (T::zero() - self.1.y) * self.2.x
+            + self.0.y * (self.2.x - self.1.x)
+            + self.0.x * (self.1.y - self.2.y)
+            + self.1.x * self.2.y;
 
-        (sign_1 == sign_2) && (sign_2 == sign_3)
+        if a == T::zero() {
```
michaelkirk, line 359: I haven't analyzed the math of the solution, but only looking at the transcription...

what's this `a==T::zero()` branch about? It's not present in the linked example: https://stackoverflow.com/questions/2049582/how-to-determine-if-a-point-is-in-a-2d-triangle/20861130#20861130

mbrobbel: My initial thinking: the area of a degenerate triangle (formed by collinear points) is zero, therefor a point is never contained by a degenerate triangle (return `false`). However, reading the suggestion from #473, checking if the given point contains any of the triangle points, I realize that makes more sense. How do you suggest to solve this?

michaelkirk: I think this makes sense, and the unit test verifies the behavior. Thanks!

michaelkirk:
Rebased and merged in 8f9c1e22

Thank you!

PR #31, opened 2016-05-27
New Algoritms added

azime:
New algorithms added: Contains, Intersects, Area and distance.

PR #480, opened 2020-07-20
Add Concave Hull algorithm

RestitutorOrbis:
As part of https://github.com/dabreegster/abstreet/issues/198, I've been working on adding an implementation of the Concave Hull algorithm primarily inspired by https://github.com/mapbox/concaveman. This implementation doesn't completely copy concaveman but it does ape the notion of using R trees in order to avoid a linear search through the points although my use of R trees relies on a circular envelope rather than the rectangular envelope that concaveman seems to use.


urschrei:
Thanks for the addition, we'd be delighted to have it - let us know when you're ready for a review, or if you have any other questions in the mean time!

RestitutorOrbis:
@urschrei I think I'm close. Unfortunately, I'm struggling with setting up a search function for the R tree. I have the following:
```rust
impl<T, W> SelectionFunction<T> for SelectWithinDistanceOfGeomFunction<W>
where
    T: RTreeObject + EuclideanDistance<W, Line<W>>,
    W: Float + RTreeNum {
    fn should_unpack_parent(&self, envelope: &T::Envelope) -> bool {
        envelope.euclidean_distance(&self.geometry) < self.max_dist
    }

    fn should_unpack_leaf(&self, leaf: &T) -> bool {
        leaf.euclidean_distance(&self.geometry) < self.max_dist
    }
}
```

This results in the following compile error:
```
error[E0599]: no method named `euclidean_distance` found for reference `&<T as rstar::RTreeObject>::Envelope` in the current scope
   --> geo/src/algorithm/concavehull.rs:153:18
    |
153 |         envelope.euclidean_distance(&self.geometry) < self.max_dist
    |                  ^^^^^^^^^^^^^^^^^^ method not found in `&<T as rstar::RTreeObject>::Envelope`
    |
    = help: items from traits can only be used if the trait is implemented and in scope
note: `algorithm::euclidean_distance::EuclideanDistance` defines an item `euclidean_distance`, perhaps you need to implement it
   --> geo/src/algorithm/euclidean_distance.rs:17:1
    |
17  | pub trait EuclideanDistance<T, Rhs = Self> {
```

As you can imagine, EuclideanDistance isn't implemented on the envelope but it is implemented on the generic T. Is it possible to convert the envelope into T? If not, do you have an alternative formulation of the types and the implementation that would be better than this one? Any help would be appreciated.

urschrei:
@RestitutorOrbis Just a brief comment as I'm pushed for time: not that I know of, but I'm not much of an authority on types – one possibility may be to use the `Envelope`'s points (assuming it's an `AABB`) to construct a `geo-types` `Rect` instance, and use that?

RestitutorOrbis:
@urschrei Do you happen to have the time to review this any time soon? No worries if not.

urschrei:
If by soon you mean "this week", then yes!

geo/src/algorithm/concavehull.rs
```diff
@@ -0,0 +1,383 @@
...
+    T: Float + RTreeNum,
+{
+    let w = line.euclidean_length() + max_dist.clone() + max_dist.clone();
+    let h = max_dist.clone() + max_dist.clone();
+    let two = T::add(T::one(), T::one());
+    let search_dist = T::div(T::sqrt(T::powi(w, 2) + T::powi(h, 2)), two);
+    let mut candidates = interior_points_tree
+        .locate_within_distance(line.centroid(), search_dist)
+        .peekable();
+    let peek = candidates.peek();
```
michaelkirk, line 26: would the rest of this method be cleaner implemented using `candidates.min_by(...`?

Maybe not once you handle `NaN`... so take it or leave it.

RestitutorOrbis:
@michaelkirk Thanks for the review! I'll address your feedback as soon as I can.

geo/src/algorithm/concavehull.rs
```diff
@@ -0,0 +1,418 @@
...
+
+fn find_point_closest_to_line<T>(
+    interior_points_tree: &RTree<Point<T>>,
+    line: Line<T>,
+    max_dist: &T,
+) -> Option<Point<T>>
+where
+    T: Float + RTreeNum,
+{
+    let h = max_dist.add(max_dist.clone());
```
RestitutorOrbis, line 125: @michaelkirk Is there a way to avoid cloning here? When I take it out, I get the following error:
```
error[E0308]: mismatched types
   --> geo/src/algorithm/concavehull.rs:125:26
    |
117 | fn find_point_closest_to_line<T>(
    |                               - this type parameter
...
125 |     let h = max_dist.add(max_dist);
    |                          ^^^^^^^^
    |                          |
    |                          expected type parameter `T`, found `&T`
    |                          help: consider dereferencing the borrow: `*max_dist`
    |
    = note: expected type parameter `T`
                    found reference `&T`
    = help: type parameters must be constrained to match other types
    = note: for more information, visit https://doc.rust-lang.org/book/ch10-02-traits.html#traits-as-parameters
```

geo/src/algorithm/concavehull.rs
```diff
@@ -0,0 +1,418 @@
...
+    }
+}
+
+// This takes significant inspiration from:
+// https://github.com/mapbox/concaveman/blob/54838e1/index.js#L11
+fn concave_hull<T>(points: &Vec<Point<T>>, concavity: T) -> Vec<Point<T>>
+where
+    T: Float + RTreeNum,
+{
+    let multipoint: MultiPoint<T> = points.clone().into();
```
RestitutorOrbis, line 151: Similarly, is there a way to avoid cloning here? Once again, the error is below:
```
error[E0277]: the trait bound `geo_types::Point<T>: std::convert::From<&std::vec::Vec<geo_types::Point<T>>>` is not satisfied
   --> geo/src/algorithm/concavehull.rs:151:44
    |
151 |     let multipoint: MultiPoint<T> = points.into();
    |                                            ^^^^ the trait `std::convert::From<&std::vec::Vec<geo_types::Point<T>>>` is not implemented for `geo_types::Point<T>`
    |
    = help: the following implementations were found:
              <geo_types::Point<T> as std::convert::From<(T, T)>>
              <geo_types::Point<T> as std::convert::From<[T; 2]>>
              <geo_types::Point<T> as std::convert::From<geo_types::Coordinate<T>>>
    = note: required because of the requirements on the impl of `std::convert::Into<geo_types::Point<T>>` for `&std::vec::Vec<geo_types::Point<T>>`
    = note: required because of the requirements on the impl of `std::convert::From<&std::vec::Vec<geo_types::Point<T>>>` for `geo_types::MultiPoint<T>`
    = note: required because of the requirements on the impl of `std::convert::Into<geo_types::MultiPoint<T>>` for `&std::vec::Vec<geo_types::Point<T>>`
```

geo/src/algorithm/test_fixtures/norway_concave_hull.rs
```diff
@@ -0,0 +1,150 @@
+vec![
+[5.227792433014884,59.468030338041025],
```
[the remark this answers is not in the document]
RestitutorOrbis: This is a regression check. I ran the algorithm against the Norway data and exported the output to determine if the algorithm has changed in the future.

geo/src/algorithm/concavehull.rs
```diff
@@ -0,0 +1,418 @@
...
+            LineString::from(concave_hull(&self.0.clone(), concavity)),
+            vec![],
+        )
+    }
+}
+
+fn find_point_closest_to_line<T>(
+    interior_points_tree: &RTree<Point<T>>,
+    line: Line<T>,
+    max_dist: &T,
```
michaelkirk, line 120:
```suggestion
    max_dist: T,
```

geo/src/algorithm/concavehull.rs
michaelkirk, line 125:
```suggestion
    let h = max_dist + max_dist;
```

geo/src/algorithm/concavehull.rs
michaelkirk: I don't see an easy way to get rid of this one.

geo/src/algorithm/concavehull.rs
```diff
@@ -0,0 +1,418 @@
+use crate::algorithm::convexhull::ConvexHull;
```
michaelkirk, line 1: one more nit: could you rename this `concave_hull.rs` to be in line with the rest of the crate?

geo/src/algorithm/concavehull.rs
michaelkirk: I've never used GH code "suggestion" feature... let's see if it works, I tried to suggest the way that I would write this bit. T is Float, so that means it's also `Copy`. I'd probably lean into that here and not borrow T at all.

geo/src/algorithm/concavehull.rs
```diff
@@ -0,0 +1,418 @@
...
+use crate::prelude::Centroid;
+use crate::utils::partial_min;
+use crate::{Line, LineString, MultiLineString, MultiPoint, MultiPolygon, Point, Polygon};
+use num_traits::Float;
+use rstar::{RTree, RTreeNum};
+use std::collections::VecDeque;
+use std::ops::Add;
+
+pub trait ConcaveHull<T> {
+    /// Returns the concave hull of a Polygon.
```
michaelkirk, line 13: Thank you for adding some doc comments. 

Since you've implemented this for more than just Polygons, maybe clearer to say something like:

> Returns the concave hull of a **Geometry**.

Also bonus points if you can avoid using the term "concave hull" in your description. Maybe something like...

> Returns a polygon which covers the geometry. Unlike _convex_ hulls, which also cover their geometry, a _concave_ hull does so while trying to minimize its area by "shrink wrapping" around the geometry it covers.

Heh - do people know what "shrink wrapping" means? Anyway, good docs are tough, and somewhat subjective, so take it or leave it.

RestitutorOrbis: What do you think about this?
'''
///Returns a polygon which covers a geometry. Unlike convex hulls, which also cover
/// their geometry, a concave hull does so while trying to further minimize its area by
/// constructing edges such that the exterior of the polygon incorporates points that would
/// be interior points in a convex hull.
'''

michaelkirk: 👏 👏 👏

rmanoka:
@RestitutorOrbis This is an interesting algorithm and implementation;  I have used [alpha hulls](https://en.wikipedia.org/wiki/Alpha_shape) before, but this idea is new to me.  A couple of questions on the algo/implementation:

1.  When passing through the lines in order, when processing a segment, say AB,  suppose the algo decides to add a point C.  Then, you are pushing CB to the queue to be processed next.  Does that mean we're skipping the processing of AC?  For instance, if there was a D that could've been added between AC (making the final hull ADCB), then this D would be ignored right?

1.  The output of the algo. seems to generally depend on the order in which the convex hull is obtained.  For eg. if we obtain the convex hull to be a square ABCD, and there is a point E somewhere near A, but closer to D than B.  Ideally, the natural "concave hull" would have been ABCDE, but this algo might return AEBCD.   Further, if the convex hull algo. had returned DABC (same hull, just rotated cyclically), then the algo would now give DEABC (which is the same shape as ABCDE).   It would be nice if the output is more robust, depending only on the shape of the input points, and not their ordering / rotation.  Is it inherent to this algo?  The paper you have cited suggests that when adding a point P, we should check neighboring line segments to ensure it is not closer to them than the current one.  I don't know if that covers all the possibilities though.

RestitutorOrbis:
> @RestitutorOrbis This is an interesting algorithm and implementation; I have used [alpha hulls](https://en.wikipedia.org/wiki/Alpha_shape) before, but this idea is new to me. A couple of questions on the algo/implementation:
> 
>     1. When passing through the lines in order, when processing a segment, say AB,  suppose the algo decides to add a point C.  Then, you are pushing CB to the queue to be processed next.  Does that mean we're skipping the processing of AC?  For instance, if there was a D that could've been added between AC (making the final hull ADCB), then this D would be ignored right?
> 
>     2. The output of the algo. seems to generally depend on the order in which the convex hull is obtained.  For eg. if we obtain the convex hull to be a square ABCD, and there is a point E somewhere near A, but closer to D than B.  Ideally, the natural "concave hull" would have been ABCDE, but this algo might return AEBCD.   Further, if the convex hull algo. had returned DABC (same hull, just rotated cyclically), then the algo would now give DEABC (which is the same shape as ABCDE).   It would be nice if the output is more robust, depending only on the shape of the input points, and not their ordering / rotation.  Is it inherent to this algo?  The paper you have cited suggests that when adding a point P, we should check neighboring line segments to ensure it is not closer to them than the current one.  I don't know if that covers all the possibilities though.

Hi @rmanoka! With regard to 1, I think you're right that flaw exists within the current implementation. I need to play around with the implementation to see if I can improve that. With regard to 2, I missed the mention in the paper of checking that when adding a point P, we should check that neighbouring line segments are not closer. I suspect that adding that check would ameliorate the problem you mention but I wouldn't be surprised if there's still issues. I'll try to implement some of these ideas and see where it goes.

RestitutorOrbis:
Hey @rmanoka! I've implemented a new round of changes to address your comments. One thing I'm not sure about is that I used another RTree to determine which edge is the closest to the newly chosen point, do you happen to know if there's a more efficient solution? I ask because I have to keep modifying the tree in order to acknowledge the flexing of edges and my understanding is that tree modifications are expensive for RTrees.

rmanoka:
@RestitutorOrbis I also can't think of a cleverer idea than two r-trees.  Let me stare at your code a bit more over the weekend.

RestitutorOrbis:
@rmanoka Any luck?

rmanoka:
> @rmanoka Any luck?

No, could not find a better way to do the search.  Is the double rtree slowing it down too much?

RestitutorOrbis:
@rmanoka I think it exhibits reasonable performance. The Norway test (which should be the most computationally expensive test) runs in a range of 350 to  450 milliseconds which strikes me as reasonable. Do you think it makes sense to write a benchmark for it just like there is with the convex hull?

geo/src/algorithm/concave_hull.rs
```diff
@@ -0,0 +1,432 @@
...
+use crate::algorithm::euclidean_distance::EuclideanDistance;
+use crate::algorithm::euclidean_length::EuclideanLength;
+use crate::prelude::Centroid;
+use crate::utils::partial_min;
+use crate::{Line, LineString, MultiLineString, MultiPoint, MultiPolygon, Point, Polygon};
+use num_traits::Float;
+use rstar::{RTree, RTreeNum};
+use std::collections::VecDeque;
+
+pub trait ConcaveHull<T> {
```
rmanoka, line 11: Could you make `T` into an associated type `Scalar` like this:

``` rust
pub trait ConcaveHull {
    type Scalar: CoordinateType;
    fn concave_hull(...) -> Polygon<Self::Scalar>;
}
```

Generally, we want the trait to be generic only if the same type could implement the trait for two different `T` (eg. see `Iterator` vs `Add` traits in std).  I've generally been advocating for this change, and recently modified the `ConvexHull` trait to reflect that.

geo/src/algorithm/concave_hull.rs
```diff
@@ -0,0 +1,432 @@
...
+    /// Returns a polygon which covers a geometry. Unlike convex hulls, which also cover
```
rmanoka, line 12: I noticed that these doc comments are better visible if they are on the trait itself, rather than on the method.  Given there's only one method, it makes sense to move the comments to just above the trait declaration.

geo/src/algorithm/concave_hull.rs
```diff
@@ -0,0 +1,432 @@
...
+    ///     (x: 4.0, y: 4.0),
+    ///     (x: 0.0, y: 4.0),
+    ///     (x: 0.0, y: 0.0),
+    ///     (x: 4.0, y: 0.0),
+    /// ];
+    ///
+    /// let res = poly.concave_hull(2.0);
+    /// assert_eq!(res.exterior(), &correct_hull);
+    /// ```
+    fn concave_hull(&self, concavity: T) -> Polygon<T>
```
rmanoka, line 46: Given you are definitely going to output a `Polygon` with no interior holes (am I right?), could we make the output as `LineString<T>`?  The user can always make a `Polygon` out of it.  This may just be my personal preference, but we should generally output the simplest type that captures our output.

geo/src/algorithm/concave_hull.rs
```diff
@@ -0,0 +1,432 @@
...
+    where
+        T: Float + RTreeNum;
```
rmanoka, line 48: These trait bounds should not appear in the trait.  Really, the restriction `T: Float + RTreeNum` is not inherent to the functionality, but an artifact of the implemetation.  They should appear in the impl. of the traits for the types.

geo/src/algorithm/concave_hull.rs
```diff
@@ -0,0 +1,432 @@
...
+}
+
+fn find_point_closest_to_line<T>(
+    interior_points_tree: &RTree<Point<T>>,
+    line: Line<T>,
+    max_dist: T,
+    edge_length: T,
+    concavity: T,
+    line_tree: &RTree<Line<T>>
+) -> Option<Point<T>>
```
rmanoka, line 126: Another minor suggestion:  all the `Point<T>` could be `Coordinate<T>`.  The idea is that `Point` may in the future be more than just the coordinates, and the algos that directly work on coordinates shouldn't need to use it.  This should also simplify some of the impls, as the geometry types store `Coordinate`s

geo/src/algorithm/concave_hull.rs
```diff
@@ -0,0 +1,432 @@
...
+    }
+}
+
+// This takes significant inspiration from:
+// https://github.com/mapbox/concaveman/blob/54838e1/index.js#L11
+fn concave_hull<T>(points: &Vec<Point<T>>, concavity: T) -> Vec<Point<T>>
+where
+    T: Float + RTreeNum,
+{
+    let multipoint: MultiPoint<T> = points.clone().into();
```
rmanoka, line 187: Can this clone be avoided?  Note that the trait impl for geometry types already clones the individual coordinates, when populating the `Vec`.  I understand that we do not implement `ConvexHull` on a slice of coordinates (may be we should).  As a temporary soln., you could go for a directly use the `qhull::quick_hull` algorithm.  It just needs a mutable slice of Coordinates, which you can get from a vector of coords.

May be the function could read: `fn concave_hull<T>(points: &[Coordinate<T>], ...)` just to make it more flexible.

rmanoka (changes requested):

rmanoka:
@RestitutorOrbis I've suggested a few changes, mostly on the structuring of the trait, and not any related to the algo.   It would be great if you could add a benchmark (may be very similar to the norway benchmarks in `benches/convex_hull.rs`).   

From the numbers you mention, it seems to be a very usable algorithm.  We can compare this against just simple convex hull, and hope we find some ideas in the future to close the gap.  

An interesting question (but I don't have an easy way to measure), is to compare this with the alpha-hull / alpha-shape algorithm;  that algorithm uses some more compilcated datastructures (Delaunay triangulation), and I don't know a rust implementation of it (`Delaunay` itself is available in the `spade` crate though).

geo/src/algorithm/concave_hull.rs
RestitutorOrbis: Yeah, you're right about the current implementation avoiding interior holes. I'll switch it to a LineString.

geo/src/algorithm/concave_hull.rs
RestitutorOrbis: Since the current `RTreeObject` implementation's don't include `Coordinate` would it be sensible to keep it as `Point`  because otherwise you get casts everywhere as you convert from `Coordinate` to `Point` when you need RTrees and you convert back when you don't.

RestitutorOrbis:
> @RestitutorOrbis I've suggested a few changes, mostly on the structuring of the trait, and not any related to the algo. It would be great if you could add a benchmark (may be very similar to the norway benchmarks in `benches/convex_hull.rs`).
> 
> From the numbers you mention, it seems to be a very usable algorithm. We can compare this against just simple convex hull, and hope we find some ideas in the future to close the gap.
> 
> An interesting question (but I don't have an easy way to measure), is to compare this with the alpha-hull / alpha-shape algorithm; that algorithm uses some more compilcated datastructures (Delaunay triangulation), and I don't know a rust implementation of it (`Delaunay` itself is available in the `spade` crate though).

I like the idea of building the benchmark to mimic that one. Should building an alpha hull algorithm be added as an issue?

geo/src/algorithm/concave_hull.rs
michaelkirk: If you're feeling bold you could implement the trait for Coordinate.

I think like this in `coordinate.rs`
```
 74 #[cfg(feature = "rstar")]
 75 impl<T> ::rstar::Point for Coordinate<T>
 76 where 
 77     T: ::num_traits::Float + ::rstar::RTreeNum,
 78 {
 79     type Scalar = T;
 80 
 81     const DIMENSIONS: usize = 2;
 82 
 83     fn generate(generator: impl Fn(usize) -> Self::Scalar) -> Self {
 84         Coordinate { x: generator(0), y: generator(1) }
 85     }         
 86 
 87     fn nth(&self, index: usize) -> Self::Scalar {
 88         match index {
 89             0 => self.x, 
 90             1 => self.y,
 91             _ => unreachable!(),
 92         } 
 93     } 
 94 
 95     fn nth_mut(&mut self, index: usize) -> &mut Self::Scalar {
 96         match index {
 97             0 => &mut self.x,
 98             1 => &mut self.y,
 99             _ => unreachable!(),
100         } 
101     }     
102 }
```

I think that this is an implementation detail that doesn't affect the user facing API, so I don't feel strongly about it, but I do agree with @rmanoka's sentiment that we should be preferring Coord whenever possible. 

Unfortunately I notice that the Rstar impl's on linestring use Points (not Coord), but I'm not sure if that affects you here.

rmanoka:
> I like the idea of building the benchmark to mimic that one. Should building an alpha hull algorithm be added as an issue?

Alpha hull algo. might fit better in the spade (or related) crate;  spade is a heavy dep. to add in geo IMO.  Do create an issue, so we can discuss it further though.

geo/src/algorithm/concave_hull.rs
RestitutorOrbis: Ooh, I'll give this a try!

gak:
Hi @RestitutorOrbis I've been using your branch and it works great. Many thanks for doing this!

I'm seeing errors at the moment in my project's CI. I don't think it's on my end, and not a big deal if you know about them already, but in case:

```
#31 118.9    Compiling geo v0.14.2 (https://github.com/RestitutorOrbis/geo.git?branch=concave-hull#6dabf3ab)
#31 119.8 error[E0308]: mismatched types
#31 119.8   --> /root/.cargo/git/checkouts/geo-afd961a39c52d1ab/6dabf3a/geo/src/algorithm/concave_hull.rs:75:22
#31 119.8    |
#31 119.8 75 |         concave_hull(&mut aggregated, concavity)
#31 119.8    |                      ^^^^^^^^^^^^^^^ expected slice, found struct `Vec`
#31 119.8    |
#31 119.8    = note: expected mutable reference `&mut [geo_types::Coordinate<T>]`
#31 119.8               found mutable reference `&mut Vec<geo_types::Point<T>>`
```

Full set of errors: https://gist.github.com/gak/e7b894726aa1a8ac70d3df85407bbde6

Please let me know if you need help with the PR.

RestitutorOrbis:
> Hi @RestitutorOrbis I've been using your branch and it works great. Many thanks for doing this!
> 
> I'm seeing errors at the moment in my project's CI. I don't think it's on my end, and not a big deal if you know about them already, but in case:
> 
> ```
> #31 118.9    Compiling geo v0.14.2 (https://github.com/RestitutorOrbis/geo.git?branch=concave-hull#6dabf3ab)
> #31 119.8 error[E0308]: mismatched types
> #31 119.8   --> /root/.cargo/git/checkouts/geo-afd961a39c52d1ab/6dabf3a/geo/src/algorithm/concave_hull.rs:75:22
> #31 119.8    |
> #31 119.8 75 |         concave_hull(&mut aggregated, concavity)
> #31 119.8    |                      ^^^^^^^^^^^^^^^ expected slice, found struct `Vec`
> #31 119.8    |
> #31 119.8    = note: expected mutable reference `&mut [geo_types::Coordinate<T>]`
> #31 119.8               found mutable reference `&mut Vec<geo_types::Point<T>>`
> ```
> 
> Full set of errors: https://gist.github.com/gak/e7b894726aa1a8ac70d3df85407bbde6
> 
> Please let me know if you need help with the PR.

Hi @gak! Thanks for reaching out, these flaws are definitely not your fault, I've been in the midst of a refactor and I pushed too soon. Try giving the latest revision a shot, it should pass your CI. Let me know if it doesn't.

gak:
Thanks heaps for the fix, just had to rename the import and it worked again.

I realise a branch in a PR is a WIP and totally don't expect it to be always working. It was just a heads up. 👍

geo/src/algorithm/concave_hull.rs
```diff
@@ -0,0 +1,433 @@
...
+        .iter()
+        .filter_map(|coord| {
+            if !hull_tree.contains(coord) {
+                Some(*coord)
+            } else {
+                None
+            }
+        })
+        .collect();
+    let mut interior_points_tree: RTree<Coordinate<T>> = RTree::bulk_load(interior_coords.clone());
```
rmanoka, line 204: `interior_coords` does not seem to be used after this; the clone can be removed.

rmanoka (approved):
Had only one minor suggestion.  Otherwise, lgtm.  

Thanks for your contribution @RestitutorOrbis .  I'll give it a few days, to see if anyone else has suggestions, and otherwise merge the PR

geo/src/algorithm/concave_hull.rs
```diff
@@ -0,0 +1,433 @@
...
+    fn concave_hull(&self, concavity: Self::Scalar) -> Polygon<Self::Scalar>;
+}
+
+impl<T> ConcaveHull for Polygon<T>
+where
+    T: Float + RTreeNum + HasKernel,
+{
+    type Scalar = T;
+    fn concave_hull(&self, concavity: Self::Scalar) -> Polygon<Self::Scalar> {
+        let mut points: Vec<_> = self.exterior().points_iter().map(|point|{
```
michaelkirk, line 58:
```suggestion
        let mut points: Vec<_> = self.exterior().0.clone();
```

michaelkirk: (and delete the following 2 lines)

geo/src/algorithm/concave_hull.rs
```diff
@@ -0,0 +1,433 @@
...
+impl<T> ConcaveHull for MultiPolygon<T>
+where
+    T: Float + RTreeNum + HasKernel,
+{
+    type Scalar = T;
+    fn concave_hull(&self, concavity: Self::Scalar) -> Polygon<Self::Scalar> {
+        let mut aggregated: Vec<Coordinate<Self::Scalar>> = self
+            .0
+            .iter()
+            .flat_map(|elem| elem.exterior().0.iter().map(|c| *c))
```
michaelkirk, line 74:
```suggestion
            .flat_map(|elem| elem.exterior().0.clone()
```

geo/src/algorithm/concave_hull.rs
```diff
@@ -0,0 +1,433 @@
...
+    }
+}
+
+impl<T> ConcaveHull for LineString<T>
+where
+    T: Float + RTreeNum + HasKernel,
+{
+    type Scalar = T;
+    fn concave_hull(&self, concavity: Self::Scalar) -> Polygon<Self::Scalar> {
+        Polygon::new(concave_hull(&mut self.clone().0, concavity), vec![])
```
michaelkirk, line 86:
```suggestion
        Polygon::new(concave_hull(&mut self.0.clone(), concavity), vec![])
```

Pretty nitpicky - but since we only need the coords, we can clone just those, rather than cloning the entire Geometry.

geo/src/algorithm/concave_hull.rs
```diff
@@ -0,0 +1,433 @@
...
+impl<T> ConcaveHull for MultiLineString<T>
+where
+    T: Float + RTreeNum + HasKernel,
+{
+    type Scalar = T;
+    fn concave_hull(&self, concavity: T) -> Polygon<T> {
+        let mut aggregated: Vec<Coordinate<T>> = self
+            .0
+            .iter()
+            .flat_map(|elem| elem.clone().0)
```
michaelkirk, line 99:
```suggestion
            .flat_map(|elem| elem.0.clone())
```

michaelkirk (commented):
I have some small feedback, but I think they should be quick changes, otherwise looks good!

rmanoka:
There is one useful optimization that is possible in the beginning of this algo.  The implementation of quick hull actually guarantees that unless there are at most three coords in the input, the coords slice is mutated so that the first n entries are the hull points, and the rest are the interior (were n is size of conv hull).

Since there's no concave-ness to add if we have < 4 points, we could handle that case and just return the conv hull.  Then, the first part of figuring which are interior points becomes trivial (and we don't need the rtree for it).

The guarantee I've mentioned above of the qhull implementation is not documented, so I'm okay with not doing that optimization in this PR.  Just noting this so we may get to it at a later point.

RestitutorOrbis:
> There is one useful optimization that is possible in the beginning of this algo. The implementation of quick hull actually guarantees that unless there are at most three coords in the input, the coords slice is mutated so that the first n entries are the hull points, and the rest are the interior (were n is size of conv hull).
> 
> Since there's no concave-ness to add if we have < 4 points, we could handle that case and just return the conv hull. Then, the first part of figuring which are interior points becomes trivial (and we don't need the rtree for it).
> 
> The guarantee I've mentioned above of the qhull implementation is not documented, so I'm okay with not doing that optimization in this PR. Just noting this so we may get to it at a later point.

I've implemented the early return on less than 4 coordinates but I didn't have any luck implementing the optimization with regard to using the earlier portion of `coords` as the hull. From looking at the `one_flex` test in particular with a debugger, I don't see the `coords` mutated that way.

rmanoka:
@RestitutorOrbis Sorry, I didn't consider some of the recursive calls;  the observation doesn't hold in the current impl.  Please ignore the suggestion.

RestitutorOrbis:
No worries!

Sent with [ProtonMail](https://protonmail.com) Secure Email.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

geo/benches/concave_hull.rs
```diff
@@ -0,0 +1,43 @@
...
+    size: usize,
+    rng: &mut R,
+) -> Vec<Coordinate<S>> {
+    (0..size)
+        .map(|_| (rng.gen_range(-range, range), rng.gen_range(-range, range)).into())
+        .collect()
+}
+
+fn criterion_benchmark(c: &mut Criterion) {
+    c.bench_function("concave hull f32", |bencher| {
```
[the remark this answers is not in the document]
michaelkirk: Thanks for addressing this!

michaelkirk (approved):
This LGTM - I'll wait a couple days before merging in case anyone else wants a final review.

Thanks for your contribution and your patience with the review process @RestitutorOrbis!

michaelkirk:
Do you know why this new test might be failing @RestitutorOrbis?

from https://github.com/georust/geo/runs/1190445056
```
---- algorithm::concave_hull::test::concave_hull_norway_test stdout ----
thread 'algorithm::concave_hull::test::concave_hull_norway_test' panicked at 'assertion failed: `(left == right)`
```

RestitutorOrbis:
> Do you know why this new test might be failing @RestitutorOrbis?
> 
> from https://github.com/georust/geo/runs/1190445056
> 
> ```
> ---- algorithm::concave_hull::test::concave_hull_norway_test stdout ----
> thread 'algorithm::concave_hull::test::concave_hull_norway_test' panicked at 'assertion failed: `(left == right)`
> ```

@michaelkirk It seems when I updated the impl, the Norway output changed as well. From looking at the SVG file, it doesn't seem visually distinct so I'm not entirely sure how concerned I should be about that. Do you think it would make sense to remove the Norway test? I was hoping it would be an effective regression test but it seems so brittle that any slight change to the algorithm perturbs it. For reference, from running `git bisect`, it seems to have been switching to coordinates from points that caused this breakage.

geo/src/algorithm/concave_hull.rs
```diff
@@ -0,0 +1,464 @@
...
+/// Returns a polygon which covers a geometry. Unlike convex hulls, which also cover
+/// their geometry, a concave hull does so while trying to further minimize its area by
+/// constructing edges such that the exterior of the polygon incorporates points that would
+/// be interior points in a convex hull.
+///
+/// This implementation is inspired by https://github.com/mapbox/concaveman
+/// and also uses ideas from the following paper:
+/// www.iis.sinica.edu.tw/page/jise/2012/201205_10.pdf
+///
+/// # Example
```
frewsxcv, line 22:
```suggestion
/// # Examples
///
```

https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#using-markdown

geo/src/algorithm/concave_hull.rs
```diff
@@ -0,0 +1,464 @@
...
+    let search_dist = T::div(T::sqrt(T::powi(w, 2) + T::powi(h, 2)), two);
+    let centroid = line.centroid();
+    let centroid_coord = Coordinate {
+        x: centroid.x(),
+        y: centroid.y(),
+    };
+    let mut candidates = interior_coords_tree
+        .locate_within_distance(centroid_coord, search_dist)
+        .peekable();
+    let peek = candidates.peek();
```
frewsxcv, line 134: if you add a `?` here, you can avoid doing the big `match` below

```suggestion
    let peek = candidates.peek()?;
```

geo/src/algorithm/concave_hull.rs
```diff
@@ -0,0 +1,464 @@
...
+
+    let interior_coords: Vec<Coordinate<T>> = coords
+        .iter()
+        .filter_map(|coord| {
+            if !hull_tree.contains(coord) {
+                Some(*coord)
+            } else {
+                None
+            }
+        })
```
frewsxcv, line 210: i think this is the same?

```suggestion
        .filter(|coord| !hull_tree.contains(coord))
```

frewsxcv (commented):
thanks so much for adding this! none of my comments are blocking

geo/src/algorithm/concave_hull.rs
RestitutorOrbis: When I try to do this, I get borrow checker errors that I'll need to work through so I'm going to elect to not proceed with this one.

geo/src/algorithm/concave_hull.rs
RestitutorOrbis: Thanks for catching that! I've made a change to account for this.

rmanoka:
@RestitutorOrbis Thanks for this effort, and your patience!  Let's leave the norway test in for now.

bors r+

michaelkirk:
Woohoo, thanks @RestitutorOrbis!

PR #482, opened 2020-07-26
Fix numerical stability in computing area

### LINKED ISSUE #481: Numerical stability of area
The current implementation of area is not the most numerically stable.  The issue surfaces particularly in cases where the centroid of the polygon is far from the origin.  Eg. a square centered around (1e10, 1e10).   Please see below for a concrete test

``` rust
    #[test]
    fn area_polygon_numerical_stability() {
        use std::f64::consts::PI;
        use std::f64::MAX;
        use geo_types::*;

        let polygon = {
            let n = 4;
            let angle_increment = 2. * PI / n as f64;

            let mut coords = vec![];
            for i in 0..n {
                let angle = i as f64 * angle_increment;
                coords.push(Coordinate{x: angle.cos(), y: angle.sin()});
            }
            Polygon::new(coords.into(), vec![])
        };
        use crate::map_coords::MapCoords;
        use crate::area::Area;
        let area = polygon.signed_area();

        let shift_x = 1.5e8;
        let shift_y = 1.5e8;
        let polygon = polygon.map_coords(
            |&(x, y)| (x + shift_x, y + shift_y)
        );
        let new_area = polygon.signed_area();
        let err = (area - new_area).abs();
        eprintln!("Error: {}", err);
        eprintln!("Relative error: {}", err / area);
        eprintln!("Area: {}", area);
        eprintln!("Shifted Area: {}", new_area);
        assert!(err < 1e-2);
    }
```

The output is:

```
running 1 test
Error: 2
Relative error: 1
Area: 2
Shifted Area: 0
```

Note that the shift used above is fairly small compared to `f64::MAX`.  With much larger shifts, the shifted area even becomes NaN as the formula for the area is a degree-2 polynomial in the coordinates.

A good fix would be to alter `twice_signed_ring_area` to compute the centroid, and subtract it out from the line-string segments before computing the determinant.  Alternatively, we could also mention the stability in the docs, and expect the user to do the same.   In GIS usage, we've encountered this as an issue when working with UTM projected coordinates, where the numbers tend to be far from the origin.

Please let me know if you would consider fixing this, and I can provide a PR to alter `twice_signed_ring_area`.

rmanoka:
We modify `fn twice_signed_ring_area` in `winding_order.rs` to first shift the coordinates by the min over the geometry, before computing the area.  The logic is correct as long as the input `LineString` is closed.  To the best of my understanding, this is the case.

Also added a test to verify stability.  Closes #481

### DIFF
--- a/geo/src/algorithm/area.rs
+++ b/geo/src/algorithm/area.rs
@@ -261,6 +261,42 @@ mod test {
         assert_relative_eq!(polygon.signed_area(), 30.);
     }
     #[test]
+    fn area_polygon_numerical_stability() {
+        let polygon = {
+            use std::f64::consts::PI;
+            const NUM_VERTICES: usize = 10;
+            const ANGLE_INC: f64 = 2. * PI / NUM_VERTICES as f64;
+
+            Polygon::new(
+                (0..NUM_VERTICES).map(|i| {
+                    let angle = i as f64 * ANGLE_INC;
+                    Coordinate {
+                        x: angle.cos(),
+                        y: angle.sin(),
+                    }
+                })
+                    .collect::<Vec<_>>()
+                    .into(),
+                vec![],
+            )
+        };
+
+        let area = polygon.signed_area();
+
+        let shift_x = 1.5e8;
+        let shift_y = 1.5e8;
+
+        use crate::map_coords::MapCoords;
+        let polygon = polygon.map_coords(
+            |&(x, y)| (x + shift_x, y + shift_y)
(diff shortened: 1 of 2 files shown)


rmanoka:
As per discussions with @urschrei , we now check and return 0 in `twice_signed_ring_area` if the LineString is not closed.  The only visible change due to this is that open `LineString` will not return `None` for `winding_order`;  this seems like the correct behavior.  Also added a test for this case.

PR #485, opened 2020-08-05
implement centroid for multilinestring

mcassels:


geo/src/algorithm/centroid.rs
```diff
@@ -129,6 +129,54 @@ where
     }
 }
 
+impl<T> Centroid<T> for MultiLineString<T>
+where
+    T: Float + FromPrimitive + Sum,
+{
+    type Output = Option<Point<T>>;
+
+    // The Centroid of a MultiLineString is the mean of the middles of all the constituent linestrings,
```
urschrei, line 138: Could you reword this slightly into "…the mean of the _centroids_…", and turn it into a doc comment (`///`)?

mcassels: done!

PR #492, opened 2020-08-17
addendum to #490: fixup tests 

michaelkirk:
Rect's, which are defined by two points, maintain the invariant that `min` is up and left of `max`, else we panic.

#490, as I understand it, makes the point that there is no reason to assume that mapping the min point (e.g. via `map_coords`) will output a point which is still the min, so normalizes the output to produce a valid rect.

However, there were two tests explicitly checking for the old failing behavior. This updates those tests.


michaelkirk:
CI is currently broken for some features, but merging after tests passed locally.

PR #493, opened 2020-08-17
Improve efficiency of Area calculation for closed LineString

rmanoka:


geo/src/algorithm/winding_order.rs
```diff
@@ -27,15 +27,7 @@ where
     // of the coordinates, but it is not fool-proof to
     // divide by the length of the linestring (eg. a long
     // line-string with T = u8)
```
michaelkirk, line 29: super nitpicky, but maybe we should update the comment? We're no longer "Comput[ing] a reasonable shift" we're just picking one.

Also, and not new to this particular change, but I don't understand how this technique works. That's no really surprising, since I don't know much about numerical methods, but is there a link or a well known name for this technique where I could learn more?

rmanoka: Good point, will change the wording.

On numerical stability, the general rule of thumb is to ensure the operands are similar in magnitude to the output.  Without the shift, the operands could be much larger than the final answer, and was causing rounding errors (see test `area_polygon_numerical_stability` in area.rs).

For the shoelace formula, our shifting approach mimics what [CGAL does](https://github.com/CGAL/cgal/blob/releases/CGAL-5.1-beta2/Cartesian_kernel/include/CGAL/Cartesian/function_objects.h#L919).  They compute the area by [summing up the area of the constituent triangles](https://github.com/CGAL/cgal/blob/releases/CGAL-5.1-beta2/Polygon/include/CGAL/Polygon_2_algorithms.h#L191), but the shifting is the same as ours (using the first point).

frewsxcv:
@rmanoka thanks for all the recent contributions! i sent you an invite to join the github org if you're interested

rmanoka:
@frewsxcv Thanks for the invite!  Happy to join and contribute to the org.

geo/src/algorithm/winding_order.rs
michaelkirk: Thanks for the explanation.

PR #494, opened 2020-08-17
Fix Haversine formulas yet again

urschrei:
We were slightly off with our mean earth radius constant.
The IUGG defines it as 6,371.0088 km (https://doi.org/10.1007%2Fs001900050278)

### DIFF
--- a/geo/src/lib.rs
+++ b/geo/src/lib.rs
@@ -65,7 +65,7 @@ mod utils;
 extern crate approx;
 
 // Mean radius of Earth in meters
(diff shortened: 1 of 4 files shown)


geo/src/lib.rs
```diff
@@ -65,7 +65,7 @@ mod utils;
 extern crate approx;
 
 // Mean radius of Earth in meters
-const MEAN_EARTH_RADIUS: f64 = 6_371_000.0;
+const MEAN_EARTH_RADIUS: f64 = 6371008.8;
```
frewsxcv, line 68: can we add a comment here explaining where this value comes from?

PR #496, opened 2020-08-17
Bump proj version to remove OpenSSL dependency

urschrei:


michaelkirk:
bors try

geo/CHANGES.md
```diff
@@ -1,5 +1,10 @@
 # Changes
 
+## Unreleased
+
+* Bump proj version to 0.20.3
```
michaelkirk, line 5: In the past, I haven't really paid attention to the release process, but shouldn't this should go in the 0.14.2 section which isn't yet released?

urschrei: Oh, I didn't realise that we'd already bumped the version number in Cargo in preparation!

PR #501, opened 2020-08-21
Switch to the Robust crate for predicate checks

urschrei:
We're currently always up-converting to `f64`, but I couldn't see a clean way of making `CoordinateType` and `Into<f64>` interact, so any feedback is welcome.


urschrei:
Closing in favour of #502

PR #502, opened 2020-08-21
Robust Winding Order

rmanoka:
This PR introduces a few traits to abstract the predicates logic for different scalar types.

+ Introduces `Kernel`, and `HasKernel` traits to allow customizing the predicates used for different scalars (integer vs float)
+ Defines `RobustKernel` meant for float types (`f32`, `f64`)
+ Defines `SimpleKernel` meant for exact arithmetic types (`i32`, `i64`; can be extended to support arbitrary precision types)

Also port a few key, but simple algos to use the predicates.  This improves the correctness of these algos.

+ `Winding`:  shift from area logic into using the `orient2d` logic.
+ Convex Hull:  provide an alternative `graham_hull` (Graham Scan implementation) using the predicates.  This has the advantage of being more correct, and also works on geoms,with int. coords.


rmanoka:
Requesting review / discussion and merging of this PR.  Currently, the new code is fairly non-intrusive (separate traits, that could later replace existing ones).  Here are the key additions:

1.  `Kernel` trait to abstract the predicate logic.  We currently have one naive implementation, used for `i32`, `i64`; and the robust one used for `f32`, `f64`.  Includes fairly comprehensive test of robustness.
2.  `RobustWinding` trait:  same interface as `Winding` trait, but uses the new goodies, and so robust.  If all is good, we can replace the existing `Winding` trait with this.
3.  `graham_hull` algorithm for computing convex_hull (it can be used for integer coords too).  Currently not yet a trait, but we may consider using this algo for `ConvexHull` trait so that it works for all scalar types.

@michaelkirk @frewsxcv @urschrei Please provide feedback.

geo/src/algorithm/kernels/test.rs
```diff
@@ -0,0 +1,168 @@
...
+#[test]
+#[ignore]
+fn test_naive() {
+    let data = orient2d_tests::<SimpleFloat>(
+        12.0, 12.0,
+        0.5, 0.5,
+        24.0, 24.0,
+        256, 256
+    );
+    write_png(&data, Path::new("naive-orientation-map.png"), 256, 256);
```
michaelkirk, line 155: Maybe adding a new [example](https://github.com/georust/geo/tree/master/geo/examples) would be a better fit than a test, since we're not actually `assert`ing anything here.

geo/src/algorithm/mod.rs
```diff
@@ -1,3 +1,6 @@
+/// Kernels to compute various predicates
+pub mod kernels;
```
michaelkirk, line 2: Is it useful for this to be public? 

If we keep it private we have more leverage to refactor the API later, it's fine for it to be public if there's a reason, just wanted to make sure.

geo/src/utils.rs
```diff
@@ -127,6 +127,25 @@ where
     }
 }
 
+/// Compute index of the lexicographically least point.
```
michaelkirk, line 130: For my own understanding: lexicographically as opposed to?

geo/src/algorithm/mod.rs
rmanoka: The main use of making Kernel public is for implementing the functionality for custom scalar types.  I think it should be fine to make it `pub(crate)` for now; and expose it once the traits mature.

geo/src/utils.rs
rmanoka: Unlike real numbers, there is no de-facto ordering of points.  Lexicographic is to emphasize that we sort by x coord, and then break ties using y coord.

geo/src/algorithm/kernels/test.rs
rmanoka: Good point;  unfortunately we can't make it an example once Kernel is made private.  However, I think this test / example belongs better in the robust crate;  I'll remove it in favor of adding it there.

rmanoka:
> It seems like there are some correctness wins with the new implementations. Is there a benefit to keeping the old ones at all? If not, I'd be in favor of replacing them.

Makes sense;  will replace the existing Winding trait as this is a more correct impl.

> 
> If there is some legit benefit to keeping the old ones, is there a reason we shouldn't _default_ to the new, more accurate, implementations?
> 
> (similar concern for graham_hull vs. the existing convex hull impl)

Currently `graham_hull` is provided as a free-standing algo, and not as a trait.  I didn't want to introduce too many changes in a single PR.  Also, I'm not completely sure if they have the same performance in practice, so we should do some benchmarks before switching.  It is in this PR to basically show-case the `Kernel` ideas and to provide the functionality for integer geometries (which wasn't previously possible).

> Is this comment outdated? Is there still some question around traits outstanding, and if so, could you restate the question?

Yeah :).  Will edit to top comment and update the current status.

geo/src/utils.rs
urschrei: > Unlike real numbers, there is no de-facto ordering of points. Lexicographic is to emphasize that we sort by x coord, and then break ties using y coord.

Could we make this a comment for future selves / maintainers?

geo/src/algorithm/kernels/mod.rs
```diff
@@ -0,0 +1,66 @@
...
+        (p.x - q.x) * (p.x - q.x) + (p.y - q.y) * (p.y - q.y)
+    }
+}
+
+/// Marker trait to assign Kernel for scalars
+pub trait HasKernel: CoordinateType {
+    type Ker: Kernel<Scalar = Self>;
+}
+
+#[macro_export]
```
urschrei, line 48: Could we explain what this macro does?

urschrei:
This is really great work, @rmanoka. Apart from my two minor comments, I don't think there's anything else that needs to be added here. There are typos here and there, but let's defer those to a separate PR. I have a strong preference to retain the QuickHull implementation, but it's based purely on sentimentality and I wouldn't want to do so at the expense of the Graham implementation and they're both `O(n log n)` on average, so ¯\_(ツ)_/¯

geo-types/src/coordinate.rs
```diff
@@ -9,7 +9,7 @@ use approx::{AbsDiffEq, RelativeEq, UlpsEq};
 /// as an envelope, a precision model, and spatial reference system
 /// information), a `Coordinate` only contains ordinate values and accessor
 /// methods.
-#[derive(Eq, PartialEq, Clone, Copy, Debug, Hash)]
+#[derive(Eq, PartialEq, PartialOrd, Clone, Copy, Debug, Hash)]
```
frewsxcv, line 12: where does `PartialOrd` get used?

rmanoka: It is used in fn `lexicographically_least_index` in utils.rs;  that is used for `Winding` and `graham_hull`.

rmanoka: Thinking about it,  it does makes things look a bit more magical.  Removed it in favor of describing the order (x then y) explicitly in the utility function.

rmanoka:
> This is really great work, @rmanoka. Apart from my two minor comments, I don't think there's anything else that needs to be added here. There are typos here and there, but let's defer those to a separate PR. I have a strong preference to retain the QuickHull implementation, but it's based purely on sentimentality and I wouldn't want to do so at the expense of the Graham implementation and they're both `O(n log n)` on average, so ¯_(ツ)_/¯

@urschrei The graham scan algorithm has two added benefits:  it uses robust predicates, and also works on integer coords.   The quick hull impl. could be made robust (by fixing `point_location`), but do not see how to get it working on integer coords.  That said, I'm okay with leaving graham scan as a stand-alone algo. that users can import and use if needed.

urschrei:
Noo the point of all this is to integrate robust predicates to any existing traits that can use them, so the convex hull algorithm should be using Graham by default! We can fix up / revisit Quickhull later on if need be.

rmanoka:
> Noo the point of all this is to integrate robust predicates to any existing traits that can use them, so the convex hull algorithm should be using Graham by default! We can fix up / revisit Quickhull later on if need be.

Sounds good!  Let's do that in a separate issue / PR; see a few simple improvements, and bug-fixes to the trait.

rmanoka:
Should we merge this?  @michaelkirk @urschrei @frewsxcv

michaelkirk:
I believe so - thanks for your work @rmanoka !

bors r+

PR #504, opened 2020-08-31
Robustify convex hull impl

rmanoka:
Changes

+ use robust predicates in quick hull

+ port quick hull to work with integer types too

+ fix trait to not use `<T>`
   Many of our algo. traits have a generic parameter, that is not really necessary.  It makes sense when the same type could implement both `Algo<T1>` and `Algo<T2>`, example `Contains`.  In other cases, we just need an associated type.  See for eg. [Iterator] trait in comparison to [Add] trait in rust std.

+ fix quick hull min/max computation logic: see test `quick_hull_test_colinear`

+ fix quick hull on 3 or fewer points: ensure it returns ccw hull and without repeated points

+ add tests for integer coords, and bugfixes

File level changes

+ rename module `convexhull` as `convex_hull`
+ move quick hull and graham hull as submodules
+ move common test utils, trait tests to test submodule

[Iterator]: https://doc.rust-lang.org/std/iter/trait.Iterator.html
[Add]: https://doc.rust-lang.org/nightly/core/ops/trait.Add.html


urschrei:
> fix quick hull min/max computation logic: see test quick_hull_test_colinear

Could you (briefly!) explain how this works? I assume it's related to the lexicographically-least-index function and tie-breaking?

rmanoka:
> Could you (briefly!) explain how this works? I assume it's related to the lexicographically-least-index function and tie-breaking?

@urschrei We were looking at the x coordinate alone for min and max.  If the input has many points with same minimum x coordinate, then the convex hull would sometimes output the collinear points in the middle, depending on the order in which the input is provided.  The collinear points are not _strictly convex_ and the fix ensures the algo. always outputs strictly convex hull which is usually desirable.

I'm not sure if we can also tweak the algo. to always provide the collinear points if requested;  may be `point_location` test can accept `Equal` (based on a extra parameter)?  I'm currently leaving the graham hull implementation for this minor, but useful case.

geo/src/algorithm/convex_hull/graham.rs
```diff
@@ -1,48 +1,58 @@
...
 
     // Allocate output vector
     let mut output = Vec::with_capacity(points.len());
 
-    // Find lexicographically least point
-    use crate::utils::lexicographically_least_index;
-    let min_idx = lexicographically_least_index(points);
+    // Find lexicographically least point and add to hull
+    use crate::utils::least_or_greatest_index;
+    let min_idx = least_or_greatest_index(points, false);
```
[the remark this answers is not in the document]
rmanoka: The function is already private to the crate (`utils` module).  Would you prefer the following syntax that makes the second parameter more clear?

``` rust
let min_idx = least_or_greatest_index(points, Ordering::Less)
```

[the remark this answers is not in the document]
michaelkirk: Yeah that's somewhat clearer.

Comprehension wise, I still think it's harder to understand what's going on here:
```
let biggest = min_or_max(10, 2, Ordering::Less);
let smallest = min_or_max(10, 2, Ordering::More);
```

vs:
```
let biggest = min(10, 2);
let smallest = max(10, 2);
```

geo/src/algorithm/convex_hull/graham.rs
```diff
@@ -0,0 +1,165 @@
...
+                .enumerate()
+                .filter_map(|(i, pt)| {
+                    // Do not care if first and last are
+                    // same, as we anyway close the linestring
+                    if i == 0 || pt != &points[i - 1] {
+                        Some(*pt)
+                    } else {
+                        None
+                    }
+                })
```
frewsxcv, line 32: is this the same?

```suggestion
                .filter(|(i, pt)| {
                    // Do not care if first and last are
                    // same, as we anyway close the linestring
                    i == 0 || pt != &points[i - 1]
                })
```

geo/src/algorithm/convex_hull/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+            .enumerate()
+            .filter_map(|(i, pt)| {
+                // Do not care if first and last are
+                // same, as we anyway close the linestring
+                if i == 0 || pt != &points[i - 1] {
+                    Some(*pt)
+                } else {
+                    None
+                }
+            })
```
frewsxcv, line 134: See my previous comment about using `filter` here

geo/src/algorithm/convex_hull/graham.rs
rmanoka: Yeah, the logic is the same.  However, if we just filter, we'll end up with Vec<(usize, Coord)>, but we want only the second component.  

Good catch of the repeated code.  I'd abstracted it out as `trivial_hull` but forgot to change the graham impl.

frewsxcv: Ah yeah gotcha. Also, is this the same as using https://doc.rust-lang.org/std/vec/struct.Vec.html#method.dedup ?

rmanoka: Yeah, exactly;  never knew this method!   #TIL

geo/src/algorithm/convex_hull/graham.rs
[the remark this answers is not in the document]
rmanoka: I didn't like to pollute utils with too many functions.  In this case, it turns out `greatest_index` is not necessary anymore, so have reverted back to a single `least_index` function.

rmanoka:
@michaelkirk @frewsxcv Trying to merging this PR as the suggestions were mostly cosmetic, and have been addressed.  Please revert if this is a mistake!

bors r+

rmanoka:
Hmm, bors doesn't like me!  Merging manually.

michaelkirk:
> Hmm, bors doesn't like me! Merging manually.

Just tried to add you @rmanoka and looks like someone beat me to it. Let me know if it's still not working for you!

PR #505, opened 2020-09-04
Robust ExtremePoint trait

rmanoka:
This PR introduces `IsConvex` trait, and partially revamps the `ExtremePoint` trait.

1. add `IsConvex` trait and documentation of predicates.

1. add vector space ops in `Coordinate`. These were earlier
available in `Point`, but it makes more sense to be
available for `Coordinate` too.

1. many minor doc improvements:  explain corner cases,
better references, fix spellings (collinear).

1. make `ExtremePoint` robust
1. Remove unneeded trait generic parameter `T`, and requirement `T: Float`

Things to revamp / discuss in `ExtremePoint`

1. FIgure out what the trait is about.  Should we really compute `convex_hull` in the impl?
1. Replace `Result<Extremes, ()>` with `Option`

### DIFF
--- a/geo-types/src/coordinate.rs
+++ b/geo-types/src/coordinate.rs
@@ -76,17 +76,9 @@ where
     pub fn x_y(&self) -> (T, T) {
         (self.x, self.y)
     }
-
-    /// Scale the coordinates by a scalar value.
-    pub fn scale_by(&self, scale: T) -> Self {
-        Coordinate {
-            x: scale * self.x,
-            y: scale * self.y,
-        }
-    }
 }
 
-use std::ops::{Add, Neg, Sub};
+use std::ops::{Add, Neg, Sub, Mul, Div};
 impl<T> Neg for Coordinate<T>
 where
     T: CoordinateType + Neg<Output = T>,
@@ -158,6 +150,54 @@ where
     /// ```
     fn sub(self, rhs: Coordinate<T>) -> Coordinate<T> {
         (self.x - rhs.x, self.y - rhs.y).into()
+    }
+}
+
+impl<T> Mul<T> for Coordinate<T>
+where
+    T: CoordinateType,
+{
+    type Output = Coordinate<T>;
+
+    /// Add a point to the given point.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use geo_types::Coordinate;
+    ///
+    /// let p: Coordinate<_> = (1.25, 2.5).into();
+    /// let q: Coordinate<_> = p * 4.;
+    ///
+    /// assert_eq!(q.x, 5.0);
+    /// assert_eq!(q.y, 10.0);
+    /// ```
+    fn mul(self, rhs: T) -> Coordinate<T> {
+        (self.x * rhs, self.y * rhs).into()
+    }
+}
+
+impl<T> Div<T> for Coordinate<T>
+where
+    T: CoordinateType,
+{
+    type Output = Coordinate<T>;
+
+    /// Add a point to the given point.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use geo_types::Coordinate;
+    ///
+    /// let p: Coordinate<_> = (5., 10.).into();
+    /// let q: Coordinate<_> = p / 4.;
+    ///
+    /// assert_eq!(q.x, 1.25);
+    /// assert_eq!(q.y, 2.5);
+    /// ```
+    fn div(self, rhs: T) -> Coordinate<T> {
+        (self.x / rhs, self.y / rhs).into()
     }
 }
 
--- a/geo-types/src/coordinate.rs
+++ b/geo-types/src/coordinate.rs
@@ -10,10 +10,9 @@ use approx::{AbsDiffEq, RelativeEq, UlpsEq};
 /// information), a `Coordinate` only contains ordinate values and accessor
 /// methods.
 ///
-/// This type obeys the typical [vector space] structure:
-/// implements the [`Add`], [`Sub`], [`Neg`], [`Zero`]
-/// traits and allows [`scaling`][`Coordinate::scale_by`] by
-/// a scalar.
+/// This type implements the [vector space] operations:
+/// [`Add`], [`Sub`], [`Neg`], [`Zero`],
+/// [`Mul<T>`][`Mul`], and [`Div<T>`][`Div`] traits.
 ///
 /// [vector space]: //en.wikipedia.org/wiki/Vector_space
 #[derive(Eq, PartialEq, Clone, Copy, Debug, Hash)]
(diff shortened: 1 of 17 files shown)


rmanoka:
@urschrei @frewsxcv The `Polygon` type has an `is_convex` method that is somewhat buggy:  is not robust, and handles collinear points in an unexpected manner (it allows collinear pts if oriented cw, but not otherwise).   We can not directly make it robust as the kernels are part of `geo` crate, and not `geo-types`.   

Should we remove that functionality from `geo-types` in favor of introducing it as a trait in `geo`?  This way, we can add a more exhaustive list of (robust) predicates:  `is_strongly_ccw_convex` `is_strongly_cw_convex` `is_strongly_convex` `is_ccw_convex` `is_cw_convex` `is_convex`.

It is only used in a couple of `geo` traits, and we can fix that.  Is a breaking change though as it is a `pub fn`.

michaelkirk:
> It is only used in a couple of geo traits, and we can fix that. Is a breaking change though as it is a pub fn.

This makes sense to me. It would be kinder to deprecate it, change the internal impls to use the robust versions, and remove it after some time has passed.

geo-types/src/coordinate.rs
```diff
@@ -69,6 +69,90 @@ where
     pub fn x_y(&self) -> (T, T) {
         (self.x, self.y)
     }
+
+    /// Construct a point at the origin, that is, with both
+    /// `x` and `y` coordinates as zero.
+    pub fn zero() -> Self {
```
michaelkirk, line 75: ❤️ I've wanted this little method many times.

rmanoka:
@michaelkirk @urschrei  I'm quite confused about the `ExtremePoint` trait.  The implementation seems to simply compute the coordinates with [min-x, min-y, max-x, and max-y](https://github.com/georust/geo/blob/eec553b52554e70f30bdcbbeb6178a42ca82fbe2/geo/src/algorithm/extremes.rs#L58).  However, for doing this, it seems to be running the [convex hull algorithm](https://github.com/georust/geo/blob/eec553b52554e70f30bdcbbeb6178a42ca82fbe2/geo/src/algorithm/extremes.rs#L133)!  Note that convex hull is O(n logn) while just direct search would just be O(n), so I have no idea what is happening.

For now, I'm going to leave this trait as is, and just make a few calls robust.  Some clarity on the purpose of this trait would be very helpful.

urschrei:
Two things:
1. ExtremePoint was an add-on to ExtremeIndices, and the fact that it could be implemented generically is actually a big disadvantage here, as you indirectly point out: in many cases, there's no need to ensure convexity.
2. When it comes to collinearity, are you referring to sequential collinear points, or within some epsilon? I ask because the geometry model (JTS and GEOS, again) allows you to assume no repeating adjacent points in a geometry. Of course, we kicked the can down the road by writing non-validating constructors because we were worried about performance…

rmanoka:
@urschrei Got it.  We currently use `polymax_naive` but could switch to the binary search version at some point.  Still, we can't compute `convex_hull` just for this purpose, as we might as well just do a loop through instead.  The binary search version could probably be provided as an independent algo. that expects (but does not verify) its input to be a convex linestring.

rmanoka:
On a side note:  the check whether to use the "rotating calipers" fast path may be wrong:

```rust
if poly2.is_convex() || !self.is_convex()
```

Shouldn't it check if both are convex?  It seems to want one of them to not be convex.

geo/src/algorithm/winding_order.rs
```diff
@@ -29,13 +29,20 @@ pub enum WindingOrder {
     CounterClockwise,
 }
 
+/// Determine and operate on how a [`LineString`] is
```
michaelkirk, line 32: Thanks for all the good docs.

geo-types/src/polygon.rs
```diff
@@ -387,6 +387,7 @@ where
     // The polygon is convex if the z-components of the cross products are either
     // all positive or all negative. Otherwise, the polygon is non-convex.
     // see: http://stackoverflow.com/a/1881201/416626
+    #[deprecated(since = "0.6.1", note = "Please use `geo::is_convex` instead")]
```
michaelkirk, line 390: Seems like currently there is no `impl IsConvex for Polygon` in `geo::algorithm::IsConvex`, so we can't call `geo::is_convex(&poly)`.

Is there a reason not to add that impl?

And if there is some reason to omit that Impl, then is the suggestion here, more explicitly, to do something like `self.exterior.is_convex() && self.interios.is_empty()`?

geo/src/algorithm/extremes.rs
```diff
@@ -1,91 +1,87 @@
...
+    use crate::is_convex::IsConvex;
+    if !polygon.exterior().is_convex() {
         return Err(());
     }
-    let directions = vec![
-        Point::new(T::zero(), -T::one()),
-        Point::new(T::one(), T::zero()),
-        Point::new(T::zero(), T::one()),
-        Point::new(-T::one(), T::zero()),
+    let directions: Vec<Coordinate<_>> = vec![
```
michaelkirk, line 54: no biggie so take it or leave it, but since this is fixed size we could make this an array instead of a vec.

geo/src/algorithm/is_convex.rs
```diff
@@ -0,0 +1,248 @@
...
+///
+/// - A closed `LineString` with at most three coordinates
+/// (including the possibly repeated first coordinate) is
+/// both convex and collinear. However, the strict convexity
+/// is _unspecified_ and must not be relied upon.
+///
+/// [convex combination]: //en.wikipedia.org/wiki/Convex_combination
+/// [convex set]: //en.wikipedia.org/wiki/Convex_set
+/// [`ConvexHull`]: crate::algorithm::convex_hull::ConvexHull
+/// [`graham_hull`]: crate::algorithm::convex_hull::graham_hull
```
michaelkirk, line 41: excellent docs!

geo/src/algorithm/is_convex.rs
```diff
@@ -0,0 +1,248 @@
...
+                }
+            }
+            _ => Some((i, orientation)),
+        }
+    });
+
+    let (i, first_non_collinear) = if let Some((i, orientation)) = find_first_non_collinear {
+        match orientation {
+            Orientation::Collinear => {
+                // Only happens if !allow_collinear
```
michaelkirk, line 172: take it or leave it: `assert!(!allow_collinear);`

geo/src/algorithm/is_convex.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    let find_first_non_collinear = (0..n).map(orientation_at).find_map(|(i, orientation)| {
+        match orientation {
+            Orientation::Collinear => {
+                // If collinear accepted, we skip, otherwise
+                // stop.
+                if allow_collinear {
+                    None
+                } else {
+                    Some((i, orientation))
```
michaelkirk, line 162: A little clarity nit: 

At this point,  find_first_non_collinear is in fact collinear, right?

To me that reads like `let not_a_dog = 🐕;` 

Would something like `starting_orientation` be clearer than `find_first_non_collinear`?

No biggie though, it just tripped me up as I was trying to understand.

michaelkirk (approved):
LGTM!

All my feedback was optional.

geo-types/src/polygon.rs
rmanoka: Great catch!  The mathematical definition of convex for polygons is a bit different from what is provided as `is_convex` in geo-types.  In particular, the inner holes are not considered.  I've updated the doc to suggest is_convex on the exterior.

geo-types/src/coordinate.rs
```diff
@@ -9,6 +9,13 @@ use approx::{AbsDiffEq, RelativeEq, UlpsEq};
 /// as an envelope, a precision model, and spatial reference system
 /// information), a `Coordinate` only contains ordinate values and accessor
 /// methods.
+///
+/// This type obeys the typical [vector space] structure:
+/// implements the [`Add`], [`Sub`], [`Neg`], [`Zero`]
+/// traits and allows [`scaling`][`Coordinate::scale_by`] by
```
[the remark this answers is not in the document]
rmanoka: Indeed, have updated the docs

geo/src/algorithm/euclidean_distance.rs
```diff
@@ -456,7 +456,8 @@ where
             }
             return mindist;
         }
-        if poly2.is_convex() || !self.is_convex() {
+        use super::is_convex::IsConvex;
+        if !poly2.exterior().is_convex() || !self.exterior().is_convex() {
```
michaelkirk, line 460: 👍 I'll close #509

geo/src/algorithm/is_convex.rs
rmanoka: So, the first loop is in fact finding the first non-collinear orientation;  we return collinear in that loop only if !allow_collinear, and as a fast exit path.  The apparent inversion is unfortunately how `find` / `find_map` works:  it shorts at the first `Some()` return from the closure.

geo/src/algorithm/euclidean_distance.rs
rmanoka: Oh yeah, I figured it's a typo and fixed it as even the comments above the code mention that both should be convex.

geo/src/algorithm/extremes.rs
rmanoka: I don't think fixed size works here because it is not an iterator.  We can re-visit this once we have the `ForEachCoord` trait, and simplify the logic here.

rmanoka:
bors r=michaelkirk

PR #506, opened 2020-09-06
proj network feature

michaelkirk:
Once https://github.com/georust/proj/pull/39 is merged, if someone opts into the `proj` feature, we'll no longer build the recently added network/grid stuff (it has a lot of deps).

This new flag let's them opt-in to that.

To be clear, the default build doesn't include proj - this is only relevant to folks who want to use the optional proj features.

### DIFF
--- a/geo/Cargo.toml
+++ b/geo/Cargo.toml
@@ -29,6 +29,7 @@ robust = { version = "0.2.2" }
 [features]
 default = []
 use-proj = ["proj"]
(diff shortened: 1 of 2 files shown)


michaelkirk:
Don't know why it took me so long, but this is ready for review.

urschrei:
This lgtm!

michaelkirk:
bors r=urschrei

PR #507, opened 2020-09-07
scalar mul/div for point

### LINKED ISSUE #171: Primitive operations on Points

michaelkirk:
fixes #171

### DIFF
--- a/geo-types/src/point.rs
+++ b/geo-types/src/point.rs
@@ -1,8 +1,6 @@
 use crate::{Coordinate, CoordinateType};
 use num_traits::Float;
-use std::ops::Add;
-use std::ops::Neg;
-use std::ops::Sub;
+use std::ops::{Add, Div, Mul, Neg, Sub};
 
 /// A single point in 2D space.
 ///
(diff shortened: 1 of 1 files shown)


rmanoka:
@michaelkirk I'm not against this, but I think it is best to provide "vector space" operations (add, sub, mul/div by scalar) on `Coordinate` itself and not on `Point`.  This is more of a preference, because it usually doesn't mean much to add points.  There is also some relevant discussion in #15 (esp. https://github.com/georust/geo/issues/15#issuecomment-282563375).

Even if you want to provide these on `Point`, we should also provide these operations on `Coordinate` as almost all composite geometries (LineString, etc) store `Coordinate` and not `Point` (I'm already doing this in #505).  Providing the same on both seems repetitive, and may possibly prevent us from some future functionality on `Point` (eg. coordinate system attached to it, so it no longer makes sense to just add them only by coords).

michaelkirk:
>  I'm not against this, but I think it is best to provide "vector space" operations (add, sub, mul/div by scalar) on Coordinate itself and not on Point.

Good call. I'd support adding: `add`, `sub`, `mul`, `div`, `neg` to Coordinate.

> may possibly prevent us from some future functionality on Point (eg. coordinate system attached to it, so it no longer makes sense to just add them only by coords).

That's a reasonable concern, but unfortunately that ship has already sailed. `add`, `sub`, `neg` exist for Point for a while now.

Since `add`, `sub`, `neg` for Point already exist, `div` `mul` seemed like an obvious complement. 

I want to avoid a situation where Point and Coord have separate subsets of vectors methods available to them, so I think either:

1. we have redundant operations defined on both of them (which is my slight preference, breaking things the least for current users)

-or- 

2. deprecate the existing ones on Point, and eventually only support them on Coordinate

michaelkirk:
> we have redundant operations defined on both of them (which is my slight preference, breaking things the least for current users)

Actually, I don't know that I prefer this method more. I could be talked out of it. It's just lame to break API's all the time. 😛

michaelkirk:
Looking again at #505, I'm happy to use what you have @rmanoka and just close this.

rmanoka:
>     1. we have redundant operations defined on both of them (which is my slight preference, breaking things the least for current users)
> 

This sounds fine to me.  We can deprecate the ops for `Point` at a later point (ha!) if it makes sense.

Regarding #505, I missed out Div by scalar, so I'll copy your changes into that PR.

PR #510, opened 2020-09-08
Fix Centroid numerical stability

rmanoka:
This issue, and the fix is similar to the one we fixed earlier with `Area`.  The centroid formula for polygons is quadratic in the coordinates, and has cancellation errors.

Also, added a test to verify stability.


michaelkirk (approved):
bors r+

LGTM - sorry for the slow review!

Issue #415, opened 2020-01-19
Use robust crate for orientation tests


urschrei:
See previous discussion: #380,

We'll need to resolve the following issues around `num` and compatibility first: https://github.com/georust/robust/issues/3, https://github.com/georust/robust/issues/5

---

Update: this issue now tracks the progress in converting the algo. traits to use the robust predicates.

- [x] Kernel abstraction (#502)
- [x] Orient and WindingOrder (#502)
- [x] ConvexHull (#504)
- [x] ExtremePoints (#505)
- [x] Contains (#511)
- [x] Intersects (#511)

frewsxcv:
Was this completed in https://github.com/georust/geo/pull/502? cc @rmanoka

urschrei:
I _think_ we've got some instances left to track down, but need to have a careful look.

rmanoka:
Yeah, we use it now for Winding and Orient traits.  Contains , Intersects are the prominent ones that need to use it yet.

rmanoka:
Converting this issue into a tracker for making the existing traits use robust preds.

PR #511, opened 2020-09-13
Robust Intersect and Contains

rmanoka:
Use robust predicates in Intersects and Contains traits

+ use robust preds (doesn't need Float anymore)
+ define interior, boundary, and exterior of geometry types (similar to geos/jts)
+ define Intersect and Contains using DE-9IM model
+ fix contains logic in a few corner cases

Future work:

The polygon contains line / polygon logic is not yet fully correct.   I've added a few TODO comments, and an ignored test for cases where our predicate doesn't match the DE-9IM definition.  Unfortunately, it doesn't seem to be a simple change to fix these corner cases.  For instance, geos seems to use a concept called `GeometryGraph` for these cases which I'm yet to understand.

### DIFF
--- a/geo-types/CHANGES.md
+++ b/geo-types/CHANGES.md
@@ -1,5 +1,10 @@
 # Changes
 
+## Master (not released)
+
+* Add vector-space operations to `Coordinate` and `Point`
+  * <https://github.com/georust/geo/pull/505>
+
 ## 0.6.0
 
 * Remove `COORD_PRECISION` which was an arbitrary constant of 0.1m
--- a/geo-types/src/line.rs
+++ b/geo-types/src/line.rs
@@ -1,7 +1,7 @@
 use crate::{Coordinate, CoordinateType, Point};
 
 /// A line segment made up of exactly two
-/// [`Point`s](struct.Point.html). The interior and
+/// [`Coordinate`s](struct.Coordinate.html). The interior and
 /// boundaries are defined as with a `LineString` with the
 /// two end points.
 #[derive(Eq, PartialEq, Clone, Copy, Debug, Hash)]
--- a/geo-types/src/line.rs
+++ b/geo-types/src/line.rs
@@ -1,6 +1,9 @@
 use crate::{Coordinate, CoordinateType, Point};
 
-/// A line segment made up of exactly two [`Point`s](struct.Point.html).
+/// A line segment made up of exactly two
+/// [`Point`s](struct.Point.html). The interior and
+/// boundaries are defined as with a `LineString` with the
+/// two end points.
 #[derive(Eq, PartialEq, Clone, Copy, Debug, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct Line<T>
--- a/geo-types/src/line.rs
+++ b/geo-types/src/line.rs
@@ -37,6 +37,11 @@ where
         }
     }
 
+    /// Calculate the difference in coordinates (Δx, Δy).
+    pub fn delta(&self) -> Coordinate<T> {
+        self.end - self.start
+    }
+
     /// Calculate the difference in ‘x’ components (Δx).
     ///
     /// Equivalent to:
@@ -53,7 +58,7 @@ where
     /// # );
     /// ```
     pub fn dx(&self) -> T {
-        self.end.x - self.start.x
+        self.delta().x
     }
 
     /// Calculate the difference in ‘y’ components (Δy).
@@ -72,7 +77,7 @@ where
     /// # );
     /// ```
     pub fn dy(&self) -> T {
-        self.end.y - self.start.y
+        self.delta().y
     }
 
     /// Calculate the slope (Δy/Δx).
--- a/geo-types/src/line_string.rs
+++ b/geo-types/src/line_string.rs
@@ -15,9 +15,9 @@ use std::ops::{Index, IndexMut};
 ///
 /// # Validity
 ///
-/// A `LineString` is valid if it does not self-intersect
-/// except possibly at the end points, and is either empty
-/// or contains 2 or more coordinates. Note that the
+/// A `LineString` is valid if it is either empty or
+/// contains 2 or more coordinates. Further, a closed
+/// `LineString` must not self intersect. Note that the
 /// validity is not enforced, and the operations and
 /// predicates are undefined on invalid linestrings.
 ///
--- a/geo-types/src/line_string.rs
+++ b/geo-types/src/line_string.rs
@@ -2,8 +2,24 @@ use crate::{Coordinate, CoordinateType, Line, Point, Triangle};
 use std::iter::FromIterator;
 use std::ops::{Index, IndexMut};
 
-/// An ordered collection of two or more [`Coordinate`s](struct.Coordinate.html), representing a
+/// An ordered collection of two or more
+/// [`Coordinate`s](struct.Coordinate.html), representing a
 /// path between locations.
+///
+/// A `LineString` is _closed_ if it is empty, or if the
+/// first and last coordinates are the same. The _boundary_
+/// of a `LineString` is empty if closed, and otherwise the
+/// end points. The interior is the (infinite) set of all
+/// points along the linestring _not including_ the
+/// boundary.
+///
+/// # Validity
+///
+/// A `LineString` is valid if it does not self-intersect
+/// except possibly at the end points, and is either empty
+/// or contains 2 or more coordinates. Note that the
+/// validity is not enforced, and the operations and
+/// predicates are undefined on invalid linestrings.
 ///
 /// # Examples
 ///
@@ -180,11 +196,9 @@ impl<T: CoordinateType> LineString<T> {
         self.0.len()
     }
 
-    /// Checks if the linestring is closed; i.e. the first
-    /// and last points have the same coords. Note that a
-    /// single point is considered closed, but the output on
-    /// an empty linestring is _unspecified_ and must not be
-    /// relied upon.
+    /// Checks if the linestring is closed; i.e. it is
+    /// either empty or, the first and last points are the
+    /// same.
     ///
     /// # Examples
     ///
--- a/geo/src/algorithm/contains/line.rs
+++ b/geo/src/algorithm/contains/line.rs
@@ -53,13 +53,32 @@ where
             return false;
         }
 
-        // A closed linestring contains the start point in
-        // its interior and hence, must be within the line
-        // too.
-        if linestring.is_closed() && !self.contains(&linestring.0[0]) {
-            return false;
-        }
+        // The interior of the linestring should have some
+        // intersection with the interior of self. Two cases
+        // arise:
+        //
+        // 1. There are at least two distinct points in the
+        // linestring. Then, if both intersect, the interior
+        // between these two must have non-empty intersection.
+        //
+        // 2. Otherwise, all the points on the linestring
+        // are the same. In this case, the interior is this
+        // specific point, and it should be contained in the
+        // line.
+        let first = linestring.0.first().unwrap();
+        let mut all_equal = true;
 
-        linestring.0.iter().all(|c| self.intersects(c))
+        // If all the vertices of the linestring intersect
+        // self, then the interior or boundary of the
+        // linestring cannot have non-empty intersection
+        // with the exterior.
+        let all_intersects = linestring.0.iter().all(|c| {
+            if c != first {
+                all_equal = false;
+            }
+            self.intersects(c)
+        });
+
+        all_intersects && (!all_equal || self.contains(first))
     }
 }
--- a/geo/src/algorithm/contains/line.rs
+++ b/geo/src/algorithm/contains/line.rs
@@ -12,7 +12,11 @@ where
     T: HasKernel,
 {
     fn contains(&self, coord: &Coordinate<T>) -> bool {
-        self.intersects(coord)
+        if self.start == self.end {
+            &self.start == coord
+        } else {
+            coord != &self.start && coord != &self.end && self.intersects(coord)
+        }
     }
 }
 
@@ -30,7 +34,11 @@ where
     T: HasKernel,
 {
     fn contains(&self, line: &Line<T>) -> bool {
-        self.contains(&line.start) && self.contains(&line.end)
+        if line.start == line.end {
+            self.contains(&line.start)
+        } else {
+            self.intersects(&line.start) && self.intersects(&line.end)
+        }
     }
 }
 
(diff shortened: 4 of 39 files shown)


rmanoka:
A couple of clarifications needed:

1. These two tests seem inconsistent with definition of `Contains` which says it does not include the boundary.
    https://github.com/georust/geo/blob/a6a3218823c440e9f6d52406b9630fea5cf7f395/geo/src/algorithm/contains.rs#L778
    https://github.com/georust/geo/blob/a6a3218823c440e9f6d52406b9630fea5cf7f395/geo/src/algorithm/contains.rs#L785

2.  Implementation of `Rect: Contains<Point>` also seems to allow the point to be on the edges.  Again, not consistent with definition, and implementation on `Polygon`.  

Also checked against shapely/geos which seems to concur with our definition (so points on edges / vertices of the polygon are not considered to be "contained").  Shall we go with the definition (so fix `Rect` and `Triangle`)?

@michaelkirk @urschrei

michaelkirk:
> Shall we go with the definition (so fix Rect and Triangle)?

Yes. Apparently I wrote that broken triangle implementation 😞: b3944caaa

For my own benefit, I referred to http://portal.opengeospatial.org/files/?artifact_id=25355 (thanks @urschrei :wink:)

> **Contains**
>   `a.Contains(b) ⇔ b.Within(a)`

So it's defined it terms of `within`, which itself is defined as:

> **Within**
> The Within relationship is defined as
> `a.Within(b) ⇔ (a∩b=a) ∧ (I(a)∩E(b)=∅)`
> Expressed in terms of the DE-9IM:
>   ```
> a.Within(b) ⇔ [ I(a)∩I(b)≠∅ ∧ I(a)∩E(b)=∅ ∧ B(a)∩E(b)=∅ ]
>       ⇔ a.Relate(b, “T*F**F***”) 
> ```

And from the [jts impl](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/geom/Geometry.java#L858):
```
contains
...
[T*****FF*]
```

So for `a.contains(b)`, boundary contact is not sufficient; The *interior* of `b` must reside (wholly) within the *interior* of `a`.

In the future, I'll be careful to be rigorous about the cases represented in DE-9IM.

rmanoka:
Yeah, DE-9IM based rigorous definitions of our geometry types, and traits definitely seems useful. I believe jts, geos (hence postgis), shapely all use the same definition of Intersects, and Contains so makes sense for us to with the same for interop with the community.  Will add some doc to the types trying to formalize the definitions.

geo-types/src/line_string.rs
```diff
@@ -2,9 +2,25 @@ use crate::{Coordinate, CoordinateType, Line, Point, Triangle};
...
+/// boundary.
+///
+/// # Validity
+///
+/// A `LineString` is valid if it is either empty or
+/// contains 2 or more coordinates. Further, a closed
+/// `LineString` must not self intersect. Note that the
+/// validity is not enforced, and the operations and
+/// predicates are undefined on invalid linestrings.
+///
```
michaelkirk, line 23: Thanks for continuing to fleshing out the docs. :+1:

geo/src/algorithm/intersects/mod.rs
```diff
@@ -1,278 +1,114 @@
...
-use num_traits::Float;
-
-/// Checks if the geometry A intersects the geometry B.
-
+use crate::*;
+
+/// Checks if the geometry A intersects the geometry B. More
+/// formally, either boundary or interior of A has non-empty
+/// (set-theoretic) intersection with the boundary or
+/// interior of B. In other words, the [DE-9IM] intersection
```
michaelkirk, line 6: Thanks for documenting this! I hope we can get all the traits to this level of specification.

geo/src/algorithm/intersects/mod.rs
```diff
@@ -1,278 +1,114 @@
...
+mod polygon;
+mod rect;
+
+// Since `Intersects` is symmetric, we use a macro to
+// implement `T: Intersects<S>` if `S: Intersects<T>` is
+// available. As a convention, we provide explicit impl.
+// whenever the Rhs is a "simpler geometry" than the target
+// type, and use the macro for the reverse impl.
+#[macro_use]
+macro_rules! symmetric_intersects_impl {
```
michaelkirk, line 52: Nice!

geo/src/algorithm/intersects/mod.rs
```diff
@@ -1,278 +1,114 @@
...
 
-impl<T> Intersects<Line<T>> for Polygon<T>
-where
-    T: Float,
-{
-    fn intersects(&self, line: &Line<T>) -> bool {
-        line.intersects(self)
-    }
-}
+symmetric_intersects_impl!(Line<T>, LineString<T>, HasKernel);
```
michaelkirk, line 72: Not a big deal, so take it or leave it, but for someone trying to find how the intersection logic is actually implemented, it might be helpful to collocate the macro invocation next to the implementation.

i.e. this one could go next to the `impl<T> Intersects<Line<T>> for LineString<T>`

Another benefit to organizing it that way is it might be easier for the implementor (or reviewer 😉)to audit that each `impl` is preceded by the macro invocation than it is to cross reference that the sum of all appropriate calls exists in this file.

geo/src/utils.rs
```diff
@@ -83,43 +84,83 @@ pub enum CoordPos {
...
+            coord,
+            Coordinate {
+                x: max_x,
+                y: coord.y,
+            },
+        );
+        if ray.intersects(&line) {
+            crossings += 1;
         }
+        eprintln!("current crossing @ iter {} = {}", i, crossings);
```
michaelkirk, line 163: did you mean to commit the `eprintln!`?

geo/src/algorithm/intersects/polygon.rs
```diff
@@ -0,0 +1,91 @@
...
+        if self.exterior().intersects(linestring)
+            || self
+                .interiors()
+                .iter()
+                .any(|inner| inner.intersects(linestring))
+        {
+            true
+        } else {
+            // or if it's contained in the polygon
+            linestring.points_iter().any(|point| self.contains(&point))
```
michaelkirk, line 56: Can we iterate over the coords to avoid building a Point in the point iter? Not a big deal, so take it or leave it.

geo/src/algorithm/intersects/polygon.rs
```diff
@@ -0,0 +1,91 @@
...
+    fn intersects(&self, p: &Point<T>) -> bool {
+        self.intersects(&p.0)
+    }
+}
+
+impl<T> Intersects<Line<T>> for Polygon<T>
+where
+    T: HasKernel,
+{
+    fn intersects(&self, p: &Line<T>) -> bool {
```
michaelkirk, line 33: nit: can we call this `l` (or `line`) instead of `p` since it's a **L**ine?

michaelkirk (approved):
I have yet to review the `Contains` portion, but I ran out of time for now. 

I have reviewed the `Intersecs` and other changes, and they look great! Just some optional feedback for clarity. 

I'll finish reviewing up hopefully yet tonight, or if not, tomorrow.

geo-types/src/line.rs
```diff
@@ -1,6 +1,9 @@
 use crate::{Coordinate, CoordinateType, Point};
 
-/// A line segment made up of exactly two [`Point`s](struct.Point.html).
+/// A line segment made up of exactly two
+/// [`Point`s](struct.Point.html). The interior and
```
frewsxcv, line 4:
```suggestion
/// [`Coordinate`s](struct.Coordinate.html). The interior and
```

geo-types/src/line_string.rs
frewsxcv: +1

geo/src/algorithm/contains/mod.rs
```diff
@@ -1,434 +1,48 @@
...
-        self.contains(&point.0)
-    }
-}
+mod geometry;
+mod line;
+mod line_string;
+mod point;
+mod polygon;
+mod rect;
+mod triangle;
```
frewsxcv, line 45: good idea splitting this file up into multiple modules 🏆

geo/src/algorithm/intersects/mod.rs
rmanoka: Good point; will do.

geo/src/utils.rs
rmanoka: Oops!

rmanoka:
Thanks for the comments @frewsxcv @michaelkirk ; I have addressed them, also added a few lines in `CHANGES.md` to highlight the recent robust pred. changes.  Will wait a couple more days to see there are any minor fixes we spot, and merge this PR.

closes #415

geo-types/CHANGES.md
```diff
@@ -1,5 +1,10 @@
 # Changes
 
+## Master (not released)
```
michaelkirk, line 3: Yes, thank you for doing this! Tracking as we go will make the eventual release easier.

geo/src/algorithm/contains/polygon.rs
```diff
@@ -0,0 +1,91 @@
...
+where
+    T: HasKernel,
+{
+    fn contains(&self, p: &Point<T>) -> bool {
+        self.contains(&p.0)
+    }
+}
+
+// TODO: ensure DE-9IM compliance: esp., when
+// line.start and line.end is on the boundaries
```
michaelkirk, line 34: Right as your TODO alludes, it seems like the current logic would fail for a case like:

image

Where the line touches, but does not pass, the interior.
Or if the line touches, but does not pass, the exterior.

geo/src/algorithm/contains/polygon.rs
```diff
@@ -0,0 +1,91 @@
...
+        // both endpoints are contained in the polygon and the line
+        // does NOT intersect the exterior or any of the interior boundaries
+        self.contains(&line.start)
+            && self.contains(&line.end)
+            && !self.exterior().intersects(line)
+            && !self.interiors().iter().any(|inner| inner.intersects(line))
+    }
+}
+
+// TODO: also check interiors
```
michaelkirk, line 49: Right, this one also looks wrong. I wouldn't want to include new broken implementations, but I believe you are just moving these, right?

geo/src/algorithm/contains/polygon.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+}
+
+// ┌──────────────────────────────────┐
+// │ Implementations for MultiPolygon │
+// └──────────────────────────────────┘
+impl<G, T> Contains<G> for MultiPolygon<T>
+where
+    T: CoordinateType,
+    Polygon<T>: Contains<G>,
```
michaelkirk, line 86: It looks like we're newly accepting *any* type which implements the Contains trait, whereas before MultiPolygon only supported Coordinate and Point, is that right?

Does this existing implementation work for types with more dimensionality than a point?

e.g. If the two squares are part of a single multipolygon, even though the black line is contained by neither square individually, is it contained by the multipolygon?

image

geo/src/algorithm/contains/rect.rs
```diff
@@ -0,0 +1,40 @@
...
+    }
+}
+
+impl<T> Contains<Rect<T>> for Rect<T>
+where
+    T: CoordinateType,
+{
+    fn contains(&self, bounding_rect: &Rect<T>) -> bool {
+        // All points of LineString must be in the polygon ?
+        self.min().x <= bounding_rect.min().x
```
michaelkirk, line 35: So a degenerate Rect which is either a Line or a Point will be `contained` if it's on the boundary - I don't think that's correct, is it?

geo/src/algorithm/contains/triangle.rs
```diff
@@ -0,0 +1,27 @@
...
+// └──────────────────────────────┘
+
+impl<T> Contains<Coordinate<T>> for Triangle<T>
+where
+    T: HasKernel,
+{
+    fn contains(&self, coord: &Coordinate<T>) -> bool {
+        let ls = LineString(vec![self.0, self.1, self.2, self.0]);
+        use utils::*;
+        coord_pos_relative_to_ring(*coord, &ls) == CoordPos::Inside
```
michaelkirk, line 16: Thank you for fixing this!

geo/src/algorithm/contains/line.rs
```diff
@@ -0,0 +1,84 @@
...
+// │ Implementations for Line │
+// └──────────────────────────┘
+
+impl<T> Contains<Coordinate<T>> for Line<T>
+where
+    T: HasKernel,
+{
+    fn contains(&self, coord: &Coordinate<T>) -> bool {
+        if self.start == self.end {
+            &self.start == coord
```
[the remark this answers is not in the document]
rmanoka: Yeah, it is confusing, but that is the precise expected behavior.  I cross verified with shapely (geos), and it does the same.

geo/src/algorithm/contains/polygon.rs
rmanoka: Yeah.  Also see the test `line_in_polygon_edgecases` in `contains/mod.rs`.  We might be able to use the winding order trait and a few additions to contains to solve these cases, but wanted to attempt this in a different PR.

geo/src/algorithm/contains/polygon.rs
rmanoka: Yeah, these implementations were just moved into this file.  We should fix the two traits for de9im compliance in another pr

geo/src/algorithm/contains/polygon.rs
rmanoka: So, [MultiPolygon](https://locationtech.github.io/jts/javadoc/org/locationtech/jts/geom/MultiPolygon.html) should not contain overlapping constituent polys: 

>As per the OGC SFS specification, the Polygons in a MultiPolygon may not overlap, and may only touch at single points. This allows the topological point-set semantics to be well-defined.

With the above defn., I think we can lift any Polygon: Contains<T> to work on MultiPolygon

geo/src/algorithm/contains/rect.rs
rmanoka: Good catch!  The previous implementation was the same with strict checking, which would also miss some edge cases.  Want to address these in a different PR (needs a few extra impls, utils).   Let me add a todo comment here so we get back to it.

michaelkirk:
Ok, this LGTM. 

I agree it’s fine to address the pre-existing issues in a follow up PR.

Thanks for making this better! I’d say merge this whenever you’re ready.

michaelkirk:
Ah, I didn’t realize polygons in a multi-polygon couldn’t overlap. Thank you for explaining.

rmanoka:
> Ah, I didn’t realize polygons in a multi-polygon couldn’t overlap. Thank you for explaining.

Actually, you're right.  The polygons can touch at a single point which introduces some edge cases.  For eg. two rectangles touching at a single point, and a line passing through that point.  It won't be in any of the individual polygons but is contained in the multipolygon.

Will leave a comment, and we can get to this when we ensure correctness of these two traits.

rmanoka:
bors r=michaelkirk

PR #514, opened 2020-09-24
intersects: implementation for MultiPolygon

remi-dupre:
Implement `Intersects` for MultiPolygons together with any type that already implement it with Polygons.


frewsxcv (approved):
thanks!

PR #517, opened 2020-10-06
fix ci

michaelkirk:
We recently did some shuffling around of docker containers to facilitate some changes in docker, and the result was that the `--features proj` build for geo was failing.

Fixing that required two steps:

1. I fixed the automatic docker-hub container builds, which I'd inadvertently broken: https://github.com/georust/docker-images/commit/2464e6740e65f9b7d19013fa5a5a636d28ba9cf5

2. This PR, which updates to use the latest available geo-ci container rather than a hardcoded one.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ jobs:
       run:
         working-directory: geo-types
     container:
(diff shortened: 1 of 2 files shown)


rmanoka (approved):
LGTM

rmanoka:
Bors r+

PR #518, opened 2020-10-07
Releasing geo-0.15.0, geo-types-0.6.1

rmanoka:
Published geo 0.15.0, and geo-types 0.6.1.  Also added instructions on publishing for future ref.


michaelkirk:
>  Also added instructions on publishing for future ref.

Thank you for doing that!

PR #519, opened 2020-10-08
more forgiving rect constructor

michaelkirk:
Is there a reason for our constructor to be so brutal?

Rather than force the programmer to order points by min/max, let the computer do it.

### DIFF
--- a/geo-types/Cargo.toml
+++ b/geo-types/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "geo-types"
-version = "0.6.1"
+version = "0.6.2"
 authors = ["Corey Farwell <coreyf@rwell.org>"]
 license = "MIT/Apache-2.0"
 repository = "https://github.com/georust/geo"
--- a/geo-types/src/rect.rs
+++ b/geo-types/src/rect.rs
@@ -64,27 +64,37 @@ impl<T: CoordinateType> Rect<T> {
     ///     Coordinate { x: 10., y: 20. }
     /// );
     /// ```
-    pub fn new<C>(min: C, max: C) -> Rect<T>
+    pub fn new<C>(c1: C, c2: C) -> Rect<T>
     where
         C: Into<Coordinate<T>>,
     {
-        Rect::try_new(min, max).unwrap()
+        let c1 = c1.into();
+        let c2 = c2.into();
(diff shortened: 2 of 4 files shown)


geo-types/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "geo-types"
-version = "0.6.1"
+version = "0.6.2"
```
michaelkirk, line 3: I *think* this would be considered a minor, non-breaking change as far as semver goes, but I'm willing to be corrected.

geo-types/src/rect.rs
```diff
@@ -64,27 +64,37 @@ impl<T: CoordinateType> Rect<T> {
     ///     Coordinate { x: 10., y: 20. }
     /// );
     /// ```
-    pub fn new<C>(min: C, max: C) -> Rect<T>
```
michaelkirk, line 67: whoops, I need to update these docs...

michaelkirk: ok, done!

geo-types/src/rect.rs
```diff
@@ -46,48 +46,54 @@ where
...
-        if rect.has_valid_bounds() {
-            Ok(rect)
-        } else {
-            Err(InvalidRectCoordinatesError)
-        }
+        Ok(Rect::new(c1, c2))
     }
 
-    /// Returns the minimum `Coordinate` (bottom-left corner) of the `Rect`.
+    /// Returns the minimum `Coordinate` of the `Rect`.
```
michaelkirk, line 96: There are lots of cases (e.g. many graphics libraries) that have a top-left as origin, rather than a bottom-left. I'm sure there are esoteric cases with a top-right and bottom-right origin too, so maybe better to omit this bit from the comments all together.

rmanoka:
@michaelkirk Could we do the changes to Cargo.toml at the time of release?  Otherwise, we may increment in more than once before release.  Otherwise, LGTM!

michaelkirk:
> Could we do the changes to Cargo.toml at the time of release?

Sure - reverted that part.

frewsxcv (approved):
great idea!

michaelkirk:
bors r=rmanoka

Issue #515, opened 2020-10-01
DE-9IM Proposal


michaelkirk:
per https://github.com/georust/geo/issues/513 I've written up a proposal about an addition I'd like to make to geo - introducing a DE-9IM module, and using it to help drive and correct some of our relation traits:  e.g. `Contains`, `Intersects`. It would also make it much easier to add future traits like `Disjoint`, `Touches`, `Crosses`, etc.

Read the full proposal here:
https://gist.github.com/michaelkirk/3ac4afe651fe24ad229dec0b56ec8f9d

(a gist allows easier editing than updating issue text in situ).

Please take a look!  I spent some time adding context, in hope that it would be accessible to you even if you don't know what DE-9IM is.

The idea is that after getting a little feedback on the proposal, I'll add some specific issues to this tracking issue.

rmanoka:
Great proposal @michaelkirk; would indeed be awesome to get all the predicate traits available here!  Also, love the concise unicode symbol!  In the interest of helping you out on this rather substantial effort (hope you don't mind :) ), here is my suggested breakdown of the general effort into as many independent pieces as I can imagine.  I'm just thinking aloud, and completely open to suggestions, or counter-arguments!

1. **Geomgraph.**  Probably the crux of the new functionalities.  Given its usefulness for above traits, I feel it should be useful for other use-cases too.  I suggest we make this a public data-structure that can be built and used by the end-user too.  We could brainstorm a bit on how we model this in our crate, to best re-use computations.

1. **Tests and Benchmarks.**  Very useful as we proceed to aim for correctness, efficiency and inter-op with other libs.  Are there any comprehensive suites available (say in geos / JTS)?

1. **Definitions.**  Formally define the semantics of each of the geometry types and traits.  The traits are fairly well defined in terms of the 👖-matrix.  In the past few PRs, we have added the defn. of many of the geometry types too.  Would be good to round-up the definitions for all the types, edge cases, and esp. the differences from other libs.

1. **👖-Matrix.**  Your proposal pretty much nails the requirements here.  I suppose we leave the existing adhoc implementations as long as they're provably correct?

urschrei:
> Geomgraph. Probably the crux of the new functionalities. Given its usefulness for above traits, I feel it should be useful for other use-cases too. I suggest we make this a public data-structure that can be built and used by the end-user too. We could brainstorm a bit on how we model this in our crate, to best re-use computations.

Completely agree, and would love to do this collaboratively.

[...]

PR #522, opened 2020-10-16
Empty trait

michaelkirk:
Same as ec69add2df63ec4e2a9f04c52bc824d8427cef54, this is a small bit of work peeled from my geometry graph progress (see #515), which seemed useful independent from the geometry graph.


michaelkirk:
Sorry for the noise - I'm actually going to close this for a bit, and reopen with some additional related functionality.

PR #523, opened 2020-10-20
Add `is_closed` for MultiLineString and fix `is_closed` for empty LineString

michaelkirk:
Some commits extracted from my geomgraph work

### DIFF
--- a/geo-types/src/multi_line_string.rs
+++ b/geo-types/src/multi_line_string.rs
@@ -3,7 +3,7 @@ use std::iter::FromIterator;
 
 /// A collection of
 /// [`LineString`s](line_string/struct.LineString.html). Can
-/// be created from a `Vec` of `LineString`s, or from an
+/// be created from a `Vec` of `LineString`s or from an
 /// Iterator which yields `LineString`s. Iterating over this
 /// object yields the component `LineString`s.
 ///
--- a/geo-types/src/multi_line_string.rs
+++ b/geo-types/src/multi_line_string.rs
@@ -35,8 +35,8 @@ where
     T: CoordinateType;
 
 impl<T: CoordinateType> MultiLineString<T> {
-    /// True if the MultiLineString is non-empty and each of its LineStrings is closed - that is,
-    /// that the first and last coordinates of each line string are the same.
+    /// True if the MultiLineString is empty or if all of its LineStrings are closed - see
+    /// [`LineString::is_closed`].
     ///
     /// # Examples
     ///
@@ -56,10 +56,7 @@ impl<T: CoordinateType> MultiLineString<T> {
     /// assert!(!MultiLineString::<f32>(vec![]).is_closed());
     /// ```
     pub fn is_closed(&self) -> bool {
-        if self.0.is_empty() {
-            return false;
-        }
-
+        // Note: Unlike JTS et al, we consider an empty MultiLineString as closed.
         self.0.iter().all(LineString::is_closed)
     }
 }
--- a/geo-types/src/multi_line_string.rs
+++ b/geo-types/src/multi_line_string.rs
@@ -34,6 +34,41 @@ pub struct MultiLineString<T>(pub Vec<LineString<T>>)
 where
     T: CoordinateType;
 
+impl<T: CoordinateType> MultiLineString<T> {
+    /// True if the MultiLineString is non-empty and each of its LineStrings is closed - that is,
+    /// that the first and last coordinates of each line string are the same.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use geo_types::{MultiLineString, LineString, line_string};
+    ///
+    /// let open_line_string: LineString<f32> = line_string![(x: 0., y: 0.), (x: 5., y: 0.)];
+    /// assert!(!MultiLineString(vec![open_line_string.clone()]).is_closed());
+    ///
+    /// let closed_line_string: LineString<f32> = line_string![(x: 0., y: 0.), (x: 5., y: 0.), (x: 0., y: 0.)];
+    /// assert!(MultiLineString(vec![closed_line_string.clone()]).is_closed());
+    ///
+    /// // MultiLineString is not closed if *any* of it's LineStrings are not closed
+    /// assert!(!MultiLineString(vec![open_line_string, closed_line_string]).is_closed());
+    ///
+    /// // An empty MultiLineString is not closed
+    /// assert!(!MultiLineString::<f32>(vec![]).is_closed());
+    ///
+    /// // Because an empty LineString is not closed, a MultiLineString containing an empty
+    /// // LineString cannot be closed.
+    /// let empty_line_string: LineString<f32> = line_string![];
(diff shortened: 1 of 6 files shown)


geo/src/utils.rs
```diff
@@ -98,7 +98,7 @@ where
     //
     // See: https://en.wikipedia.org/wiki/Point_in_polygon
 
-    assert!(linestring.is_closed());
+    debug_assert!(linestring.is_closed());
```
michaelkirk, line 101: I'm not sure if this is controversial - we have some "undefined behavior" for invalid geometries, yet we don't (yet) have a way of asserting validity of geometries. 

I think we should be a little more conservative about crashing in release builds.

geo/src/algorithm/contains/polygon.rs
```diff
@@ -12,6 +12,12 @@ where
     T: HasKernel,
 {
     fn contains(&self, coord: &Coordinate<T>) -> bool {
+        // Officially, we require the exterior to be a valid ring, but this isn't curently enforced
+        // by the Polygon contructor. We bail out here rather than trigger later debug asserts.
+        if self.exterior().0.is_empty() {
+            return false;
```
michaelkirk, line 18: This is to avoid hitting the assert in `coord_pos_relative_to_ring`.

Without this, a unit test exercises that path and fails.

There's arguably other approaches - like deleting that unit test. I think ultimately a more blessed path for validating geometries or perhaps having Result types that return Err where we're currently asserting. But I didn't particularly feel like taking that larger change on yet.

frewsxcv (approved):
lgtm! should we consider this a breaking change since it's a change in logic for a public method? i don't feel strongly one way or the other, especially since things are probably gonna be broken in general for folks working with empty linestrings

michaelkirk:
> should we consider this a breaking change since it's a change in logic for public methods?

I'd say no, because there's no change to the existing API, and the behavioral change is strictly a bug fix. 

And a behavior which was even documented, though not implemented:

> Close the `LineString`. Specifically, **if the `LineString` has at least one coordinate**, and
> the value of the first coordinate does not equal the value of the last coordinate, then a
> new coordinate is added to the end with the value of the first coordinate.

michaelkirk:
Though actually I have another PR coming up in a minute which is maybe a big enough new feature to bump the minor version so it's maybe a moot point.

pre-1.0 semver is... "hard" to reason about.

rmanoka:
@michaelkirk the new logic doesn't consider an empty linestring as closed?  I think this is not consistent with other libs.  See for eg. [JTS linear ring](https://locationtech.github.io/jts/javadoc/org/locationtech/jts/geom/LinearRing.html) which accepts empty ones.  Any reason for considering empty as not-closed?

michaelkirk:
> @michaelkirk the new logic doesn't consider an empty linestring as closed? I think this is not consistent with other libs. See for eg. JTS linear ring which accepts empty ones. Any reason for considering empty as not-closed?

Thanks for taking a look @rmanoka. That's interesting. I was mostly just going off our own documentation which says:

Indeed - JTS allows empty LinearRing's and considers empty LinearRings as closed:

```
  public boolean isClosed() {
    if (isEmpty()) {
    	// empty LinearRings are closed by definition
      return true;
    }
    return super.isClosed();
  }
```

And yet, looking a little deeper, JTS does the opposite for their [`LineString#isClosed`](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/geom/LineString.java#L142) to this, which considers empty linestrings as *not* closed.

```java
  public boolean isClosed() {
    if (isEmpty()) {
      return false;
    }
    return getCoordinateN(0).equals2D(getCoordinateN(getNumPoints() - 1));
  }
```

Since we don't have a LinearRing type, our LineString type sometimes does double duty as LinearRings. I guess I don't yet understand the consequences of having it behave like JTS's LineString or whether it should behave like JTS's LinearRing in all cases...

I'm going to keep looking.

rmanoka:
@michaelkirk Interesting find on JTS LineString.  I'm inclined to argue that LineString logic is a bug: SFS defines LinearRing as a closed and simple LineString (so empty is automatically a LinearRing).  From SFS defn, I'm also hoping we can get away with LineString also proxying for LinearRing when closed.  

Can you think of any reason for requiring an empty LineString that's not a LinearRing?

michaelkirk:
> @michaelkirk Interesting find on JTS LineString. I'm inclined to argue that LineString logic is a bug: SFS defines LinearRing as a closed and simple LineString (so empty is automatically a LinearRing). From SFS defn, I'm also hoping we can get away with LineString also proxying for LinearRing when closed.

Thanks for your thoughts there. I'm compelled!

> Can you think of any reason for requiring an empty LineString that's not a LinearRing?

After digging in more, I can see no reason for an empty LineString to behave differently from a LinearRing. Other than just "thinking about it", mainly my approach was to dig through the JTS code base and see where they used the method call. It seems like it wouldn't change any of the libraries internal functionality (though a unit test which specifically tests this behavior would fail).

Thanks for catching this! PTAL @rmanoka

geo-types/src/multi_line_string.rs
```diff
@@ -34,6 +34,36 @@ pub struct MultiLineString<T>(pub Vec<LineString<T>>)
...
+    /// assert!(MultiLineString(vec![closed_line_string.clone()]).is_closed());
+    ///
+    /// // MultiLineString is not closed if *any* of it's LineStrings are not closed
+    /// assert!(!MultiLineString(vec![open_line_string, closed_line_string]).is_closed());
+    ///
+    /// // An empty MultiLineString is not closed
+    /// assert!(!MultiLineString::<f32>(vec![]).is_closed());
+    /// ```
+    pub fn is_closed(&self) -> bool {
+        if self.0.is_empty() {
```
michaelkirk, line 59: Note this bit @rmanoka - it mirrors [JTS's MultiLineString#is_closed](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/geom/MultiLineString.java#L74), but seems to be on similarly shaky ground.

rmanoka: @michaelkirk Ya, not a fan of returning `false` if empty here either.  Usually, it is mathematically neat to extend predicates on list that are defined as a for-all over elements in the list to be true on the empty list.  See for instance [std::Iterator::all](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.all).  Since is_closed is such a predicate, I would have liked it to be true on empty list.  I have no idea why JTS chose the other way.  There is no MultiLinearRing in JTS, so I don't have a compelling argument either :( .

michaelkirk: Ultimately, I can't find a reason why it'd matter, so I'm willing to defer to your intuition here.

michaelkirk:
Ok I think I've addressed all your feedback @rmanoka. PTAL

rmanoka (approved):
LGTM! Thanks @michaelkirk

michaelkirk:
bors r=rmanoka

PR #524, opened 2020-10-20
Geometry dimensions via HasDimensions trait

michaelkirk:
Some efforts extracted from the geomgraph work that seemed generally useful, and thus worth separating.

(I'm also trying to do what I can to keep the upcoming geomgraph PR just a little bit smaller.)

*Note this PR represents separate functionality from #523, but it relies on #523 for correct behavior, so I'm targeting #523 as the base branch. Before merging we should update the base to target master.*

### DIFF
--- a/geo/src/algorithm/dimensions.rs
+++ b/geo/src/algorithm/dimensions.rs
@@ -1,0 +1,452 @@
+use crate::{
+    CoordinateType, Geometry, GeometryCollection, Line, LineString, MultiLineString, MultiPoint,
+    MultiPolygon, Point, Polygon, Rect, Triangle,
+};
+
+/// Geometries can have 0, 1, or two dimensions. Or, in the case of an [`empty`](#is_empty)
+/// geometry, a special `Empty` dimensionality.
+///
+/// ## Examples
+///
+/// ```
+/// use geo_types::{Point, Rect, line_string};
+/// use geo::algorithm::dimensions::{HasDimensions, Dimensions};
+///
+/// let point = Point::new(0.0, 5.0);
+/// let line_string = line_string![(x: 0.0, y: 0.0), (x: 5.0, y: 5.0), (x: 0.0, y: 5.0)];
+/// let rect = Rect::new((0.0, 0.0), (10.0, 10.0));
+/// assert_eq!(Dimensions::ZeroDimensional, point.get_dimensions());
+/// assert_eq!(Dimensions::OneDimensional, line_string.get_dimensions());
+/// assert_eq!(Dimensions::TwoDimensional, rect.get_dimensions());
+///
+/// assert!(point.get_dimensions() < line_string.get_dimensions());
+/// assert!(rect.get_dimensions() > line_string.get_dimensions());
+/// ```
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Ord, PartialOrd)]
+pub enum Dimensions {
+    /// Some geometries, like a `MultiPoint` or `GeometryColletion` may have no elements - thus no
+    /// dimensions. Note that this is distinct from being `ZeroDimensional`, like a `Point`.
+    Empty,
+    /// Dimension of a point
+    ZeroDimensional,
+    /// Dimension of a line or curve
+    OneDimensional,
+    /// Dimension of a surface
+    TwoDimensional,
+}
+
+/// Operate on the dimensionality of geometries.
+pub trait HasDimensions {
+    /// Some geometries, like a `MultiPoint`, can have zero coordinates - we call these `empty`.
+    ///
+    /// Types like `Point` and `Rect`, which have at least one coordinate by construction, can
+    /// never be considered empty.
+    /// ```
+    /// use geo_types::{Point, Coordinate, LineString};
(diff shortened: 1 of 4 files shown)


geo/src/algorithm/dimensions.rs
```diff
@@ -0,0 +1,452 @@
...
+/// let rect = Rect::new((0.0, 0.0), (10.0, 10.0));
+/// assert_eq!(Dimensions::ZeroDimensional, point.get_dimensions());
+/// assert_eq!(Dimensions::OneDimensional, line_string.get_dimensions());
+/// assert_eq!(Dimensions::TwoDimensional, rect.get_dimensions());
+///
+/// assert!(point.get_dimensions() < line_string.get_dimensions());
+/// assert!(rect.get_dimensions() > line_string.get_dimensions());
+/// ```
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Ord, PartialOrd)]
+pub enum Dimensions {
```
michaelkirk, line 26: "Dimensionality" would probably be a more "correct" term for this enum, but I thought maybe it was too verbose. 

I could be persuaded!

geo/src/algorithm/dimensions.rs
```diff
@@ -0,0 +1,452 @@
...
+    /// Some geometries, like a `MultiPoint` or `GeometryColletion` may have no elements - thus no
+    /// dimensions. Note that this is distinct from being `ZeroDimensional`, like a `Point`.
+    Empty,
```
michaelkirk, line 29: I considered ommitting this and instead representing Empty as a `None` `Option<Dimensions>`, but it seemed to work better with `Empty` as an explicit member of the enum. 

Primarily because I want to use it in comparisons:

```
if a.get_dimensions() < Dimensions::OneDimensional {
    do_something()
} else {
    do_something_else()
}
```

is preferable to something like:

```
match a.get_dimensions() {
    Some(dimension) => {
        if dimensions < Dimensions::OneDimensional {
            do_something()
        } else {
            do_something_else()
        } 
    }
    None => {  do_something_else() }
}
```

geo/src/algorithm/dimensions.rs
urschrei: It's our code, and we do what we feel like. (There's no ambiguity in "dimensions", so there's no problem imo)

geo/src/algorithm/dimensions.rs
```diff
@@ -0,0 +1,452 @@
...
+    ///
+    /// let point = Point::new(10.0, 10.0);
+    /// assert_eq!(Dimensions::ZeroDimensional, point.get_dimensions());
+    ///
+    /// // An `Empty` dimensionality is distinct from, and less than, being 0-dimensional
+    /// let empty_collection = GeometryCollection::<f32>(vec![]);
+    /// assert_eq!(Dimensions::Empty, empty_collection.get_dimensions());
+    /// assert!(empty_collection.get_dimensions() < point.get_dimensions());
+    /// ```
+    fn get_dimensions(&self) -> Dimensions;
```
frewsxcv, line 97: thoughts on removing the `get_` prefix? getters in rust tend to not have 'get'

https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

geo/src/algorithm/dimensions.rs
```diff
@@ -0,0 +1,452 @@
...
+        let first = self.0[0];
+        if self.0.iter().any(|&coord| first != coord) {
+            Dimensions::OneDimensional
+        } else {
+            // all coords are the same - i.e. a point
+            Dimensions::ZeroDimensional
+        }
+    }
+
+    ///```
```
frewsxcv, line 239:
```suggestion
    /// ```
```

frewsxcv (approved):
🎉

urschrei (approved):
This PR is a _delight_

geo/src/algorithm/dimensions.rs
michaelkirk: Love it! Honestly I assumed you couldn't have a method and property with the same name. 😅

geo/src/algorithm/dimensions.rs
```diff
@@ -0,0 +1,452 @@
...
+    }
+
+    fn boundary_dimensions(&self) -> Dimensions {
+        Dimensions::Empty
+    }
+}
+
+impl<C: CoordinateType> HasDimensions for MultiLineString<C> {
+    fn is_empty(&self) -> bool {
+        if self.0.is_empty() {
```
rmanoka, line 296: Why is this needed?  Iterator::all returns true on empty iterator (Iterator::any returns false).

michaelkirk: Good point - removed!

michaelkirk:
I think I've addressed all your feedback here as well @rmanoka. PTAL

michaelkirk:
@rmanoka - I believe I've addressed your feedback. Unless I hear that you'd like more time to review, I'll merge in 24 hours.

rmanoka (approved):

rmanoka:
@michaelkirk lgtm, please go ahead.

michaelkirk:
bors r=rmanoka

Issue #527, opened 2020-11-07
Create a fuzz target to test correctness of rust-geo algorithms


frewsxcv:
Steps:

1. Implement the [`Arbitrary`](https://docs.rs/arbitrary/0.4.7/arbitrary/trait.Arbitrary.html) trait for `geo::Geometry` so we can generate arbitrary geometries from a byte slice
1. Create a new cargo-fuzz project
1. Add geo and gdal as dependencies
1. For each fuzz iteration:
    1. Generate identical geometries for geo and gdal
    1. Perform operations on both geometries using geo and gdal algorithms
    1. Ensure the outputs of the algorithms are identical

michaelkirk:
Love this idea!

frewsxcv:
First fuzz target was added in https://github.com/georust/geo/commit/1e36d1e07bd7b83b6132940c66ebedb6204c3c8e

PR #532, opened 2020-11-10
Add fuzz targets

frewsxcv:
- https://github.com/georust/geo/issues/527
- [Video](https://www.youtube.com/watch?v=N00_yt4_kNQ)

TODO:

- [ ] Check validity of outputted geometries
- [ ] Include NaN values
- [ ] Fuzz against the geos and gdal crates

### DIFF
--- a/geo-types/src/lib.rs
+++ b/geo-types/src/lib.rs
@@ -199,7 +199,7 @@ impl<T: num_traits::Float + arbitrary::Arbitrary + CoordinateType> arbitrary::Ar
             2 => Geometry::Polygon(u.arbitrary()?),
             3 => Geometry::MultiPoint(u.arbitrary()?),
             4 => Geometry::MultiLineString(u.arbitrary()?),
-            5 => Geometry::MultiPoint(u.arbitrary()?),
+            5 => Geometry::MultiPolygon(u.arbitrary()?),
             6 => Geometry::GeometryCollection(u.arbitrary()?),
             7 => Geometry::Triangle(u.arbitrary()?),
             8 => Geometry::Rect(u.arbitrary()?),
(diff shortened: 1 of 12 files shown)


geo-types/src/lib.rs
```diff
@@ -81,6 +81,133 @@ mod macros;
...
+    fn arbitrary(u: &mut arbitrary::Unstructured<'_>) -> arbitrary::Result<Self> {
+        let n = u.int_in_range(0..=8)?;
+
+        Ok(match n {
+            0 => Geometry::Point(u.arbitrary()?),
+            1 => Geometry::LineString(u.arbitrary()?),
+            2 => Geometry::Polygon(u.arbitrary()?),
+            3 => Geometry::MultiPoint(u.arbitrary()?),
+            4 => Geometry::MultiLineString(u.arbitrary()?),
+            5 => Geometry::MultiPoint(u.arbitrary()?),
```
frewsxcv, line 202: oops this is wrong

geo-types/src/lib.rs
```diff
@@ -81,6 +81,133 @@ mod macros;
...
+    for Coordinate<T>
+{
+    fn arbitrary(u: &mut arbitrary::Unstructured<'_>) -> arbitrary::Result<Self> {
+        let x = u.arbitrary::<T>()?;
+        if x.is_nan() {
+            return Err(arbitrary::Error::IncorrectFormat);
+        }
+
+        let y = u.arbitrary::<T>()?;
+        if y.is_nan() {
```
michaelkirk, line 94: I wonder if there would actually be some value in fuzzing with NaN some times. 

Generally we expect valid input, but ultimately our methods could be downstream from something that outputs NaN, and we should try to respond as best we can (e.g. probably not crash).

PR #537, opened 2020-11-11
Don't panic on negative epsilon

### LINKED ISSUE #536: Passing a negative value for `simplify`'s epsilon will result in a stack overflow
Discovered while fuzzing

urschrei:
Fixes #536

### DIFF
--- a/geo/src/algorithm/simplify.rs
+++ b/geo/src/algorithm/simplify.rs
@@ -19,6 +19,10 @@ fn rdp<T>(points: &[Point<T>], epsilon: &T) -> Vec<Point<T>>
 where
     T: Float,
 {
+    // Epsilon must be greater than zero for any meaningful simplification to happen
+    if *epsilon <= T::zero() {
+        points.to_vec();
+    }
     compute_rdp(
         &points
             .into_iter()
--- a/geo/src/algorithm/simplify.rs
+++ b/geo/src/algorithm/simplify.rs
@@ -91,6 +91,8 @@ where
 /// rings. This may result in invalid Polygons, and has no guarantee of preserving topology.
 ///
 /// Multi* objects are simplified by simplifing all their constituent geometries individually.
+///
+/// An epsilon less than or equal to zero will return an unaltered version of the geometry.
 pub trait Simplify<T, Epsilon = T> {
     /// Returns the simplified representation of a geometry, using the [Ramer–Douglas–Peucker](https://en.wikipedia.org/wiki/Ramer–Douglas–Peucker_algorithm) algorithm
     ///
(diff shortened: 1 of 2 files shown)


frewsxcv (approved):
your last commit looks good! but there are also some other commits in here!

urschrei:
🤦‍♂️

geo/src/algorithm/simplify.rs
```diff
@@ -19,6 +19,10 @@ fn rdp<T>(points: &[Point<T>], epsilon: &T) -> Vec<Point<T>>
 where
     T: Float,
 {
+    // Epsilon must be greater than zero for any meaningful simplification to happen
+    if *epsilon <= T::zero() {
```
michaelkirk, line 23: This implies some kind of programmer error right? Should we add a debug assertion failure here?

michaelkirk: I'm *assuming* the fuzzer runs against release builds and thus won't trigger the debug assertion? But I might be wrong...

michaelkirk (approved):
LGTM!

Feedback was optional, so take it or leave it.

geo/src/algorithm/simplify.rs
frewsxcv: cargo-fuzz builds in release mode _with_ debug assertions

https://github.com/rust-fuzz/cargo-fuzz/blob/799bf534b889859b090b76f2ca1cffdca4ee7455/src/main.rs#L45-L47

https://github.com/rust-fuzz/cargo-fuzz/blob/bde819e87a86e6b1c5c0320e94343a6f97d063bb/src/options.rs#L68-L70

i don't know where the default value for debug-assertions gets set in cargo-fuzz, but i verified locally that debug assertions are enabled

michaelkirk: Yeah, I don't feel confident about how to resolve this.

I can imagine some cases where it would be interesting if it were possible to reach an assert via the fuzzer, and others, like here, where it wouldn't be interesting.

We *could* have the convention that `debug_assert` is the former while `assert` is the latter, but... that seems pretty limiting.

We could make our own assert macros, like `debug_assert_unless_fuzzing!`, but that seems a bit much.

michaelkirk: To restate, I'm really OK with just merging this as is. 👍

urschrei: I am feeling very indecisive too.

frewsxcv: +1 to merging as-is

frewsxcv (approved):

PR #538, opened 2020-11-12
Determine the position of a Coordinate relative to a geometry.

michaelkirk:
# Examples
```rust
use geo::{polygon, Coordinate};
use geo::coordinate_position::{CoordinatePosition, CoordPos};
let square_poly = polygon![(x: 0.0, y: 0.0), (x: 2.0, y: 0.0), (x: 2.0, y: 2.0), (x: 0.0, y: 2.0), (x: 0.0, y: 0.0)];
let inside_coord = Coordinate { x: 1.0, y: 1.0 };
assert_eq!(square_poly.coordinate_position(&inside_coord), CoordPos::Inside);
let boundary_coord = Coordinate { x: 0.0, y: 1.0 };
assert_eq!(square_poly.coordinate_position(&boundary_coord), CoordPos::OnBoundary);
let outside_coord = Coordinate { x: 5.0, y: 5.0 };
assert_eq!(square_poly.coordinate_position(&outside_coord), CoordPos::Outside);
```

Some work extracted from the geomgraph effort.

### DIFF
--- a/geo-types/src/coordinate.rs
+++ b/geo-types/src/coordinate.rs
@@ -222,14 +222,19 @@ use num_traits::Zero;
 /// assert_eq!(p.x, 0.);
 /// assert_eq!(p.y, 0.);
 /// ```
-impl<T: CoordinateType> Zero for Coordinate<T> {
-    fn zero() -> Self {
+impl<T: CoordinateType> Coordinate<T> {
+    pub fn zero() -> Self {
         Coordinate {
             x: T::zero(),
             y: T::zero(),
         }
     }
+}
 
+impl<T: CoordinateType> Zero for Coordinate<T> {
+    fn zero() -> Self {
+        Coordinate::zero()
+    }
     fn is_zero(&self) -> bool {
         self.x.is_zero() && self.y.is_zero()
     }
--- a/geo/src/algorithm/coordinate_position.rs
+++ b/geo/src/algorithm/coordinate_position.rs
@@ -1,0 +1,763 @@
+use crate::algorithm::{
+    bounding_rect::BoundingRect, dimensions::HasDimensions, intersects::Intersects,
+    kernels::HasKernel,
+};
+use crate::{
+    Coordinate, Geometry, GeometryCollection, Line, LineString, MultiLineString, MultiPoint,
+    MultiPolygon, Point, Polygon, Rect, Triangle,
+};
+
+/// The position of a `Coordinate` relative to a `Geometry`
+#[derive(PartialEq, Clone, Debug)]
+pub enum CoordPos {
+    OnBoundary,
+    Inside,
+    Outside,
+}
+
+/// Determine whether a `Coordinate` lies inside, outside, or on the boundary of a geometry.
+///
+/// # Examples
+///
+/// ```rust
+/// use geo::{polygon, Coordinate};
+/// use geo::coordinate_position::{CoordinatePosition, CoordPos};
+///
+/// let square_poly = polygon![(x: 0.0, y: 0.0), (x: 2.0, y: 0.0), (x: 2.0, y: 2.0), (x: 0.0, y: 2.0), (x: 0.0, y: 0.0)];
+///
+/// let inside_coord = Coordinate { x: 1.0, y: 1.0 };
+/// assert_eq!(square_poly.coordinate_position(&inside_coord), CoordPos::Inside);
+///
+/// let boundary_coord = Coordinate { x: 0.0, y: 1.0 };
+/// assert_eq!(square_poly.coordinate_position(&boundary_coord), CoordPos::OnBoundary);
+///
+/// let outside_coord = Coordinate { x: 5.0, y: 5.0 };
+/// assert_eq!(square_poly.coordinate_position(&outside_coord), CoordPos::Outside);
+/// ```
+pub trait CoordinatePosition {
+    type Scalar: HasKernel;
+    fn coordinate_position(&self, coord: &Coordinate<Self::Scalar>) -> CoordPos {
+        let mut is_inside = false;
+        let mut boundary_count = 0;
+
+        self.calculate_coordinate_position(coord, &mut is_inside, &mut boundary_count);
+
+        // “The boundary of an arbitrary collection of geometries whose interiors are disjoint
+        // consists of geometries drawn from the boundaries of the element geometries by
+        // application of the ‘mod 2’ union rule”
+        //
+        // ― OpenGIS Simple Feature Access § 6.1.15.1
+        if boundary_count % 2 == 1 {
+            CoordPos::OnBoundary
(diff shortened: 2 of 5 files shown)


geo-types/src/coordinate.rs
```diff
@@ -222,14 +222,19 @@ use num_traits::Zero;
 /// assert_eq!(p.x, 0.);
 /// assert_eq!(p.y, 0.);
 /// ```
-impl<T: CoordinateType> Zero for Coordinate<T> {
-    fn zero() -> Self {
+impl<T: CoordinateType> Coordinate<T> {
+    pub fn zero() -> Self {
```
michaelkirk, line 226: ...tangential quality of life improvement.

geo/src/algorithm/coordinate_position.rs
```diff
@@ -0,0 +1,749 @@
...
+            Geometry::GeometryCollection(geometry_collection) => {
+                geometry_collection.calculate_coordinate_position(coord, is_inside, boundary_count)
+            }
+        }
+    }
+}
+
+/// Calculate the position of a `Coordinate` relative to a
+/// closed `LineString`.
+pub fn coord_pos_relative_to_ring<T>(coord: Coordinate<T>, linestring: &LineString<T>) -> CoordPos
```
michaelkirk, line 362: This method was moved from utils unmodified.

In an effort to keep `utils` from becoming a big junk drawer, this seemed like a more specific home for it.

geo/src/algorithm/coordinate_position.rs
```diff
@@ -0,0 +1,749 @@
...
+    kernels::HasKernel,
+};
+use crate::{
+    Coordinate, Geometry, GeometryCollection, Line, LineString, MultiLineString, MultiPoint,
+    MultiPolygon, Point, Polygon, Rect, Triangle,
+};
+
+/// The position of a `Coordinate` relative to a `Geometry`
+#[derive(PartialEq, Clone, Debug)]
+pub enum CoordPos {
```
michaelkirk, line 12: moved this from utils - reusing form the "coord_pos_relative_to_ring".

@rmanoka - do you see any issue with re-using this enum from coord_pos_relative_to_ring? 

At first I introduced a new "Interior/Exterior/Boundary" enum, but it seemed redundant.

rmanoka: Sounds fine to me

geo/src/algorithm/coordinate_position.rs
```diff
@@ -0,0 +1,749 @@
...
+/// let inside_coord = Coordinate { x: 1.0, y: 1.0 };
+/// assert_eq!(square_poly.coordinate_position(&inside_coord), CoordPos::Inside);
+///
+/// let boundary_coord = Coordinate { x: 0.0, y: 1.0 };
+/// assert_eq!(square_poly.coordinate_position(&boundary_coord), CoordPos::OnBoundary);
+///
+/// let outside_coord = Coordinate { x: 5.0, y: 5.0 };
+/// assert_eq!(square_poly.coordinate_position(&outside_coord), CoordPos::Outside);
+/// ```
+pub trait CoordinatePosition<T>
```
rmanoka, line 37: @michaelkirk Do you prefer the `CoordinatePosition<T>` in favor of `CoordinatePosition` with a associate type `Scalar: CoordinateType` ?  My personal pref. is the latter;  optional comment.

geo/src/algorithm/coordinate_position.rs
```diff
@@ -0,0 +1,749 @@
...
+    T: HasKernel,
+{
+    fn calculate_coordinate_position(
+        &self,
+        coord: &Coordinate<T>,
+        is_inside: &mut bool,
+        boundary_count: &mut usize,
+    ) {
+        // PERF TODO: I'm sure there's a better way to calculate than converting to a polygon
+        Polygon::from(self.clone()).calculate_coordinate_position(coord, is_inside, boundary_count);
```
frewsxcv, line 184: i think you can do:

```suggestion
        self.to_polygon().calculate_coordinate_position(coord, is_inside, boundary_count);
```

geo/src/algorithm/coordinate_position.rs
```diff
@@ -0,0 +1,749 @@
...
```
frewsxcv, line 199:
```suggestion
        self.to_polygon().calculate_coordinate_position(coord, is_inside, boundary_count);
```

geo/src/algorithm/coordinate_position.rs
```diff
@@ -0,0 +1,749 @@
...
+where
+    T: HasKernel,
+{
+    fn coordinate_position(&self, coord: &Coordinate<T>) -> CoordPos {
+        let mut is_inside = false;
+        let mut boundary_count = 0;
+
+        self.calculate_coordinate_position(coord, &mut is_inside, &mut boundary_count);
+
+        // SFS "mod 2" boundary rule
```
frewsxcv, line 47: thoughts on adding a quote here?

```suggestion
        // “The boundary of an arbitrary collection of geometries whose interiors are disjoint
        // consists of geometries drawn from the boundaries of the element geometries by
        // application of the ‘mod 2’ union rule”
        //
        // ― OpenGIS Simple Feature Access § 6.1.15.1
```

frewsxcv (approved):
looks great, thanks!

geo/src/algorithm/coordinate_position.rs
michaelkirk: I believe I've done this here: a60dc27 - can you take a look @rmanoka?

Despite reading a couple articles on this, I only kind of theoretically understand why this distinction is good, and still don't have a very intuitive grasp for when to apply it.

Is your rule of thumb to prefer associated types and only use full-blown generic type params when associated types are insufficient?

michaelkirk: I'll plan to merge this in a couple days if I don't hear otherwise.

michaelkirk:
bors r=frewsxcv

michaelkirk:
bors retrt

michaelkirk:
bors retry

geo/src/algorithm/coordinate_position.rs
rmanoka: @michaelkirk Sorry for the late response;  the changes look great!   My pref. is mostly similar to `Iterator` vs `AddAsign` traits in std.  I go for an associated type unless the same type could implement the trait with two possible param.  Here, we're implementing the trait `CoordinatePosition` with only `T = Ty` for `Line<Ty>`, so prefer it to be associated than make the trait `CoordinatePosition<T>`.  

May be in the future, our scalar types could have additional sophistication that tells us when `T1` can be converted to `T2` without loss in precision, and we can make some of our traits like `CoordinatePosition`, `ForEach` into generics.

PR #539, opened 2020-11-13
add missing `iter`, `iter_mut`, and `IntoIter<&'a mut, T>` methods

michaelkirk:
This is just a small quality of life thing - you can see its application in the final "apply" commit.

GeometryCollection had the full suite of iterator methods, but although MultiLineString, MultiPolygon, and MultiPoint had the consuming `IntoIter<T>` trait, they did *not* have the  conventional `iter()`, `iter_mut()` methods nor the non-consuming IntoIter methods.

### DIFF
--- a/geo-types/src/multi_line_string.rs
+++ b/geo-types/src/multi_line_string.rs
@@ -81,3 +81,116 @@ impl<T: CoordinateType> IntoIterator for MultiLineString<T> {
         self.0.into_iter()
     }
 }
+
+impl<'a, T: CoordinateType> IntoIterator for &'a MultiLineString<T> {
+    type Item = &'a LineString<T>;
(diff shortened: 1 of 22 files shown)


geo-types/src/multi_line_string.rs
```diff
@@ -81,3 +81,116 @@ impl<T: CoordinateType> IntoIterator for MultiLineString<T> {
         self.0.into_iter()
     }
 }
+
+impl<'a, T: CoordinateType> IntoIterator for &'a MultiLineString<T> {
```
michaelkirk, line 85: FYI: the implementation and tests were more or less copy+pasted across MultiLineString, MultiPolygon, and MultiPoint.

frewsxcv (approved):

Issue #533, opened 2020-11-10
Add Debug implementations for all structs


frewsxcv:
1. Add `#![warn(missing_debug_implementations)]` to the root of each of our crates
1. Resolve all warnings

martinfrances107:
I have a PR for this issue when handling Iterator enum .. I have resorted to outputting a indication of what is being debugged.

so the debug string looks like 

"GeometryCoordsIter::Point() Iterator"

maybe there is something better ?

michaelkirk:
Thanks for taking a look at this @martinfrances107!

I think I'd expect something akin to [`impl Debug for Vec::Iter`](https://doc.rust-lang.org/src/core/slice/iter.rs.html#73) which includes details about the vec contents, which is more likely what you're interested in rather than just the type.

```
fn main() {
    let v = vec!["a", "b", "c"];
    println!("v.iter() debug: {:?}", v.iter());
    println!("v debug: {:?}", v);
}
```

Which outputs:
```
v.iter() debug: Iter(["a", "b", "c"])
v debug: ["a", "b", "c"]
```

michaelkirk:
Fixed in #547

PR #547, opened 2020-12-02
#533 Add Debug for implementations for all structs

martinfrances107:
---

Where appropriate, I have added 

`#[derive(Debug)]`

It is more complicated for the enumeration

GeometryCoordsIter

For this iterator thing I think it would be useful to just output a string to effect that 

this is a Iterator for type x, y or z

So Debug::fmt() just outputs string of the type ""GeometryCoordsIter::Point() Iterator"

For other iterators,  my response has just been to individually allow "No Debug" as I think that is  valid.

PS 
This is my first PR here, so I will be responsive to reviews as I may have done something awkward

### DIFF
--- a/geo/src/algorithm/coords_iter.rs
+++ b/geo/src/algorithm/coords_iter.rs
@@ -2,6 +2,7 @@ use crate::{
     Coordinate, CoordinateType, Geometry, GeometryCollection, Line, LineString, MultiLineString,
     MultiPoint, MultiPolygon, Point, Polygon, Rect, Triangle,
 };
+use std::fmt;
 use std::{iter, marker, slice};
 
 /// Iterate over geometry coordinates.
@@ -212,6 +213,7 @@ impl<'a, T: CoordinateType + 'a> CoordsIter<'a, T> for Geometry<T> {
 
 // Utility to transform Iterator<CoordsIter> into Iterator<Iterator<Coordinate>>
 #[doc(hidden)]
+#[derive(Debug)]
 pub struct MapCoordsIter<
     'a,
     T: 'a + CoordinateType,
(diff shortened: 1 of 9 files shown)


geo/src/algorithm/coords_iter.rs
```diff
@@ -278,6 +280,35 @@ impl<'a, T: CoordinateType> Iterator for GeometryCoordsIter<'a, T> {
     }
 }
 
+impl<'a, T: CoordinateType> fmt::Debug for GeometryCoordsIter<'a, T> {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
```
michaelkirk, line 284: I think I'd expect something akin to [`impl Debug for Vec::Iter`](https://doc.rust-lang.org/src/core/slice/iter.rs.html#73) which includes details about the vec contents, which is more likely what you're interested in rather than just the type.

```
fn main() {
    let v = vec!["a", "b", "c"];
    println!("v.iter() debug: {:?}", v.iter());
    println!("v debug: {:?}", v);
}
```

Which outputs:
```
v.iter() debug: Iter(["a", "b", "c"])
v debug: ["a", "b", "c"]
```

martinfrances107: Thanks for the tip... that does look better, more standard.

The fix was simple for everything but the "GeometryCollection" item .. I am having touble dealing with the match type (i)

which in this case is of type

&Box< dyn Iterator<item...>>

I played with it and could not find a good way of dynamically casting into something that would compile.

Any advice would be welcome.

frewsxcv: @martinfrances107 what's code have you tried, and what error do you get?

martinfrances107: > what's the code you have tried

inside the implmentation of "fmt::Debug for GeometryCoordsIter<'a, T>"

            GeometryCoordsIter::GeometryCollection(i) => {
                fmt.debug_tuple("GeometryCollection").field(i).finish()
            }

I will post the error below ... 

speaking conversationally .. I think maybe I need to alter a definition in geometry_collection.rs 
enforcing the type contract  ( "+ std::fmt::Debug" ) I just can't seem to get a working version.

or maybe I need to implement Debug for the GeometryCoordsIter .. and just have a simple loop which reports the debug info for each element ?

That is where I am stuck 
```

`dyn std::iter::Iterator<Item = geo_types::Coordinate<T>>` doesn't implement `std::fmt::Debug`

`dyn std::iter::Iterator<Item = geo_types::Coordinate<T>>` cannot be formatted using `{:?}` because it doesn't implement `std::fmt::Debug`

help: the trait `std::fmt::Debug` is not implemented for `dyn std::iter::Iterator<Item = geo_types::Coordinate<T>>`
note: required because of the requirements on the impl of `std::fmt::Debug` for `std::boxed::Box<dyn std::iter::Iterator<Item = geo_types::Coordinate<T>>>`
note: required for the cast to the object type `dyn std::fmt::Debug`


```

frewsxcv: the `Iterator` itself doesn't implement `Debug`, but each of the items do so we need to loop through the iterator.

does something like this work?

```rust
let debug_tuple = fmt.debug_tuple("GeometryCollection");

for geometry in i {
    debug_tuple.field(i);
}

debug_tuple.finish();
```

martinfrances107:
Thanks for talking this through with me...

here is the definition of the object input to the function.


```
impl<'a, T: CoordinateType + 'a> CoordsIter<'a, T> for GeometryCollection<T> {
     type Iter = Box<dyn Iterator<Item = Coordinate<T>> + 'a>;

     fn coords_iter(&'a self) -> Self::Iter {
         Box::new(self.0.iter().flat_map(|geometry| geometry.coords_iter()))
     }
 }
```

I think we need rust to accept that elements of that definition are going to have to be dynamically determined 

by careful insertion of the dyn keyword into a variable I have called dynamicIter

It does not compile .. but I think is a step in the right direction is 

```
            GeometryCoordsIter::GeometryCollection(i) => {                
                let dynamicIter: Box<dyn CoordsIter<T, Iter=Box<dyn Iterator<Item = dyn CoordsIter<T, Iter=Box<dyn CoordsIter<T, Iter=dyn Iterator<Item = Coordinate<T>>>>>>>>> = Box::new(i);                
                let mut debug_tuple = fmt.debug_tuple("GeometryCollection");
                for geometry in dynamicIter.into_iter() {
                    debug_tuple.field(&geometry);
                }
                debug_tuple.finish()
            }
```

frewsxcv:
thanks for looking into this @martinfrances107. one thing i forgot is that to iterate over an `Iterator`, we need a `&mut` iterator, but we have a `&` iterator. so trying to iterate with a `for` loop probably won't work. maybe we should keep this boring for now and do:

```rust
            GeometryCoordsIter::GeometryCollection(i) => {
                fmt.debug_tuple("GeometryCollection")
                    .field(&String::from("..."))
                    .finish()
            }
```

in the future, if we find a way to make this work, we can do that in a follow-up pull request

martinfrances107:
Ok that is a good idea let us be boring...

martinfrances107:
I have update the PR

michaelkirk (approved):
Thanks @martinfrances107 - this LGTM! 

I'll leave it open another day in case @frewsxcv or anyone else had any final feedback, and otherwise I'll merge it tomorrow.

frewsxcv (approved):
thanks @martinfrances107!

michaelkirk:
Looking at some of our current debug output while reviewing this PR inspired this issue: https://github.com/georust/geo/issues/564

Issue #548, opened 2020-12-02
Clippy: remove the clone on Copy type warnings.


martinfrances107:
In the past on other projects I have seen unexpected performance boost when following this clippy directive/warning... it all depends on the complexity of the data structure but I think it best to sweep the following  'clippy' warning from our code.

`warning: using "clone" on a "Copy"  type`

Here is a relevant link to the clippy documentation site [clippy warning ](https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_copy)

There a 2 warnings in the code base I want to remove .. just out of caution
I am about submit a PR

martinfrances107:
Thanks for the quick review.

PR #549, opened 2020-12-02
#548 Clippy: Remove the "clone on Copy type" warnings

martinfrances107:
---

This is the PR associated with #548


michaelkirk (approved):
Thanks!

bors r+

Issue #553, opened 2020-12-03
Clippy: Don't call into_iter() on types that support iter()


martinfrances107:
I just notices this clippy warning being generated on our code

here is the link to the clippy documentation.

[clippy warning](https://rust-lang.github.io/rust-clippy/master/index.html#into_iter_on_ref)

If the type support iter() then we should call it directly.

we have this in 6 places.

I am about to upload my PR

michaelkirk:
Fixed in #554

PR #554, opened 2020-12-03
#553 Clippy: Don't call into_iter on types that support iter

martinfrances107:


martinfrances107:
This is is support of https://github.com/georust/geo/issues/553

frewsxcv (approved):
thanks!

Issue #558, opened 2020-12-04
clippy in CI?


michaelkirk:
All of @martinfrances107's recent PR's have got me wondering if we should incorporate clippy into our CI.

What do you think?

And if we did incorporate, should it *fail* CI or just be a warning? 

IIRC @urschrei pointed out elsewhere that sometimes these kinds of hurdles can be off-putting to new contributors who aren't already familiar with the large variety of tools.

urschrei:
Ah, yep, big fan of Clippy _warning_ on CI, for exactly the reason above.

frewsxcv:
+1 for warning, +0 for failing

martinfrances107:
There are 20 or so clippy warnings with the project ... 

I will generate a new PR to fix them 

Just a little warning  there is are lots of warnings but one I suggest we ignore

#[allow(needless_resturn)]

This project has lots of instances of 

return(x);

which I think is perfectly reasonable, but clippy will tell you that is is not idiomatic enough.

martinfrances107:
In the review of the PR 

[Followup ](https://github.com/georust/geo/pull/559#discussion_r536938856)

That given `return A `

We should follow the rust specific idom, dropping both the return and the semicolon.

This is contraversial and makes the blood boil of some C programmer I know
but hey I am going to follow the convention.

I am about to submit a new PR

Finally

"**Should we also follow up with a git hook that rejects  PRs with clippy warning?**"

martinfrances107:
Here is the followup PR

https://github.com/georust/geo/pull/560

michaelkirk:
> Should we also follow up with a git hook that rejects PRs with clippy warning?

I'm not sure what you mean by "git hook" in this context. Assuming you mean something like: https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks - I'm pretty sure GitHub doesn't support server side hooks.

The closest I think we have to server-side enforcement is what I'm proposing here - to add something to our GitHub action workflow. 

It seems like most people would rather not *reject* a PR outright for clippy violations. The thinking being that it may discourage new contributors, but rather just print the warnings.

In addition to that though - you've got me thinking - I'd personally like to have a client-side git-hook to run clippy on commit. Shipping a "blessed" set of git hooks and documentation in the repository which can be quickly installed on the developer's machine might be worthwhile.

martinfrances107:
Sorry for letting this conversation become stale...

In terms of adding to our github action work flow ...

[...]

PR #559, opened 2020-12-05
Resolved all clippy issues.

martinfrances107:
This PR is in relation to 

https://github.com/georust/geo/issues/558

Naming things is hard ... 

there are a couple of new type names, such as 

RectChainIter<T> 

which people might object to .. I am happy to make changes.


geo/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(clippy::needless_return)]
```
frewsxcv, line 1: we should fix this in the future!

frewsxcv (approved):
thanks @martinfrances107 🙇

PR #561, opened 2020-12-07
FIX MultiPoint#dimensions

michaelkirk:
Simply wrong.

---

### DIFF
--- a/geo/src/algorithm/dimensions.rs
+++ b/geo/src/algorithm/dimensions.rs
@@ -283,7 +283,7 @@ impl<C: CoordinateType> HasDimensions for MultiPoint<C> {
             return Dimensions::Empty;
         }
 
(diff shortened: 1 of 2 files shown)


geo/src/algorithm/dimensions.rs
```diff
@@ -283,7 +283,7 @@ impl<C: CoordinateType> HasDimensions for MultiPoint<C> {
             return Dimensions::Empty;
         }
 
-        Dimensions::OneDimensional
+        Dimensions::ZeroDimensional
```
michaelkirk, line 286: straight up bug. 😞

martinfrances107:
This looks like a good change to me.

frewsxcv (approved):
good catch 🐛

Issue #412, opened 2020-01-09
MultiPolygon::rotate should use centroid as pivot


anlumo:
Right now, `MultiPolygon::rotate` rotates all its polygons individually. However, I'd expect it to rotate them around its own centroid. Otherwise, I'd just iterate over the inner polygons and rotate them myself.

Obviously, this would be a breaking change, so it'd probably be better to add a separate function for this. Ideally, it'd be nice to have a variant where you just pass the pivot explicitly.

urschrei:
I could have sworn we used the same logic as Shapely when we implemented this. Hmm.

>  it'd be nice to have a variant where you just pass the pivot explicitly.

Does https://docs.rs/geo/0.12.2/geo/algorithm/rotate/trait.RotatePoint.html work for you?

anlumo:
Yes, that one does exactly what I need, thank you! So, disregard the second paragraph.

The point about rotate arguably not behaving as you'd expect still stands, though.

andrei-papou:
Is something like this sufficient?
```rust
pub trait RotateCentroid<T>: RotatePoint<T> + Centroid {
    fn rotate_around_centroid(&self, angle: T) -> <Self as Centroid>::Output
    where
        T: Float,
        <Self as Centroid>::Output: Float,
    {
        self.rotate_around_point(angle, self.centroid())
    }
}
```
Seems like it gives the ability to rotate the multipolygon around the centroid.

frewsxcv:
@andrei-papou that looks good to me!

a couple relevant links: https://github.com/georust/geo/pull/151 https://github.com/georust/geo/issues/287

martinfrances107:
The code fragment form above does not compile

https://github.com/georust/geo/issues/412#issuecomment-725923942

`<Self as Centroid>::Output: Float,`

Float is a 1-D scalar type, the value is needed is a 2D type like Point<T> .. but that also errors out a as Trait is needed struct type given. 

the solution is listed below .. I have bundled this up with a  test and I am about to post a PR.

```
pub trait RotateCentroid<T: CoordinateType>: RotatePoint<T> + Centroid<Output = Point<T>> {
    fn rotate_around_centroid(&self, angle: T) -> Self
    where
        T: Float,
        Self: Sized,
    {
        self.rotate_around_point(angle, self.centroid())
    }
}
```

michaelkirk:
I know I'm late to be giving design input - sorry.

[...]

PR #562, opened 2020-12-08
#412 MultiPolygon::rotate should use centroid as pivot

martinfrances107:
Here is the solution with tests.

### DIFF
--- a/geo/src/algorithm/rotate.rs
+++ b/geo/src/algorithm/rotate.rs
@@ -1,9 +1,10 @@
 use crate::algorithm::centroid::Centroid;
 use crate::algorithm::map_coords::MapCoords;
-use crate::{Line, LineString, MultiLineString, MultiPoint, MultiPolygon, Point, Polygon};
+use crate::{
+    CoordinateType, Line, LineString, MultiLineString, MultiPoint, MultiPolygon, Point, Polygon,
+};
 use num_traits::{Float, FromPrimitive};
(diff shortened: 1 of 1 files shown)


martinfrances107:
Here is the associated issue 

https://github.com/georust/geo/issues/412

martinfrances107:
To criticize my own work...

I have copied the project wide practise of copying over numerical rounding error into the expected test output 

for example the value -0.9999999999999999 when -1 was desired 

In a separate issue I want to fix this .. hence I have created a new issue 

https://github.com/georust/geo/issues/565

geo/src/algorithm/rotate.rs
```diff
@@ -217,6 +218,16 @@ where
     }
 }
 
+pub trait RotateCentroid<T: CoordinateType>: RotatePoint<T> + Centroid<Output = Point<T>> {
+    fn rotate_around_centroid(&self, angle: T) -> Self
+    where
+        T: Float,
+        Self: Sized,
+    {
+        self.rotate_around_point(angle, self.centroid())
```
michaelkirk, line 227: I had some input which I thought would be better to discuss in the issue rather than the PR, please see here: https://github.com/georust/geo/issues/412#issuecomment-740996625

Curious about your opinion here.

michaelkirk:
As discussed in #412 we took a slightly different approach - merged in #651

PR #567, opened 2020-12-11
relative_eq()

martinfrances107:
Added num_coords() method to MultiPoint. To Point, Line, LineString and MultiPoint test helpers and associated tests were added to assert that two structs were relatively equal. That is identical within the limits of numerical rounding accuracy.

---

### DIFF
--- a/geo-types/src/line.rs
+++ b/geo-types/src/line.rs
@@ -172,9 +172,8 @@ impl<T: CoordinateType> From<[(T, T); 2]> for Line<T> {
 #[cfg(test)]
 impl<T> RelativeEq for Line<T>
 where
-    T: AbsDiffEq<Epsilon = T> + CoordinateType + RelativeEq
+    T: AbsDiffEq<Epsilon = T> + CoordinateType + RelativeEq,
 {
-
     #[inline]
     fn default_max_relative() -> Self::Epsilon {
         T::default_max_relative()
@@ -187,11 +186,8 @@ where
         epsilon: Self::Epsilon,
         max_relative: Self::Epsilon,
     ) -> bool {
-        self.start
-            .relative_eq(&other.start, epsilon, max_relative)
-            && self
-                .end
-                .relative_eq(&other.end, epsilon, max_relative)
+        self.start.relative_eq(&other.start, epsilon, max_relative)
+            && self.end.relative_eq(&other.end, epsilon, max_relative)
     }
 }
 
@@ -206,9 +202,7 @@ impl<T: AbsDiffEq<Epsilon = T> + CoordinateType> AbsDiffEq for Line<T> {
 
     #[inline]
     fn abs_diff_eq(&self, other: &Self, epsilon: Self::Epsilon) -> bool {
-        self.start
-            .abs_diff_eq(&other.start, epsilon)
-            && self.end.abs_diff_eq(&other.end, epsilon)
+        self.start.abs_diff_eq(&other.start, epsilon) && self.end.abs_diff_eq(&other.end, epsilon)
     }
 }
 
--- a/geo-types/src/line.rs
+++ b/geo-types/src/line.rs
@@ -1,7 +1,7 @@
 use crate::{Coordinate, CoordinateType, Point};
-#[cfg(any(feature = "relative_eq", test))]
+#[cfg(feature = "relative_eq")]
 use approx::AbsDiffEq;
-#[cfg(any(feature = "relative_eq", test))]
+#[cfg(feature = "relative_eq")]
 use approx::RelativeEq;
 
 /// A line segment made up of exactly two
--- a/geo-types/src/line.rs
+++ b/geo-types/src/line.rs
@@ -3,8 +3,7 @@ use crate::{Coordinate, CoordinateType, Point};
 use approx::AbsDiffEq;
 #[cfg(test)]
 use approx::RelativeEq;
-#[cfg(test)]
-use num_traits::Float;
+
 /// A line segment made up of exactly two
 /// [`Coordinate`s](struct.Coordinate.html).
 ///
@@ -173,18 +172,18 @@ impl<T: CoordinateType> From<[(T, T); 2]> for Line<T> {
 #[cfg(test)]
 impl<T> RelativeEq for Line<T>
 where
-    T: AbsDiffEq<Epsilon = T> + CoordinateType + Float + RelativeEq
+    T: AbsDiffEq<Epsilon = T> + CoordinateType + RelativeEq
 {
 
     #[inline]
     fn default_max_relative() -> Self::Epsilon {
-        T::epsilon()
+        T::default_max_relative()
     }
 
     #[inline]
     fn relative_eq(
         &self,
-        other: &Line<T>,
+        other: &Self,
         epsilon: Self::Epsilon,
         max_relative: Self::Epsilon,
     ) -> bool {
@@ -197,16 +196,16 @@ where
 }
 
 #[cfg(test)]
-impl<T: AbsDiffEq<Epsilon = T> + CoordinateType + Float> AbsDiffEq for Line<T> {
+impl<T: AbsDiffEq<Epsilon = T> + CoordinateType> AbsDiffEq for Line<T> {
     type Epsilon = T;
 
     #[inline]
     fn default_epsilon() -> Self::Epsilon {
-        T::epsilon()
+        T::default_epsilon()
     }
 
     #[inline]
-    fn abs_diff_eq(&self, other: &Line<T>, epsilon: Self::Epsilon) -> bool {
+    fn abs_diff_eq(&self, other: &Self, epsilon: Self::Epsilon) -> bool {
         self.start
             .abs_diff_eq(&other.start, epsilon)
             && self.end.abs_diff_eq(&other.end, epsilon)
--- a/geo-types/src/line.rs
+++ b/geo-types/src/line.rs
@@ -197,7 +197,7 @@ where
 }
 
 #[cfg(test)]
-impl<T: AbsDiffEq + CoordinateType + Float> AbsDiffEq for Line<T> {
+impl<T: AbsDiffEq<Epsilon = T> + CoordinateType + Float> AbsDiffEq for Line<T> {
     type Epsilon = T;
 
     #[inline]
@@ -207,9 +207,9 @@ impl<T: AbsDiffEq + CoordinateType + Float> AbsDiffEq for Line<T> {
 
     #[inline]
     fn abs_diff_eq(&self, other: &Line<T>, epsilon: Self::Epsilon) -> bool {
-        self.start_point()
-            .abs_diff_eq(&other.start_point(), epsilon)
-            && self.end_point().abs_diff_eq(&other.end_point(), epsilon)
+        self.start
+            .abs_diff_eq(&other.start, epsilon)
+            && self.end.abs_diff_eq(&other.end, epsilon)
     }
 }
 
--- a/geo-types/src/line.rs
+++ b/geo-types/src/line.rs
@@ -1,5 +1,10 @@
 use crate::{Coordinate, CoordinateType, Point};
-
+#[cfg(test)]
+use approx::AbsDiffEq;
+#[cfg(test)]
+use approx::RelativeEq;
+#[cfg(test)]
+use num_traits::Float;
 /// A line segment made up of exactly two
 /// [`Coordinate`s](struct.Coordinate.html).
 ///
@@ -165,6 +170,49 @@ impl<T: CoordinateType> From<[(T, T); 2]> for Line<T> {
     }
 }
 
+#[cfg(test)]
+impl<T> RelativeEq for Line<T>
+where
+    T: CoordinateType + Float + AbsDiffEq,
+{
+
+    #[inline]
+    fn default_max_relative() -> Self::Epsilon {
+        T::epsilon()
+    }
+
+    #[inline]
+    fn relative_eq(
+        &self,
+        other: &Line<T>,
+        epsilon: Self::Epsilon,
+        max_relative: Self::Epsilon,
+    ) -> bool {
+        self.start_point()
+            .relative_eq(&other.start_point(), epsilon, max_relative)
+            && self
+                .end_point()
+                .relative_eq(&other.end_point(), epsilon, max_relative)
+    }
+}
+
+#[cfg(test)]
+impl<T: AbsDiffEq + CoordinateType + Float> AbsDiffEq for Line<T> {
+    type Epsilon = T;
+
+    #[inline]
+    fn default_epsilon() -> Self::Epsilon {
+        T::epsilon()
+    }
+
+    #[inline]
+    fn abs_diff_eq(&self, other: &Line<T>, epsilon: Self::Epsilon) -> bool {
+        self.start_point()
+            .abs_diff_eq(&other.start_point(), epsilon)
+            && self.end_point().abs_diff_eq(&other.end_point(), epsilon)
+    }
+}
+
 #[cfg(feature = "rstar")]
 impl<T> ::rstar::RTreeObject for Line<T>
 where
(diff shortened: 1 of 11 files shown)


geo-types/src/point.rs
```diff
@@ -409,6 +414,79 @@ where
...
+            || other.x().is_infinite()
+            || other.y().is_infinite()
+        {
+            return false;
+        }
+
+        let diff_x = self.x() - other.x();
+        let diff_y = self.y() - other.y();
+        let abs_diff = (diff_x * diff_x + diff_y * diff_y).sqrt();
+        println!("compute largest");
```
martinfrances107, line 452: I need to remove this debug

geo-types/src/line_string.rs
```diff
@@ -281,6 +288,61 @@ impl<T: CoordinateType> IndexMut<usize> for LineString<T> {
     }
 }
 
+#[cfg(test)]
+impl<T> RelativeEq for LineString<T>
+where
+    T: CoordinateType + Float + AbsDiffEq,
+{
+    // type Epsilon = T;
```
michaelkirk, line 296: Does this comment mean something or just leftover and can can be removed?

geo-types/src/line_string.rs
```diff
@@ -281,6 +288,61 @@ impl<T: CoordinateType> IndexMut<usize> for LineString<T> {
...
+impl<T: AbsDiffEq + CoordinateType + Float> AbsDiffEq for LineString<T> {
+    type Epsilon = T;
+
+    #[inline]
+    fn default_epsilon() -> Self::Epsilon {
+        T::epsilon()
+    }
+
+    #[inline]
+    fn abs_diff_eq(&self, other: &LineString<T>, epsilon: Self::Epsilon) -> bool {
```
michaelkirk, line 335: Can we do the num_coords check optimization here too like you did in relative_eq?

geo-types/src/multi_point.rs
```diff
@@ -91,6 +98,78 @@ impl<T: CoordinateType> MultiPoint<T> {
...
+{
+    type Epsilon = T;
+
+    #[inline]
+    fn default_epsilon() -> Self::Epsilon {
+        T::epsilon()
+    }
+
+    #[inline]
+    fn abs_diff_eq(&self, other: &MultiPoint<T>, epsilon: Self::Epsilon) -> bool {
```
michaelkirk, line 162: Can we do the `if self.num_coords() != other.num_coords() {` check?

geo-types/src/point.rs
```diff
@@ -409,6 +414,79 @@ where
     }
 }
 
+#[cfg(test)]
+impl<T> RelativeEq for Point<T>
+where
+    T: CoordinateType + Float,
+{
+    // type Epsilon = T;
```
michaelkirk, line 422: Is this comment vestigial?

geo-types/src/point.rs
```diff
@@ -409,6 +414,79 @@ where
     }
 }
 
+#[cfg(test)]
+impl<T> RelativeEq for Point<T>
```
michaelkirk, line 418: Please put the implementation on `Coordinate` and change `Point`'s implementation to delegate to that - like:
```
impl<T> RelativeEq for Point<T> {
    fn relative_eq(self, other, epsilon) -> bool {
        self.0.relative_eq(other.0, epsilon)
    }
   ...
}
```

geo-types/src/multi_point.rs
```diff
@@ -91,6 +98,78 @@ impl<T: CoordinateType> MultiPoint<T> {
...
+    /// # Examples
+    ///
+    /// ```
+    /// use geo_types::MultiPoint;
+    ///
+    /// let mut coords = vec![(0., 0.), (5., 0.), (7., 9.)];
+    /// let multi_point: MultiPoint<f32> = coords.into_iter().collect();
+    /// assert_eq!(3, multi_point.num_coords());
+    /// ```
+    pub fn num_coords(&self) -> usize {
```
michaelkirk, line 113: This is a helpful method. 

Since @frewsxcv has a WIP implementation for the same functionality here: https://github.com/georust/geo/pull/563, I think it'd be better to avoid the public API churn - can you make this method private or `pub(crate)` for now, and add a comment linking to the as-of-yet unmerged PR to explain why it's not public.

geo-types/src/point.rs
```diff
@@ -409,6 +414,79 @@ where
...
+    }
+
+    #[inline]
+    fn relative_eq(
+        &self,
+        other: &Point<T>,
+        epsilon: Self::Epsilon,
+        max_relative: Self::Epsilon,
+    ) -> bool {
+        // Handle same infinities.
```
michaelkirk, line 436: I was initially surprised that we weren't simply delegating to the `approx::relative_eq`, but I think I understand why.

It might be helpful for future readers if there were a comment that explained this - something like:

```
// this implementation is based on the [approx crates `relative_eq`](some link) 
// implementation, but rather than comparing the difference
// between two scalars, we compare the euclidean distance between two 
// points
```
(feel free to improve)

geo-types/src/line_string.rs
martinfrances107: Thanks .. I will fix this

geo-types/src/point.rs
martinfrances107: Thanks, it is a mistake... fix.

geo-types/src/line_string.rs
```diff
@@ -281,6 +288,61 @@ impl<T: CoordinateType> IndexMut<usize> for LineString<T> {
...
+        let points_zipper = self.points_iter().zip(other.points_iter());
+        for (lhs, rhs) in points_zipper {
```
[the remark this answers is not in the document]
martinfrances107: Thank you, that is a really nice touch... I have learn't something... fixed.

geo-types/src/multi_point.rs
martinfrances107: Thanks fixed. Also added test_abs_diff_eq() which is almost a copy of test_relative_eq() because that,  mistake of mine, should have been caught by the tests.

martinfrances107:
I am slowly fixing up the code in response to the review. I will post again when I think I have gotten to all the problems.

geo-types/src/point.rs
martinfrances107: > please put the implementation on 

I think when I submit a fix... there is going to be a radical change here. So I want to signal it ahead of time

in the main issue .. I talked about euclidean distance versus 'Manhattan' distance. 

Coordinate{x: T, y: T} has such a similar structure to Complex{ re: T, im: T}

I think I made a over-complication when I look in approx-0.4..0 at the implementation of Complex I see this ..

The implementation forgets the euclidean distance and just ensures that the both re and im meet the constraint individually. Because I copy/modifed the default implementation .. the same infinity checks for example will be done in the same order .. etc  

```
#[cfg(feature = "num-complex")]
impl<T: RelativeEq> RelativeEq for Complex<T>
where
    T::Epsilon: Clone,
{
    #[inline]
    fn default_max_relative() -> T::Epsilon {
        T::default_max_relative()
    }

    #[inline]
    fn relative_eq(
        &self,
        other: &Complex<T>,
        epsilon: T::Epsilon,
        max_relative: T::Epsilon,
    ) -> bool {
        T::relative_eq(&self.re, &other.re, epsilon.clone(), max_relative.clone())
            && T::relative_eq(&self.im, &other.im, epsilon.clone(), max_relative.clone())
    }
}
```

michaelkirk: Honestly - I think using manhattan distance is fine — it might even be preferred because it's so simple.

geo-types/src/line.rs
```diff
@@ -165,6 +170,49 @@ impl<T: CoordinateType> From<[(T, T); 2]> for Line<T> {
...
+    }
+
+    #[inline]
+    fn relative_eq(
+        &self,
+        other: &Line<T>,
+        epsilon: Self::Epsilon,
+        max_relative: Self::Epsilon,
+    ) -> bool {
+        self.start_point()
```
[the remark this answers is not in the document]
martinfrances107: I have fixed this in the commit, to this PR, labelled 'points.rs Simplify RelativeEq for Point'

geo-types/src/line.rs
```diff
@@ -165,6 +170,49 @@ impl<T: CoordinateType> From<[(T, T); 2]> for Line<T> {
...
+    type Epsilon = T;
+
+    #[inline]
+    fn default_epsilon() -> Self::Epsilon {
+        T::epsilon()
+    }
+
+    #[inline]
+    fn abs_diff_eq(&self, other: &Line<T>, epsilon: Self::Epsilon) -> bool {
+        self.start_point()
```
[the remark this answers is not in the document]
martinfrances107: Thanks you fixed in : -
lines.rs Simplify "AbsDiffEq for Line<T>" to operate of coords not points.

geo-types/src/line_string.rs
martinfrances107: Thanks this was fixed in the commit 

line_string.rs -  abs_diff_eq() Now rejects line_strings with unequal lengths. (with tests)

geo-types/src/point.rs
martinfrances107: Great -- so this has been fixed. The implementation was not moved into coords .. it was simply dissolved away.

michaelkirk: I'd actually still prefer if this were moved into Coordinate as described, even if the implantation is much simpler now.

Reason being that `Coordinate` is kind of the currency of lots of geo algorithms, so it'd be great if we could start to leverage this functionality in those tests too.

michaelkirk:
Looks like you're making steady progress @martinfrances107! Please tag me when you're ready for another round of review.

geo-types/src/point.rs
martinfrances107: Thank you .. yep I had only sorted this out for RelativeEq... While I was fixing this - I noticed that I was systematically over constraining this stuff with 'Float' ... fixed.

this commit.

Removed type constraint on RelativeEq and AbsDiffEq that the type must be Float. Also AbsDiffEq for Point now uses AbsDiffEq in Coords.

geo-types/src/line.rs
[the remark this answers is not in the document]
martinfrances107: I want to respond to this question 

> With these traits implemented, we can now use something like:
> approx::assert_relative_eq!(line_string_1, line_string_2);
> Right?

I have added a new commit here "Added new feature relative_eq"
It does not work as I want yet... so there is more to come. 

But I am at the point where I should discuss some architectural re-plumbing in public. 

I actually want this to work in a module I am developing. This project rightly wants to keep approx as an internal dev-dependency. So If I want to see in in my third party module then I need to optionally expose approx when the other build environment ( Cargo.toml) enables  the geo module's feature "relative_eq" This might explain some of my changes.

My current issue is that I can ONLY demonstrate approx::assert_relative_eq working  in a limited number of build environments. It works in 'test' and 'doctest' ONLY and ONLY inside geo-types. When I try and demo this in geo sub-crate ( ie geo/src/algorithm/rotate.rs)  or my third party module.. then it fails to compile..

geo-types/src/multi_point.rs
michaelkirk: BTW https://github.com/georust/geo/pull/563 has since merged.

Could you please rebase or merge against master and use the newly added `CoordsIter#coords_count` method instead of this one?

martinfrances107: There is a question of placement, I would like to discuss..

CoordsItrer#coords_count is defined in the geo module.

And there is a security conflict. The implementation of the trait must be in the module defining the trait or the module containing the struct upon which the trait operates. 

The conflict is that I can't 

A) use coords_count() in geo_types.

or 

B) I can't define RelatveEq for Point<T> in geo because it it viewed as a unsafe third party module.

I will post and update when I have found it ... but I can't do what you ask.

martinfrances107: So I have rebased and resolved the deprection notices. here 

"After rebase, drop the use of the deprecated num_coords() method."

The solution is that coords_count is a simple one-liner - so I opted to break the DRY principle and inline the function is a few places .. no biggie

martinfrances107:
@michaelkirk , Hi

I think I have fixed all the outstanding issues  - and this is PR is ready for  a review again

I have modified geo/src/algorithm/rotate.rs to use the new assertions and avoid hardcoding numerical rounding errors in the expected test results.

I have checked cargo doc ... and looked at the HTML output .. example of how to used the new methods are documented.

I wanted to be able to demonstrate these test / dependencies features being used my third party module but I think I want to leave that as a followup.

Thanks for the detailed review it really did greatly simplify the code and caught to holes in the test strategy

Merry Christmas!

geo-types/src/multi_point.rs
michaelkirk: It think this will lead to frustration for our users down the road to have methods with identical functionality but different names. I don't think people will be able to remember that  *all* geo::Geometries can use `coords_count` but that only *some* geo::Geometries can use `num_coords`. 

As you rightly point out, we unfortunately can't use the `geo::CoordsIter` trait in the `geo-types` crate, so instead I think it's best to delete this `num_coords` method and use the more verbose `self.0.len()` in the `geo-types` crate.

geo-types/src/multi_point.rs
```diff
@@ -91,6 +96,99 @@ impl<T: CoordinateType> MultiPoint<T> {
...
+    /// # Examples
+    ///
+    /// ```
+    /// use geo_types::MultiPoint;
+    /// use geo_types::point;
+    ///
+    /// let a = MultiPoint(vec![point![x: 0., y: 0.], point![x: 10., y: 10.]]);
+    /// let b = MultiPoint(vec![point![x: 0., y: 0.], point![x: 10.001, y: 10.]]);
+    ///
+    /// approx::assert_relative_eq!(a, b, epsilon=0.1)
```
michaelkirk, line 179: ditto about assert_abs_diff_eq and the comment

geo/Cargo.toml
```diff
@@ -22,19 +22,20 @@ geographiclib-rs = { version = "0.2" }
 
 proj = { version = "0.20.3", optional = true }
 
-geo-types = { version = "0.6.2", path = "../geo-types", features = ["rstar"] }
+geo-types = { version = "0.6.2", optional = true, path = "../geo-types", features = ["rstar"] }
```
michaelkirk, line 25: I'm not 100% I understand the changes in this file. Surely geo cannot be built without geo-types, so making it an optional dependency seems fraught.

I'm guessing this is related to enabling the `relative_eq` by default only as a devDependency... Is that right? I see why that might be tricky.

A maybe radical alternative which avoids the problem is to get rid of the `relative_eq` feature you've introduced and simply have the new functionality you've written enabled by default. 

I appreciate the intent, because we try to keep geo-types svelte, but since geo-types already unconditionally relies on the `approx` crate, including your new code for the approx impls seems like an acceptable addition to me.

Anyone else care to weigh in? 
/cc @georust/geo-publishers

geo/src/algorithm/rotate.rs
```diff
@@ -334,16 +335,13 @@ mod test {
     #[test]
     fn test_rotate_line() {
         let line0 = Line::from([(0., 0.), (0., 2.)]);
-        let line1 = Line::from([(1., 0.9999999999999999), (-1., 1.)]);
-        assert_eq!(line0.rotate(90.), line1);
+        let line1 = Line::from([(1., 1.), (-1., 1.)]);
+        assert_relative_eq!(line0.rotate(90.0), line1);
```
michaelkirk, line 339: 🎉 yay!

geo/src/algorithm/rotate.rs
```diff
@@ -334,16 +335,13 @@ mod test {
...
     }
     #[test]
     fn test_rotate_line_around_point() {
         let line0 = Line::new(Point::new(0., 0.), Point::new(0., 2.));
-        let line1 = Line::new(
-            Point::new(0., 0.),
-            Point::new(-2., 0.00000000000000012246467991473532),
```
michaelkirk, line 345: 🎉 yay! good riddance!

michaelkirk (changes requested):
This is looking very close! 

I had some additional feedback and am hoping to solicit some input from other georust folks in one comment in particular...

michaelkirk:
And thank you for all the good docs and tests!

geo/Cargo.toml
martinfrances107: Thanks for the review, this is a tricky one 

yes to a radical alternatives... this frustrates me ..Let me state the problem in very simplistic terms just so someone who has seen this before can say .. Oh that is easy.

I want different features to be present on (A) the release build and (B) the dev build.

Making geo-types optional is bad when it must always be included. I was constrained here ... there are long threads on stackoverflow about this -  and meandering discussions amongst rustlang  developers  with lots of well meaning comments along the line of .. we can do better but no action.  I hope I am just making a silly mistake.

martinfrances107:
Thanks for the carefully review again... I have fixed all my silly mistakes. 

The major problem concerning how to pull in different features into different build environment persists.

geo/Cargo.toml
michaelkirk: To be clear, I'm just waiting a few days to solicit feedback from @georust/geo-publishers

The specific question is:

Since we already include approx by default in geo-types, are we ok with adding these implementations of approx traits on geo-types by default as well.

frewsxcv: This looks like the feature we'd want: https://github.com/rust-lang/cargo/issues/7916

frewsxcv: > This looks like the feature we'd want: [rust-lang/cargo#7916](https://github.com/rust-lang/cargo/issues/7916)

And if so, it looks like this feature should stabilize soon: https://github.com/rust-lang/cargo/pull/8997

So it seems like we have a few options:

1. Wait to merge this until https://github.com/rust-lang/cargo/issues/7916 lands on stable Rust
2. Make geo-types optional, which is what this pull request currently does. When https://github.com/rust-lang/cargo/issues/7916 lands, we can make `geo-types` required again
3. Keep geo-types required, but always include the `relative_eq` feature. Even on non-test builds. When https://github.com/rust-lang/cargo/issues/7916 lands, we can turn on `relative_eq` only in `dev-dependencies`

I'm fine with any of these options, with a slight preference for option 3. It's unfortunate we introduce a new dependency for everyone, but it'll hopefully be temporary.

If we go with options 2 or 3, we should add a comment or open a GitHub Issue saying we can improve the situation when https://github.com/rust-lang/cargo/issues/7916 lands.

michaelkirk: I'm working on a slightly different proposal... hopefully I'll have something to share in 20min.

michaelkirk: please see https://github.com/martinfrances107/geo/pull/1

michaelkirk: What do you think of this solution @frewsxcv (et al)?

Currently in master:

- geo-types dependency on approx is not optional
- the methods in geo_types::private_utils are unconditionally built, but only used within geo-types if the `rstar` feature is enabled.

After this PR:

- geo-types dependency on approx is optional
- These new implementations for RelativeEq and AbsDiffEq are enabled with the now optional `approx` feature
- There is "public" API breakage, in that methods in geo_types::private_utils are now skipped unless the rstar feature which requires them is enabled (`geo` builds `geo-types` w/ `rstar` feature, so this won't affect `geo`).

So technically, if some other crate is using the methods in private_utils, and not enabling the `rstar` feature, this would break for them. Even though this technically breaks a public API, I _think_ it's OK to treat this as a minor change since the changes are in a module with the word "private" in its name. WDYT?

frewsxcv: SGTM! Great solution

michaelkirk (approved):
bors r+

Ok, everything seems to be in order! Thanks for working through the lengthy review process with us @martinfrances107, I'm pretty happy with where it ended up. Thanks for your contribution!

Issue #573, opened 2020-12-20
Update rand version number


martinfrances107:
This is a minor house keeping issue.

In the language of Semver. The minor version of rand has been bumped.

The rand dependency occurs only in one place. I am about to create a one line PR

PR #574, opened 2020-12-20
#573 Update rand version to 0.8.0

martinfrances107:
---

This is in answer to this PR #573

### DIFF
--- a/geo/Cargo.toml
+++ b/geo/Cargo.toml
@@ -1,4 +1,4 @@
-[package]
+    [package]
 name = "geo"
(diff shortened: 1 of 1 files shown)


geo/Cargo.toml
```diff
@@ -1,4 +1,4 @@
-[package]
+    [package]
```
michaelkirk, line 1: weird indent here

martinfrances107: Opps... fixed.

michaelkirk:
thanks!

bors r+

PR #581, opened 2020-12-29
Fix bench compiler error; use black_box in more places.

frewsxcv:
---

Fix a couple lines that were causing compiler errors. And extended `black_box` to be used on all parameters and return values. I'm not sure if we need to use it on the return value (see https://github.com/rust-lang/rust/issues/80480)

### DIFF
--- a/geo/benches/concave_hull.rs
+++ b/geo/benches/concave_hull.rs
@@ -16,7 +16,7 @@ pub fn uniform_points_in_range<S: CoordinateType + SampleUniform + Signed, R: Rn
     rng: &mut R,
 ) -> Vec<Coordinate<S>> {
     (0..size)
-        .map(|_| (rng.gen_range(-range, range), rng.gen_range(-range, range)).into())
+        .map(|_| (rng.gen_range(-range..=range), rng.gen_range(-range..=range)).into())
         .collect()
 }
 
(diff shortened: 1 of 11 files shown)


geo/benches/concave_hull.rs
```diff
@@ -16,7 +16,7 @@ pub fn uniform_points_in_range<S: CoordinateType + SampleUniform + Signed, R: Rn
     rng: &mut R,
 ) -> Vec<Coordinate<S>> {
     (0..size)
-        .map(|_| (rng.gen_range(-range, range), rng.gen_range(-range, range)).into())
```
frewsxcv, line 19: This was a compilation error

geo/benches/convex_hull.rs
```diff
@@ -15,7 +15,7 @@ pub fn uniform_points_in_range<S: CoordinateType + SampleUniform + Signed, R: Rn
     rng: &mut R,
 ) -> Vec<Coordinate<S>> {
     (0..size)
-        .map(|_| (rng.gen_range(-range, range), rng.gen_range(-range, range)).into())
```
frewsxcv, line 18: This was a compilation error

michaelkirk:
Good catch!

bors r+

Issue #577, opened 2020-12-28
Make HasKernel public or introduce geo::Float trait?


michaelkirk:
A lot of our algorithms depend on Float + HasKernel. 

When downstream users want to implement methods that are generic across these same types they are stymied because HasKernel is not public outside of the geo crate.

We could make HasKernel public, and then they could implement for Float+HasKernel, but another option (and my current preference) would be to export a combined trait like:

`pub trait geo::Float: geo::CoordinateType + geo::HasKenel + num_traits::Float { }`

Note, I came across this issue while working on https://github.com/urschrei/polylabel-rs/pull/6 which is a downstream user of geo.

**public `HasKernel` approach looks like:**

geo: https://github.com/georust/geo/compare/mkirk/pub-has-kernel?expand=1

downstream geo user (polylabel): https://github.com/urschrei/polylabel-rs/compare/master...michaelkirk:mkirk/update-geo?expand=1

**`geo::Float` approach (my preference) looks like:**

geo: https://github.com/georust/geo/compare/mkirk/geo-float?expand=1
downstream geo user (polylabel): https://github.com/urschrei/polylabel-rs/compare/master...michaelkirk:mkirk/update-geo-2?expand=1

The reason I prefer the `geo::Float` approach is:

1. it's a bit tidier from the client perspective (rather than `num_traits::Float+geo::HasKernel` they just have  `geo::Float` (in fact we could tidy up the geo code base a bit with this)

2. The `HasKernel` type is a bit unintuitive for a casual reader. I think the concept of a Float will be more approachable for people just "trying to make things work".

It's very possible there's a better approach, or that I'm missing something in my analysis, so curious what other folks think.

urschrei:
I'm in favour of the new public geo::Float trait approach. We can clearly and succinctly document what it's for and what it does in a few lines, and spare downstream users the (entirely necessary internal) complexity of HasKernel IMO. @rmanoka what do you think?

rmanoka:
TL;DR:  I suggest we do both: create a "alias" trait `geo::Float`, as well as make `HasKernel` public.  Actually just making the alias trait public, would still be unusable as the end-user would anyway need to implement `HasKernel` to implement the alias trait on a new type.

[...]

PR #583, opened 2020-12-29
Allow external crates to implement generic float methods for Geometry<T>

michaelkirk:
---

Solves #577 (see discussion there for context)

You can see it in use here: https://github.com/urschrei/polylabel-rs/pull/6/files

### DIFF
--- a/geo/src/algorithm/closest_point.rs
+++ b/geo/src/algorithm/closest_point.rs
@@ -1,6 +1,6 @@
 use crate::prelude::*;
 use crate::{
-    Closest, Float, Line, LineString, MultiLineString, MultiPoint, MultiPolygon, Point, Polygon,
+    Closest, GeoFloat, Line, LineString, MultiLineString, MultiPoint, MultiPolygon, Point, Polygon,
 };
 use std::iter;
 
@@ -23,7 +23,7 @@ use std::iter;
 /// let closest = horizontal_line.closest_point(&p);
 /// assert_eq!(closest, Closest::SinglePoint(Point::new(0.0, 0.0)));
 /// ```
-pub trait ClosestPoint<F: Float, Rhs = Point<F>> {
+pub trait ClosestPoint<F: GeoFloat, Rhs = Point<F>> {
     /// Find the closest point between `self` and `p`.
     fn closest_point(&self, p: &Rhs) -> Closest<F>;
 }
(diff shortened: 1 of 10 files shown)


michaelkirk:
Actually, going to apply this internally in some places first.

michaelkirk:
Forgive the flapping...

michaelkirk:
Ok this is ready for review.

geo/src/algorithm/closest_point.rs
```diff
@@ -49,7 +49,7 @@ impl<F: Float> ClosestPoint<F> for Point<F> {
 }
 
 #[allow(clippy::many_single_char_names)]
-impl<F: Float + HasKernel> ClosestPoint<F> for Line<F> {
+impl<F: Float> ClosestPoint<F> for Line<F> {
```
michaelkirk, line 52: It's possible that people will assume this Float refers to num_traits::Float. I could easily be convinced to call it something more explicit like `GeoFloat`.

rmanoka: May be call it `geo::Float`, but `use geo::Float as GeoFloat` to clarify?  Don't have any hard opinions here either.

rmanoka (approved):
Looks good to me!

geo/src/algorithm/closest_point.rs
urschrei: I vote for GeoFloat, because it's unambiguous. Using Float will lead to people assuming that they know what it is, whereas with GeoFloat they'll have to look it up and read the docstring.

michaelkirk: Ok, I've renamed geo::Float -> geo::GeoFloat and added some doc comments.

geo/src/lib.rs
```diff
@@ -122,3 +122,35 @@ pub mod prelude {
...
+///                 if next.euclidean_distance(point) > farthest.euclidean_distance(point) {
+///                     Some(next)
+///                 } else {
+///                     Some(farthest)
+///                 }
+///             }
+///         }
+///     })
+/// }
+/// ```
```
urschrei, line 154: 👏👏👏

urschrei (approved):

michaelkirk:
bors r=rmanoka,urschrei

PR #585, opened 2020-12-30
ci bench

### LINKED ISSUE #582: Build (but don't run) benchmarks in CI
Benchmarks are currently failing on `master` (will be fixed in https://github.com/georust/geo/pull/581). But we should build benchmarks in CI so we don't hit this again

michaelkirk:
---

Fixes #582

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,8 +48,7 @@ jobs:
       - run: cargo install cargo-all-features
       - run: cargo build-all-features
       - run: cargo test-all-features
-
-  test:
+  coverage:
     name: coverage
     runs-on: ubuntu-latest
     if: "!contains(github.event.head_commit.message, '[skip ci]')"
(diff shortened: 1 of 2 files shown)


michaelkirk:
We *could* actually even run the benches without slowing down our overall build (see https://github.com/georust/geo/runs/1624487390) they're faster than our `cargo test-all-features`).

Though I expect the results would be somewhat noisy on GH actions.

frewsxcv (approved):

frewsxcv:
> We _could_ actually even run the benches without slowing down our overall build since they're run in parallel with and faster than our slowest step `cargo test-all-features` - see https://github.com/georust/geo/actions/runs/451936580.
> 
> Though I expect the results would be somewhat noisy on GH actions.

They might be noisy, and if GitHub's CI workers are using shared resources (there's a labor pun somewhere in here), then I don't think we can rely on the performance results that the CI produces, so not sure how much benefit we'd get out of it, besides detecting runtime errors

PR #592, opened 2021-01-02
Rewrite Extremes algorithm to iteratively find the extreme coordinates using CoordsIter

frewsxcv:
---

* Added
    * `geo::algorithm::extremes::Extremes` (trait)
    * `geo::algorithm::extremes::Outcome` (struct)
    * `geo::algorithm::extremes::Extreme` (struct)
* Removed
    * `geo::algorithm::extremes::ExtremeIndices` (trait)
    * `geo::algorithm::extremes::ExtremePoints` (trait)
    * `geo::Extremes` (struct)

The `Extremes` algorithm is now based on `CoordsIter` instead of `ConvexHull`. The trait now runs 6x faster than the previous implementation:

Screen Shot 2021-01-06 at 12 26 13 PM

This speeds up any algorithms that incorporate `Extremes`, like `Polygon` to `Polygon` distance:

Screen Shot 2021-01-06 at 12 28 59 PM

### DIFF
--- a/geo/src/algorithm/extremes.rs
+++ b/geo/src/algorithm/extremes.rs
@@ -28,11 +28,13 @@ pub trait Extremes<'a, T: CoordinateType> {
     fn extremes(&'a self) -> Option<Outcome<T>>;
 }
 
+#[derive(Debug, PartialEq)]
 pub struct Extreme<T: CoordinateType> {
     pub index: usize,
     pub coord: Coordinate<T>,
 }
 
+#[derive(Debug, PartialEq)]
 pub struct Outcome<T: CoordinateType> {
     pub x_min: Extreme<T>,
     pub y_min: Extreme<T>,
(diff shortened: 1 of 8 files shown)


geo/src/algorithm/extremes.rs
```diff
@@ -1,274 +1,131 @@
...
+///     (x: 1.0, y: 2.0),
+///     (x: 0.0, y: 1.0),
+///     (x: 1.0, y: 0.0),
+/// ];
+///
+/// let extremes = polygon.extremes().unwrap();
+///
+/// assert_eq!(extremes.y_max.index, 2);
+/// assert_eq!(extremes.y_max.coord.x, 1.);
+/// assert_eq!(extremes.y_max.coord.y, 2.);
```
michaelkirk, line 25: It just occurred to me, but this is effectively the same thing as a bounding box, except that we also return the indices, right?

frewsxcv: @michaelkirk `Extremes` returns the most extreme coordinates of the geometry. In the case of a diamond, it will return each of the four coordinates. `BoundingRect` returns a minimum bounding `Rect` encompassing the geometry. The coordinates of the `Rect` are not necessarily vertices. In the case of a diamond, it will not return any of the four coordinates. The minimum bounding rectangle can be deduced from the extremes, but the reverse is not true. Does that answer your question?

michaelkirk: > In the case of a diamond, it will return each of the four coordinates

Ah, it returns the entire coord for each extreme, not just the extreme component of the coord. Thanks for explaining!

We could similarly use `exterior_coords` for bounding_rect. It seems like it's already doing everything optimally, so it'd just be a little code cleanup.

I think the only case where it would be less optimal is for Rects, for which we already know the min/max. You could apply a similar optimization for `Rect` in `extremes` to save yourself a couple comparisons if you're feeling like shaving off some nanograms, but seems like small potets.

frewsxcv:
This is ready for review! I'll squash the commits in a little bit

frewsxcv:
I'm just gonna squash with GitHub when we're ready to merge.

bors try

geo/src/algorithm/extremes.rs
```diff
@@ -1,274 +1,131 @@
...
 where
-    T: Signed + HasKernel,
+    G: CoordsIter<'a, Scalar = T>,
+    T: CoordinateType,
 {
-    fn extreme_indices(&self) -> Result<Extremes, ()> {
-        find_extreme_indices(polymax_naive_indices, self)
-    }
-}
+    fn extremes(&'a self) -> Option<Outcome<T>> {
```
michaelkirk, line 50: I like the new API! 

How arduous would it be to maintain a deprecated version of the old API for compatibility? 

I know we're pre-v1 so technically we can "do whatever we want", but whenever we bump semver (for pre v1 like us, that's the `X` in `0.X.Y`), people get left behind for a while, which means it's more likely they're using old code until their crate maintainers update, or are stuck with multiple incompatible versions.

That's why when it's easy, I'd prefer to do things in a way that uses deprecation so downstream users don't get stranded or need multiple versions of geo for compatibility reasons. Then we can batch all the breaking changes together, by deleting the deprecated methods, only once in a while.

I also think in-line deprecation notices are more helpful upgrade documentation than release notes for the people who update before the deprecated methods are removed.

(If it's not easy, then I'm OK with merging as is and bumping the minor version)

michaelkirk (approved):
Nice perf increase!

geo/src/algorithm/extremes.rs
frewsxcv: Before I started implementing this, I took a glance at these GitHub queries: [1](https://github.com/search?l=rust&q=geo+extremepoints&type=code) [2](https://github.com/search?l=rust&q=geo+extremeindices&type=code). Assuming these are doing the right thing, there's (essentially) no open source project utilizing the old traits. There could be users, but we at least have some indication there's minimal breakage in the GeoRust ecosystem. And I also took a look at the [current changelog](https://github.com/georust/geo/blob/master/geo/CHANGES.md) which already has at least [one breaking change](https://github.com/georust/geo/pull/593), so no matter what we're gonna need to bump to 0.17.0 with the next release (unless we were to branch off master). So I'd say it's not difficult to do the deprecation dance, and I do understand the benefits, but it still takes time for myself and other maintainers. So all in all I'm neutral about doing it in this case, but I have capacity to do it if you still prefer.

michaelkirk:
Eh, since we’re already breaking in the next release, I’m fine with skipping it.

michaelkirk:
Also, cool trick about using the GH search that way. I should utilize it more.

Issue #597, opened 2021-01-08
CoordinatesType should extend Debug


frewsxcv:
https://github.com/georust/geo/issues/596#issuecomment-756618060

michaelkirk:
I'm working on implementing this now for the upcoming geo-type 0.7 release

frewsxcv:
https://github.com/georust/geo/pull/602

PR #602, opened 2021-01-11
+Debug for CoordinateType

michaelkirk:
- [TODO] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

This is intended to address https://github.com/georust/geo/issues/597

I took a broader approach, and wonder if people are interested in it. The tldr is:

### geo-types
- deprecate `CoordinateType`; (now includes `std::fmt::Debug`)
- Add `trait CoordNum: CoordinateType;` 
- Add `trait CoordFloat: CoordNum + num_traits::Float`

### geo
Add `trait GeoNum: CoordNum + HasKernel`
Add `trait GeoFloat: CoordFloat + HasKernel`

**downstream adoption**

If your crate interops with geo-types, you may need  to make changes like this:

```
pub fn foo<T>(line_string: &geo_types::LineString<T>)
where
-    T: num_traits::Float,
+    T: geo_types::CoordFloat,
{
```

or this (this one is actually only a deprecation, so you could leave it as is for now, but eventually you'll need to make the change):
```
pub fn foo<T>(line_string: &geo_types::LineString<T>)
where
-    T: geo_types::CoordinateType
+    T: geo_types::CoordNum
{
```

**alternative considered**

If I went with the minimal approach to adding Debug trait, changes would instead need to look like this:

This one, for non-floats, could remain untouched:
```
pub fn foo<T>(line_string: &geo_types::LineString<T>)
where
    T: geo_types::CoordinateType
{
```

But anything restricted to floats would need to explicitly include Debug

```
pub fn foo<T>(line_string: &geo_types::LineString<T>)
where
-    T: num_traits::Float,
+    T: num_traits::Float + std::fmt::Debug,
{
```

**motivation:**

I initially completed #597 with the most conservative approach, achieved in the first two commits. The first simply adds Debug to CoordinateType and the second fixes the resulting compiler errors. 

This is a breaking change for anyone constraining generic implementations to just `num_traits::Float`, e.g. like [geojson crate does](https://github.com/georust/geojson/blob/master/src/conversion/to_geo_types.rs#L31) (fix here: https://github.com/georust/geojson/compare/mkirk/geo-types-0.7?expand=1), and polylabel (fix here: https://github.com/urschrei/polylabel-rs/compare/master...michaelkirk:mkirk/geo-types-0.7?expand=1)

Compare this to a crate like [geo-booleanop](https://github.com/21re/rust-geo-booleanop/blob/master/lib/src/boolean/helper.rs#L61), for which things "just work" because they targeted CoordinateType.

This got me thinking that, similar to #577 (yet unreleased), if we want people to write code targeting geo-types in a generic way, it might be helpful to make it clearer and easier how to do so.

So, I've added a CoordFloat and a CoordNum which is basically just a rename of CoordinateType, and applied them liberally within our own code base. I'm hoping this would drive others to reach for these traits rather than specific num_traits when implementing code intended for generic geo-types interop. 

For crates like geo-booleanop, which targeted CoordinateType, things "just work" after the update, but they'll see a deprecation notice like:
```
$ cargo build
warning: use of deprecated trait `geo_types::CoordinateType`: use `CoordFloat` or `CoordNum` instead
 --> lib/src/boolean/helper.rs:2:29
  |
2 | use geo_types::{Coordinate, CoordinateType};
  |                             ^^^^^^^^^^^^^^
  |
  = note: `#[warn(deprecated)]` on by default
```

Which I think is acceptably clear.

It's a large number of lines changed change overall, I know! But it's mostly mechanical, and I thought, since we're breaking geo-types, maybe this is a good time to bite the bullet and avoid some future pain. 

If it's too controversial, as I said, the minimal amount of change we need to enable Debug for CoordinateType is the first three commits (through and including "address fallout of "Add Debug to CoordinateType"). It's still a pretty big change with just that, and note that this will still break the crates I mentioned. I intend to fix the ones I know about, but there's a long tail of crates, which is why I hate breaking geo-types. 

In the few that I've checked it's been quick to adapt to the change, but it might be worth announcing and letting sit open for a while in case downstream geo-types users want to weigh in.

### DIFF
--- a/geo-types/src/lib.rs
+++ b/geo-types/src/lib.rs
@@ -35,7 +35,7 @@
 //! [JTS]: https://github.com/locationtech/jts
 //! [GEOS]: https://trac.osgeo.org/geos
 extern crate num_traits;
-use num_traits::{Num, NumCast};
+use num_traits::{Float, Num, NumCast};
 use std::fmt::Debug;
 
 #[cfg(feature = "serde")]
@@ -57,13 +57,17 @@ impl<T: Num + Copy + NumCast + PartialOrd + Debug> CoordinateType for T {}
 
 /// The type of an x or y value of a point/coordinate.
 ///
-/// Floats (`f32` and `f64`) and Integers (`u8`, `i32` etc.)
-/// implement this. Many algorithms only make sense for
-/// Float types (like area, or length calculations).
+/// Floats (`f32` and `f64`) and Integers (`u8`, `i32` etc.) implement this.
+///
+/// For algorithms which only make sense for floating point, like area or length calculations,
+/// see [CoordFloat](trait.CoordFloat.html).
 #[allow(deprecated)]
 pub trait CoordNum: CoordinateType + Debug {}
 #[allow(deprecated)]
 impl<T: CoordinateType + Debug> CoordNum for T {}
+
+pub trait CoordFloat: CoordNum + Float {}
+impl<T: CoordNum + Float> CoordFloat for T {}
 
 mod coordinate;
 pub use crate::coordinate::Coordinate;
--- a/geo-types/src/lib.rs
+++ b/geo-types/src/lib.rs
@@ -51,7 +51,6 @@ extern crate approx;
 
 #[deprecated(since = "0.7", note = "use `CoordFloat` or `CoordNum` instead")]
 pub trait CoordinateType: Num + Copy + NumCast + PartialOrd + Debug {}
-// Little bit of a hack to make to make this work
 #[allow(deprecated)]
 impl<T: Num + Copy + NumCast + PartialOrd + Debug> CoordinateType for T {}
 
--- a/geo-types/src/lib.rs
+++ b/geo-types/src/lib.rs
@@ -36,6 +36,7 @@
 //! [GEOS]: https://trac.osgeo.org/geos
 extern crate num_traits;
 use num_traits::{Num, NumCast};
+use std::fmt::Debug;
 
 #[cfg(feature = "serde")]
 #[macro_use]
@@ -53,9 +54,9 @@ extern crate approx;
 /// Floats (`f32` and `f64`) and Integers (`u8`, `i32` etc.)
 /// implement this. Many algorithms only make sense for
 /// Float types (like area, or length calculations).
-pub trait CoordinateType: Num + Copy + NumCast + PartialOrd {}
+pub trait CoordinateType: Num + Copy + NumCast + PartialOrd + Debug {}
 // Little bit of a hack to make to make this work
-impl<T: Num + Copy + NumCast + PartialOrd> CoordinateType for T {}
+impl<T: Num + Copy + NumCast + PartialOrd + Debug> CoordinateType for T {}
 
 mod coordinate;
 pub use crate::coordinate::Coordinate;
--- a/geo-types/src/lib.rs
+++ b/geo-types/src/lib.rs
@@ -49,14 +49,21 @@ extern crate rstar;
 #[macro_use]
 extern crate approx;
 
+#[deprecated(since = "0.7", note = "use `CoordFloat` or `CoordNum` instead")]
+pub trait CoordinateType: Num + Copy + NumCast + PartialOrd + Debug {}
+// Little bit of a hack to make to make this work
+#[allow(deprecated)]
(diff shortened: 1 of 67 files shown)


geo-types/src/lib.rs
```diff
@@ -48,14 +49,25 @@ extern crate rstar;
 #[macro_use]
 extern crate approx;
 
+#[deprecated(since = "0.7", note = "use `CoordFloat` or `CoordNum` instead")]
+pub trait CoordinateType: Num + Copy + NumCast + PartialOrd + Debug {}
+// Little bit of a hack to make to make this work
```
frewsxcv, line 54: Not sure we need this comment anymore 🤷🏻 

```suggestion
```

geo-types/src/lib.rs
```diff
@@ -48,14 +49,25 @@ extern crate rstar;
...
-/// Float types (like area, or length calculations).
-pub trait CoordinateType: Num + Copy + NumCast + PartialOrd {}
-// Little bit of a hack to make to make this work
-impl<T: Num + Copy + NumCast + PartialOrd> CoordinateType for T {}
+/// Floats (`f32` and `f64`) and Integers (`u8`, `i32` etc.) implement this.
+///
+/// For algorithms which only make sense for floating point, like area or length calculations,
+/// see [CoordFloat](trait.CoordFloat.html).
+#[allow(deprecated)]
+pub trait CoordNum: CoordinateType + Debug {}
```
frewsxcv, line 65: Great idea! I was thinking recently about renaming `CoordinateType` recently to something like `CoordScalar`, but I wasn't sure how accurate that is, or if it makes sense since "scalar" is usually used in vectors. But I like `CoordNum`! Seems much simpler

geo/src/lib.rs
```diff
@@ -123,12 +123,13 @@ pub mod prelude {
...
-/// A common floating point trait used for geo algorithms.
+/// A common numeric trait used for geo algorithms.
 ///
 /// Different numeric types have different tradeoffs. `geo` strives to utilize generics to allow
 /// users to choose their numeric types. If you are writing a function which you'd like to be
-/// generic over the floating point types supported by geo, you probably want to constrain
-/// your function input to GeoFloat.
+/// generic over all the numeric types supported by geo, you probably want to constraint
+/// your function input to `GeoFloat`. For methods which work for integers, and not just floating
+/// point, see [`GeoNum`](trait.GeoFloat.html).
```
frewsxcv, line 132: Don't need to change it, but FYI I think you should be able to just do:

```
/// point, see [`GeoNum`].
```

and rustdoc will make it link to the appropriate docs

frewsxcv (approved):
love this!

geo/src/lib.rs
michaelkirk: Done. Thanks!

nyurik:
@michaelkirk hi, what's the status of the deprecations related to this PR? What kind of changes should be done to this and other repos that haven't been done yet?  This may simplify adding extra dimensions if I don't have to support deprecated features. Thx!

michaelkirk:
The deprecated code here is for using CoordinateType as an alias of CoordNum. I don't immediately understand how this could substantially complicate things for your implementation.

But in general, I'd advocate for deleting deprecated code the next time we have a breaking release. No breaking release is scheduled because we haven't had sufficient reason for breaking geo-types. 

We want to break geo-types rarely, and preferably accrue any less important breaking changes so we can break all at once (so that we hopefully won't have reason to break again right away).

nyurik:
@michaelkirk thx, makes sense about the breaking changes. There are a number of deprecated features, like the ones in https://github.com/georust/geo/blob/39bc4635f5d965c75f19f79960178726e6b73044/geo-types/src/geometry.rs#L90 -- so with my dimensional support, I wouldn't want to try to upgrade them only to delete them shortly after. Plus I am pretty sure the multi-dimension support will be a breaking change, so for simplicity it would make it easier to delete deprecated features as part of that.  Perhaps I should start an "official" branch in the main geo repo for multidimensional efforts, and do all the work there that would include the deprecated stuff.

WRT CoordinateType being an alias of CoordNum -- so I guess I should just delete the CoordinateType from the repo as part of this effort for simplicity sake (and to keep all breaking changes together).

What do you think?

nyurik:
P.S. I created a https://github.com/georust/geo/pull/772 that does the deprecation, and proposes a path forward

PR #611, opened 2021-01-26
macro for delegating Geometry impl to it's inner cases

michaelkirk:
- [TODO] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

Note there are some cases *not* yet covered by the macro, where the delegation isn't *quite* trivial enough (see CoordsIter). It's possible that the macro could be expanded, but my macro-fu is pretty weak and I'm trying to avoid proc_macro+syn as a dependency unless it's really worthwhile.

This does clean up some code, but the higher comprehension cost might be controversial. I'd be ok with not merging this for now if people aren't excited about it.

As an aside:

There is a crate `enum_dispatch`, which does something similar, and I
actually initially intended to use it, but I steered away for a couple
reasons:

1. enum_dispatch seems intended for the cases where "I have a trait and
   an arbitrary subset of impls. I'd like to group that arbitrary subset of impls
   into a **new enum per trait**"
   Whereas here we're really only targeting one enum: geo-types::Geometry across many traits.
2. I'd like to avoid relying on `syn`. It's a big dependency. Our dev
   dependencies (rayon) and some of our features already use it (postgres),
   but our base feature set does not.
3. enum_dispatch (and thus `syn`) would have to be added to *both* geo
   and geo-types. In particular we want to keep geo-types really light.

There's also a `delegate` crate with a similar macro. It's more sophisticated, but also relies on syn.

I could be wrong about any of the above, but that was my understanding after trying to make it work for a while.

### DIFF
--- a/geo/src/algorithm/area.rs
+++ b/geo/src/algorithm/area.rs
@@ -236,34 +236,9 @@ impl<T> Area<T> for Geometry<T>
 where
     T: CoordFloat,
 {
-    fn signed_area(&self) -> T {
-        match self {
-            Geometry::Point(g) => g.signed_area(),
-            Geometry::Line(g) => g.signed_area(),
-            Geometry::LineString(g) => g.signed_area(),
-            Geometry::Polygon(g) => g.signed_area(),
-            Geometry::MultiPoint(g) => g.signed_area(),
-            Geometry::MultiLineString(g) => g.signed_area(),
-            Geometry::MultiPolygon(g) => g.signed_area(),
-            Geometry::GeometryCollection(g) => g.signed_area(),
-            Geometry::Rect(g) => g.signed_area(),
-            Geometry::Triangle(g) => g.signed_area(),
-        }
-    }
-
-    fn unsigned_area(&self) -> T {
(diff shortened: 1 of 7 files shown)


rmanoka:
Cool macro. +1 for not wanting to add syn crate; it's indeed a large dependecy.  Would it make sense to add support for GeometryCollection too in the same macro?  These two types typically go together in impls anyway.

michaelkirk:
> Would it make sense to add support for GeometryCollection too in the same macro? These two types typically go together in impls anyway.

I'm not sure what you mean. 

Can you provide an example of what you'd envision the call site of this expanded macro looking like for GeometryCollection?

rmanoka:
> I'm not sure what you mean.
> 
> Can you provide an example of what you'd envision the call site of this expanded macro looking like for GeometryCollection?

I'm sorry, I didn't think through this carefully enough.  My point was that most GeometryCollection trait impls are boilerplate delegation to Geometry, but of course the reduction/folding to the output type is different for each method.  Pls ignore the question.

PR #613, opened 2021-01-29
export more proj types

jdroenner:
---

### DIFF
--- a/geo/src/algorithm/proj.rs
+++ b/geo/src/algorithm/proj.rs
@@ -1,1 +1,1 @@
-pub use proj::Proj;
+pub use proj::{Area, Info, Proj, ProjBuilder, ProjError, Projinfo};


michaelkirk:
The change itself seems reasonable: if we are going to re-export `proj::Proj`, it makes sense to re-export the other types you need to use `proj::Proj` in practice.

But I'm actually a little confused about why we re-export `proj::Proj` in geo at all. 

Is it easier for you to include geo-types w/ the `use-proj` than it is to include the `proj` crate w/ the `geo-types` feature?

michaelkirk:
Really curious to see how people use proj+geo in practice if anyone can share links to public projects.

urschrei:
> But I'm actually a little confused about why we re-export proj::Proj in geo at all.

The dependency graph used to go the other way: we re-exported geo-types members in proj. In practice, it ended up being really annoying because if you were depending on both proj and geo(-types) you'd get type errors if you didn't use the right crate versions, because different crate versions are considered separate types.
I thought that using the present system, you have two options:

1. Depend directly on proj from geo, and use its exported types
2. Depend on proj AND geo-types, which "automatically" activates the optional geo-types proj feature.

It's possible I've misunderstood (2), though.

jdroenner:
i tried both ways:
1. did work but i could not handle errors.
2. did not work for me because the types are not the same (however, this is the way that feels right)

jdroenner:
so i guess it would be better to remove proj from geo and import the proj crate separately (without version conflict).

michaelkirk:
> so i guess it would be better to remove proj from geo and import the proj crate separately (without version conflict).

Sorry for the lack of response. It's not because I don't care, but because I'm not really sure what the best course of action is.

I'd be ok with either merging this, or alternatively, deprecating the re-export... assuming that's a thing.

urschrei:
> I'd be ok with either merging this, or alternatively, deprecating the re-export... assuming that's a thing.

Let's merge this, and see whether deprecating the re-export in `geo` is feasible.

michaelkirk:
bors r+

SGTM

PR #614, opened 2021-01-29
Specify the details of conversion failures in an exported error enum

michaelkirk:
---

I came across this while trying to improve ergonomics of the WKT crate, which relies on this conversion logic. (see https://github.com/georust/wkt/pull/57)

I *think* this is not a breaking change (see comment inline), but would appreciate confirmation. If I'm wrong, and it indeed is a breaking change, I'd prefer to hold off on merging it for now.

### DIFF
--- a/geo-types/src/error.rs
+++ b/geo-types/src/error.rs
@@ -1,0 +1,44 @@
+use std::fmt;
+
+#[derive(Debug)]
+pub enum Error {
+    MismatchedGeometry {
+        expected: &'static str,
+        found: &'static str,
+    },
+}
+
+impl std::error::Error for Error {}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Error::MismatchedGeometry { expected, found } => {
+                write!(f, "Expected a {}, but found a {}", expected, found)
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use crate::{Geometry, Point, Rect};
+    use std::convert::TryFrom;
+
+    #[test]
+    fn error_output() {
+        let point = Point::new(1.0, 2.0);
+        let point_geometry = Geometry::from(point);
+
+        let rect = Rect::new(Point::new(1.0, 2.0), Point::new(3.0, 4.0));
+        let rect_geometry = Geometry::from(rect);
+
+        Point::try_from(point_geometry).expect("failed to unwrap inner enum Point");
+
+        let failure = Point::try_from(rect_geometry).unwrap_err();
+        assert_eq!(
+            format!("{}", failure),
+            "Expected a geo_types::point::Point<f64>, but found a geo_types::rect::Rect<f64>"
+        );
+    }
+}
--- a/geo-types/src/geometry.rs
+++ b/geo-types/src/geometry.rs
@@ -1,10 +1,9 @@
 use crate::{
-    CoordNum, GeometryCollection, Line, LineString, MultiLineString, MultiPoint, MultiPolygon,
-    Point, Polygon, Rect, Triangle,
+    CoordNum, Error, GeometryCollection, Line, LineString, MultiLineString, MultiPoint,
+    MultiPolygon, Point, Polygon, Rect, Triangle,
 };
+use core::any::type_name;
 use std::convert::TryFrom;
-use std::error::Error;
-use std::fmt;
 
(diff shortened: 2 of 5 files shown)


geo-types/src/rect.rs
```diff
@@ -264,11 +265,17 @@ impl<T: CoordFloat> Rect<T> {
...
+    since = "0.6.2",
+    note = "Use `Rect::new` instead, since `Rect::try_new` will never Error"
+)]
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub struct InvalidRectCoordinatesError;
 
+#[allow(deprecated)]
 impl std::error::Error for InvalidRectCoordinatesError {}
 
+#[allow(deprecated)]
```
michaelkirk, line 278: This deprecation stuff is sort of orthogonal. We deprecated the only method that produces this error, so we should also deprecate the error itself - in hopes that we can delete them both one day.

geo-types/src/geometry.rs
```diff
@@ -182,94 +181,56 @@ impl<T: CoordNum> Geometry<T> {
     }
 }
 
-#[derive(Debug)]
-pub struct FailedToConvertError;
```
michaelkirk, line 186: I deleted this FailedToConvertError.

My understanding is that this is *not* a breaking change, because, though it was public, it was never exported - so noone could have typed in "geo_types::FailedToConvertError" or be matching on that explicit case... so non-breaking, right?

geo-types/src/geometry.rs
```diff
@@ -182,94 +181,56 @@ impl<T: CoordNum> Geometry<T> {
...
+                }
+            }
         }
+        )+
     }
 }
 
-impl<T: CoordNum> TryFrom<Geometry<T>> for Line<T> {
-    type Error = FailedToConvertError;
+try_from_geometry_impl!(
```
michaelkirk, line 208: I'm starting to leverage more macros to DRY up our code. I'm pretty new to using macros, so if this feels sophomoric or something, let me know - I'm not actually that bothered by the mostly copy-pasted implementations that I replaced.

geo-types/src/error.rs
```diff
@@ -0,0 +1,44 @@
...
+
+        let rect = Rect::new(Point::new(1.0, 2.0), Point::new(3.0, 4.0));
+        let rect_geometry = Geometry::from(rect);
+
+        Point::try_from(point_geometry).expect("failed to unwrap inner enum Point");
+
+        let failure = Point::try_from(rect_geometry).unwrap_err();
+        assert_eq!(
+            format!("{}", failure),
+            "Expected a geo_types::point::Point<f64>, but found a geo_types::rect::Rect<f64>"
```
michaelkirk, line 41: here's what the error output actually looks like.

michaelkirk: An example of where I'd like to take advantage of these error methods, is in code like this:
```
 let wkt = wkt::Wkt::from_str(include_str!("../resources/test_fixtures/louisiana.wkt"))
            .expect("invalid test data");
        let ls = geo::LineString::<f32>::try_from(wkt).expect("unexpected geometry in wkt");
```

An error message like this is helpful!

> Expected a geo_types::line_string::LineString<f64>, but found a geo_types::geometry_collection::GeometryCollection<f64>

urschrei: This is great!

geo-types/src/geometry.rs
urschrei: I am completely clueless when it comes to writing macros but this one is readable and I understand what it's doing, so I think that's a good sign?

geo-types/src/geometry.rs
```diff
@@ -182,94 +181,56 @@ impl<T: CoordNum> Geometry<T> {
...
+    Point,
+    Line,
+    LineString,
+    Polygon,
+    MultiPoint,
+    MultiLineString,
+    MultiPolygon,
+    Rect,
+    Triangle
+);
```
frewsxcv, line 218: ~Should `GeometryCollection` be in here?~ Nvm nvm nvm

geo-types/src/rect.rs
frewsxcv: Good catch!

frewsxcv (approved):

rmanoka (approved):
Cool macros!  Changes look good to me.

michaelkirk:
bors retry

(rebased and fixed conflict in CHANGES.md)

PR #576, opened 2020-12-28
Flatten algorithm module layout

frewsxcv:
- [ ] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/master/CODE_OF_CONDUCT.md).
- [ ] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

Fixes https://github.com/georust/geo/issues/572

PR #619, opened 2021-02-04
Rewrite `geo` crate documentation.

frewsxcv:
---

Extracted out of https://github.com/georust/geo/pull/576

Screen Shot 2021-02-22 at 6 12 29 PM

### DIFF
--- a/geo/src/lib.rs
+++ b/geo/src/lib.rs
@@ -80,7 +80,7 @@
 //!
 //! ## Winding
 //!
-//! - **[`Orient`](algorithm::orient::Orient)**: Orients the exterior and interior rings of a [`Polygon`] according to convention
+//! - **[`Orient`](algorithm::orient::Orient)**: Apply a specified [`Winding`](algorithm::winding_order::Winding) to a [`Polygon`]’s interior and exterior rings
 //! - **[`Winding`](algorithm::winding_order::Winding)**: Calculate and manipulate the winding order of a [`LineString`]
 //!
 //! ## Iteration
--- a/geo/src/lib.rs
+++ b/geo/src/lib.rs
@@ -20,6 +20,9 @@
 //! - **[`GeometryCollection`]**: A collection of [`Geometry`]s
 //! - **[`Geometry`]**: An enumeration of all geometry types, excluding [`Coordinate`]
 //!
+//! The preceding types are reexported from the [`geo-types`] crate. Consider using that crate
+//! if you only need access to these types and no other `geo` functionality.
+//!
 //! ## Semantics
 //!
 //! The geospatial types provided here aim to adhere to the [OpenGIS Simple feature access][OGC-SFA]
@@ -126,6 +129,7 @@
 //! - `use-proj`: Enables coordinate conversion and transformation of `Point` geometries using the [`proj` crate]
 //! - `use-serde`: Allows geometry types to be serialized and deserialized with [Serde]
 //!
+//! [`geo-types`]: https://crates.io/crates/geo-types
 //! [`proj` crate]: https://github.com/georust/proj
 //! [Cargo features]: https://doc.rust-lang.org/cargo/reference/features.html
 //! [GEOS]: https://trac.osgeo.org/geos
--- a/geo/src/lib.rs
+++ b/geo/src/lib.rs
@@ -129,8 +129,27 @@
 //! - `use-proj`: Enables coordinate conversion and transformation of `Point` geometries using the [`proj` crate]
 //! - `use-serde`: Allows geometry types to be serialized and deserialized with [Serde]
 //!
+//! # Ecosystem
+//!
+//! There’s a wide variety of `geo`-compatible crates in the ecosystem that offer functionality not
+//! included in the `geo` crate, including:
+//!
+//! * Reading and writing file formats (e.g. [GeoJSON][geojson crate], [WKT][wkt crate],
+//!   [shapefile][shapefile crate])
+//! * [Latitude and longitude parsing][latlng crate]
+//! * [Label placement][polylabel crate]
+//! * [Geocoding][geocoding crate]
+//! * [and much more...][georust website]
+//!
 //! [`geo-types`]: https://crates.io/crates/geo-types
 //! [`proj` crate]: https://github.com/georust/proj
+//! [geojson crate]: https://crates.io/crates/geojson
+//! [wkt crate]: https://crates.io/crates/wkt
+//! [shapefile crate]: https://crates.io/crates/shapefile
+//! [latlng crate]: https://crates.io/crates/latlon
+//! [polylabel crate]: https://crates.io/crates/polylabel
+//! [geocoding crate]: https://crates.io/crates/geocoding
+//! [georust website]: https://georust.org
 //! [Cargo features]: https://doc.rust-lang.org/cargo/reference/features.html
 //! [GEOS]: https://trac.osgeo.org/geos
 //! [JTS]: https://github.com/locationtech/jts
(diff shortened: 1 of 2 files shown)


michaelkirk:
Seems like a big improvement!

geo/src/lib.rs
```diff
@@ -1,4 +1,121 @@
...
+//! # Algorithms
+//!
+//! ## Area
+//!
+//! - **[`Area`](algorithm::area::Area)**: Calculate the planar area of a geometry
+//! - **[`ChamberlainDuquetteArea`](algorithm::chamberlain_duquette_area::ChamberlainDuquetteArea)**: Calculate the geodesic area of a geometry
+//!
+//! ## Distance
+//!
+//! - **[`EuclideanDistance`](algorithm::euclidean_distance::EuclideanDistance)**: Calculate the euclidean distance between geometries
```
urschrei, line 32: Maybe _minimum_ euclidean distance?

geo/src/lib.rs
```diff
@@ -1,4 +1,121 @@
...
+//! - **[`GeodesicLength`](algorithm::geodesic_length::GeodesicLength)**: Calculate the geodesic length of a geometry using the algorithm presented in _Algorithms for geodesics_ by Charles Karney (2013)
+//! - **[`HaversineLength`](algorithm::haversine_length::HaversineLength)**: Calculate the geodesic length of a geometry using the haversine formula
+//! - **[`VincentyLength`](algorithm::vincenty_length::VincentyLength)**: Calculate the geodesic length of a geometry using Vincenty’s formula
+//!
+//! ## Simplify
+//!
+//! - **[`Simplify`](algorithm::simplify::Simplify)**: Simplify a geometry using the Ramer–Douglas–Peucker algorithm
+//! - **[`SimplifyIdx`](algorithm::simplify::SimplifyIdx)**:
+//! - **[`SimplifyVW`](algorithm::simplifyvw::SimplifyVW)**: Simplify a geometry using the Visvalingam-Whyatt algorithm
+//! - **[`SimplifyVWPreserve`](algorithm::simplifyvw::SimplifyVWPreserve)**: Simplify a geometry with the using a topology-preserving variant of the Visvalingam-Whyatt algorithm
```
urschrei, line 49: Typo: no need for "with the"

geo/src/lib.rs
```diff
@@ -1,4 +1,121 @@
...
+//! - **[`LineInterpolatePoint`](algorithm::line_interpolate_point::LineInterpolatePoint)**:
+//! - **[`LineLocatePoint`](algorithm::line_locate_point::LineLocatePoint)**:
+//!
+//! ## Similarity
+//!
+//! - **[`FrechetDistance`](algorithm::frechet_distance::FrechetDistance)**: Calculate the similarity between [`LineString`]s using the Fréchet distance
+//!
+//! ## Winding
+//!
+//! - **[`Orient`](algorithm::orient::Orient)**: Orients the exterior and interior rings of a [`Polygon`] according to convention
```
urschrei, line 69: We should probably clarify what convention means here for non-GIS users

geo/src/lib.rs
```diff
@@ -1,4 +1,121 @@
...
+//! - **[`Winding`](algorithm::winding::Winding)**: Calculate and manipulate the winding order of a [`LineString`]
+//!
+//! ## Map
```
urschrei, line 72: "Map" is a bit ambiguous in this context, although I can't immediately think of something as succint

michaelkirk: "Iteration" migh be less ambiguous, albeit not as precise.

geo/src/lib.rs
frewsxcv: Should we clarify that here or in the `Orient` trait docs?

geo/src/lib.rs
```diff
@@ -1,43 +1,139 @@
...
-//! ## Optional Features (these can be activated in your `cargo.toml`)
-//! The following optional features are available:
-//! - `use-proj`: enable coordinate conversion and transformation of `Point` geometries using the [`proj`](https://docs.rs/proj) crate
-//! - `proj-network`: enables functionality for `proj` crate's network grid. After enabling
-//! this feature, some [further
-//! configuration](https://docs.rs/proj/0.20.5/proj/#grid-file-download) is
-//! required to actually use the network grid.
-//! - `use-serde`: enable serialisation of geometries using `serde`.
-//!
-//! ## GeoJSON
```
frewsxcv, line 34: Should we keep the GeoJSON section?

michaelkirk: I'd wager almost everyone will want to use geo to operate on data derived outside of geo, so I think it's worth including and also simultaneously introducing the context of the `geo-types` crate.

Here's some off-the-cuff copy:

> [geojson](https://crates.io/crates/geojson), [wkt](https://crates.io/crates/wkt), [shapefile](https://crates.io/crates/shapefile), and [other crates](#link to addendum with exhaustive list) which support `geo-types` allow you to use `geo` with your favorite spatial format (insert joke about shapefile).
> ...
> ## Formats that support geo-types
> - postgis
> - flatgeobuffer
> - any others??? 
It's a little tricky because it'd be nice to just gloss over the fact that geo-types exists for users of geo, but in reality, people using these formats may need to build the crate with `--feature geo-types` so it'd be good to prime them to look for the word `geo-types` in that crate's documentation.

geo/src/lib.rs
michaelkirk: How about something like:

> Apply a specified [`Winding`] to a [`Polygon`]'s interior and exterior rings.

I realize that you just pulled this doc from the existing `Orient` docs, and so it's kind of unfair to put you on the hook for also improving all the docs just because you decided to helpfully consolidate them... so feel free to take or leave this suggestion.

geo/src/lib.rs
urschrei: > I'd wager almost everyone will want to use geo to operate on data derived outside of geo

This was my rationale for the original inclusion of the GeoJSON section. I think it's useful to try to explicitly point out the various components of the Geo ecosystem to people as early as possible rather than relying on them to figure it out using the context from the docs.

urschrei (approved):
Just two minor things, but this is a big improvement in clarity overall. Thank you for doing this 👏👏❤️

geo/src/lib.rs
frewsxcv: Addressed in https://github.com/georust/geo/pull/619/commits/54160329232ae16c0b9fa1aac1f40d96ef27d1f4

frewsxcv:
@michaelkirk @urschrei How do those last few commits look? Very open to suggestions for the Ecosystem section I added

michaelkirk:
> @michaelkirk @urschrei How do those last few commits look? Very open to suggestions for the Ecosystem section I added

Copy looks good! Presumably you'll need to plug in actual URL's for the ecosystem links.

Thanks for working on this!

michaelkirk (approved):

frewsxcv:
> Presumably you'll need to plug in actual URL's for the ecosystem links.

they're linked [below](https://github.com/georust/geo/pull/619/files#diff-7def06d46be6dd3a599a2327202b3b701237c2b06ee56304499fb5fc554ac965R144-R159). i don't see the [markdown link reference feature i'm using](https://spec.commonmark.org/0.29/#example-523) used often and am happy to switch to inline links if people prefer that.

michaelkirk:
> they're linked below. i don't see the markdown link reference feature i'm using used often and am happy to switch to inline links if people prefer that.

My bad! I thought they were just place holders. I think the link reference is great.

urschrei:
> @michaelkirk @urschrei How do those last few commits look? Very open to suggestions for the Ecosystem section I added

Lgtm!

frewsxcv:
bors=michaelkirk,urschrei

frewsxcv:
bors r=michaelkirk,urschrei

PR #628, opened 2021-02-20
approx for all Geometry types

michaelkirk:
---

Follow up to https://github.com/georust/geo/pull/567 which implemented for Coord, Point, LineString, and a few others, this PR implements it for the remaining types, and `Geometry` itself.

FYI this is a precursor to an integration I'm working on with some of the JTS test suite.

### DIFF
--- a/geo-types/src/geometry.rs
+++ b/geo-types/src/geometry.rs
@@ -268,77 +268,32 @@ where
         epsilon: Self::Epsilon,
         max_relative: Self::Epsilon,
     ) -> bool {
-        match self {
-            Geometry::Point(g) => {
-                if let Geometry::Point(other) = other {
-                    g.relative_eq(other, epsilon, max_relative)
-                } else {
-                    false
-                }
+        match (self, other) {
+            (Geometry::Point(g1), Geometry::Point(g2)) => g1.relative_eq(g2, epsilon, max_relative),
+            (Geometry::Line(g1), Geometry::Line(g2)) => g1.relative_eq(g2, epsilon, max_relative),
+            (Geometry::LineString(g1), Geometry::LineString(g2)) => {
+                g1.relative_eq(g2, epsilon, max_relative)
             }
-            Geometry::Line(g) => {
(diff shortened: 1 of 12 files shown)


martinfrances107:
So I had a mental checklist whilst reviewing this...

After a general review of the code ... and I satisfied myself it all looked good.

1) When zip method is used, it is always preceded by a len() equality check... otherwise its invalid.
2) Are documented tests in place... yes
3) Stage by stage - is the implementation of AbsDiffEq always paired with an RelativeEq.
4) Inspection of the auto generated documentation website -- 
    a - the style looks professional
    b - the formatting is perfect.
5) From last time, talking only about my own work  ... those #[cfg(any(feature="approx"))]
  statement take a lot of care and time to get working .. so I had re-review and double check.
  .. I am confidant all is OK.
6) is "cargo clippy" happy -- well there was a unexpected item which also appears in the main branch.

In short there is a lot of good work here...let us get this in.

michaelkirk++

geo-types/src/geometry.rs
```diff
@@ -234,3 +237,206 @@ where
...
+        max_relative: Self::Epsilon,
+    ) -> bool {
+        match self {
+            Geometry::Point(g) => {
+                if let Geometry::Point(other) = other {
+                    g.relative_eq(other, epsilon, max_relative)
+                } else {
+                    false
+                }
+            }
```
frewsxcv, line 278: Another way to format this, if you prefer it:

```suggestion
        match (self, other) {
            (Geometry::Point(g1), Geometry::Point(g2)) => g1.relative_eq(g2, epsilon, max_relative),
            ...
            _ => false,
        }
```

frewsxcv (approved):

geo-types/src/geometry.rs
michaelkirk: This is better! I'll apply this to the rest of the cases.

michaelkirk: done!

michaelkirk:
bors r=frewsxcv

PR #629, opened 2021-02-20
centroid fixups

michaelkirk:
---

- Fixes the implementation for a MultiLineString with zero length lines
- `impl Centroid for Geometry` and it's remaining variants.

FYI: I found these via the JTS test suite runner I'm working on here: https://github.com/michaelkirk/jts-test-runner/pull/1/files 

Status:
- [x] integrate jts-test-runner centroid tests
- [x] needs review http://github.com/michaelkirk/jts-test-runner/pull/1
- [x] move http://github.com/michaelkirk/jts-test-runner/pull/1 to georust
- [ ] publish jts-test-runner? (alternatively, we could leave it unpublished and just reference GH for this unique crate)
- [x] publish wkt
- [x] remove unpublished references from `patch.crates-io`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,2 +1,8 @@
 [workspace]
 members = ["geo", "geo-types", "geo-postgis"]
+
+[patch.crates-io]
+
+# Ensure any transitive dependencies also use the local geo/geo-types
+geo = { path = "geo" }
+geo-types = { path = "geo-types" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,2 +1,6 @@
 [workspace]
 members = ["geo", "geo-types", "geo-postgis"]
+
+[patch.crates-io]
+geo = { path = "geo" }
+geo-types = { path = "geo-types" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,2 @@
 [workspace]
 members = ["geo", "geo-types", "geo-postgis"]
-
-[patch.crates-io]
-geo = { path = "geo" }
-geo-types = { path = "geo-types" }
--- a/geo/Cargo.toml
+++ b/geo/Cargo.toml
@@ -36,7 +36,7 @@ approx = "0.4.0"
 criterion = { version = "0.3" }
 # jts-test-runner is an internal crate which exists only to be part of the geo test suite.
 # As such it's kept unpublished. It's in a separate repo primarily because it's kind of large.
-jts-test-runner = { git = "https://github.com/georust/jts-test-runner", commit = "bf873f7858792498b70451c2818c40d4e0975264" }
+jts-test-runner = { git = "https://github.com/georust/jts-test-runner", rev = "3294b9af1d3e64fcc9caf9646a93d0116f7d8321" }
 rand = "0.8.0"
 
 [[bench]]
--- a/geo/Cargo.toml
+++ b/geo/Cargo.toml
@@ -22,7 +22,7 @@ geographiclib-rs = { version = "0.2" }
 
 proj = { version = "0.20.3", optional = true }
 
-geo-types = { version = "0.7.0", path = "../geo-types", features = ["approx", "use-rstar"] }
+geo-types = { version = "0.7.0", features = ["approx", "use-rstar"] }
 
 robust = { version = "0.2.2" }
 
@@ -34,6 +34,9 @@ use-serde = ["serde", "geo-types/serde"]
 [dev-dependencies]
 approx = "0.4.0"
 criterion = { version = "0.3" }
+# jts-test-runner is an internal crate which exists only to be part of the geo test suite.
+# As such it's kept unpublished. It's in a separate repo primarily because it's kind of large.
+jts-test-runner = { git = "https://github.com/georust/jts-test-runner", commit = "bf873f7858792498b70451c2818c40d4e0975264" }
 rand = "0.8.0"
 
 [[bench]]
--- a/geo/src/algorithm/centroid.rs
+++ b/geo/src/algorithm/centroid.rs
@@ -355,7 +355,7 @@ where
 
 impl<T> Centroid for Geometry<T>
 where
-    T: CoordFloat + FromPrimitive + Sum,
+    T: CoordFloat + FromPrimitive + Sum + Default,
 {
     type Output = Option<Point<T>>;
 
@@ -366,12 +366,80 @@ where
 
 impl<T> Centroid for GeometryCollection<T>
 where
-    T: CoordFloat + FromPrimitive + Sum,
+    T: CoordFloat + FromPrimitive + Sum + Default,
 {
     type Output = Option<Point<T>>;
     fn centroid(&self) -> Self::Output {
-        let centroids = self.iter().flat_map(Centroid::centroid).collect();
-        MultiPoint(centroids).centroid()
+        use crate::algorithm::dimensions::{Dimensions, HasDimensions};
+
+        let mut centroid_0d_accum: Option<(Point<T>, T)> = None;
+        let mut centroid_1d_accum: Option<(Point<T>, T)> = None;
+        let mut centroid_2d_accum: Option<(Point<T>, T)> = None;
+
+        for geometry in self.iter() {
+            let centroid = geometry.centroid();
+            match (geometry.dimensions(), centroid) {
+                (Dimensions::Empty, _) | (_, None) => continue,
+                (Dimensions::ZeroDimensional, Some(centroid)) => {
+                    let mut centroid_accum = centroid_0d_accum.unwrap_or_default();
+                    let weight = T::one();
+                    centroid_accum.0 = centroid_accum.0 + centroid * weight;
+                    centroid_accum.1 = centroid_accum.1 + weight;
+                    centroid_0d_accum = Some(centroid_accum);
+                }
+                (Dimensions::OneDimensional, Some(centroid)) => {
+                    let mut centroid_accum = centroid_1d_accum.unwrap_or_default();
+
+                    fn length<T>(geometry: &Geometry<T>) -> T
+                    where
+                        T: CoordFloat + FromPrimitive + Sum + Default,
+                    {
+                        match geometry {
+                            Geometry::Line(l) => l.euclidean_length(),
+                            Geometry::LineString(l) => l.euclidean_length(),
+                            Geometry::MultiLineString(l) => l.euclidean_length(),
+                            Geometry::GeometryCollection(geometry_collection) => {
+                                geometry_collection.iter().map(length).sum()
+                            }
+                            Geometry::Point(_)
+                            | Geometry::MultiPoint(_)
+                            | Geometry::Polygon(_)
+                            | Geometry::MultiPolygon(_)
+                            | Geometry::Rect(_)
+                            | Geometry::Triangle(_) => T::zero(),
+                        }
+                    }
+                    let weight = length(geometry);
+
+                    if weight == T::zero() {
+                        continue;
+                    }
+
+                    centroid_accum.0 = centroid_accum.0 + centroid * weight;
+                    centroid_accum.1 = centroid_accum.1 + weight;
+                    centroid_1d_accum = Some(centroid_accum);
+                }
+                (Dimensions::TwoDimensional, Some(centroid)) => {
+                    let mut centroid_accum = centroid_2d_accum.unwrap_or_default();
+
+                    // REVIEW: unsigned?
+                    let weight = geometry.unsigned_area();
+
+                    if weight == T::zero() {
+                        continue;
+                    }
+
+                    centroid_accum.0 = centroid_accum.0 + centroid * weight;
+                    centroid_accum.1 = centroid_accum.1 + weight;
+                    centroid_2d_accum = Some(centroid_accum);
+                }
+            };
+        }
+
+        centroid_2d_accum
+            .or(centroid_1d_accum)
+            .or(centroid_0d_accum)
+            .map(|(centroid, weight)| centroid / weight)
     }
 }
 
--- a/geo/src/algorithm/centroid.rs
+++ b/geo/src/algorithm/centroid.rs
@@ -4,7 +4,8 @@ use std::iter::Sum;
 use crate::algorithm::area::{get_linestring_area, Area};
 use crate::algorithm::euclidean_length::EuclideanLength;
 use crate::{
-    CoordFloat, Line, LineString, MultiLineString, MultiPoint, MultiPolygon, Point, Polygon, Rect,
+    CoordFloat, Geometry, GeometryCollection, Line, LineString, MultiLineString, MultiPoint,
+    MultiPolygon, Point, Polygon, Rect, Triangle,
 };
 
 /// Calculation of the centroid.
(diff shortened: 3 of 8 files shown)


Cargo.toml
```diff
@@ -1,2 +1,6 @@
 [workspace]
 members = ["geo", "geo-types", "geo-postgis"]
+
+[patch.crates-io]
+geo = { path = "geo" }
+geo-types = { path = "geo-types" }
```
frewsxcv, line 6: What overrides does this effectively do? IIRC, `geo/Cargo.toml` already specifies `path = "../geo-types"`

geo/src/algorithm/centroid.rs
```diff
@@ -173,14 +174,12 @@ where
...
-                        return Some(Point(linestring[0]));
-                    }
-                }
-                None // this should never happen, since all linestrings being empty was previously checked
+                // All line strings were 0 length - dimensionally equivalent to a multi point.
+                let points = self
+                    .iter()
+                    .flat_map(|line_string| Some(Point::from(*line_string.0.first()?)))
+                    .collect();
+                MultiPoint(points).centroid()
```
frewsxcv, line 182: Something to explore in the future is extracting the body of the `impl Centroid for MultiPoint` implementation into a separate function that takes an iterator of `Point`s. Then we can call it here to avoid the `Vec` allocation

geo/src/algorithm/centroid.rs
```diff
@@ -354,14 +353,113 @@ where
     }
 }
 
+impl<T> Centroid for Geometry<T>
+where
+    T: CoordFloat + FromPrimitive + Sum + Default,
+{
+    type Output = Option<Point<T>>;
+
+    crate::geometry_delegate_impl! {
```
frewsxcv, line 362: Niiiiceeee ✨

frewsxcv (commented):
Ran out of time and still need to read through `impl<T> Centroid for GeometryCollection<T>`, but the rest looks good! Would be great if we had tests for the new implementations, but no worries if you don't have capacity

Cargo.toml
michaelkirk: Good point, I guess this ended up not being  relevant. 

The context is that I was working on integrating the JTS test suite which relied on the WKT crate. WKT depends on geo-types, but because the WKT crate's Cargo.toml doesn't specify the local path to geo-types (nor should it) we were using the local geo-types for geo, but the remote geo-types for WKT, so different versions.

I ended up not including WKT (yet!) so I'll remove this for now, but I expect to revisit this soon.

> Would be great if we had tests for the new implementations, but no worries if you don't have capacity

In case you missed it, there are tests (in fact this work was driven by it), but I agree it's admittedly awkward that the tests are not including in this PR. 

My plan is, once https://github.com/michaelkirk/jts-test-runner is reviewed, to add it as a dev dependency and run the tests as part of geo, I originally planned to do this as a followup.

But I'd also be happy with this course of action:

- https://github.com/michaelkirk/jts-test-runner/pull/1 is reviewed
- jts-test-runner is moved to georust, and 0.1 published
- this pr is updated to run the centroid tests from jts-test-runner
- this pr is merged

geo/src/algorithm/centroid.rs
michaelkirk: Good point, I've gone ahead and pushed a commit to this PR to achieve this.

geo/src/algorithm/centroid.rs
```diff
@@ -340,28 +339,136 @@ where
...
+{
+    type Output = Point<T>;
+    fn centroid(&self) -> Self::Output {
+        let coord = self.0 + self.1 + self.2 / T::from_usize(3).unwrap();
+        Point::from(coord)
+    }
+}
+
+fn centroid_of_coords<T>(
+    mut coords_iter: impl Iterator<Item = Coordinate<T>>,
```
michaelkirk, line 448: I considered implementing this as `impl Centroid for CoordsIter` but thought it might be a footgun, since, e.g. Polygon implements CoordsIter, but we'd never want to use the `impl CoordsIter for Centroid` for a `Polygon`.

geo/src/algorithm/centroid.rs
```diff
@@ -454,6 +444,25 @@ where
     }
 }
 
+fn centroid_of_coords<T>(
+    mut coords_iter: impl Iterator<Item = Coordinate<T>>,
```
frewsxcv, line 448: Looks great as-is, though if you wanted you could use `impl ExactSizeIterator` and use `len()` instead of calculating length yourself

geo/src/algorithm/centroid.rs
```diff
@@ -340,28 +339,136 @@ where
...
+}
+
+impl<T> Centroid for GeometryCollection<T>
+where
+    T: CoordFloat + FromPrimitive + Sum + Default,
+{
+    type Output = Option<Point<T>>;
+    fn centroid(&self) -> Self::Output {
+        use crate::algorithm::dimensions::{Dimensions, HasDimensions};
+
```
frewsxcv, line 364: Can we add a comment here explaining the contents of the tuple? Something like:

```
// (Option<T>, T) represents (centroid, weight)
```

or

```
// (centroid: Option<T>, weight: T)
```

geo/src/algorithm/centroid.rs
```diff
@@ -340,28 +339,136 @@ where
...
+                    }
+
+                    centroid_accum.0 = centroid_accum.0 + centroid * weight;
+                    centroid_accum.1 = centroid_accum.1 + weight;
+                    centroid_1d_accum = Some(centroid_accum);
+                }
+                (Dimensions::TwoDimensional, Some(centroid)) => {
+                    let mut centroid_accum = centroid_2d_accum.unwrap_or_default();
+
+                    // REVIEW: unsigned?
```
frewsxcv, line 415: I think so? It hurts my brain thinking about it

geo/src/algorithm/centroid.rs
```diff
@@ -340,28 +339,136 @@ where
...
+                            Geometry::MultiLineString(l) => l.euclidean_length(),
+                            Geometry::GeometryCollection(geometry_collection) => {
+                                geometry_collection.iter().map(length).sum()
+                            }
+                            Geometry::Point(_)
+                            | Geometry::MultiPoint(_)
+                            | Geometry::Polygon(_)
+                            | Geometry::MultiPolygon(_)
+                            | Geometry::Rect(_)
+                            | Geometry::Triangle(_) => T::zero(),
```
frewsxcv, line 399: Do we ever expect this branch to get hit? If not, do you have opinions about using `unreachable!()` here?

frewsxcv (approved):
Ty!

geo/src/algorithm/centroid.rs
michaelkirk: alas:
```
error[E0277]: the trait bound `FlatMap<impl Iterator, Option<geo_types::Coordinate<T>>, [closure@geo/src/algorithm/centroid.rs:180:35: 180:78]>: ExactSizeIterator` is not satisfied
   --> geo/src/algorithm/centroid.rs:179:21
    |
179 | /                     self.iter()
180 | |                         .flat_map(|line_string| Some(*line_string.0.first()?)),
    | |______________________________________________________________________________^ the trait `ExactSizeIterator` is not implemented for `FlatMap<impl Iterator, Option<geo_types::Coordinate<T>>, [closure@geo/src/algorithm/centroid.rs:180:35: 180:78]>`
...
447 |   fn centroid_of_coords<T>(
    |      ------------------ required by a bound in this
448 |       mut coords_iter: impl ExactSizeIterator<Item = Coordinate<T>>,
    |                             --------------------------------------- required by this bound in `centroid_of_coords`

```

geo/src/algorithm/centroid.rs
michaelkirk: Good call - I added a doc for the tuple as well as an overview of the algorithm, because I think it's not obvious.

geo/src/algorithm/centroid.rs
michaelkirk: It's complicated 😅 

For `Geometry::Point(_)`, and `Geometry::MultiPoint(_)`, I'm confident that it should never happen and could add an unreachable.

HasDimensions for Rect and Triangle supports degenerating to 1-D.

```
impl<C: CoordNum> HasDimensions for Triangle<C> {
    ...
    fn dimensions(&self) -> Dimensions {
        if self.0 == self.1 && self.1 == self.2 {
            // degenerate triangle is a point
            Dimensions::ZeroDimensional
        } else if self.0 == self.1 || self.1 == self.2 || self.2 == self.0 {
            // degenerate triangle is a line
            Dimensions::OneDimensional
        } else {
            Dimensions::TwoDimensional
        }
    }
    ...
}
```
So I guess it's expected for some inputs.

For Polygon/MultiPolygon... it's certainly impossible for valid Polygons (closed linestring with 3+ points), but am not confident that we'd want to crash a production instance.

How about a debug_assert for those?

Cargo.toml
frewsxcv: This all sgtm! Thanks for the explanation

geo/src/algorithm/centroid.rs
frewsxcv: Oh yeah we've got a flat map in there. Nvm

geo/src/algorithm/centroid.rs
frewsxcv: > How about a debug_assert for those?

Up to you! My preference is to have assertions (or a comment) since it'll indicate to the reader that this is not a branch we expect to get hit, and could surface bugs

geo/src/algorithm/centroid.rs
michaelkirk: I verified that some test cases fail if I switch this to `signed_area`.

michaelkirk:
Ok! I've completed my TODOS:

Status:
- [x] integrate jts-test-runner centroid tests
- [x] needs review http://github.com/michaelkirk/jts-test-runner/pull/1
- [x] move http://github.com/michaelkirk/jts-test-runner/pull/1 to georust
- [ ] publish jts-test-runner? (alternatively, we could leave it unpublished and just reference GH for this unique crate)
- [x] publish wkt
- [x] remove unpublished references from `patch.crates-io`

I rebased, but here are the changes since your last review @frewsxcv:
https://github.com/georust/geo/pull/629/files/4b5ab0862bb06e19501534035414513c9053c1df..c3b3ff2a19a9db14709a708e4a3aeb8436f58675

geo/Cargo.toml
```diff
@@ -34,6 +34,9 @@ use-serde = ["serde", "geo-types/serde"]
 [dev-dependencies]
 approx = "0.4.0"
 criterion = { version = "0.3" }
+# jts-test-runner is an internal crate which exists only to be part of the geo test suite.
+# As such it's kept unpublished. It's in a separate repo primarily because it's kind of large.
+jts-test-runner = { git = "https://github.com/georust/jts-test-runner", commit = "bf873f7858792498b70451c2818c40d4e0975264" }
```
michaelkirk, line 39: This might be controversial... but it seemed like a lot of ceremony to go through the publish dance for something which is basically an internal dependency.

I could be convinced otherwise!

michaelkirk: If anything, I'd reconsider just including the crate in the main geo repo - It's 15MB of xml.

Cargo.toml
michaelkirk: Now that I've integrated the jts-test-runner into this PR the patch lines are back.

geo/Cargo.toml
urschrei: One thing to note is that the current cargo package size limit is 10mb…

michaelkirk: Ah, good to know! My currently solution is indeed to just agglomerate the test fixtures into the jts-test-runner crate. 

I'm including a pretty small subset at this point, but good to know that we may eventually need to do something else.

Still, I'm now more inclined to leave it as a separate crate.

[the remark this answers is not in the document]
michaelkirk: Whoops, thanks!

geo/src/algorithm/centroid.rs
```diff
@@ -340,28 +339,153 @@ where
...
+        let mut centroid_1d_accum: Option<(Point<T>, T)> = None;
+        let mut centroid_2d_accum: Option<(Point<T>, T)> = None;
+
+        for geometry in self.iter() {
+            let centroid = geometry.centroid();
+            match (geometry.dimensions(), centroid) {
+                (Dimensions::Empty, _) | (_, None) => continue,
+                (Dimensions::ZeroDimensional, Some(centroid)) => {
+                    let mut centroid_accum = centroid_0d_accum.unwrap_or_default();
+                    let weight = T::one();
```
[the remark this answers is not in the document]
michaelkirk: ooph good catch. I'm going to rework this, but might not have for a few days. 

Thank you so much for catching this mistake!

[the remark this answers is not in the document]
michaelkirk: Ok, I believe I've addressed this, but it ended being non-trivial. In the process I found a few more bugs and fleshed out some more tests (almost always the issues were with degenerate geometries).

> I think we should extend our Centroid trait to add a method that returns both the centroid Point, and the total weight of the geometry.

I started with this approach, and it would've worked fine, but I felt things ended up cleaner when I bundled the state and logic into a separate private `CentroidOperation` accumulation struct.

Even in cases where there is a single geometry, we'll still utilize the aggregating operation, which at first seems weird.  This is primarily to avoid redundantly having to recompute the centroids weight in the case the geometry *does* appear in a collection. Since computing the centroid essentially involves computing its weight anyway, there's not much lost by tracking it in this way, except some small extra stack allocations in some cases.

Since I've essentially re-written the trait at this point, it'd be great if I could get another review.

geo/src/algorithm/centroid.rs
```diff
@@ -335,42 +163,324 @@ where
...
+                self.add_centroid(TwoDimensional, rect.centroid().0, rect.unsigned_area())
+            }
+            Empty => unreachable!("Rect dimensions cannot be empty"),
+        }
+    }
+
+    fn add_triangle(&mut self, triangle: &Triangle<T>) {
+        match triangle.dimensions() {
+            ZeroDimensional => self.add_coord(triangle.0),
+            OneDimensional => {
```
rmanoka, line 338: This seems different from how we are handling a degenerate ring.  I think we should simply add all three line segments in this case rather than just the longest one.  The centroid point itself doesn't change, but the weight would be double what it is now.  

These sort of degenerate cases are not really valid as per SFS, so we just have to be internally consistent.  Below is a natural test case that users would expect to pass (IMO).

``` rust
#[test]
fn degenerate_triangle_like_ring() {
    let triangle = Triangle(c(0., 0.), c(1., 1.), c(2., 2.));
    let poly: Polygon<_> = triangle.clone().into();

    let line = Line::new(c(0., 1.), c(1., 3.));

    let g1 = GeometryCollection(vec![triangle.into(), line.into()]);
    let g2 = GeometryCollection(vec![poly.into(), line.into()]);
    assert_eq!(g1.centroid(), g2.centroid());
}
```

rmanoka (approved):
Cool work @michaelkirk !   There's only one minor change I've requested in one of the degeneracy handling logic (1-D triangle vs closed line string).

geo/src/algorithm/centroid.rs
michaelkirk: Yeah, that's a great point. 

Plus your suggestion is simpler. =)

I've rebased and pushed up one new commit to address this for Triangles and Rects (which had the same issue).

michaelkirk:
Thanks for the review @rmanoka!

Since it was a big change, I'm going to leave it open for a couple days before merging in case @frewsxcv or others want to take another look.

PR #635, opened 2021-03-04
Adds new algorithm for concave hull calculation - k nearest neighbours approach

Maximkaaa:
The short description of the algorithm is in the docs for the trait.

---

### DIFF
--- a/geo/src/algorithm/k_nearest_concave_hull.rs
+++ b/geo/src/algorithm/k_nearest_concave_hull.rs
@@ -74,7 +74,7 @@ fn concave_hull<T>(points: Vec<Point<T>>, k: u32) -> Polygon<T>
 
 const DELTA: f32 = 0.000000001;
 
-// Removes duplicate points from the dataset.
+/// Removes duplicate points from the dataset.
 fn prepare_dataset<T>(points: &Vec<Point<T>>) -> rstar::RTree<Point<T>>
     where T: GeoFloat + RTreeNum
 {
@@ -93,16 +93,18 @@ fn prepare_dataset<T>(points: &Vec<Point<T>>) -> rstar::RTree<Point<T>>
     dataset
 }
 
+/// The points are considered equal, if both coordinate values are same with 0.0000001% range
+/// (see the value of DELTA constant).
 fn points_are_equal<T>(p1: &Point<T>, p2: &Point<T>) -> bool
     where T: GeoFloat + RTreeNum
 {
-    float_equal(p1.x(), p2.x()) && float_equal(p1.x(), p2.y())
+    float_equal(p1.x(), p2.x()) && float_equal(p1.y(), p2.y())
 }
 
 fn float_equal<T>(a: T, b: T) -> bool
     where T: GeoFloat
 {
-    let da = a * T::from(DELTA).expect("Conversion from constant is always valid.");
+    let da = a * T::from(DELTA).expect("Conversion from constant is always valid.").abs();
     b > (a - da) && b < (a + da)
 }
 
@@ -140,7 +142,7 @@ fn concave_hull_inner<T>(original_dataset: rstar::RTree<Point<T>>, k: u32) -> Po
     let mut current_point = first_point;
     dataset.remove(&first_point);
     
-    let mut prev_angle = T::zero();
+    let mut prev_point = current_point;
     let mut curr_step = 2;
     while (current_point != first_point || curr_step == 2) && dataset.size() > 0 {
         if curr_step == 5 {
@@ -148,14 +150,14 @@ fn concave_hull_inner<T>(original_dataset: rstar::RTree<Point<T>>, k: u32) -> Po
         }
 
         let mut nearest_points = get_nearest_points(&dataset, &current_point, k_adjusted);
-        sort_by_angle(&mut nearest_points, &current_point, prev_angle);
+        sort_by_angle(&mut nearest_points, &current_point, &prev_point);
 
         let selected = nearest_points.iter().find(|x| !intersects(&hull, &[&current_point, x]));
 
         if let Some(sel) = selected {
+            prev_point = current_point;
             current_point = **sel;
             hull.push(current_point);
-            prev_angle = get_angle(&[&hull[hull.len() - 1], &hull[hull.len() - 2]]);
             dataset.remove(&current_point);
 
             curr_step += 1;
@@ -210,16 +212,38 @@ where T: GeoFloat + RTreeNum
     dataset.nearest_neighbor_iter(base_point).take(candidate_no as usize).collect()
 }
 
-fn sort_by_angle<'a, T>(points: &'a mut Vec<&Point<T>>, curr_point: &Point<T>, prev_angle: T)
+fn sort_by_angle<'a, T>(points: &'a mut Vec<&Point<T>>, curr_point: &Point<T>, prev_point: &Point<T>)
 where T: GeoFloat
 {
+    let base_angle = pseudo_angle(prev_point.x() - curr_point.x(), prev_point.y() - curr_point.y());
     points.sort_by(|a, b| {
-        let mut angle_a = get_angle(&[curr_point, a]) - prev_angle;
-        let mut angle_b = get_angle(&[curr_point, b]) - prev_angle;
-        if angle_a < T::zero() { angle_a = angle_a + two_pi(); }
-        if angle_b < T::zero() { angle_b = angle_b + two_pi(); }
+        let mut angle_a = pseudo_angle(a.x() - curr_point.x(), a.y() - curr_point.y()) - base_angle;
+        if angle_a < T::zero() {
+            angle_a = angle_a + T::from(4.0).unwrap();
+        }
+
+        let mut angle_b = pseudo_angle(b.x() - curr_point.x(), b.y() - curr_point.y()) - base_angle;
+        if angle_b < T::zero() {
+            angle_b = angle_b + T::from(4.0).unwrap();
+        }
+
         angle_a.partial_cmp(&angle_b).unwrap().reverse()
     });
+}
+
+fn pseudo_angle<T>(dx: T, dy: T) -> T
+    where T: GeoFloat
+{
+    if dx == T::zero() && dy == T::zero() {
+        return T::zero();
+    }
+
+    let p = dx / (dx.abs() + dy.abs());
+    if dy < T::zero() {
+        T::from(3.).unwrap() + p
+    } else {
+        T::from(1.).unwrap() - p
+    }
 }
 
 fn intersects<T>(hull: &Vec<Point<T>>, line: &[&Point<T>; 2]) -> bool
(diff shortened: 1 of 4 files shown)


geo/src/algorithm/k_nearest_concave_hull.rs
```diff
@@ -0,0 +1,348 @@
...
+    &point_set[min_i]
+}
+
+fn get_nearest_points<'a, T>(dataset: &'a rstar::RTree<Point<T>>, base_point: &Point<T>, candidate_no: u32) -> Vec<&'a Point<T>>
+where T: GeoFloat + RTreeNum
+{
+    dataset.nearest_neighbor_iter(base_point).take(candidate_no as usize).collect()
+}
+
+fn sort_by_angle<'a, T>(points: &'a mut Vec<&Point<T>>, curr_point: &Point<T>, prev_angle: T)
```
lnicola, line 192: Would it be possible to use a cross product here (signed area) instead of computing an angle?

michaelkirk:
I was hoping to hold off on reviewing this until this was addressed:

> The first one though looks like an obvious bug. There's actially a check in the code to see if the resulting hull contains all the input points. It probably has to do something with getting points out of the multipolygons. If you send me the geometries, I'll check it out. Any format would be fine. Json is good.

@dabreegster - have you sent the json?

dabreegster:
Sorry for the delay! These're serialized `Multipolygon<f64>`s.

The buggy case:
buggy
[buggy.json.txt](https://github.com/georust/geo/files/6104086/buggy.json.txt)

Some desired "shrinkwrapping" missing:
shrink
[large.json.txt](https://github.com/georust/geo/files/6104091/large.json.txt)

`k=10` for both, but I also tried 3.

Maximkaaa:
Thank you, @dabreegster, your application helped a lot in debugging this algorithm. There were a couple of bugs that I've found and fixed. After fixing, the result in case building concave polygons around the buildings looks like this:
image
image
image

There are still some issues that are not due to implementation but the limitation of the algorithm itself.

1. In some cases polygon looks like it could go more tightly around the objects:

image

The reason for this is that small house in the middle of the park. The algorithm is designed to always build a polygon with single contour. In case of clustered input objects, a rather larger K must be chosen to surround all input objects. One option is to let the method to return multipolygons, but that would break the logic for automatically adjust K, so it would have to be adjusted by hand in many cases.

2. If i switch to surround lanes instead of building, another issue appears:
image
image

The reason for this is that the algorithm in its core only works with points. So to surround other shapes we just take all the vertexes and use them as an input. In the case when there are many little objects (like when we use buildings as input) it's not a problem as all vertexes are close to each other.

In the case when we have big long lines without any points, the algorithm may find some closer points to snap to, so not the whole polygon or line ends up to be inside of the hull. Although all the vertexes are actually inside of the hull.

One way to deal with this issue is to insert virtual vertexes on the long sides of polygons (or lines). But I failed to come up with a good generic way to choose how close I should insert those points.

Another options is to remove this treat from the Lines and Polygons altogether, to make it obvious that the algorithm with deal only with points.

What are your thoughts on this guys?

dabreegster:
> the result in case building concave polygons around the buildings looks like this:

These look much more "shrink-wrapped" -- thanks for iterating on this!

>  In case of clustered input objects, a rather larger K must be chosen to surround all input objects

This limitation makes sense; I don't think manually adjusting K would be feasible for my particular case, so I'm not opinionated about trying it.

> Another options is to remove this treat from the Lines and Polygons altogether, to make it obvious that the algorithm with deal only with points.

This limitation also makes sense, and yes, I would vote for only taking `Point`s in the method signature. I've been a bit confused by other geometric algorithms like generating Voronoi diagrams in the past, because from API documentation, it's not always clear that they don't work on entire polygons, and that you can't just feed in all of the polygon's boundary points.

michaelkirk:
> Another options is to remove this treat from the Lines and Polygons altogether, to make it obvious that the algorithm with deal only with points.

I'd prefer the approach of a signature which only accepts points for now.

To be correct, a hull needs to `Cover` its input. It might be surprising to people that the output could be incorrect in this way if the input is Polygons, especially since other implementations (e.g. PostGIS) do not behave this way. Note that our current concave_hull algorithm also has this limitation, and yet still accepts 1D and 2D geometries, but in hindsight I think this was a mistake.

For followup work, including a recipe in the documentation (and potentially helper methods) for interpolating points along a `line_string` and feeding that into `concave_hull` might be a good middle ground between making things not too tedious while also making clearer the limitations on this algorithm.

Maybe something like:
```
# note: this method doesn't currently exist
let subsampled_linestring = polygon.exterior().interpolated_with_max_distance(0.1);
let points = subsampled_linestring.0;
concave_hull(points)
```

(maybe relevant [`line_interpolate_point`](https://docs.rs/geo/0.18.0/geo/algorithm/line_interpolate_point/trait.LineInterpolatePoint.html))

And also, we should aspire to one day get a fancy concave hull algorithm that also works with 1D and 2D geometries. I'll open a separate issue for this...

geo/src/algorithm/k_nearest_concave_hull.rs
```diff
@@ -0,0 +1,393 @@
...
+    use crate::coords_iter::CoordsIter;
+    use num_traits::FloatConst;
+    use std::f32::consts::PI;
+
+    #[test_case(0., 0., 1., 0., 0.)]
+    #[test_case(0., 0., 0., 1., PI / 2.)]
+    #[test_case(0., 0., -1., 0., PI)]
+    #[test_case(0., 0., 0., -1., PI / 2. * 3.)]
+    #[test_case(0., 0., -1., 1., PI / 4. * 3.)]
+    #[test_case(0., 0., -1., -1., PI / 4. * 5.)]
```
michaelkirk, line 311: This is admittedly kind of nice, but I'd prefer to avoid adding an extra dependency for it.

(Also, curious if, as @lnicola mentioned, we could remove get_angle altogether)

geo/src/algorithm/k_nearest_concave_hull.rs
```diff
@@ -0,0 +1,393 @@
...
+const K_MULTIPLIER: f32 = 1.5;
+
+/// Another approach for [concave hull](trait.algorithm.ConcaveHull.html). This algorithm is based
+/// on a [k nearest neighbours approach](https://pdfs.semanticscholar.org/2397/17005c3ebd5d6a42fc833daf97a0edee1ce4.pdf)
+/// by Adriano Moreira and Maribel Santos.
+/// 
+/// The idea of the algorithm is simple:
+/// 1. Find a point on a future hull (e. g. a point with the smallest Y coordinate).
+/// 2. Find K nearest neighbours to the chosen point.
+/// 3. As the next point on the hull choose one of the nearest points, that would make the largest
```
michaelkirk, line 18:
```suggestion
/// 3. As the next point on the hull chose one of the nearest points, that would make the largest
```

geo/src/algorithm/k_nearest_concave_hull.rs
```diff
@@ -0,0 +1,393 @@
...
+///    left hand turn from the previous segment.
+/// 4. Repeat 2-4.
+/// 
+/// In cases when the hull cannot be calculated for the given K, a larger value is chosen and
+/// calculation starts from the beginning.
+/// 
+/// In the worst case scenario, when no K can be found to build a correct hull, convex hull is
```
michaelkirk, line 25:
```suggestion
/// In the worst case scenario, when no K can be found to build a correct hull, the convex hull is
```

geo/src/algorithm/k_nearest_concave_hull.rs
```diff
@@ -0,0 +1,393 @@
...
+/// returned.
+/// 
+/// This algorithm is generally several times slower then the one used in
```
michaelkirk, line 28:
```suggestion
/// This algorithm is generally several times slower than the one used in the
```

geo/src/algorithm/k_nearest_concave_hull.rs
```diff
@@ -0,0 +1,393 @@
...
+/// In the worst case scenario, when no K can be found to build a correct hull, convex hull is
+/// returned.
+/// 
+/// This algorithm is generally several times slower then the one used in 
+/// [ConcaveHull](trait.algorithm.ConcaveHull.html) trait, but gives better results and
+/// does not require manual coefficient adjustment.
+/// 
+/// The larger K is given to the algorithm, the more "smooth" the hull will generally be, but the
+/// longer calculation may take. If performance is not critical, K=3 is a safe value to set
+/// (lower values do not make sense for this algorithm). If K is equal or larger then the number of
```
michaelkirk, line 34:
```suggestion
/// (lower values do not make sense for this algorithm). If K is equal or larger than the number of
```

geo/src/algorithm/k_nearest_concave_hull.rs
```diff
@@ -0,0 +1,393 @@
...
+/// input points, convex hull will be produced.
```
michaelkirk, line 35:
```suggestion
/// input points, the convex hull will be produced.
```

geo/src/algorithm/k_nearest_concave_hull.rs
```diff
@@ -0,0 +1,393 @@
...
+    where T: GeoFloat + RTreeNum
+{
+    (p1.x() - p2.x()).abs() + (p2.y() - p2.y()).abs()
+}
+
+fn polygon_from_tree<T>(dataset: &rstar::RTree<Point<T>>) -> Polygon<T>
+    where T: GeoFloat + RTreeNum
+{
+    let mut points: Vec<Coordinate<T>> = dataset.iter().map(|p| p.0).collect();
+    points.push(points[0]);
```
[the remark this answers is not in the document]
michaelkirk: Upon further inspection... it seems like this method is only used when the original input has 3 or less points, so orderings here are all equivalent, but that's not obvious by reading this method.

So two alternatives I'd be OK with:
- do an `assert!(points.len() <= 3)` here to document this

or 

- delete this method and bail out earlier in the algorithm when we still have access to the initial point ordering rather than an RTree of those points.

michaelkirk (changes requested):

geo/src/algorithm/k_nearest_concave_hull.rs
```diff
@@ -0,0 +1,393 @@
...
+}
+
+fn concave_hull<T>(points: Vec<Point<T>>, k: u32) -> Polygon<T>
+    where T: GeoFloat + RTreeNum
+{
+    let dataset = prepare_dataset(&points);
+    concave_hull_inner(dataset, k)
+}
+
+const DELTA: f32 = 0.000000001;
```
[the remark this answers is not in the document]
Maximkaaa: I've changed the logic in points_equal() method and added a comment about this constant.

geo/src/algorithm/k_nearest_concave_hull.rs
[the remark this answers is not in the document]
Maximkaaa: To use the second option we would have to change the algorithm signature to return Option<Polygon<T>>, so I prefer the first option.

geo/src/algorithm/k_nearest_concave_hull.rs
Maximkaaa: I've tried using dot-product method, but finally I've found even faster algorithm that works just fine.

Maximkaaa:
Hi, @michaelkirk . Couldn't return to this earlier because of lack of time. I've fixed all the issues, mentioned in the review and updated it to the current master. Please, check if everything fine.

michaelkirk:
Hi @Maximkaaa. I made some time to review and QA the latest version, and things look good to me. Sorry for the slow turn around time!

I'll plan to merge this tomorrow unless I hear from someone else that they'd like to review it.

Maximkaaa:
@michaelkirk NP, totally understand you. It took me a month to make the fixes.

michaelkirk:
Squashed, formatted, and merged in https://github.com/georust/geo/commit/2c4261a4d413e18f3dfbcbc83ad551ae53e775cb

PR #636, opened 2021-03-08
Add `line_intersection` to compute point or segment intersection of two Lines.

### LINKED ISSUE #288: Line - Line intersection
An attempt was made in https://github.com/georust/rust-geo/pull/122

michaelkirk:
---

Fixes #288 

I pulled this out of my [Geometry Graph](https://github.com/georust/geo/tree/mkirk/geomgraph) work (which is getting close, I swear!).

### DIFF
--- a/geo/src/algorithm/line_intersection.rs
+++ b/geo/src/algorithm/line_intersection.rs
@@ -1,0 +1,604 @@
+use crate::{Coordinate, GeoFloat, Line};
+
+use crate::algorithm::bounding_rect::BoundingRect;
+use crate::algorithm::intersects::Intersects;
+
+#[derive(PartialEq, Eq, Clone, Copy, Debug)]
+pub enum LineIntersection<F: GeoFloat> {
+    /// Lines intersect in a single point
+    SinglePoint {
+        intersection: Coordinate<F>,
+        /// For Lines which intersect in a single point, that point may be either an endpoint
+        /// or in the interior of each Line.
+        /// If the point lies in the interior of both Lines, we call it a _proper_ intersection.
+        is_proper: bool,
+    },
+
+    /// Overlapping Lines intersect in a line segment
+    Collinear { intersection: Line<F> },
+}
+
+impl<F: GeoFloat> LineIntersection<F> {
+    pub fn is_proper(&self) -> bool {
+        match self {
+            Self::Collinear { .. } => false,
+            Self::SinglePoint { is_proper, .. } => *is_proper,
+        }
+    }
+}
+
+/// Returns the intersection between two [`Lines`](Line).
+///
+/// Lines can intersect in a Point or, for Collinear lines, in a Line. See [`LineIntersection`]
+/// for more details about the result.
+///
+/// # Examples
+///
+/// ```
+/// use geo::{Line, Coordinate};
+/// use geo::algorithm::line_intersection::{line_intersection, LineIntersection};
+///
+/// let line_1 = Line::new(Coordinate {x: 0.0, y: 0.0}, Coordinate { x: 5.0, y: 5.0 } );
+/// let line_2 = Line::new(Coordinate {x: 0.0, y: 5.0}, Coordinate { x: 5.0, y: 0.0 } );
+/// let expected = LineIntersection::SinglePoint { intersection: Coordinate { x: 2.5, y: 2.5 }, is_proper: true };
+/// assert_eq!(line_intersection(line_1, line_2), Some(expected));
+///
+/// let line_1 = Line::new(Coordinate {x: 0.0, y: 0.0}, Coordinate { x: 5.0, y: 5.0 } );
+/// let line_2 = Line::new(Coordinate {x: 0.0, y: 1.0}, Coordinate { x: 5.0, y: 6.0 } );
+/// assert_eq!(line_intersection(line_1, line_2), None);
+///
+/// let line_1 = Line::new(Coordinate {x: 0.0, y: 0.0}, Coordinate { x: 5.0, y: 5.0 } );
+/// let line_2 = Line::new(Coordinate {x: 5.0, y: 5.0}, Coordinate { x: 5.0, y: 0.0 } );
+/// let expected = LineIntersection::SinglePoint { intersection: Coordinate { x: 5.0, y: 5.0 }, is_proper: false };
+/// assert_eq!(line_intersection(line_1, line_2), Some(expected));
+///
+/// let line_1 = Line::new(Coordinate {x: 0.0, y: 0.0}, Coordinate { x: 5.0, y: 5.0 } );
+/// let line_2 = Line::new(Coordinate {x: 3.0, y: 3.0}, Coordinate { x: 6.0, y: 6.0 } );
+/// let expected = LineIntersection::Collinear { intersection: Line::new(Coordinate { x: 3.0, y: 3.0 }, Coordinate { x: 5.0, y: 5.0 })};
+/// assert_eq!(line_intersection(line_1, line_2), Some(expected));
+/// ```
+/// Strongly inspired by, and meant to produce the same results as, [JTS's RobustLineIntersector](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/algorithm/RobustLineIntersector.java#L26).
+pub fn line_intersection<F>(p: Line<F>, q: Line<F>) -> Option<LineIntersection<F>>
+where
+    F: GeoFloat,
+{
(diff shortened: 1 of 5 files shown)


geo/src/algorithm/line_intersection.rs
```diff
@@ -0,0 +1,663 @@
...
+    if !(p.bounding_rect().intersects(&int_pt) && q.bounding_rect().intersects(&int_pt)) {
+        // compute a safer result
+        // copy the coordinate, since it may be rounded later
+        int_pt = nearest_endpoint(p, q);
+    }
+    return int_pt;
+}
+
+#[cfg(test)]
+mod test {
```
michaelkirk, line 360: As well as these tests, my main confidence for correctness comes transitively from using this code and passing the [JTS suite of "Relate" tests](https://github.com/georust/jts-test-runner/tree/mkirk/relate).

frewsxcv (commented):
short on time this evening so I can't review the whole thing! but could you add a new entry in the index for the `geo` crate root documentation (that was recently added)?

urschrei:
Also pushed for time so haven't thoroughly reviewed, but this lgtm.
Am I interpreting some of these test comments correctly in that the tests are correctly passing in geo, but have failed in JTS in the past? That's interesting (in a good way).

michaelkirk:
> short on time this evening so I can't review the whole thing! but could you add a new entry in the index for the geo crate root documentation (that was recently added)?

Thanks for calling this out. Done!

I slightly reorganized some items into a new "Topology" subheading. WDYT?

michaelkirk:
> Am I interpreting some of these test comments correctly in that the tests are correctly passing in geo, but have failed in JTS in the past? That's interesting (in a good way).

I have perhaps confused you by my commenting convention. Those tests are mostly just ported from the JTS suite. I tried to capture the comments *from* those tests to capture at least a little context. I tried to indicate which lines were quoted directly from JTS by prefixing them with `>`, as in:

```
/// Based on JTS test `testLeduc_1`
///
/// > Test involving two non-almost-parallel lines.
/// > Does not seem to cause problems with basic line intersection algorithm.
```

So JTS worked out all the kinks, and the geo code also passes these ported tests because the georust code ported from the modern JTS code.

michaelkirk:
> short on time this evening

> Also pushed for time so haven't thoroughly reviewed


giphy
paging @rmanoka 🥺

geo/src/algorithm/line_intersection.rs
```diff
@@ -0,0 +1,666 @@
...
+fn collinear_intersection<F: GeoFloat>(p: Line<F>, q: Line<F>) -> Option<LineIntersection<F>> {
+    let p_rect = p.bounding_rect();
+    let p1q1p2 = p_rect.intersects(&q.start);
+    let p1q2p2 = p_rect.intersects(&q.end);
+
+    let q_rect = q.bounding_rect();
+    let q1p1q2 = q_rect.intersects(&p.start);
+    let q1p2q2 = q_rect.intersects(&p.end);
+
+    if p1q1p2 && p1q2p2 {
```
frewsxcv, line 158: This check could go before the `q_rect` assignment above, if we wanted to save a few calls. Breaks the flow a little so it's a tradeoff  🤷🏻

geo/src/algorithm/line_intersection.rs
```diff
@@ -0,0 +1,666 @@
...
+            return Some(LineIntersection::SinglePoint {
+                intersection: q.end,
+                is_proper: false,
+            });
+        } else {
+            return Some(LineIntersection::Collinear {
+                intersection: Line::new(q.end, p.end),
+            });
+        }
+    }
```
frewsxcv, line 211: I played around with formatting these conditional checks a different way:

```rust
    Some(match (p1q1p2, p1q2p2, q1p1q2, q1p2q2) {
        (true, true, _, _) => LineIntersection::Collinear { intersection: q },
        (_, _, true, true) => LineIntersection::Collinear { intersection: q },
        (true, false, true, false) if q.start == p.start => {
            LineIntersection::SinglePoint {
                intersection: q.start,
                is_proper: false,
            }
        }
        (true, _, true, _) => {
            LineIntersection::Collinear {
                intersection: Line::new(q.start, p.start),
            }
        }
        ...
        _ => return None,
    })
```

As always, take it or leave it. Was just experimenting

geo/src/algorithm/line_intersection.rs
```diff
@@ -0,0 +1,666 @@
...
+    let dist = point_line_euclidean_distance(q.end, p);
+    if dist < min_dist {
+        nearest_pt = q.end;
+    }
+    nearest_pt
+}
+
+fn raw_line_intersection<F: GeoFloat>(p: Line<F>, q: Line<F>) -> Option<Coordinate<F>> {
+    let min_x0 = if p.start.x < p.end.x {
```
frewsxcv, line 249: Is this the same as:

```rust
    let min_x0 = p.start.x.min(p.end.x);
```

https://docs.rs/num/0.4.0/num/trait.Float.html#tymethod.min

frewsxcv (approved):
Looks good! Just commented on some nonblocking minor stuff

geo/src/lib.rs
```diff
@@ -81,6 +75,18 @@
...
+//! ## Topology
+//!
+//! - **[`Contains`](algorithm::contains::Contains)**: Calculate if a geometry contains another
+//!   geometry
+//! - **[`CoordinatePosition`](algorithm::coordinate_position::CoordinatePosition)**: Calculate
+//!   the position of a coordinate relative to a geometry
+//! - **[`HasDimensions`](algorithm::dimensions::HasDimensions)**: Determine the dimensions of a geometry
+//! - **[`Intersects`](algorithm::intersects::Intersects)**: Calculate if a geometry intersects
+//!   another geometry
+//! - **[`line_intersection`](algorithm::line_intersection::line_intersection)**: Calculates the
```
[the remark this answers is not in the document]
michaelkirk: The reason it's capitalized like this is because `line_intersection` is a free function, not a trait. I should have called this out explicitly, because I went back and fourth on it.

My first thought was on one end of the spectrum: "eventually we want everything to intersect with everything else, so I should just introduce a super generic intersection trait" like:
```
trait Intersection<T> {
    type Output;
    fn intersection(&self, other: T) -> Self::Output
}

impl Intersection<Line<F>> for Line<F> {
    type Output = Option<LineIntersection<F>>
    fn intersection(&self, other: Line<F>) -> Self::Output {
        ...
    }
}
```

But since there aren't any near-term plans to introduce impls other than `Intersection<Line> for Line`, I thought suggesting that we support a generic concept of `Intersection` might just end up confusing people.

So then I thought maybe a more specific trait:
```
trait LineIntersection {
    fn line_intersection(&self, other: Line<T>) -> LineIntersection<T>;
}

impl LineIntersection<T> for Line<T> {
    fn line_intersection(&self, other: Line<T>) -> LineIntersection<T>;
        ...
    }
}
```

I'd be OK with this, but I think I would only ever want to have one thing (Line<T>) implement that trait... so it seems like a lot of machinery, and if we *did* eventually introduce a more generic `Intersection` trait, this trait would likely become wholly redundant.

So that's why I eventually settled on just a bare non-trait function, and why it's capitalized that way in the docs. It seemed like a stable way to expose the functionality that would be less likely to surprise people.

If we did eventually add intersections for more than just lines, I'd be in favor of adding a generic `Intersecion<T>` trait.

WDYT?

geo/src/algorithm/line_intersection.rs
michaelkirk: Indeed it is! I've updated to use that, and it's much more concise.

geo/src/algorithm/line_intersection.rs
michaelkirk: It's easier to see the patterns when laid out as pattern matching -- good call. I've done that (and introduced a couple helper functions for brevity).

michaelkirk:
I think I've addressed all your feedback @frewsxcv. I'll wait another day before merging incase you or anyone else wants to re-review.

rmanoka:
@michaelkirk  Looks great!  I'm a bit pressed for time this week, but can take a look at the details of the core logic next week.

michaelkirk:
> @michaelkirk Looks great! I'm a bit pressed for time this week, but can take a look at the details of the core logic next week.

@rmanoka: No problem - would you prefer that I leave this open until then?

rmanoka:
@michaelkirk Pls feel free to merge it.

michaelkirk:
bors r=frewsxcv

(rebased and squashed the review commits)

PR #638, opened 2021-03-10
Don't redundantly sort Rect inputs (cleanup - no change in behavior)

michaelkirk:
- [ ] ~~I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.~~
---

Since ec69add2df63ec4e2a9f04c52bc824d8427cef54, `Rect::new` already sorts it's input by min/max, and so there's no longer a need to manually sort min/max points.


frewsxcv (approved):
good finds!

michaelkirk:
bors r=frewsxcv

Issue #596, opened 2021-01-07
Add `log` crate?


michaelkirk:
What would folks think of starting to use the [`log`](https://crates.io/crates/log) crate in geo?

In general, I think high quality permanent logging is a worthwhile addition to the software, especially as more outside people use the crate and encounter errors.

But for me this was spurred by the work I'm doing in #515, some of which is pretty complicated. Consequently I've introduced some non-trivial logging while I work out the bugs. I think some of this would be valuable to keep around if future problems arise or changes made. Typically I just delete my `print` statements once I'm confident everything is working, but since this is a larger more complicated component, my preference would be to leave some of it in  - mostly at `debug!` or lower log levels,  so people won't typically see it.

The `log` crate is already used transitively by some of our features (postgres, proj_network) but it's not currently required by the default geo build.

The biggest shortcoming I see is that I know of only awkward ways to utilize it while developing locally and while testing, e.g. https://stackoverflow.com/questions/30177845/how-to-initialize-the-logger-for-integration-tests

rmanoka:
I like this idea.  Have felt the need to debug by printing messages while implementing a few algos in this crate.

On a related note,  should we add Debug to the traits defining [`CoordinateType`](https://github.com/georust/geo/blob/master/geo-types/src/lib.rs#L56) ?  I found it very difficult to quickly add a few `eprintln` lines to print coordinates from within algorithms while debugging.  I can't imagine scalar type that wouldn't support `Debug`.

michaelkirk:
> should we add Debug to the traits defining CoordinateType 

I'm in favor of that. It's technically a geo-types breaking change though 😢.

urschrei:
> I'm in favor of that. It's technically a geo-types breaking change though 😢.

Bah. We should do both, though. Debugging is enough of a pain as it is.

michaelkirk:
> should we add Debug to the traits defining CoordinateType

Just in case I wasn't clear, I agree that we should do it.

I wonder if there's some slightly graceful way we could introduce this feature in a way that allows deprecation first.

Like introduce a new trait that implements Debug - maybe by way of https://github.com/georust/geo/issues/540?

frewsxcv:
Created a new issue for the `Debug` conversation. +1 from me. https://github.com/georust/geo/issues/597

@michaelkirk Back to this issue, do you have an example of where logging in this crate would be helpful? I'm fine with this, just want to get an idea of how it'll get implemented.

michaelkirk:
[...]

PR #552, opened 2020-12-02
line_segement_distance works for Coordinates too

michaelkirk:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/master/CODE_OF_CONDUCT.md).
- [x] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users. 
---

A tiny bit of functionality cleaved off while working on https://github.com/georust/geo/projects/1#card-46950138

PR #639, opened 2021-03-12
Introduce the geomgraph module for DE-9IM Relate trait

### LINKED ISSUE #513: Correctness issues with Contains
There are some correctness issues with our Contains implementation - discussed in https://github.com/georust/geo/pull/511

We should fix them.

I can find time to fix these this week, unless @rmanoka was already planning to do the work (I'm happy either way, so long as it gets fixed).

michaelkirk:
---

Fixes #513, #515

(I'm sorry it's so large)

~~I'm going to leave it as a draft (edit: 🤦 I failed to actually open the PR as a draft) while I wait to merge #636 and #638 and then do some rebasing, but I don't anticipate doing other large changes before review.~~ *update: ready for review!*

Here's some of the earlier work in pursuit of this:

#514
#516
#523
#524 
#538
#552
#561
#611
#628
#629
#636 

Primarily, this introduces the geomgraph module for a DE-9IM `Relate` trait.

geomgraph implements a topology graph largely inspired by JTS's module of the same name:
https://github.com/locationtech/jts/tree/jts-1.18.1/modules/core/src/main/java/org/locationtech/jts/geomgraph

You can see some of the reference code if you omit the "REMOVE JTS COMMENTS" commit. In some places the implementation is quite close to the JTS source. 

The overall "flow" is pretty similar to that of JTS, but in the small, there were some divergences. It's not easy (or desirable) to literally translate a Java codebase making heavy use of inheritance and pointers to rust. Additionally, I chose to take advantage of `Option` and rust's enums with associated data to make some case analysis more explicit.

There is a corresponding PR in our [jts-test-runner](https://github.com/georust/jts-test-runner/pull/6) crate which includes the bulk of the tests for the new Relate trait.

## Algorithm Overview 

This functionality is accessed on geometries, via the `Relate` trait, e.g. `line.relate(point)` which returns a DE-9IM [`IntersectionMatrix`](https://en.wikipedia.org/wiki/DE-9IM#Matrix_model).

The `Relate` trait is driven by the `RelateOperation`. The `RelateOperation` builds a `GeometryGraph` for each of the two geometries being related. 

A `GeometryGraph` is a systematic way to organize the "interesting" parts of a geometry's structure - e.g. where its vertices, lines, and areas lie relative to one another.

Once the `RelateOperation` has built the two `GeometryGraph`s, it uses them to efficiently compare the two Geometries's structures, outputting the `IntesectionMatrix`.

### DIFF
--- a/geo/Cargo.toml
+++ b/geo/Cargo.toml
@@ -19,6 +19,7 @@ num-traits = "0.2"
 serde = { version = "1.0", optional = true, features = ["derive"] }
 rstar = { version = "0.8" }
 geographiclib-rs = { version = "0.2" }
+log = "0.4.11"
 
 proj = { version = "0.20.3", optional = true }
 
@@ -32,11 +33,12 @@ proj-network = ["use-proj", "proj/network"]
 use-serde = ["serde", "geo-types/serde"]
 
 [dev-dependencies]
+pretty_env_logger = "0.4"
 approx = "0.4.0"
 criterion = { version = "0.3" }
 # jts-test-runner is an internal crate which exists only to be part of the geo test suite.
 # As such it's kept unpublished. It's in a separate repo primarily because it's kind of large.
-jts-test-runner = { git = "https://github.com/georust/jts-test-runner", rev = "3294b9af1d3e64fcc9caf9646a93d0116f7d8321" }
+jts-test-runner = { git = "https://github.com/georust/jts-test-runner", rev = "4a69a552546bd3d1dec574a51280dbc82bb417ce" }
 rand = "0.8.0"
 
 [[bench]]
@@ -73,6 +75,10 @@ harness = false
 
 [[bench]]
 name = "rotate"
+harness = false
+
+[[bench]]
+name = "relate"
 harness = false
 
 [[bench]]
--- a/geo/benches/relate.rs
+++ b/geo/benches/relate.rs
@@ -1,0 +1,36 @@
+#[macro_use]
+extern crate criterion;
+extern crate geo;
+
+use crate::geo::relate::Relate;
+use criterion::Criterion;
+use geo::{LineString, Polygon};
+
+fn criterion_benchmark(c: &mut Criterion) {
+    c.bench_function("relate overlapping 50-point polygons", |bencher| {
+        let points = include!("../src/algorithm/test_fixtures/norway_main.rs");
+
+        let sub_polygon = {
+            let points = points[0..50].to_vec();
+            let mut exterior = LineString::<f32>::from(points);
+            exterior.close();
+            Polygon::new(exterior, vec![])
+        };
+
+        let polygon = {
+            let points = points[40..90].to_vec();
+            let mut exterior = LineString::<f32>::from(points);
+            exterior.close();
+            Polygon::new(exterior, vec![])
+        };
+
+        bencher.iter(|| {
+            criterion::black_box(
+                criterion::black_box(&polygon).relate(criterion::black_box(&sub_polygon)),
+            );
+        });
+    });
+}
+
+criterion_group!(benches, criterion_benchmark);
+criterion_main!(benches);
--- a/geo/src/algorithm/contains/geometry.rs
+++ b/geo/src/algorithm/contains/geometry.rs
@@ -9,8 +9,8 @@ impl<T> Contains<Coordinate<T>> for Geometry<T>
 where
     T: GeoNum,
 {
-    geometry_delegate_impl! {
-        fn contains(&self, coord: &Coordinate<T>) -> bool;
+    fn contains(&self, coord: &Coordinate<T>) -> bool {
+        self.contains(&Point::from(*coord))
     }
 }
 
@@ -18,8 +18,8 @@ impl<T> Contains<Point<T>> for Geometry<T>
 where
     T: GeoNum,
 {
-    fn contains(&self, point: &Point<T>) -> bool {
-        self.contains(&point.0)
+    geometry_delegate_impl! {
+        fn contains(&self, point: &Point<T>) -> bool;
     }
 }
 
--- a/geo/src/algorithm/coordinate_position.rs
+++ b/geo/src/algorithm/coordinate_position.rs
@@ -2,12 +2,12 @@ use crate::algorithm::{
     bounding_rect::BoundingRect, dimensions::HasDimensions, intersects::Intersects,
 };
 use crate::{
-    Coordinate, GeoNum, Geometry, GeometryCollection, Line, LineString, MultiLineString,
-    MultiPoint, MultiPolygon, Point, Polygon, Rect, Triangle,
+    Coordinate, GeoNum, Geometry, GeometryCollection, GeometryCow, Line, LineString,
+    MultiLineString, MultiPoint, MultiPolygon, Point, Polygon, Rect, Triangle,
 };
 
 /// The position of a `Coordinate` relative to a `Geometry`
-#[derive(PartialEq, Clone, Debug)]
+#[derive(PartialEq, Clone, Copy, Debug)]
 pub enum CoordPos {
     OnBoundary,
     Inside,
@@ -327,48 +327,23 @@ where
     T: GeoNum,
 {
     type Scalar = T;
-    fn calculate_coordinate_position(
-        &self,
-        coord: &Coordinate<T>,
-        is_inside: &mut bool,
-        boundary_count: &mut usize,
-    ) {
-        if self.is_empty() {
-            return;
-        }
+    crate::geometry_delegate_impl! {
+        fn calculate_coordinate_position(
+            &self,
+            coord: &Coordinate<T>,
+            is_inside: &mut bool,
+            boundary_count: &mut usize) -> ();
+    }
+}
 
-        match self {
-            Geometry::Point(point) => {
-                point.calculate_coordinate_position(coord, is_inside, boundary_count)
-            }
-            Geometry::Line(line) => {
-                line.calculate_coordinate_position(coord, is_inside, boundary_count)
-            }
-            Geometry::LineString(line_string) => {
-                line_string.calculate_coordinate_position(coord, is_inside, boundary_count)
-            }
-            Geometry::Polygon(polygon) => {
-                polygon.calculate_coordinate_position(coord, is_inside, boundary_count)
-            }
-            Geometry::Rect(rect) => {
-                rect.calculate_coordinate_position(coord, is_inside, boundary_count)
-            }
-            Geometry::Triangle(triangle) => {
-                triangle.calculate_coordinate_position(coord, is_inside, boundary_count)
-            }
-            Geometry::MultiPoint(multi_point) => {
-                multi_point.calculate_coordinate_position(coord, is_inside, boundary_count)
-            }
-            Geometry::MultiLineString(multi_line_string) => {
-                multi_line_string.calculate_coordinate_position(coord, is_inside, boundary_count)
-            }
-            Geometry::MultiPolygon(multi_polygon) => {
-                multi_polygon.calculate_coordinate_position(coord, is_inside, boundary_count)
-            }
-            Geometry::GeometryCollection(geometry_collection) => {
-                geometry_collection.calculate_coordinate_position(coord, is_inside, boundary_count)
-            }
-        }
+impl<'a, T: GeoNum> CoordinatePosition for GeometryCow<'a, T> {
+    type Scalar = T;
+    crate::geometry_cow_delegate_impl! {
+        fn calculate_coordinate_position(
+            &self,
+            coord: &Coordinate<T>,
+            is_inside: &mut bool,
+            boundary_count: &mut usize) -> ();
     }
 }
 
--- a/geo/src/algorithm/relate/edge_end_builder.rs
+++ b/geo/src/algorithm/relate/edge_end_builder.rs
@@ -1,0 +1,125 @@
+use super::geomgraph::{Edge, EdgeEnd, EdgeIntersection};
+use crate::GeoFloat;
+
+use std::cell::RefCell;
+use std::rc::Rc;
+
+/// Computes the [`EdgeEnd`]s which arise from an [`Edge`] who has had its `edge_intersections`
+/// populated with self and proper [`EdgeIntersection`]s.
+///
+/// Based on [JTS's EdgeEndBuilder as of 1.18.1](https://github.com/locationtech/jts/blob/jts-1.18.1/modules/core/src/main/java/org/locationtech/jts/operation/relate/EdgeEndBuilder.java)
+pub(crate) struct EdgeEndBuilder<F: GeoFloat> {
+    _marker: std::marker::PhantomData<F>,
+}
+
+impl<F: GeoFloat> EdgeEndBuilder<F> {
+    pub fn new() -> Self {
+        EdgeEndBuilder {
+            _marker: std::marker::PhantomData,
+        }
+    }
+
+    pub fn compute_ends_for_edges(&self, edges: &[Rc<RefCell<Edge<F>>>]) -> Vec<EdgeEnd<F>> {
+        let mut list = vec![];
+        for edge in edges {
+            self.compute_ends_for_edge(&mut edge.borrow_mut(), &mut list);
+        }
+        list
+    }
+
+    /// Creates stub edges for all the intersections in the [`Edge`] (if any) and inserts them into
+    /// the graph's `list`.
+    fn compute_ends_for_edge(&self, edge: &mut Edge<F>, list: &mut Vec<EdgeEnd<F>>) {
+        edge.add_edge_intersection_list_endpoints();
+
+        let mut ei_iter = edge.edge_intersections().iter();
+        let mut ei_prev;
+        let mut ei_curr = None;
+        let mut ei_next = ei_iter.next();
+        if ei_next.is_none() {
+            return;
+        }
+
+        loop {
+            ei_prev = ei_curr;
+            ei_curr = ei_next;
+            ei_next = ei_iter.next();
+            if let Some(ei_curr) = ei_curr {
+                self.create_edge_end_for_prev(edge, list, ei_curr, ei_prev);
+                self.create_edge_end_for_next(edge, list, ei_curr, ei_next);
+            }
+
+            if ei_curr.is_none() {
+                break;
+            }
+        }
+    }
+
+    /// Adds a `EdgeEnd`, if any, to `list` for the edge before the intersection ei_curr.
+    ///
+    /// The previous intersection is provided in case it is the endpoint for the stub edge.
+    /// Otherwise, the previous point from the parent edge will be the endpoint.
+    fn create_edge_end_for_prev(
+        &self,
+        edge: &Edge<F>,
+        list: &mut Vec<EdgeEnd<F>>,
+        ei_curr: &EdgeIntersection<F>,
+        ei_prev: Option<&EdgeIntersection<F>>,
+    ) {
+        let mut i_prev = ei_curr.segment_index();
+        if ei_curr.distance().is_zero() {
+            // if at the start of the edge there is no previous edge
+            if i_prev == 0 {
+                return;
+            }
+            i_prev -= 1;
+        }
+
+        let mut coord_prev = edge.coords()[i_prev];
+        // if prev intersection is past the previous vertex, use it instead
+        if let Some(ei_prev) = ei_prev {
+            if ei_prev.segment_index() >= i_prev {
+                coord_prev = ei_prev.coordinate();
+            }
+        }
+
+        let mut label = edge.label().clone();
+        // since edgeStub is oriented opposite to its parent edge, have to flip sides for edge label
+        label.flip();
+
+        let edge_end = EdgeEnd::new(ei_curr.coordinate(), coord_prev, label);
+        list.push(edge_end);
+    }
+
+    /// Adds a `EdgeEnd`, if any, to `list` for the edge after the intersection ei_curr.
+    ///
+    /// The next intersection is provided in case it is the endpoint for the stub edge.
+    /// Otherwise, the next point from the parent edge will be the endpoint.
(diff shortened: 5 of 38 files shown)


geo/Cargo.toml
```diff
@@ -19,6 +19,7 @@ num-traits = "0.2"
 serde = { version = "1.0", optional = true, features = ["derive"] }
 rstar = { version = "0.8" }
 geographiclib-rs = { version = "0.2" }
+log = "0.4.11"
```
michaelkirk, line 22: I've been waffling on doing this for a while, but finally bit the bullet. 

FIXES https://github.com/georust/geo/issues/596 if you're on board.

geo/Cargo.toml
```diff
@@ -32,11 +33,12 @@ proj-network = ["use-proj", "proj/network"]
 use-serde = ["serde", "geo-types/serde"]
 
 [dev-dependencies]
+pretty_env_logger = "*"
```
michaelkirk, line 36: It's helpful to tune the logging when running tests. This is my go-to, but I could be convinced otherwise.

geo/Cargo.toml
```diff
@@ -32,11 +33,12 @@ proj-network = ["use-proj", "proj/network"]
...
 approx = "0.4.0"
 criterion = { version = "0.3" }
 # jts-test-runner is an internal crate which exists only to be part of the geo test suite.
 # As such it's kept unpublished. It's in a separate repo primarily because it's kind of large.
-jts-test-runner = { git = "https://github.com/georust/jts-test-runner", rev = "3294b9af1d3e64fcc9caf9646a93d0116f7d8321" }
+jts-test-runner = { git = "https://github.com/georust/jts-test-runner", branch = "mkirk/relate" }
```
michaelkirk, line 41: See: https://github.com/georust/jts-test-runner/pull/6

geo/src/algorithm/contains/geometry.rs
```diff
@@ -7,19 +7,19 @@ use crate::*;
 
 impl<T> Contains<Coordinate<T>> for Geometry<T>
 where
-    T: GeoNum,
+    T: GeoFloat,
```
michaelkirk, line 10: Note this is a breaking change - I'm dropping support for `Contains` for integer Polygons, and thus integer Geometries.

If this is controversial, I could revert this part, but IMO it's more important to fix #513 than to support integer geometries.

geo/src/algorithm/coordinate_position.rs
```diff
@@ -2,12 +2,12 @@ use crate::algorithm::{
...
-    Coordinate, GeoNum, Geometry, GeometryCollection, Line, LineString, MultiLineString,
-    MultiPoint, MultiPolygon, Point, Polygon, Rect, Triangle,
+    Coordinate, GeoNum, Geometry, GeometryCollection, GeometryCow, Line, LineString,
+    MultiLineString, MultiPoint, MultiPolygon, Point, Polygon, Rect, Triangle,
 };
 
 /// The position of a `Coordinate` relative to a `Geometry`
-#[derive(PartialEq, Clone, Debug)]
+#[derive(PartialEq, Clone, Copy, Debug)]
 pub enum CoordPos {
```
michaelkirk, line 11: This might be a point of confusion if you are comparing to the reference JTS code...

There are two enums JTS uses ubiquitously in the geomgraph module:

`JTS::Location: { Interior, Boundary, Exterior }` and `JTS::Position: { On, Left, Right }`

Somewhat confusingly `JTS::Location` maps to `geo::CoordPos` and we didn't yet have a concept of `JTS::Position`, and I wanted to call it something other than `Position` to differentiate it from our `CoordPos`, so I've introduced `pub(crate) enum geomgraph::Direction { On, Left, Right }`. 

Unfortunately, this somewhat collides with `pub enum orient::Direction { Default, Reversed }`, but I don't have a better solution at this point. FWIW it's not `pub`, so we can always revisit later without breaking public APIs.

geo/src/algorithm/coordinate_position.rs
```diff
@@ -327,48 +327,23 @@ where
...
-            Geometry::MultiPolygon(multi_polygon) => {
-                multi_polygon.calculate_coordinate_position(coord, is_inside, boundary_count)
-            }
-            Geometry::GeometryCollection(geometry_collection) => {
-                geometry_collection.calculate_coordinate_position(coord, is_inside, boundary_count)
-            }
-        }
+impl<'a, T: GeoNum> CoordinatePosition for GeometryCow<'a, T> {
+    type Scalar = T;
+    crate::geometry_cow_delegate_impl! {
```
michaelkirk, line 341: See the new `enum GeometryCow` for details...

Some relevant discussion https://github.com/georust/geo/pull/421#issuecomment-763162741

geo/src/algorithm/relate/geomgraph/index/simple_edge_set_intersector.rs
```diff
@@ -0,0 +1,59 @@
...
+    fn compute_intersections_within_set(
+        &mut self,
+        edges: &[Rc<RefCell<Edge<F>>>],
+        check_for_self_intersecting_edges: bool,
+        segment_intersector: &mut SegmentIntersector<F>,
+    ) {
+        for edge0 in edges.iter() {
+            for edge1 in edges.iter() {
+                if check_for_self_intersecting_edges || edge0.as_ptr() != edge1.as_ptr() {
+                    self.compute_intersects(edge0, edge1, segment_intersector);
```
michaelkirk, line 41: I believe there were two reasons for the unfortunate use of `Rc<RefCell<>>`.

The lesser one is that RelateOperation grabs a duplicated reference to `Edge` at some point. I think we could get rid of that usage by storing an arena-style index-of-edge instead of a pointer-to-edge.

The bigger reason is here: `compute_intersection` calls `segment_intersector.add_intersection` which, as well as *computing* the intersections, also mutates the edge to record those intersections **and it might be intersecting an edge with itself**.

This all makes for some pretty unremarkable Java code, but rust makes it painfully clear that this is less-than-nice.

I think it would be *possible* to remove the `Rc<RefCell<>>` stuff with more work, but I decided to leave it as is for now.

FWIW I did some profiling and the RefCell/Rc stuff does not significantly contribute to runtime.

geo/src/algorithm/relate/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+    // pop b through all of b with head of a
+    ([$($out_pairs:tt)*] [$a_car:ty, $($a_cdr:ty,)*] [$b_car:ty, $($b_cdr:ty,)*] $init_b:tt $macro_name:ident) => {
+        cartesian_pairs_helper!{[$($out_pairs)* ($a_car, $b_car),] [$a_car, $($a_cdr,)*] [$($b_cdr,)*] $init_b $macro_name}
+    };
+}
+
+// Implement Relate for every combination of Geometry. Alternatively we could do something like
+// `impl Relate<Into<GeometryCow>> for Into<GeometryCow> { }`
+// but I don't know that we want to make GeometryCow public (yet?).
+cartesian_pairs!(relate_impl, [Geometry<F>, Point<F>, Line<F>, LineString<F>, Polygon<F>, MultiPoint<F>, MultiLineString<F>, MultiPolygon<F>, Rect<F>, Triangle<F>]);
```
michaelkirk, line 111: These insane macros are all in service of keeping GeometryCow private (for now at least).

michaelkirk (commented):
Ok, this is ready for review!

geo/Cargo.toml
frewsxcv: sgtm. can we lock it to a version so our tests don't break if there's ever breaking changes?

michaelkirk: done!

frewsxcv (commented):
not done reviewing, but here's some early comments

geo/src/types.rs
```diff
@@ -57,38 +57,7 @@ impl<F: GeoFloat> Closest<F> {
...
-/// }
-/// impl<T: GeoNum> Foo<T> for GeometryCollection<T> {
-///     fn foo_1(&self, coord: Coordinate<T>)  -> bool { false }
-///     fn foo_2(&self)  -> i32 { 2 }
-/// }
-/// impl<T: GeoNum> Foo<T> for MultiLineString<T> {
-///     fn foo_1(&self, coord: Coordinate<T>)  -> bool { false }
-///     fn foo_2(&self)  -> i32 { 1 }
-/// }
+/// ... and so on
```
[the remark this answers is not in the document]
michaelkirk: Good comments about these macros! They were kind of a mess. I've done some juggling which I think addresses all your feedback:

- I wanted an example of how to use it in the docs, so I fleshed out the entire example, and removed the `ignore`.
- This revealed a bug... I need to prefix the inner helper macro with `$crate::` not `crate::`.
- I also moved the doc comment from this helper method to the public `geometry_delegate_impl!` macro.
- I prefixed the helper method with `__` to indicate that it's intended to be called directly.
- I marked `geometry_cow_delegate_impl` as `doc(hidden)`. Since `GeometryCow` is `pub(crate)` I want `geometry_cow_delegate_impl` to have similarly restricted visibility, but I don't know enough about macros to know if there's a better way than just hiding it from the docs. Is there a better way?

geo/src/algorithm/relate/geomgraph/topology_position.rs
```diff
@@ -0,0 +1,197 @@
...
+/// If the graph component is an _area_ edge, there is a position for each [`Direction`]:
+/// - [`On`](Direction::On): on the edge
+/// - [`Left`](Direction::Left): left-hand side of the edge
+/// - [`Right`](Direction::Right): right-hand side
+///
+/// If the parent component is a _line_ edge or a node (a point), there is a single
+/// topological relationship attribute for the [`On`](Direction::On) position.
+///
+/// See [`CoordPos`] for the possible values.
+#[derive(Clone)]
```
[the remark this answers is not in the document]
michaelkirk: Done!

geo/src/algorithm/relate/geomgraph/quadrant.rs
```diff
@@ -0,0 +1,32 @@
+use crate::GeoNum;
+
+/// Utility functions for working with quadrants of the cartesian plane,
+/// which are labeled as follows:
+///          (+)
+///        NW ┃ NE
+///    (-) ━━━╋━━━━ (+)
+///        SW ┃ SE
+///          (-)
```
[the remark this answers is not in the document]
michaelkirk: Done! 

(fyi this and much of the geomgraph docs are hidden but you can see with `cargo doc --document-private-items`)

geo/src/types.rs
```diff
@@ -129,18 +99,28 @@ macro_rules! geometry_delegate_impl {
...
         };
 }
+
+#[macro_export]
+macro_rules! geometry_cow_delegate_impl {
+    ($($a:tt)*) => { crate::geometry_delegate_impl_helper!{ GeometryCow, $($a)* } }
+}
+
+#[macro_export]
+macro_rules! geometry_delegate_impl {
```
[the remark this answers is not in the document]
michaelkirk: responded in https://github.com/georust/geo/pull/639#discussion_r598317268

geo/src/types.rs
```diff
@@ -129,18 +99,28 @@ macro_rules! geometry_delegate_impl {
...
+                        $enum::Rect(g) => g.$func_name($($arg_name),*).into(),
+                        $enum::Triangle(g) => g.$func_name($($arg_name),*).into(),
                     }
                 }
             )+
         };
 }
+
+#[macro_export]
+macro_rules! geometry_cow_delegate_impl {
```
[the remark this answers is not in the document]
michaelkirk: responded in https://github.com/georust/geo/pull/639#discussion_r598317268

geo/src/algorithm/relate/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+        relate_computer.compute_intersection_matrix()
+    }
+}
+
+macro_rules! relate_impl {
+    ($(($k:ty, $t:ty),)*) => {
+        $(
+            impl<F: GeoFloat> Relate<F, $t> for $k {
+                fn relate(&self, other: &$t) -> IntersectionMatrix {
+                    GeometryCow::from(self).relate(&GeometryCow::from(other))
```
michaelkirk, line 69: FYI: this line is why I introduced `GeometryCow`.

Unlike much of our other code, the GeometryGraph was implemented to deal only with `Geometry`, rather than have a bunch of different implementations for each Geometry variant.

However, if we want to be able to call something like: 

```
multi_line_string.relate(polygon)
```

We'd actually have to do something like:
```
Geometry::from(multi_line_string.clone()).relate(Geometry::from(polygon.clone()))
```

Those clones are really unfortunate - hence `GeometryCow`, which is an enum like `Geometry` but can be made from a variant without cloning.

geo/src/algorithm/relate/geomgraph/geometry_graph.rs
```diff
@@ -0,0 +1,387 @@
...
+        if line_string.is_empty() {
+            return;
+        }
+
+        let mut coords: Vec<Coordinate<F>> = vec![];
+        for coord in &line_string.0 {
+            if coords.last() != Some(coord) {
+                coords.push(*coord)
+            }
+        }
```
frewsxcv, line 222: is the conditional in the loop to avoid adding the endpoint? in most cases, the filter won't apply, so we could preallocate the vector on line 217 with `Vec::with_capacity`

geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
```diff
@@ -0,0 +1,233 @@
...
+            for b in &[CoordPos::Inside, CoordPos::OnBoundary, CoordPos::Outside] {
+                match chars.next().expect("already validated length is 9") {
+                    '0' => self.0[*a][*b] = self.0[*a][*b].max(Dimensions::ZeroDimensional),
+                    '1' => self.0[*a][*b] = self.0[*a][*b].max(Dimensions::OneDimensional),
+                    '2' => self.0[*a][*b] = self.0[*a][*b].max(Dimensions::TwoDimensional),
+                    'F' => {}
+                    other => {
+                        let message = format!("expected '0', '1', '2', or 'F'. Found: {}", other);
+                        // We should make this return Err if this method becomes public
+                        return Err(InvalidInputError::new(message));
```
frewsxcv, line 181:
```
// We should make this return Err if this method becomes public
```

doesn't this return `Err`?

geo/src/algorithm/relate/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+    Point, Polygon, Rect, Triangle,
+};
+
+mod edge_end_builder;
+mod geomgraph;
+mod relate_operation;
+
+/// Topologically relate two geometries based on [DE-9IM](https://en.wikipedia.org/wiki/DE-9IM) semantics.
+///
+/// See [`IntersectionMatrix`] for details.
```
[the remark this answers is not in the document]
michaelkirk: Done!

geo/src/algorithm/relate/mod.rs
michaelkirk: Oh, good catch. This was actually two bugs. 

1. GeometryCollection _should_ be included here.


2. Geometry should _not_ be included here.

In line with our other traits, we only want a single `impl Relate<Geometry> for Geometry`, rather than a `impl Relate<Point> for Geometry`, `impl Relate<Line> for Geometry`, etc.

geo/src/algorithm/relate/geomgraph/geometry_graph.rs
michaelkirk: This was copied from JTS, but presumably there's no topological value in having redundant points adjacent in a vec.

I think you're right though - typically the filter won't apply.

I've done two things:
1. `Vec::with_capacity` here (and similarly in add_polygon)
2. `coords.shrink_to_fit()` in Edge construction, since these coord vec's never change len after this point.

geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
michaelkirk: Whoops - old TODO that got TODONE.

michaelkirk:
I am available over the weekend if anyone would like a review docent.

geo/src/algorithm/contains/geometry.rs
michaelkirk: @rmanoka  - You've previously convinced me that integer geometries are neat, so I'd like to support them, but it seems unreasonable to leave things broken for the mainstream use case in order to hold on to a possibility of something more esoteric.

Assuming [specialization](https://github.com/rust-lang/rust/issues/31844) won't be finalized any time soon, do you have any thoughts on how we could support both? 

I'd be curious to see how well something like this geometry graph approach lends itself to integer geometries. I took a brief spelunking just to see what would break from a code POV: https://github.com/georust/geo/commit/cf56814202c3aefd297ad1febaef541fe1c20640

The large commented out portions are the things that don't compile as GeoNum.

I don't have any experience working with integer geometries - some of it seems readily addressable, but the biggest immediate hangup for me is that the approach in this PR relies heavily on knowing where lines intersect - is their a corollary in integer geometries? I'm assuming you can't just round.

michaelkirk: I should say, it is my preference to get this mega PR merged without adding even more changes to support Relate for integers. I think the Relate functionality is useful even if it's just for Floats (which I expect is almost all users).

As I said, if it's controversial to drop support for Contains<Integer>, I could revert this portion. I'd still add the new relate trait, just for Floats, but revert the fix for the Contains edge cases. I'd be OK with that approach if that's your preference @rmanoka.

rmanoka: @michaelkirk Relate is a brilliant trait and I think we should definitely add it even if can only supports `GeoFloat`.  What I was wondering is whether we could leave the previous implementations of `Contains` as long as they pass all test cases as `GeoNum` subsumes `GeoFloat`.

For the above case, is it possible to use our hand-rolled `Contains<Coordinate>` and `Contains<Point>` implementations for the geometries?  From what I recall, we did handle the edge cases, and hopefully the JTS test suite will point out any edge cases we missed out for these.  I do agree `Contains<LineString>`, and `Contains<Polygon>` should use the geomgraph techniques you've implemented as there were corner cases we couldn't handle easily earlier.

michaelkirk: And just to be clear, I'm still removing the API for other GeoNum backed variants, like `LineString<GeoNum>`, which now only support `LineString<GeoFloat>`

geo/benches/relate.rs
```diff
@@ -0,0 +1,33 @@
...
+
+        let polygon = {
+            let points = points[30..90].to_vec();
+            let mut exterior = LineString::<f32>::from(points);
+            exterior.close();
+            Polygon::new(exterior, vec![])
+        };
+
+        bencher.iter(|| {
+            criterion::black_box(criterion::black_box(&polygon).relate(&sub_polygon));
```
frewsxcv, line 27: I recall in the depths of my brain that every parameter of the function your testing should be wrapped in a `black_box`, but I can't find a source for that now 🤷🏻 

Related to https://github.com/rust-lang/rust/issues/80480

geo/src/algorithm/relate/edge_end_builder.rs
```diff
@@ -0,0 +1,125 @@
+use super::geomgraph::{Edge, EdgeEnd, EdgeIntersection};
+use crate::GeoFloat;
+
+use std::cell::RefCell;
+use std::rc::Rc;
+
+/// Computes the [`EdgeEnd`]s which arise from an [`Edge`] who has had it's `edge_intersections`
```
frewsxcv, line 7:
```suggestion
/// Computes the [`EdgeEnd`]s which arise from an [`Edge`] who has had its `edge_intersections`
```

geo/src/algorithm/relate/geomgraph/geometry_graph.rs
```diff
@@ -0,0 +1,387 @@
...
+                // PERF: avoid this conversion/clone?
+                self.add_polygon(&rect.to_polygon());
+            }
+            GeometryCow::Point(point) => {
+                self.add_point(point);
+            }
+            GeometryCow::Polygon(polygon) => self.add_polygon(polygon),
+            GeometryCow::Triangle(triangle) => {
+                // PERF: avoid this conversion/clone?
+                self.add_polygon(&Polygon::from(triangle.clone().into_owned()));
```
frewsxcv, line 123: is the `into_owned()` here redundant with `clone`?

frewsxcv (approved):
finally found some time to look through all this and wow is this over my head lol. i did try to follow the best i could given the capacity i have right now, and i think i have a rough understanding of all the pieces here! i wish i had more time to learn more about the nitty gritty details, but no reason that need for that to wait on me. this is so exciting @michaelkirk, thanks for adding all this!!

geo/src/algorithm/relate/geomgraph/geometry_graph.rs
michaelkirk: Good catch! 

I've changed this to `triangle.to_polygon()` which I think is still clearer.

michaelkirk:
> finally found some time to look through all this

Thank you so much for slogging through this @frewsxcv!

@urschrei / @rmanoka - would either of you like me to hold off on merging this a while longer until you've had a look?

geo/benches/relate.rs
michaelkirk: I've added another block_box around the argument. So now it's:

```
 criterion::black_box(criterion::black_box(&polygon).relate(criterion::black_box(&sub_polygon)));
```

rmanoka:
I think it's good to go, especially given all the JTS tests, and our edge case tests (prev. ignored) all pass!  Thanks again for the wonderful contrib.

michaelkirk:
bor r+

(rebased and squashed)

michaelkirk:
bors r-

(whoops - I want to merge the jts-tests first)

michaelkirk:
bors r=frewsxcv,rmanoka

PR #642, opened 2021-04-13
Introduce the geomgraph module for DE-9IM Relate trait

michaelkirk:
---

This is just a dupe of #639, but targeting master this time. I just now merged #639, forgetting #639 was targeting #636 as its base for the purposes of minimizing the diff (at the time, #636 was unmerged). 

**from #639**
Screen Shot 2021-04-13 at 12 01 25 PM

...I suppose an epically long PR wouldn't be complete without an epically long merge process. 😓


michaelkirk:
bors r=frewsxcv,rmanoka

(from #639)

PR #644, opened 2021-04-15
update postgis

### LINKED ISSUE #643: postgis has been updated to 0.8.0 -- update geo-postgis?
This was leading to some confusion when I was playing with geo-postgis.
Would that break anything?

michaelkirk:
---

FIXES #643

### DIFF
--- a/geo-postgis/Cargo.toml
+++ b/geo-postgis/Cargo.toml
@@ -11,5 +11,5 @@ description = "Conversion between `geo-types` and `postgis` types."
 edition = "2018"
 
 [dependencies]
-postgis = "0.7"
+postgis = { version = ">=0.7.0, <9.0" }
 geo-types = { version = "0.7", path = "../geo-types" }
--- a/geo-postgis/Cargo.toml
+++ b/geo-postgis/Cargo.toml
@@ -11,5 +11,5 @@ description = "Conversion between `geo-types` and `postgis` types."
 edition = "2018"
 
 [dependencies]
-postgis = { version = ">=0.7.0, <9.0" }
+postgis = { version = ">=0.7.0, <0.9.0" }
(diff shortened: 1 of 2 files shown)


geo-postgis/Cargo.toml
```diff
@@ -11,5 +11,5 @@ description = "Conversion between `geo-types` and `postgis` types."
 edition = "2018"
 
 [dependencies]
-postgis = "0.7"
+postgis = { version = ">=0.7.0, <9.0" }
```
lnicola, line 14:
```suggestion
postgis = { version = ">=0.7.0, <0.9.0" }
```

TBH I understand, but don't like this style of pinning, since some of the crates that use it are very slow at putting up new releases.

michaelkirk: Wow, bad typo on my part. Thank you for noticing! ❤️ 

> TBH I understand, but don't like this style of pinning, since some of the crates that use it are very slow at putting up new releases.

Interesting, but I'm not sure I understand.

My (simple) rule is to avoid unnecessary breaking changes. Requiring postgis 0.8 would be an unnecessary breaking change, so we'd have to breaking-bump semver, and anyone downstream that depends on this crate would have to breaking-bump semver on their crate.

Can you give an example or clarify why it might be better to force this cascade of updates to happen when we could instead support both the new and the old?

lnicola: No, it's fine to support both `0.7` and `0.8`, just that (as a user of crates that do this -- `diesel` comes to mind), when `0.9` comes out, I won't be able to upgrade to it until `postgis` changes the dependency to `>= 0.7, < 0.10`, which might take a long time to get in a release.

So it's correct to do this, just very annoying at times.

michaelkirk:
bors r=lnicola

PR #646, opened 2021-04-16
add rust-1.51 to ci

michaelkirk:
---

Depends on https://github.com/georust/docker-images/pull/12

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,7 +35,7 @@ jobs:
         working-directory: geo-types
     strategy:
       matrix:
-        container_image: ["georust/geo-ci:rust-1.49", "georust/geo-ci:rust-1.50"]
+        container_image: ["georust/geo-ci:rust-1.49", "georust/geo-ci:rust-1.50", "georust/geo-ci:rust-1.51"]
     container:
       image: ${{ matrix.container_image }}
     steps:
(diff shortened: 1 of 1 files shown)


frewsxcv (approved):

lnicola:
1.52 is already out, but does it make sense to test on every version?

michaelkirk:
> 1.52 is already out, but does it make sense to test on every version? stable and some MSRV (and maybe beta or nightly) should suffice.

Yes! I think ideally we'd do something like what you've described, where we target tags like "stable" an MSRV (and maybe beta/nightly).

But I think that implies some machinery to automatically rebuild stable/beta/nightly, is that what you mean @lnicola? 

I'm not very familiar with docker/docker-hub, or what the best way to do something like that is, so we've just been hobbling along with manual updates so as not to get toooo far behind.

I'd be happy if you have any advice (or better yet, if you are interested in setting it up).

lnicola:
No, I don't really have any advice about rebuilding the images automatically (though it should be scriptable, at least), but that's independent of whether we want N (=3 currently) Rust versions in the testing matrix.

lnicola:
PS: I think GitHub Actions can be triggered on a schedule, so one approach would be to rebuild the stable/beta/nightly images daily and push them to Docker Hub.

michaelkirk:
bors r=frewsxcv

PR #648, opened 2021-05-05
Add chaikins smoothing algorithm

nmandery:
This PR adds an implementation of [Chaikins smoothing](http://www.idav.ucdavis.edu/education/CAGDNotes/Chaikins-Algorithm/Chaikins-Algorithm.html) algorithm.

Thank you for creating these crates.

### DIFF
--- a/geo/src/algorithm/chaikins_smoothing.rs
+++ b/geo/src/algorithm/chaikins_smoothing.rs
@@ -1,0 +1,211 @@
+use std::ops::Mul;
+
+use num_traits::FromPrimitive;
+
+use crate::{CoordFloat, Coordinate, LineString, MultiLineString, MultiPolygon, Polygon};
+
+/// Smoothen `LineString`, `Polygon`, `MultiLineString` and `MultiPolygon` using Chaikins algorithm.
+///
+/// [Chaikins smoothing algorithm](http://www.idav.ucdavis.edu/education/CAGDNotes/Chaikins-Algorithm/Chaikins-Algorithm.html)
+///
+/// This implementation preserves the start and end vertices of an open linestring and
+/// smoothes the corner between start and end of a closed linestring.
+///
+/// The smoothing increases the number of vertices of the geometry, so in some
+/// cases it makes sense to apply a simplification afterwards.
+pub trait ChaikinsSmoothing<T>
+where
+    T: CoordFloat + FromPrimitive,
+{
+    /// create a new geometry with the Chaikins smoothing being
+    /// applied `n_iterations` times
+    fn chaikins_smoothing(&self, n_iterations: usize) -> Self;
+}
+
+impl<T> ChaikinsSmoothing<T> for LineString<T>
+where
+    T: CoordFloat + FromPrimitive,
+{
+    fn chaikins_smoothing(&self, n_iterations: usize) -> Self {
+        if n_iterations == 0 {
+            self.clone()
+        } else {
+            let mut smooth = smoothen_linestring(self);
+            for _ in 0..(n_iterations - 1) {
+                smooth = smoothen_linestring(&smooth);
+            }
+            smooth
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


geo/src/algorithm/chaikins_smoothing.rs
```diff
@@ -0,0 +1,211 @@
...
+{
+    let mut out_coords: Vec<_> = Vec::with_capacity(linestring.0.len() * 2);
+
+    if let (Some(first), Some(last)) = (linestring.0.first(), linestring.0.last()) {
+        if first != last {
+            // preserve start coordinate when the linestring is open
+            out_coords.push(*first);
+        }
+    }
+    for window_coordiates in linestring.0.windows(2) {
```
urschrei, line 97: Typo fix: `window_coordiates` --> `window_coordinates`

geo/src/algorithm/chaikins_smoothing.rs
```diff
@@ -0,0 +1,211 @@
...
+
+use crate::{CoordFloat, Coordinate, LineString, MultiLineString, MultiPolygon, Polygon};
+
+/// Smoothen `LineString`, `Polygon`, `MultiLineString` and `MultiPolygon` using Chaikins algorithm.
+///
+/// [Chaikins smoothing algorithm](http://www.idav.ucdavis.edu/education/CAGDNotes/Chaikins-Algorithm/Chaikins-Algorithm.html)
+///
+/// This implementation preserves the start and end vertices of an open linestring and
+/// smoothes the corner between start and end of a closed linestring.
+///
```
urschrei, line 13: Would it make sense to add a line explaining the effect of increasing `n_iterations`?

urschrei (changes requested):
Hi @nmandery, thanks for the PR! I've got some small changes, and one more general: while the algorithm is known as "Chaikin's smoothing algorithm" and it's fine to describe it like that in the docs, I'd prefer if the module and related functions were named e.g. `ChaikinSmoothing` and `chaikin_smoothing`, without the repeated "s", which is a bit unintuitive to remember / type.

geo/src/algorithm/chaikins_smoothing.rs
nmandery: Good catch - corrected

geo/src/algorithm/chaikins_smoothing.rs
nmandery: Also a good point - I added that and reworded that comment.

nmandery:
Thanks for the feedback. I changed the naming - it is indeed nicer without the "s".

rmanoka:
@nmandery Neat implementation!  Is it true that the number of vertices in the line string increases exponentially with number of iterations?  So m vertices becomes about `m*2^k` after k iterations?  This may be useful to document as it'll help the user pick reasonable k to not explode memory usage.  Also, you could consider initialising using  `Vec::with_capacity` if that's the case.

nmandery:
@rmanoka Thank you. Yes each iteration doubles the number of vertices - I just added that to the documentation of the trait. Like you said, this is quite essential for the user to know.

I am currently using `Vec::with_capacity` to create the vec for the new `LineString` created during one iteration. The reserved size is twice the size of the input vec. It may be more efficient to rework the algorithm to reserve a new vec only once across all iterations and operate on that. Not sure if thats worth it, though.

rmanoka:
> I am currently using `Vec::with_capacity` to create the vec for the new `LineString` created during one iteration. The reserved size is twice the size of the input vec. It may be more efficient to rework the algorithm to reserve a new vec only once across all iterations and operate on that. Not sure if thats worth it, though.

Yeah that's fair.  It's probably a micro optimization that should not be a bottle-neck at all.

rmanoka (approved):
Looks good to me

urschrei (approved):

PR #651, opened 2021-07-13
multipolygon rotate pivot

mapoulos:
---

### DIFF
--- a/geo/src/algorithm/rotate.rs
+++ b/geo/src/algorithm/rotate.rs
@@ -345,4 +345,9 @@ mod test {
         let line1 = Line::new(Point::new(0., 0.), Point::new(-2., 0.));
         assert_relative_eq!(line0.rotate_around_point(90., Point::new(0., 0.)), line1);
     }
+
+    #[test]
+    fn test_rotate_multipolygon_around_centroid() {
+        todo!("implement");
+    }
 }
--- a/geo/src/algorithm/rotate.rs
+++ b/geo/src/algorithm/rotate.rs
@@ -363,7 +363,7 @@ mod test {
                 (x: 10., y: 10.),
                 (x: 0., y: 10.),
                 (x: 0., y: 0.),
-            ], 
+            ],
             polygon![
                 (x: 0., y: 0.),
                 (x: -10., y: 0.),
@@ -371,7 +371,7 @@ mod test {
                 (x: 0., y: -10.),
                 (x: 0., y: 0.),
             ]
-        ].into(); 
+        ].into();
             
         let expected: MultiPolygon<f64> = vec![
             polygon![
@@ -389,7 +389,7 @@ mod test {
                 (x: 0., y: 0.),
             ]
         ].into();
-        
+
         // results agree with Shapely / GEOS
         // (relaxing the episilon a bit)
         assert_relative_eq!(multipolygon.rotate(45.), expected, epsilon=1e-12);
--- a/geo/src/algorithm/rotate.rs
+++ b/geo/src/algorithm/rotate.rs
@@ -220,11 +220,9 @@ where
 
 #[cfg(test)]
 mod test {
-    use std::ops::Mul;
-
     use super::*;
     use crate::{line_string, point, polygon, Coordinate, Point};
-    use approx::{RelativeEq, assert_relative_eq};
+    use approx::{assert_relative_eq};
 
     #[test]
     fn test_rotate_around_point() {
--- a/geo/src/algorithm/rotate.rs
+++ b/geo/src/algorithm/rotate.rs
@@ -194,7 +194,14 @@ where
 {
     /// Rotate the contained Polygons about their centroids by the given number of degrees
     fn rotate(&self, angle: T) -> Self {
-        MultiPolygon(self.iter().map(|poly| poly.rotate_around_point(angle, self.centroid().unwrap())).collect())
+        match self.centroid() {
+            Some(centroid) => {
+                MultiPolygon(self.iter().map(|poly| poly.rotate_around_point(angle, centroid)).collect())
+            },
+            None => {
+                self.clone()
+            }
+        }
     }
 }
 
--- a/geo/src/algorithm/rotate.rs
+++ b/geo/src/algorithm/rotate.rs
@@ -222,7 +222,7 @@ where
 mod test {
     use super::*;
     use crate::{line_string, point, polygon, Coordinate, Point};
-    use approx::{assert_relative_eq};
+    use approx::assert_relative_eq;
 
     #[test]
     fn test_rotate_around_point() {
--- a/geo/src/algorithm/rotate.rs
+++ b/geo/src/algorithm/rotate.rs
@@ -199,6 +199,7 @@ where
                 MultiPolygon(self.iter().map(|poly| poly.rotate_around_point(angle, centroid)).collect())
             },
             None => {
+                // Multipolygon was empty or otherwise degenerate and had no computable centroid
                 self.clone()
             }
         }
@@ -388,7 +389,7 @@ mod test {
                 (x: 0., y: 0.),
             ]
         ].into();
-
-        assert_relative_eq!(multipolygon.rotate(-90.), expected, max_relative=1.0);
+        
+        assert_relative_eq!(multipolygon.rotate(-90.), expected, max_relative=1.);
     }
 }
--- a/geo/src/algorithm/rotate.rs
+++ b/geo/src/algorithm/rotate.rs
@@ -390,6 +390,7 @@ mod test {
             ]
         ].into();
         
+        //TODO: make sure max_relative is doing the right thing here (or fix the test)
         assert_relative_eq!(multipolygon.rotate(-90.), expected, max_relative=1.);
     }
 }
--- a/geo/src/algorithm/rotate.rs
+++ b/geo/src/algorithm/rotate.rs
@@ -194,7 +194,7 @@ where
 {
     /// Rotate the contained Polygons about their centroids by the given number of degrees
     fn rotate(&self, angle: T) -> Self {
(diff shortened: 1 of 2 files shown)


geo/src/algorithm/rotate.rs
```diff
@@ -194,7 +194,7 @@ where
 {
     /// Rotate the contained Polygons about their centroids by the given number of degrees
     fn rotate(&self, angle: T) -> Self {
-        MultiPolygon(self.iter().map(|poly| poly.rotate(angle)).collect())
+        MultiPolygon(self.iter().map(|poly| poly.rotate_around_point(angle, self.centroid().unwrap())).collect())
```
lnicola, line 197: Maybe?

```suggestion
        let centroid = self.centroid().unwrap();
        MultiPolygon(self.iter().map(|poly| poly.rotate_around_point(&centroid)).collect())
```

The `unwrap()` doesn't seem ideal, maybe it should return a clone of `self` there's no centroid?

mapoulos: hmm, thinking about this more it seems like if we can't find a centroid, then we should probably return a None, which would require changing the function signature to return an option (as we find with `centroid()`)

mapoulos: Or put another way, is there a satisfactory failure state if a centroid can't be returned?

mapoulos: note `rotate` for polygon does do an `unwrap` on the centroid, which is why I did the same.  is that not a good precedent to follow?

```rust
        // if a polygon has holes, use the centroid of its outer shell as the rotation origin
        let centroid = if self.interiors().is_empty() {
            self.centroid().unwrap()
        } else {
            self.exterior().centroid().unwrap()
        };
        Polygon::new(
            rotate_many(angle, centroid, self.exterior().points_iter()).collect(),
            self.interiors()
                .iter()
                .map(|ring| rotate_many(angle, centroid, ring.points_iter()).collect())
                .collect(),
        )
    }
```

lnicola: I can't check right now and I'm not familiar with the code (there are probably some traits involved?) but it seems like you should be able to rotate e.g. an empty geometry.

mapoulos: (I have changed along the lines you suggested, but I confess I'm not familiar enough with a rust api user's expectations to know what we should do.  My first instinct is to expect parity between `Multipolygon` and `Polygon`, (i.e. both should return `Option<Point>` or both should return `Point`)

lnicola: I agree, being consistent is more important here.

mapoulos: Yeah I think your thought about the empty geometry is probably right. Looking at JTS centroid will return null in some cases (presumably for an empty geometry). Since the trait is shared across all types centroid always returns an option.

An empty multipolygon is a good example where a centroid perhaps wouldn't succeed and returning a clone of self would be suitable. I'll add a a test or two to test those edge cases (idk off hand if you can make an empty multipolygon but I'll try)

geo/src/algorithm/rotate.rs
```diff
@@ -194,7 +194,14 @@ where
 {
     /// Rotate the contained Polygons about their centroids by the given number of degrees
     fn rotate(&self, angle: T) -> Self {
-        MultiPolygon(self.iter().map(|poly| poly.rotate(angle)).collect())
+        match self.centroid() {
+            Some(centroid) => {
+                MultiPolygon(self.iter().map(|poly| poly.rotate_around_point(angle, centroid)).collect())
+            },
+            None => {
```
michaelkirk, line 201: Can you add something like `debug_assert!(self.is_empty())` to document why this could happen?

geo/src/algorithm/rotate.rs
```diff
@@ -345,4 +352,43 @@ mod test {
...
+            polygon![
+                (x: 0., y: 0.),
+                (x: -10., y: 0.),
+                (x: -10., y: 10.),
+                (x: 0., y: 10.),
+                (x: 0., y: 0.),
+            ]
+        ].into();
+
+        assert_relative_eq!(multipolygon.rotate(-90.), expected, max_relative=1.0);
```
michaelkirk, line 392: I'm curious, does this still pass if you delete the `max_relative=1.0`?

My understanding of `max_relative=1.0` is that you're allowing up to 100% error, which seems unnecessarily large.

michaelkirk (changes requested):
Thanks for the PR!

It mostly looks good, but I had some minor nits / questions.

geo/src/algorithm/rotate.rs
michaelkirk: Sorry - this was a bad suggestion. Having no members is only one of a handful of reasons a multipolygon might not have a centroid. Additionally it might have polygons, but *they* might be empty, or their rings otherwise degenerate.

Maybe worth a comment though:
```
// Multipolygon was empty or otherwise degenerate and had no computable centroid
```

geo/src/algorithm/rotate.rs
mapoulos: Removing the `max_relative` does indeed fail:

```
running 1 test
thread 'algorithm::rotate::test::test_rotate_multipolygon_around_centroid' panicked at 'assert_relative_eq!(multipolygon.rotate(-90.), expected, max_relative = 0.1)

    left  = MultiPolygon([Polygon { exterior: LineString([Coordinate { x: 0.0, y: 0.0 }, Coordinate { x: 0.0000000000000006123233995736766, y: -10.0 }, Coordinate { x: 10.0, y: -10.0 }, Coordinate { x: 10.0, y: 0.0000000000000006123233995736766 }, Coordinate { x: 0.0, y: 0.0 }]), interiors: [] }, Polygon { exterior: LineString([Coordinate { x: 0.0, y: 0.0 }, Coordinate { x: -0.0000000000000006123233995736766, y: 10.0 }, Coordinate { x: -10.0, y: 10.0 }, Coordinate { x: -10.0, y: -0.0000000000000006123233995736766 }, Coordinate { x: 0.0, y: 0.0 }]), interiors: [] }])
    right = MultiPolygon([Polygon { exterior: LineString([Coordinate { x: 0.0, y: 0.0 }, Coordinate { x: 10.0, y: 0.0 }, Coordinate { x: 10.0, y: -10.0 }, Coordinate { x: 0.0, y: -10.0 }, Coordinate { x: 0.0, y: 0.0 }]), interiors: [] }, Polygon { exterior: LineString([Coordinate { x: 0.0, y: 0.0 }, Coordinate { x: -10.0, y: 0.0 }, Coordinate { x: -10.0, y: 10.0 }, Coordinate { x: 0.0, y: 10.0 }, Coordinate { x: 0.0, y: 0.0 }]), interiors: [] }])
```

My knowledge of floating point comparison is sketchy enough to where I'm not sure what this a tuning param for this is. It's possible I've just missed something stupid tho, so I'll look more tomorrow.

michaelkirk: Ignoring the *small* errors for now, a quick glance shows the coords seem to be _way_ off. 

e.g. look at the second coord of the first poly:

```
a: Coordinate { x: 0.0000000000000006123233995736766, y: -10.0 }
b: Coordinate { x: 10.0, y: 0.0 }
```

One thing that could be going on... be aware that the equals check is _not_ a topology check — to be `eq`, the MultiPolygons can't just have the same **shape**, but also their member Polygons must be in the same order, and each members coords must be in the same order.

Maybe something going on there?

mapoulos: Thanks for the pointers. I tried to short cut my way to a rotation I thought I could do by hand but I either botched the expected or missed something subtle. I think I'll create a real example and then run it through geos to get the expected values.￼

mapoulos:
Okay, I've use shapely as an oracle for a multipolygon rotation. I still relaxed the epsilon a tad in the `assert_relative_eq!` macro call but just a tad so I can write `0.` instead of something like `0.000000000000887`.

I've also added a test to ensure an un-centroid-able multipolygon returns itself. should we do the same for `LineString` and `Polygon`? Those both panic if a centroid can't be called. For instance, this code produces a panic:

```rust
let empty_polygon: Polygon<f64> = polygon![];
let rotated_empty_polygon = empty_polygon.rotate(90.);
```

michaelkirk (approved):
bors r+

michaelkirk:
>  Those both panic if a centroid can't be calculated. For instance, this code produces a panic:

That sounds undesirable, but it seems like that's existing behavior, unaffected by this PR, so I'm fine with merging this as is.

I've added https://github.com/georust/geo/issues/652 to track the bug you uncovered. Feel free to work on it =)

michaelkirk:
> Build failed

Shoot, at first glance it looks like some failure in our testing code, unrelated to your change. 

Unfortunately I don't have more time to dig into this right now.

michaelkirk:
> Build failed

Ok, I took a quick peak actually. It *looks* like `arbitrary` a crate we use for testing updated their minimum supported rust version to be something newer than what we're running in CI.

I can take a look at revving CI sometime in the next few days.

urschrei:
@michaelkirk I should have some time over the weekend. Will ping you if I get to it.

mapoulos:
I more or less assume there's nothing that I can or should do to fix the CI, but if I can help with something let me know.

michaelkirk:
I'm loooking at CI now. It takes a while to build everything, but is pretty low effort provided everything "Just works"... I'll keep you updated.

michaelkirk:
@mapoulos - I've fixed CI. Would you mind rebasing or merging in master and we can give it another go?

mapoulos:
@michaelkirk done!

michaelkirk:
bors try

michaelkirk:
Thanks @mapoulos!

PR #654, opened 2021-07-24
concave hull: fix empty crash, plus some cleanup

michaelkirk:
---

Initially I just wanted to address: https://github.com/georust/geo/commit/243ad260ba9dec5fe6ab5d184ac625f6ed01b53b

But upon working on it I noticed I missed some things while reviewing #635 (sorry @Maximkaaa).

There's quite a few changes, but the majority of them are cosmetic. It's probably easier to review the individual commits in this case.

(please take a look @Maximkaaa)

### DIFF
--- a/geo-types/src/macros.rs
+++ b/geo-types/src/macros.rs
@@ -27,6 +27,32 @@ macro_rules! point {
     };
 }
 
+/// Creates a [`Coordinate`] from the given scalars.
+///
+/// ```txt
+/// coord!(x: <number>, y: <number>)
+/// ```
+///
+/// # Examples
+///
+/// Creating a [`Coordinate`], supplying x/y values:
+///
+/// ```
+/// use geo_types::coord;
+///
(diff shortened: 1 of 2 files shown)


geo-types/src/macros.rs
```diff
@@ -27,6 +27,32 @@ macro_rules! point {
...
+/// use geo_types::coord;
+///
+/// let c = coord!(x: 181.2, y: 51.79);
+///
+/// assert_eq!(c, geo_types::Coordinate { x: 181.2, y: 51.79 });
+/// ```
+///
+/// [`Coordinate`]: ./struct.Point.html
+#[macro_export]
+macro_rules! coord {
```
michaelkirk, line 50: Also worth calling out that I added a `coord!` macro, just like `point!`.. I guess the advantage over `Coordinate { x: a, y: b }` is only brevity.

Maximkaaa:
Yes, it all makes sense. Thanks.

PR #656, opened 2021-07-24
Gracefully handle rotation of empty linestring and polygon

### LINKED ISSUE #652: panic when computing centroid of some degenerate geometries
from https://github.com/georust/geo/pull/651#issuecomment-880768328

> [LineString and Polygon] both panic if a centroid can't be calculated. For instance, this code produces a panic:
>
> ```
> let empty_polygon: Polygon<f64> = polygon![];
> let rotated_empty_polygon = empty_polygon.rotate(90.);
> ```

Probably they should return a `None` centroid instead of exploding.

mapoulos:
---

Resolves #652

### DIFF
--- a/geo/src/algorithm/rotate.rs
+++ b/geo/src/algorithm/rotate.rs
@@ -164,7 +164,16 @@ where
     fn rotate(&self, angle: T) -> Self {
         match self.centroid() {
             Some(centroid) => rotate_many(angle, centroid, self.points_iter()).collect(),
-            None => self.clone(),
+            None => {
+                // LineString was empty and had no computable centroid
+                let points = self.clone().into_points();
+                debug_assert!(
+                    points.len() == 0,
+                    "Expected `None` centroid to only occur for an empty LineString, but this LineString has {} points", 
+                    points.len()
+                );
+                self.clone()
+            }
         }
     }
 }
--- a/geo/src/algorithm/rotate.rs
+++ b/geo/src/algorithm/rotate.rs
@@ -162,7 +162,10 @@ where
 {
     /// Rotate the LineString about its centroid by the given number of degrees
     fn rotate(&self, angle: T) -> Self {
-        rotate_many(angle, self.centroid().unwrap(), self.points_iter()).collect()
+        match self.centroid() {
+            Some(centroid) => rotate_many(angle, centroid, self.points_iter()).collect(),
+            None => self.clone(),
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


geo/src/algorithm/rotate.rs
```diff
@@ -400,9 +410,23 @@ mod test {
     }
 
     #[test]
-    fn test_rotate_multipolygon_errors_gracefully() {
-        // an multipolygon whose centroid cannot be found should return itself, rather
-        // than panicing (e.g. an empty multipolygon, or a multipolygon composed of empty polygons)
+    fn test_rotate_empty_geometries_error_gracefully() {
```
mapoulos, line 413: If I should also test some other degenerate geometries, I can add another test case or two.

lnicola (approved):

geo/src/algorithm/rotate.rs
```diff
@@ -162,7 +162,10 @@ where
 {
     /// Rotate the LineString about its centroid by the given number of degrees
     fn rotate(&self, angle: T) -> Self {
-        rotate_many(angle, self.centroid().unwrap(), self.points_iter()).collect()
+        match self.centroid() {
+            Some(centroid) => rotate_many(angle, centroid, self.points_iter()).collect(),
+            None => self.clone(),
```
michaelkirk, line 167: This is only for the case of an empty linestring right?

Can you add a debug_assert or a comment to document this? Otherwise it might look like we're ignoring errors due to oversight.

geo/src/algorithm/rotate.rs
```diff
@@ -174,17 +177,24 @@ where
     fn rotate(&self, angle: T) -> Self {
         // if a polygon has holes, use the centroid of its outer shell as the rotation origin
```
michaelkirk, line 178: I realize this predates your work, but I'm surprised by the behavior documented in this comment. I would have guessed we're always rotating about the centroid (including the holes). @urschrei - do you recall why we do it this way?

geo/src/algorithm/rotate.rs
michaelkirk: Thanks for the tests!

michaelkirk (changes requested):
I have a question about the behavior of rotating a polygon-with-holes, but that's existing behavior, so I'd be happy to merge provided you can add a little documentation or debug_assert in LineString::rotate.

geo/src/algorithm/rotate.rs
urschrei: My commit comment (https://github.com/georust/geo/commit/dd345c5491a09b3e45f3c08c0bc3a78f7d627bb9) isn't super informative, but I assume it's because holes shift the centroid and that's not what you want when rotating. I'll have to confirm my suspicion that Shapely / GEOS do this, and I copied them, however.

michaelkirk: > it's because holes shift the centroid and that's not what you want when rotating.

I brought this up because I think rotating around the geometry's centroid (accounting for holes - the "real" centroid) in all cases is what I'd want and expect. 

I think primarily because this is a concise and consistent explanation of what the method does.

I think I see what you're saying about how some people might want to rotate around something more like a bbox center, rather than a centroid, but this "exterior-only centroid" concept seems a little uncommon.


From shapely, their rotate method takes a "point of origin" argument with docs:

>  The point of origin can be a keyword 'center' for the bounding box center (default), 'centroid' for the geometry’s centroid, a Point object or a coordinate tuple (x0, y0).

I'd love to steal this design, though because we don't have default args the api might be a little different... something like:

```
trait Rotate {
  #[deprecate(note="use more specific rotate_around_centroid or rotate_around_center")]
    rotate_around_centroid(angle)
  }

  fn rotate_around_coord(&self, coord: Coordinate<T>, angle: T) -> Self;

  /// equivalent to:
  /// ```
  /// let center = g.bounding_rect().center();
  /// g.rotate_around_coord(center, angle)
  /// ```
  fn rotate_around_center(&self, angle: T) -> Self;

  /// ```
  /// let centroid = g.centroid();
  /// g.rotate_around_coord(centroid, angle)
  /// ```
  fn rotate_around_centroid(&self, angle: T) -> Self;
}
```

What would you think of something like that?

urschrei: I am def. in favour. I just had a look to see whether we could do something tricky with an enum and default value (center), but it seems a bit too messy atm – separate methods is a cleaner design.

michaelkirk: Ok, I'm gonna make a new issue with these details, linking to this discussion, and give @mapoulos their PR back. 😅

geo/src/algorithm/rotate.rs
mapoulos: I've added a comment: happy to convert these to `debug_asserts` if we want to make it more prominent.

michaelkirk:
thanks!

bors r+

PR #658, opened 2021-07-31
[WIP] Rotate centroid and center

### LINKED ISSUE #657: clarify Rotate trait: center vs. centroid
Per some discussion in https://github.com/georust/geo/pull/656/files#r676937398, the expected behavior of Rotate isn't super clear.

Leaning on shapely, we came upon this more explicit and flexible api:

```
trait Rotate {
  // this is the existing method. Existing implementations vary, but most rotate around the geometries centroid.
  // let's deprecate it, and (I guess?) have it call centroid since that's what most impls currently do. 
  #[deprecate(note="equivalent to `rotate_around_centroid`, call that instead, or `rotate_around_center` if you'd like to rotate around the geometry's bounding box center")]
    self.rotate_around_centroid(angle)
  }

  fn rotate_around_coord(&self, coord: Coordinate<T>, angle: T) -> Self;

  fn rotate_around_center(&self, angle: T) -> Self {
    let center = g.bounding_rect().center();
    g.rotate_around_coord(center, angle)
  }

  fn rotate_around_centroid(&self, angle: T) -> Self {
    let centroid = self.centroid();
    self.rotate_around_coord(centroid, angle)
  }
}
```

mapoulos:
---

Resolves #657 

Use shapely as source of truth for `rotate`.

Tests:
- [ ] Point rotate_around_center / centroid
- [ ] MultiPoint rotate_around_center
- [ ] MultiPoint rotate_around_centroid
- [ ] Line rotate_around_center
- [ ] Line rotate_around_centroid
- [ ] LineString rotate_around_center
- [ ] LineString rotate_around_centroid
- [ ] MultiLineString rotate_around_center
- [ ] MultiLineString rotate_around_centroid
- [ ] Polygon rotate_around_center
- [ ] Polygon rotate_around_centroid
- [ ] MultiPolygon rotate_around_center
- [ ] MultiPolygon rotate_around_centroid

### DIFF
--- a/geo/src/algorithm/rotate.rs
+++ b/geo/src/algorithm/rotate.rs
@@ -281,8 +281,26 @@ impl<T> Rotate<T> for MultiPolygon<T>
 where
     T: GeoFloat,
 {
-    /// Rotate the contained Polygons about their centroids by the given number of degrees
     fn rotate(&self, angle: T) -> Self {
+        self.rotate_around_centroid(angle)
+    }
+
+    /// Rotate the MultiPolygon about the center of its bounding rectangle by the given number of degrees
+    fn rotate_around_center(&self, angle: T) -> Self {
+        match self.bounding_rect() {
+            Some(bounding_rect) => {
+                MultiPolygon(
+                   self.iter()
+                        .map(|poly| poly.rotate_around_point(angle, bounding_rect.center().into()))
+                        .collect() 
+                )
+            },
+            None => self.clone(), // MultiPolygon was empty or otherwise degenerate and had no computable center
+        }
+    }
+
+    /// Rotate the contained Polygons about their centroids by the given number of degrees
+    fn rotate_around_centroid(&self, angle: T) -> Self {
         match self.centroid() {
             Some(centroid) => MultiPolygon(
                 self.iter()
@@ -291,7 +309,7 @@ where
             ),
             None => {
                 // Multipolygon was empty or otherwise degenerate and had no computable centroid
-                self.clone()
+                self.clone() 
             }
         }
     }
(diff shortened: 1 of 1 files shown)


mapoulos:
This is now ready for review

geo/src/algorithm/rotate.rs
```diff
@@ -141,27 +155,51 @@ where
...
             rotate_one(angle, centroid, self.start_point()),
             rotate_one(angle, centroid, self.end_point()),
         )
     }
+
+    fn rotate_around_center(&self, angle: T) -> Self {
+        let center: Point<T> = self.bounding_rect().center().into();
+        Line::new(
+            rotate_one(angle, center, self.start_point()),
+            rotate_one(angle,center, self.end_point()),
```
lnicola, line 188: Nit:

```suggestion
            rotate_one(angle, center, self.end_point()),
```

geo/src/algorithm/rotate.rs
```diff
@@ -215,7 +312,7 @@ where
             ),
             None => {
                 // Multipolygon was empty or otherwise degenerate and had no computable centroid
-                self.clone()
+                self.clone()
```
lnicola, line 315:
```suggestion
                self.clone()
```

lnicola:
Looks good on a quick pass, but I'd run `cargo fmt`.

mapoulos:
Format on save needs to be in my vimrc lol. Done.

mapoulos:
Bumping this: anything else we need to address before merging?

rmanoka:
@mapoulos Thanks for your effort!  Would you be willing to also incorporate the following clean-ups?

1. In some places `GeoFloat` is used instead of `CoordFloat`, but I think it is not necessary.  We don't really use the `Kernel`, so just `CoordFloat` should do the job.
2. I think the logic could be simplified a bit more by using `RotatePoint` trait.  Once you compute either the centroid or the center of the bounding rect, you can just delegate to `RotatePoint` trait to rotate the geometry.
3. You could make it more succinct by creating a macro to handle the two different cases:  centroid is `Option<Point>`, and just `Point` as logic is the same for all the types.  Unfortunately, a blanket impl like for `RotatePoint` doesn't seem to work as the type-system doesn't seem to understand the same geometry `G` can't both be `Centroid<Output = Point<T>>` and `Centroid<Output = Option<Point<T>>>`;  two separate macros to handle the two cases should work though.

I think 1 is a fairly simple change.  Feel free to skip 2 and 3 if it's too much work; it is mainly to streamline the code bit, and avoid redundant logic.  The PR looks good to me otherwise.

mapoulos:
I'll have a crack at 1 at least this next week. I probably won't have time for 2 or 3 right now but I'll poke around a bit first. 

Can you point me to an example of delegating behavior to a trait? I'm relatively new to Rust and haven't mastered the intricacies of the type system or the standard patterns yet.

rmanoka:
@mapoulos By delagating to `RotatePoint`, I mean just use that trait in the geometry.  So use the `geom.rotate_around_point` method once you have calculated the center / centroid;  this will remove the repetitive code to rotate the components and re-assemble the geometry as that is already being done in the other traits.

For macro, you can check out the `macro_rules` declarations used in the source-code.  A bit of macro basics would be needed though.  We could guide you on it if you get started.

rmanoka:
Thanks for the clarification @mapoulos .  The pr looks good to me; thanks for your patience and the effort.

I'll merge it in 2 days unless anyone else has some comments.

PR #659, opened 2021-08-15
Add Into implementation for Coordinate and Point

lehmanju:
This PR adds two implementations of the `Into` trait for `Coordinate` and `Point`.

As these structs can already be constructed from tuples and arrays it is expected that they can also be converted back.

### DIFF
--- a/geo-types/src/coordinate.rs
+++ b/geo-types/src/coordinate.rs
@@ -69,6 +69,18 @@ impl<T: CoordNum> From<Point<T>> for Coordinate<T> {
     }
 }
 
+impl<T: CoordNum> From<Coordinate<T>> for (T, T) {
+    fn from(coord: Coordinate<T>) -> Self {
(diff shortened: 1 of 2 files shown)


lnicola:
It's [recommended](https://doc.rust-lang.org/std/convert/trait.Into.html) to implement `From` instead of `Into`.

urschrei:
#660 bumps rstar, let's keep them separate.

lehmanju:
sorry, that commit was not supposed to be merged

michaelkirk:
> It's recommended to implement From instead of Into. I think it's allowed in this case under the coherence rules.

@lehmanju - is there a reason you can't implement the `From` instead? As @lnicola said, that's preferable because then we get `Into` for free.

lehmanju:
No reason, done now.

michaelkirk:
Thanks!

bors r+

PR #660, opened 2021-08-15
cargo: bump rstar dependency

QuentinPerez:
---

Hey 👋 
A small PR  to update rstar and fix the possible conflit between geo-types definitions.

```rust
// Cargo.toml
// ...
// geo = "0.18"
// geo-types = { version = "0.7", features = ["use-rstar"] }
// rstar = "0.9"

impl From<XXX> for AABB<geo::Point<f32>> {
                   ^^^^^^^^^^^^^^^^^^^ the trait `rstar::Point` is not implemented for `geo::Point<f32>`
        ....
}
```

### DIFF
--- a/geo-types/src/coordinate.rs
+++ b/geo-types/src/coordinate.rs
@@ -309,7 +309,7 @@ where
 
     const DIMENSIONS: usize = 2;
 
-    fn generate(generator: impl Fn(usize) -> Self::Scalar) -> Self {
+    fn generate(mut generator: impl FnMut(usize) -> Self::Scalar) -> Self {
         Coordinate {
             x: generator(0),
             y: generator(1),
--- a/geo-types/CHANGES.md
+++ b/geo-types/CHANGES.md
@@ -1,5 +1,10 @@
 # Changes
 
+
+## Unreleased
+
+* Bump `rstar` dependency
+
 ## 0.7.2
 
 * Implement `RelativeEq` and `AbsDiffEq` for fuzzy comparison of remaining Geometry Types
--- a/geo-types/Cargo.toml
+++ b/geo-types/Cargo.toml
@@ -21,7 +21,7 @@ serde = { version = "1", optional = true, features = ["derive"] }
 # Prefer `use-rstar` feature rather than enabling rstar directly.
 # rstar integration relies on the optional approx crate, but implicit features cannot yet enable other features.
 # See: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#namespaced-features
-rstar = { version = "0.8", optional = true }
+rstar = { version = "0.9", optional = true }
 
 [dev-dependencies]
 approx = "0.4.0"
--- a/geo-types/src/point.rs
+++ b/geo-types/src/point.rs
@@ -486,7 +486,7 @@ where
 
     const DIMENSIONS: usize = 2;
 
-    fn generate(generator: impl Fn(usize) -> Self::Scalar) -> Self {
+    fn generate(mut generator: impl FnMut(usize) -> Self::Scalar) -> Self {
         Point::new(generator(0), generator(1))
     }
 
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* Bump `rstar` dependency
 * Add `ChaikinSmoothing` algorithm
 * Fix `rotate` for multipolygons to rotate around the collection's centroid, instead of rotating each individual polygon around its own centroid.
   * <https://github.com/georust/geo/pull/651>
--- a/geo/Cargo.toml
+++ b/geo/Cargo.toml
@@ -17,7 +17,7 @@ travis-ci = { repository = "georust/geo" }
 [dependencies]
 num-traits = "0.2"
 serde = { version = "1.0", optional = true, features = ["derive"] }
-rstar = { version = "0.8" }
+rstar = { version = "0.9" }
 geographiclib-rs = { version = "0.2" }
 log = "0.4.11"
 


lnicola (approved):

geo-types/src/coordinate.rs
```diff
@@ -309,7 +309,7 @@ where
 
     const DIMENSIONS: usize = 2;
 
-    fn generate(generator: impl Fn(usize) -> Self::Scalar) -> Self {
+    fn generate(mut generator: impl FnMut(usize) -> Self::Scalar) -> Self {
```
michaelkirk, line 312: calling all semver lawyers: is this considered a breaking change for geo-types?

lnicola: I don't think so. It's more general and if there was a way for the argument type to leak into the outside code (*), type inference changes are generally considered benign.

(*) E.g. `x` here gets inferred as `()` by `rustc`:

```rust
fn f(g: impl Fn()) {}

fn main() {
    let x;
    f(x);
    x = todo!();
}
```

lnicola: Actually, no, it's a breaking change because `rstar::Point` is already in our public API. So if user code depends on `rstar 0.8`, they'll have to upgrade to `0.9`.

michaelkirk: Is this an argument for moving the `geo-types` trait impls to the rstar crate?

michaelkirk: Just that it's a real pain to bump geo-types, since so many depend on it.

lnicola: > Is this an argument for moving the geo-types trait impls to the rstar crate?

That's a pretty sensible idea.

rmanoka:
Summarizing the discussions so we could resolve it.  rstar has some nice additions in 0.9 that are particularly useful with geo-types (like `GeomWithData`).

The proposal is to move the rstar trait impls for geo-types into the `rstar` crate itself.

## Pros

1. No breaking change to keep up with breaking changes in rstar.

## Cons

1. rstar isn't really dependent on geo-types, so it's an artificial dependency, that should probably use a feature-gate.
2. Breaking changes here would then have to be maintained in rstar

@rmanoka 's suggestion:  I think pros > cons as `rstar` is an algo crate, and is expected to have more breaking changes than geo-types.   To signal that `geo-types` is fairly stable, could we agree to release a 1.0.0 version of it?  Cons.1 is somewhat unfortunate, but it should be okay.

Pinging a few rstar contributors (and author): @urschrei @adamreichold @Stoeoef

adamreichold:
>  I think pros > cons

I agree.

> rstar isn't really dependent on geo-types, so it's an artificial dependency, that should probably use a feature-gate.

Indeed, I would expect geo-types to become an optional dependency of rstar.

urschrei:
> I think pros > cons

I also agree. I'm not feeling any anxiety about a `geo-types` 1.0 release.

michaelkirk:
Thanks for pushing the conversation forward @rmanoka - I kind of let it stall out =/

I support either or both of these:

> 1. move rstar trait impls for geo-types into the rstar crate itself.

> 2. To signal that geo-types is fairly stable, could we agree to release a 1.0.0 version of it

adamreichold:
>  We can use features to support different rstar versions without breaking changes

Personally, I would prefer the inverted dependency where the concrete library (rstar) depends on the abstraction (geo-types) and avoiding the extra features.

I also think that in general, it is best if the ecosystem moves forward in coordinated manner instead of supporting multiple incompatible versions for an indefinite amount of time.

rmanoka:
> Personally, I would prefer the inverted dependency where the concrete library (rstar) depends on the abstraction (geo-types) and avoiding the extra features.

This is counter-intuitive to me.  `rstar` doesn't even need to know the existence of `geo-types`.  It is really the users of `geo-types` (eg. `geo`) that are attempting to use the `rstar` functionalities.

Further, this would also imply it's natural to add impls for `nalgebra::Vec*`, and `cgmath::Vec*`.  But then we're back to square one as we'll need to publish breaking releases whenever any of these crates make one.

> I also think that in general, it is best if the ecosystem moves forward in coordinated manner instead of supporting multiple incompatible versions for an indefinite amount of time.

While this is ideal, it's hard to coordinate across a large eco-system.

adamreichold:
>  It is really the users of geo-types (eg. geo) that are attempting to use the rstar functionalities.

Yes, and their rstar version could accommodate them using an optional dependency similar to how a lot of crates have an optional dependency on serde to provide a default serialization of their types. Of course, this assumes that the abstraction, geo-types resp. serde, is much more stable than the concrete library. And in the case of serde, this is often required as the serialization needs to access internals of the data structures.

The serde example also highlights a downside of this approach: It provides a default serialization variant which might not be the one a downstream user needs or at least prefers. I think the most general approach here would be to put the integration into neither crate but provide a separate "rstar-geo-types-int" crate which would depend on both crates and contain nothing but the trait impls. But I also think that the mental overhead of that is probably not worth it.

> While this is ideal, it's hard to coordinate across a large eco-system.

I would argue in good faith that all of us are currently in the process of trying to do so?

rmanoka:
> Yes, and their rstar version could accommodate them using an optional dependency similar to how a lot of crates have an optional dependency on serde to provide a default serialization of their types. Of course, this assumes that the abstraction, geo-types resp. serde, is much more stable than the concrete library. And in the case of serde, this is often required as the serialization needs to access internals of the data structures.

Yeah, `rstar` traits are similar to `serde` in that diff. crates `geo-types`, `nalgebra`, etc. can impl. it so their users can use `RTree`s on the structs.  This is why it's natural to impl the traits in the respective crates.

> 
> The serde example also highlights a downside of this approach: It provides a default serialization variant which might not be the one a downstream user needs or at least prefers. I think the most general approach here would be to put the integration into neither crate but provide a separate "rstar-geo-types-int" crate which would depend on both crates and contain nothing but the trait impls. But I also think that the mental overhead of that is probably not worth it.

Rust doesn't allow trait impls outside the crate defining the trait or the struct (which is the whole issue).

adamreichold:
> Yeah, rstar traits are similar to serde in that diff. crates geo-types, nalgebra, etc. can impl. it so their users can use RTrees on the structs. This is why it's natural to impl the traits in the respective crates.

While it is rstar's traits and serde's traits are syntactically similar, I think this is mixing up their roles: Both serde and geo-types are vocabulary crates which provide abstractions that serve as the glue between more concrete data structures and algorithms.

PR #665, opened 2021-10-12
Clippy: Remove .clone() where possible.

martinfrances107:
---

Looking at the output of `cargo clippy`

I want to cherry pick the single aspect I find most troublesome. 
In general it is better to use copy over clone.
It has been fixed in a couple of places.


lnicola:
Yeah, `assert_relative_eq` is already used in other places. I think it was an oversight.

martinfrances107:
Ok here is the new issue... I will create a patch later today 

#666

michaelkirk:
bors retry

PR #667, opened 2021-10-13
map_coords.rs: Use assert_relative_eq!() for floats.

martinfrances107:
- [ x] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/master/CODE_OF_CONDUCT.md).
---
Uniformly use assert_releative_eq!() when comparing floats.

### DIFF
--- a/geo/src/algorithm/map_coords.rs
+++ b/geo/src/algorithm/map_coords.rs
@@ -2,6 +2,7 @@
 //!
 //! ```
 //! // activate the [use-proj] feature in cargo.toml in order to access proj functions
+//! use approx::assert_relative_eq;
 //! # #[cfg(feature = "use-proj")]
 //! use geo::{Coordinate, Point};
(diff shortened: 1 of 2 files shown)


lnicola:
Looks like we need a higher tolerance here:

```
---- algorithm::map_coords::test::test_fallible_proj stdout ----
thread 'algorithm::map_coords::test::test_fallible_proj' panicked at 'assert_relative_eq!(3497301.5918027186, usa_ft.y())

    left  = 3497301.5918027186
    right = 3497301.5918027232

', geo/src/algorithm/map_coords.rs:840:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

martinfrances107:
thanks for the prompt review... it is appreciated.

Now using this pattern which I see exists in the codebase.

` assert_relative_eq!(p2, Point::new(10.0f64, 20.0f64), epsilon = 1e-6);
`

PR #671, opened 2021-10-14
New Feature "use-approx".

martinfrances107:
- [x ] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/master/CODE_OF_CONDUCT.md).
- [x ] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

Exports various implementations of the traits "AbsDiffEq" and "RelativeEq" which were previously only visible in test build, so they can also be used in production code.

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* New Feaure 'use-approx' exports the implementation of the traits AbsDiffEq and RelativeEq
+  for various geometry types, for use in third party crates.
 * Add `ChaikinSmoothing` algorithm
 * Fix `rotate` for multipolygons to rotate around the collection's centroid, instead of rotating each individual polygon around its own centroid.
   * <https://github.com/georust/geo/pull/651>
--- a/geo/Cargo.toml
+++ b/geo/Cargo.toml
@@ -18,6 +18,7 @@ travis-ci = { repository = "georust/geo" }
 num-traits = "0.2"
 serde = { version = "1.0", optional = true, features = ["derive"] }
 rstar = { version = "0.8" }
+approx = { version = "0.4.0", optional= true }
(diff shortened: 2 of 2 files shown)


lnicola:
I'm not sure I'm a fan of this. It will probably bump the CI time to one hour.

bors try

martinfrances107:
Time is not a factor for me... I have my own private branch.

I am using it out in the wild. [rust_d3_geo](https://github.com/martinfrances107/rust_d3_geo)

I would just like to keep the number of deviations a minimum.

I can't quite see how do somthing local in my third party without running into the security isolations problems.  "The trait, the struct and the implementation cannot be in more than two distinct  modules."

https://github.com/martinfrances107/rust_d3_geo/blob/main/src/clip/rejoin/mod.rs

```
            if p0.p.abs_diff_eq(&p1.p, T::from(1e-6).unwrap()) {
                if p0.m.is_none() && p1.m.is_none() {
                    stream_b.line_start();
                    // for i in 0..n {
                    for elem in segment.iter().take(n) {
                        p0 = *elem;
                        stream_b.point(&p0.p, None);
                    }
                    stream_b.line_end();
                    return;
                }
                // handle degenerate cases by moving the point
                p1.p.x = p1.p.x + T::from(2.0 * 1e-6).unwrap();
            }

```

michaelkirk:
> Instead of this PR, would it suffice for you to use the existing optional approx integration on geo-types?

@martinfrances107 - I'm still curious about this. Why do we need this feature if there's already an approx integration on geo-types?

michaelkirk:
> @martinfrances107 - I'm still curious about this. Why do we need this feature if there's already an approx integration on geo-types?

I'm going to close this since it's sat here for a while without response from @martinfrances107.

Feel free to explain and we can re-open it.

PR #673, opened 2021-10-16
Simplify CI testing matrix

### LINKED ISSUE #672: RFC: Stop testing feature combinations exhaustively
It looks like the tests take about 12 minutes or so, but the CI workflow uses `cargo-all-features`, which checks all combinations of features. We currently have three of them, which means that the tests (probably) run 8 times.

Rust features are supposed to be additive. I realize this isn't always the case (e.g. the PROJ `network` feature might influence the results), but there's not that much place for strange interactions between them:

 - `use-proj` only re-exports some `proj` types AFAICT
 - `proj-network` only enables a `proj` feature; this could presumably influence some (doc-)tests though
 - `serde` only enables the `serde` integration

So I think it would be relatively safe to only run `cargo check --no-default-features` and `cargo test --all-features`, and maybe trim down the tests where `proj-network` could have an impact.

lnicola:
---

Closes #672

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,7 +60,8 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v2
       - run: cargo check --all-targets --no-default-features
-      - run: cargo test --all-features
+      # we don't want to test `proj-network` because it only enables the `proj` feature
+      - run: cargo test --features use-proj use-serde
 
   geo_postgis:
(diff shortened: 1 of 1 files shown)


lnicola:
Updated.

rmanoka:
Lgtm!  Let's see if anyone with more knowledge of the proj integration has some comments.  Ping @michaelkirk @urschrei .

Issue #544, opened 2020-11-22
Implement ClosestPoint for Geometry


frewsxcv:
We already have ClosestPoint implemented for most geometry types, so let's implement it for the `Geometry` enum

michaelkirk:
> I'd like to take a crack at this. Is the basic idea to add an impl for the Geometry enum under geo/src/algorithm/contains/geometry.rs?

Yep, that's the idea!

> I'm not very familiar with macros, but would something like make sense?

You can use `geometry_delegate_impl!` if it's helpful.  I think it would look something more like:
```
impl<F: GeoFloat> ClosestPoint<F> for Geometry<F> {
    crate::geometry_delegate_impl! {
        fn closest_point(&self, p: &Point<F>) -> Closest<F>;
    }
}
````

In any case, you'll have to make sure that all the inner variants of Geometry `impl ClosestPoint`. I believe most do, but haven't checked that all of them do.

See also the [`geometry_delegate_impl` documentation](https://github.com/georust/geo/blob/master/geo/src/types.rs#L40)

And an example of how it's used elsewhere in geo: https://github.com/georust/geo/blob/master/geo/src/algorithm/bounding_rect.rs#L157

pkopparla:
Thank you @michaelkirk !
I've added implementations for Triangle, Rectangle, GeometryCollection and Geometry. I'm still a little confused on how the latter two work, it feels like GeometryCollection and Geometry have a circular dependency on each other. But I will make a pull request now so you guys can have a look.

callpraths:
I think this issue can be closed. #675 addressed both `Geometry` and `GeometryCollection`:

https://docs.rs/geo/latest/geo/enum.Geometry.html#impl-ClosestPoint%3CF%2C%20Point%3CF%3E%3E
https://docs.rs/geo/latest/geo/struct.GeometryCollection.html#impl-ClosestPoint%3CF%2C%20Point%3CF%3E%3E

I don't think there's any more work left here.

michaelkirk:
Indeed, thanks!

PR #675, opened 2021-11-02
ClosestPoint for Triangle, Rect, GeometryCollection types & Geometry enum

pkopparla:
---

This pull request addresses issue #544 
The ClosestPoint method was implemented for the remaining members of the `Geometry` enum ie., `Triangle, Rect and GeometryCollection` and then implemented for the enum itself.

### DIFF
--- a/geo/src/algorithm/closest_point.rs
+++ b/geo/src/algorithm/closest_point.rs
@@ -141,7 +141,7 @@ impl<F: GeoFloat> ClosestPoint<F> for Triangle<F> {
         if self.intersects(p) {
             return Closest::Intersection(*p);
         }
-        closest_of(self.to_lines(), *p)
+        closest_of(&self.to_lines(), *p)
     }
 }
 
@@ -150,7 +150,7 @@ impl<F: GeoFloat> ClosestPoint<F> for Rect<F> {
         if self.intersects(p) {
             return Closest::Intersection(*p);
         }
-        closest_of(self.to_lines(), *p)
+        closest_of(&self.to_lines(), *p)
     }
 }
 
--- a/geo/src/algorithm/closest_point.rs
+++ b/geo/src/algorithm/closest_point.rs
@@ -1,7 +1,8 @@
+use crate::coords_iter::CoordsIter;
 use crate::prelude::*;
 use crate::{
-    Closest, GeoFloat, Line, LineString, MultiLineString, MultiPoint, MultiPolygon, Point, Polygon, 
-    Rect, Triangle, Geometry, GeometryCollection,
+    Closest, Coordinate, GeoFloat, Geometry, GeometryCollection, Line, LineString, MultiLineString,
+    MultiPoint, MultiPolygon, Point, Polygon, Rect, Triangle,
 };
 use std::iter;
 
@@ -122,15 +123,21 @@ impl<F: GeoFloat> ClosestPoint<F> for Polygon<F> {
     }
 }
 
+impl<F: GeoFloat> ClosestPoint<F> for Coordinate<F> {
+    fn closest_point(&self, p: &Point<F>) -> Closest<F> {
+        Point(*self).closest_point(p)
+    }
+}
+
 impl<F: GeoFloat> ClosestPoint<F> for Triangle<F> {
     fn closest_point(&self, p: &Point<F>) -> Closest<F> {
-        self.to_polygon().closest_point(p)
+        closest_of(self.coords_iter(), *p)
     }
 }
 
 impl<F: GeoFloat> ClosestPoint<F> for Rect<F> {
     fn closest_point(&self, p: &Point<F>) -> Closest<F> {
-        self.to_polygon().closest_point(p)
+        closest_of(self.coords_iter(), *p)
     }
 }
 
--- a/geo/src/algorithm/closest_point.rs
+++ b/geo/src/algorithm/closest_point.rs
@@ -1,6 +1,7 @@
 use crate::prelude::*;
 use crate::{
-    Closest, GeoFloat, Line, LineString, MultiLineString, MultiPoint, MultiPolygon, Point, Polygon,
+    Closest, GeoFloat, Line, LineString, MultiLineString, MultiPoint, MultiPolygon, Point, Polygon, 
+    Rect, Triangle, Geometry, GeometryCollection,
 };
 use std::iter;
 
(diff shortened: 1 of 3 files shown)


geo/src/algorithm/closest_point.rs
```diff
@@ -121,6 +122,18 @@ impl<F: GeoFloat> ClosestPoint<F> for Polygon<F> {
     }
 }
 
+impl<F: GeoFloat> ClosestPoint<F> for Triangle<F> {
+    fn closest_point(&self, p: &Point<F>) -> Closest<F> {
+        self.to_polygon().closest_point(p)
```
michaelkirk, line 127: This looks like it would behave correctly, but a minor improvement request:

Creating a polygon here, will allocate a Vec (two of them actually), please instead do something like: 

`closest_of(self.coords_iter())` which will use the [CoordsIter](https://docs.rs/geo/0.18.0/geo/algorithm/coords_iter/trait.CoordsIter.html) trait without necessarily allocating on the heap.

michaelkirk (changes requested):

michaelkirk:
A couple other housekeeping things:
- please update CHANGES.md 
- please run `cargo fmt`

geo/src/algorithm/closest_point.rs
pkopparla: Thanks for the feedback @michaelkirk 
I tried this, but there's no `ClosestPoint` method for the type `Coordinate`. And it looks like `Coordinate` will need to be converted to `Point` to implement that method, in one way or another. Would that also cause a heap allocation?

michaelkirk: Ah right, good point. Converting  a `Coordinate` to a `Point` will not require a heap allocation.

michaelkirk: It's possible that this will be a larger change. I'd also be OK with merging the performance improvement in a followup since it won't affect the API. Up to you!

pkopparla: Thanks Michael.
I've added a `ClosestPoint` for `Coordinate`, switched `Triangle` and `Rect` to use `self.iter()` and updated the `CHANGES.md`.

pkopparla: I'm new to Rust, so I'm not sure if using `*self` for the `Coordinate` to `Point` transform will cause permanent type changes. Is this way of doing it alright?

michaelkirk: > 've added a ClosestPoint for Coordinate, switched Triangle and Rect to use self.iter() and updated the CHANGES.md.

The change log and the new Coordinate implementation look good, thank you!

But there's a problem of my own making: we're trying to find the point nearest to a Rect or Triangle by using only its corners. This will give incorrect results whenever the nearest point is mid line. Sorry for suggesting you go down that path, that was my fault.

So two alternative ways forward I see:
1. revert to using to_poly as you originally wanted.
2. create an implementation for Rect/Tri  that builds Lines and uses the `impl ClosestPoint for Line` to determine the closest point - this looks like it should be correct for endpoints and midline points.

Alternative 1 is simpler to write, but alternative 2 will avoid the heap allocations (and should be correct 🙈).

Also, it might be worthwhile to include a couple more test cases to catch future bad suggestions.

> I'm new to Rust, so I'm not sure if using *self for the Coordinate to Point transform will cause permanent type changes. Is this way of doing it alright?

Looks good to me. This is part of rust ownership system. `self` is a reference (you can see in the signature `&self`), but Point needs an owned `Coordinate`, not a borrowed coordinate (`&Coordinate`), so we de-reference the borrow to the owned type. Because `Coordinate` implements the `Copy` trait, rather than moving/consume `self`, a new copy of self will be made. Since copying could in theory be expensive, typically Copy is only implemented for trivial types, whereas for more complicated types (especially anything doing heap allocation) you'll have to explicitly call "thing.clone()" to get a copy.

pkopparla: Ahahaha! A seemingly easy task that grows in complexity is a nice step-by-step tutorial to learning rust. 
I'll try to do it the right way, should have something next week. Thank you also for the explanation of ownership.

michaelkirk: I was thinking about this more and realized our current implementation `for Polygon` is incorrect. For 2D geometries, we should considering points *inside* the geometry as an Intersection.

Since it's a correctness thing, I wanted to address it right away and opened a PR here: https://github.com/georust/geo/pull/679

michaelkirk: Assuming that gets merged, you should merge or rebase to incorporate that fix into your work next week.

lnicola:
Still a conflict on the changelog.

geo/src/algorithm/closest_point.rs
pkopparla: Will do. Another question there: If we use Using only lines to determine `ClosestPoint`, wouldn't a similar interior issue affect `Triangle` and `Rect`?

pkopparla:
Thanks @lnicola . This implementation is still not finalized, I will resolve conflicts when it is ready to be merged.

geo/src/algorithm/closest_point.rs
michaelkirk: Yes exactly. You''ll have to include a similar intersection check like in #679.

pkopparla: @michaelkirk I'm trying to understand: how do we determine the interior for intersection if `Triangle` and `Rect` are a set of lines joining the vertices? It's so much easier if it's converted to a polygon. What do you think?

michaelkirk: The intersection trait is implemented for Triangles and Rects too.

pkopparla:
@michaelkirk I've changed the implementation to use `to_lines` now, how about this?

michaelkirk:
The implementation LGTM! 

If you can fix up the conflict in the changelog, I think this should be good to merge.

pkopparla:
I think that should've fixed the changelog now. Thanks for reviewing @michaelkirk !

lnicola:
Still needs a rebase, I think.

pkopparla:
@lnicola I rebased it on `georust/master` before the last commit, would you mind explaining what I should now?

lnicola:
bors try

urschrei:
bors try-

urschrei:
bors try

lnicola:
bors retry

?

urschrei:
Oh ffs. This is because cargo shipping with rust < 1.56.0 doesn't know what to do with the 2021 feature, isn't it?

urschrei:
`pdqselect` 0.1.1 (an `rstar` dep) has switched to 2021, which looks to be the cause here.

urschrei:
This isn't fixable in the PR: it'll require someone to pin `pdqselect 0.1.0` in `rstar` and push out a point release, I think. We can't even fix it in our CI because there's only one stable 2021 release.

lnicola:
Does `pdqselect` even have a public repository?

urschrei:
> Does pdqselect even have a public repository?

Not that I know of. I had to go through docs.rs to see the source: https://docs.rs/crate/pdqselect/0.1.1/source/

urschrei:
bors try

urschrei:
🤷‍♂️

urschrei:
bors try

urschrei:
Sorry everyone, it's a "real" error now – looks like something iterator-related landed for arrays in 1.52

pkopparla:
Thank you @urschrei and @lnicola for figuring out that problem. I've fixed the error based on the compiler help message, hopefully it will work now.

lnicola:
bors try

lnicola:
bors r+

> would you mind explaining what I should now?

Sorry, I was confused by this GitHub message:

image

michaelkirk:
Thanks for seeing it through @pkopparla!

PR #678, opened 2021-11-04
Upgrade to rstar 0.9

rmanoka:
---

Useful to be able to use this crate's impl with `GeomWithData`


lnicola:
This is a breaking change, isn't it?

rmanoka:
Yes I think so.  I also just noticed the other [PR](#660) with the same changes and previous discussions.  Closing this.

PR #679, opened 2021-11-06
Fix ClosestPoint when internal for Polygons

michaelkirk:
---

While discussing #675, I noticed our current implementation of ClosestPoint for Polygon seems incorrect.

Previously, we were only considering the boundary of the polygon. But presumably we should behave like JTS and [PostGIS](https://postgis.net/docs/ST_ClosestPoint.html) which consider interior points to have a distance 0.


frewsxcv (approved):

urschrei (approved):

michaelkirk:
Thanks @ frewsxcv and @urschrei!

bors r=frewsxcv

PR #680, opened 2021-11-19
update CI to rust 1.56

michaelkirk:
---

Corresponding PR at https://github.com/georust/docker-images/pull/14

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -91,7 +91,7 @@ jobs:
           # We aim to support rust-stable plus (at least) the prior 3 releases,
           # giving us about 6 months of coverage.
           #
-          # Minimum supported rust versino (MSRV)
+          # Minimum supported rust version (MSRV)
           - "georust/geo-ci:rust-1.51"
           # Two most recent releases - we omit older ones for expedient CI
           - "georust/geo-ci:rust-1.55"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,7 +34,13 @@ jobs:
         working-directory: geo-types
     strategy:
       matrix:
-        container_image: ["georust/geo-ci:rust-1.51", "georust/geo-ci:rust-1.52", "georust/geo-ci:rust-1.53"]
+        container_image:
+          - "georust/geo-ci:rust-1.51"
+          - "georust/geo-ci:rust-1.52"
+          - "georust/geo-ci:rust-1.53"
+          - "georust/geo-ci:rust-1.54"
+          - "georust/geo-ci:rust-1.55"
+          - "georust/geo-ci:rust-1.56"
     container:
       image: ${{ matrix.container_image }}
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -34,7 +34,14 @@ jobs:
         working-directory: geo-types
     strategy:
       matrix:
-        container_image: ["georust/geo-ci:rust-1.51", "georust/geo-ci:rust-1.52", "georust/geo-ci:rust-1.53"]
+        container_image:
+          # We aim to support at least the last 6 months of rust toolchains, so
+          # run on MSRV and the most recent two releases.
+          #
+          # We omit some of the middle releases for faster CI runs.
+          - "georust/geo-ci:rust-1.51"
```
michaelkirk, line 42: If we had an MSRV of 1.52, it'd still be within a 6 month threshold, but it seemed like we might as well maintain support for 1.51 if tests are already passing.

Also, I realize this "6 month support" policy could be controversial. In practice we've been doing something pretty close to this anyway, but I'd be open to revisiting it if it proves arduous.

geo/fuzz/Cargo.toml
```diff
@@ -12,9 +12,11 @@ cargo-fuzz = true
 libfuzzer-sys = "0.4"
 
 [dependencies.geo]
+path = ".."
```
michaelkirk, line 15: specifying a dependency without a version or path was causing a warning (which became an error in rust-1.55).

.github/workflows/test.yml
urschrei: Very happy to throw it open for wider discussion, but the 6 month policy seems reasonable to me (as someone who admittedly doesn't depend on my distro's Rust version)

lnicola (approved):

.github/workflows/test.yml
lnicola: I would slightly prefer something more precise than "6 months", like "stable - 3".

michaelkirk: > I would slightly prefer something more precise than "6 months", like "stable - 3".

Good idea - that would be easier to think about. Updated!

.github/workflows/test.yml
```diff
@@ -71,7 +87,15 @@ jobs:
         working-directory: geo-postgis
     strategy:
       matrix:
-        container_image: ["georust/geo-ci:rust-1.51", "georust/geo-ci:rust-1.52", "georust/geo-ci:rust-1.53"]
+        container_image:
+          # We aim to support rust-stable plus (at least) the prior 3 releases,
+          # giving us about 6 months of coverage.
+          #
+          # Minimum supported rust versino (MSRV)
```
lnicola, line 94:
```suggestion
          # Minimum supported rust version (MSRV)
```

.github/workflows/test.yml
```diff
@@ -52,7 +60,15 @@ jobs:
         working-directory: geo
     strategy:
       matrix:
-        container_image: ["georust/geo-ci:rust-1.51", "georust/geo-ci:rust-1.52", "georust/geo-ci:rust-1.53"]
+        container_image:
+          # We aim to support rust-stable plus (at least) the prior 3 releases,
+          # giving us about 6 months of coverage.
+          #
+          # Minimum supported rust versino (MSRV)
```
lnicola, line 67:
```suggestion
          # Minimum supported rust version (MSRV)
```

.github/workflows/test.yml
```diff
@@ -34,7 +34,15 @@ jobs:
         working-directory: geo-types
     strategy:
       matrix:
-        container_image: ["georust/geo-ci:rust-1.51", "georust/geo-ci:rust-1.52", "georust/geo-ci:rust-1.53"]
+        container_image:
+          # We aim to support rust-stable plus (at least) the prior 3 releases,
+          # giving us about 6 months of coverage.
+          #
+          # Minimum supported rust versino (MSRV)
```
lnicola, line 41:
```suggestion
          # Minimum supported rust version (MSRV)
```

michaelkirk:
bors r=lnicola

PR #682, opened 2021-11-20
Versioned feature to support rstar 0.9

rmanoka:
---

Another approach to solve resolve #660 .  It seems this approach is not uncommon: eg. https://docs.rs/tokio-postgres/0.7.5/tokio_postgres/#features .  It does involve a bit of boiler-plate macros though :(

### DIFF
--- a/geo-types/Cargo.toml
+++ b/geo-types/Cargo.toml
@@ -12,6 +12,7 @@ edition = "2018"
 
 [features]
 use-rstar = ["rstar", "approx"]
+use-rstar-0_9 = ["rstar_0_9", "approx"]
 
 [dependencies]
 approx = { version = "0.4.0", optional = true }
@@ -22,6 +23,7 @@ serde = { version = "1", optional = true, features = ["derive"] }
 # rstar integration relies on the optional approx crate, but implicit features cannot yet enable other features.
 # See: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#namespaced-features
 rstar = { version = "0.8", optional = true }
+rstar_0_9 = { package = "rstar", version = "0.9", optional = true }
 
 [dev-dependencies]
 approx = "0.4.0"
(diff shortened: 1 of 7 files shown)


lnicola (approved):

michaelkirk (approved):
This approach makes sense to me. Thanks for workshopping with us @rmanoka.

geo-types/Cargo.toml
```diff
@@ -22,6 +23,7 @@ serde = { version = "1", optional = true, features = ["derive"] }
 # rstar integration relies on the optional approx crate, but implicit features cannot yet enable other features.
 # See: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#namespaced-features
 rstar = { version = "0.8", optional = true }
```
michaelkirk, line 25: so, if/when we eventually make a major update to geo-types, we'd presumably drop the older rstars?

rmanoka: Yes, we can drop old versions based on some agreeable policy; simplest would be to drop all but latest.

Another question is we adopt a uniform "use-{crate}-{version}" name for the feature or if there's going to be one without version.  I think the uniform naming makes more sense, but don't have a strong opinion.

lnicola: I'd say uniform. I didn't add a comment about it because it would imply a breaking change in the next release, which I assume you wanted to avoid.

rmanoka:
Thanks for the approvals.  I'll go ahead andrefactor the tests to work with either feature, and merge this if all tests pass.

Any thoughts on CI?  I wonder if too many versions would lead to too many combinations here.  Not an issue at the moment I think.

lnicola:
I'd enable all the versions, but not test the combinations.

geo-types/Cargo.toml
rmanoka: Makes sense. Will add some docs so we adopt a policy at the next breaking release.

rmanoka:
bors try

rmanoka:
bors r=@lnicola,@michaelkirk

PR #683, opened 2021-11-25
Fixes the code example in the README file.

denis2glez:
---

The code example in the README file does not compile. The problem is in the assertion 
```
can't compare `&geo::LineString<{float}>` with `geo::LineString<{float}>` the trait `PartialEq<geo::LineString<{float}>>` is not implemented for `&geo::LineString<{float}>`
```
To solve this issue, we could just borrow the right side.
Then, the assertion fails because the exterior of the convex hull has a different order.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,14 +45,17 @@ let poly = polygon![
 // Calculate the polygon's convex hull
 let hull = poly.convex_hull();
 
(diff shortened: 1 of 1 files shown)


michaelkirk (approved):
Good catch, thank you!

bors r+

PR #698, opened 2021-12-08
Bumps the two most recent Rust versions in CI

denis2glez:
---

Only bumps the two most recent Rust versions in the CI config. @michaelkirk, maybe it would be necessary to also increase the minimum supported Rust version?


michaelkirk:
bors r+

LGTM - thanks!

> maybe it would be necessary to also increase the minimum supported Rust version?

As long as tests are still passing we might as well keep the oldest version.

PR #699, opened 2021-12-10
Removes unnecessary call and improves readability.

denis2glez:
---

- Remove an unnecessary call to `map`.
- Use `std::mem::take` instead of `std::mem::replace` when the type `T` implements `Default`.
- Improve readability by using a negated `any` instead of `find` plus `is_none()`.

### DIFF
--- a/geo/src/algorithm/convex_hull/mod.rs
+++ b/geo/src/algorithm/convex_hull/mod.rs
@@ -138,7 +138,7 @@ where
 // Utility function: swap idx to head(0th position), remove
 // head (modifies the slice), and return head as a reference
 fn swap_remove_to_first<'a, T>(slice: &mut &'a mut [T], idx: usize) -> &'a mut T {
-    let tmp = std::mem::replace(slice, &mut []);
+    let tmp = std::mem::take(slice);
     tmp.swap(0, idx);
     let (h, t) = tmp.split_first_mut().unwrap();
     *slice = t;
--- a/geo/src/algorithm/is_convex.rs
+++ b/geo/src/algorithm/is_convex.rs
@@ -189,13 +189,12 @@ where
 
     // Now we have a fixed orientation expected at the rest
     // of the coords. Loop to check everything matches it.
-    if ((i + 1)..n)
+    if !((i + 1)..n)
         .map(orientation_at)
-        .find(|&(_, orientation)| match orientation {
+        .any(|(_, orientation)| match orientation {
             Orientation::Collinear => !allow_collinear,
             orientation => orientation != first_non_collinear,
         })
-        .is_none()
     {
         Some(first_non_collinear)
     } else {
--- a/geo/src/algorithm/is_convex.rs
+++ b/geo/src/algorithm/is_convex.rs
@@ -189,12 +189,13 @@ where
 
     // Now we have a fixed orientation expected at the rest
     // of the coords. Loop to check everything matches it.
-    if !((i + 1)..n)
+    if ((i + 1)..n)
         .map(orientation_at)
-        .any(|(_, orientation)| match orientation {
+        .find(|&(_, orientation)| match orientation {
             Orientation::Collinear => !allow_collinear,
             orientation => orientation != first_non_collinear,
         })
+        .is_none()
     {
         Some(first_non_collinear)
     } else {
--- a/geo/src/algorithm/k_nearest_concave_hull.rs
+++ b/geo/src/algorithm/k_nearest_concave_hull.rs
@@ -338,7 +338,7 @@ mod tests {
             coord!(x: 1.0, y: 0.0),
         ];
 
-        let mut coords_mapped: Vec<&Coordinate<f32>> = coords.iter().map(|x| x).collect();
+        let mut coords_mapped: Vec<&Coordinate<f32>> = coords.iter().collect();
 
         let center = coord!(x: 0.0, y: 0.0);
         let prev_coord = coord!(x: 1.0, y: 1.0);


geo/src/algorithm/convex_hull/mod.rs
```diff
@@ -138,7 +138,7 @@ where
 // Utility function: swap idx to head(0th position), remove
 // head (modifies the slice), and return head as a reference
 fn swap_remove_to_first<'a, T>(slice: &mut &'a mut [T], idx: usize) -> &'a mut T {
-    let tmp = std::mem::replace(slice, &mut []);
+    let tmp = std::mem::take(slice);
```
lnicola, line 141: Wow, does this actually work?

geo/src/algorithm/is_convex.rs
```diff
@@ -189,13 +189,12 @@ where
 
     // Now we have a fixed orientation expected at the rest
     // of the coords. Loop to check everything matches it.
-    if ((i + 1)..n)
+    if !((i + 1)..n)
```
lnicola, line 192: To me, this still seems hard to read :sweat_smile:.

lnicola: Does:

```rust
if !((i + 1)..n)
 
 lnicola 43 seconds ago Member

To me, this still seems hard to read.

Does something like

```rust
    if ((i + 1)..n)
        .map(orientation_at)
        .all(|(_, orientation)| match orientation {
            Orientation::Collinear => allow_collinear,
            orientation => orientation == first_non_collinear,
        })
```

work?

denis2glez: Sure, readability can be a bit subjective

geo/src/algorithm/convex_hull/mod.rs
denis2glez: Yes

geo/src/algorithm/is_convex.rs
denis2glez: > Does something like
> 
> ```rust
>     if ((i + 1)..n)
>         .map(orientation_at)
>         .all(|(_, orientation)| match orientation {
>             Orientation::Collinear => allow_collinear,
>             orientation => orientation == first_non_collinear,
>         })
> ```
> 
> work?

I didn't want to touch the match, but sure, it's equivalent

lnicola: I'm not actually sure, but hopefully it's covered by tests? I think it's worth applying DeMorgan's in this case.

denis2glez: Of course, it is a simplification

geo/src/algorithm/convex_hull/mod.rs
urschrei: I think @lnicola's reaction is informative here: it's completely non-obvious what this call does now; you have to look at the docs (or your IDE) to figure out what it's doing compared to the existing call, and if there were no alternative I'd advocate for a comment.

lnicola: I don't mind it that much, I was just surprised that `&[T]` implements `Default`.

denis2glez: That has been the case since 1.0

geo/src/algorithm/is_convex.rs
michaelkirk: As phrased in the current PR, this isn't an obvious improvement to me, and every time we touch code people have to spend time reviewing it, and of course we might break something that was working before, so I'd caution you against work that changes presumed working code for stylistic reasons unless there is a very clear and sizable readability vs. risk payoff.

(I think the other two changes meet this criteria FWIW)

I'd actually prefer you just revert this part altogether, but if other folks think this code should change, I think @lnicola's suggestion is the best so far, provided you're willing to verify the test cases cover all possible paths relevant to the change.
```
if ((i + 1)..n)
        .map(orientation_at)
        .all(|(_, orientation)| match orientation {
            Orientation::Collinear => allow_collinear,
            orientation => orientation == first_non_collinear,
        })
```

denis2glez: I could guarantee that this change is equivalent, but I think that is not the point. I fully agree not to change the code that works without objective reasons. Thanks everyone for the feedback, I'm going to revert this part.

denis2glez: Let me know anything else.

geo/src/algorithm/convex_hull/mod.rs
michaelkirk: from https://doc.rust-lang.org/std/mem/fn.take.html:

> **`take`**: Replaces dest with the default value of T, returning the previous dest value.

To my eye this change reads like pretty idiomatic usage of `take`. I'm struggling to think of what kind of clarifying comment we could add that isn't just the documentation for `take`, which seems like kind of a silly thing to do.

If anyone has a good comment suggestion, we could use it. Otherwise, I'd say it's fine as is.

michaelkirk (approved):
Thanks for the changes @denis2glez 

It LGTM, but I'm going to hold off on merging for a bit in case anyone has further feedback.

geo/src/algorithm/convex_hull/mod.rs
urschrei: My point is that what the new code does is less obvious (to me) than the existing code, only because take's purpose is less immediately obvious than replace. I also realise this is far too much time to devote to such a minor change and nothing would ever get done if this was the norm – if you're both happy to accept it then I'm happy to go along with you.

michaelkirk (approved):
bors r+

geo/src/algorithm/convex_hull/mod.rs
michaelkirk: I went ahead and merged the PR, along with a [code comment](https://github.com/georust/geo/commit/d606fc5f01a6927f5e6033b09df7d404155b5637) to cover all the bases.

PR #700, opened 2021-12-11
Minor: Remove test_all_general() from code coverage report.

martinfrances107:
- [ x] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/master/CODE_OF_CONDUCT.md).
---

I was using `cargo tarpaulin -o Html` to generate a code coverage webpage.

All looks good this is a well tested crate.
 
But I noticed the report painted the test function test_all_general() red  --- as having no coverage

This is a minor oversight .. test harnesses and test infrastructure should be excluded for the consideration.

### DIFF
--- a/geo/tests/jts_tests.rs
+++ b/geo/tests/jts_tests.rs
@@ -9,6 +9,7 @@ fn init_logging() {
 }
 
 #[test]
+#[cfg(not(tarpaulin_include))]
 fn test_all_general() {
     init_logging();
 


martinfrances107:
>but doesn't the same problem exist for every single #test method we have? 

In the tarpaulin context, 

1) functions, modules with the #test attribute are automatically ignored by default. 
2) Files in the integration test directory [./tests] are ignored.

[I did check,  but did not document that this function is the only visible outlier.]

The reason this is considered an outlier is that according to Rusts rules for what is a integration test and what is not. The following applies.

The tests/ directory is considered flat and not recusivley serched.

So  a test in file 

tests/moduleA/module_a_integration_test.rs 

Will be ignored.. it is some thing that always gets me in trouble.

In that light 
geo/tests/jts_tests.rs contains a tests helper function indistinguiable for production code  and need special casing.


I hope that  helps

michaelkirk:
bors try 

Sorry for the slow feedback cycle here. To be honest I don't really have the time to test this myself (tarpaulin doesn't work on my host machine, so I'd have to rent a VM or something.).

The change basically makes sense and seems innocuous enough to me, so I'm willing to merge it if the tests pass since it's useful to you. I'll leave it open for just a couple more days in case someone else wants to weight in.

michaelkirk:
bors try

trying again now that we fixed geo-ci...

martinfrances107:
> tarpaulin doesn't work on my host machine,

Just adding supporting background information 

`cargo install cargo-tarpaulin
`

only works on x86 machines 

here is the reference.

https://dev.to/davidadewoyin/top-rust-cargo-commands-2b70
> 
> Tarpaulin is a code coverage reporting tool for the Cargo build system.
> 
> It's installed with cargo install cargo-tarpaulin. Currently tarpaulin only works on x86_64 CPU architectures running Linux.Use cargo tarpaulin --ignore-tests to compute code coverage for your application while ignoring test functions.
> 
>

Issue #649, opened 2021-05-19
perf: speed up `Relate` trait with an R-Tree backed EdgeSetIntersector 


michaelkirk:
Extracted from our conversation here: https://github.com/georust/geo/issues/641#issuecomment-813108578

**background:**

The [Relate trait](https://github.com/georust/geo/blob/master/geo/src/algorithm/relate/mod.rs) (introduced in https://github.com/georust/geo/pull/639) topologically relates two geometries based on [DE-9IM](https://en.wikipedia.org/wiki/DE-9IM) semantics, — whether `A contains B`,  `A is within B`, `A overlaps B`, etc.

Part of the computation is finding all the intersections between the set of all segments in the  geometries, which happens via the [EdgeSetIntersector](https://github.com/georust/geo/blob/master/geo/src/algorithm/relate/geomgraph/index/edge_set_intersector.rs) trait.

We currently only have a [naive `O(n^2)` implementation](https://github.com/georust/geo/blob/master/geo/src/algorithm/relate/geomgraph/index/simple_edge_set_intersector.rs) which intersects every segment with every other. 

The Bentley-Ottman class of sweep line algorithms can be much faster than our naive approach. We don't currently have a BO implementation, but @rmanoka et al have discussed before (some previous discussion https://github.com/georust/geo/issues/515#issuecomment-778703693)

JTS/GEOS currently rely on a ["Monotonic Chain Sweeper"](https://github.com/locationtech/jts/blob/0afbfb1956ec24912a8b4dc4edff0f1200442857/modules/core/src/main/java/org/locationtech/jts/geomgraph/index/SimpleMCSweepLineIntersector.java#L37) for this functionality, which I understand to be an augmented Bentley-Ottman algorithm (though I could be mistaken).

Another option, proposed here, would be an r-tree backed approach, which, apparently can perform reasonably well in practice for both sparsely and densely intersecting sets.

And, since we already have access to the `rstar` crate, might prove simpler to implement.

Part of this work should be some benchmarks on a useful array of cases to have some kind of baseline for assessing this work.

Note that other work, like perhaps a future Buffering feature could also benefit from this.

rmanoka:
[...]

Issue #566, opened 2020-12-10
Add WKT to devDependencies for test fixtures


michaelkirk:
When testing, we'll often define inline simple geometries of a few points or lines. But for testing more geometries, we conventionally stick them in our test_fixtures:
https://github.com/georust/geo/blob/master/geo/src/algorithm/test_fixtures

Rather than writing fixtures as rust code, I think it would be better to store them in a format like WKT. As well as being able to read, it makes it easier to share tests with external projects / reference implementations.

WDYT?

So, todo:

- [ ] add WKT dep
- [ ] convert test fixtures to wkt

frewsxcv:
storing them in wkt instead of rust vecs sounds good to me!

PR #701, opened 2021-12-22
Large geometry intersection check perf speedup

urschrei:
Above some threshold, it may be faster to load the geometry or geometries into an R*-star tree and query for a subset of intersection candidates.

This is a draft PR because:
- I haven't benchmarked the tree-query logic in order to figure out even an approximate value for `MAX_NAIVE_SEGMENTS`
- I haven't figured out whether checking the number of segments in order to decide to switch to trees is an unacceptable perf hit 

---


urschrei:
Screenshot 2021-12-22 at 13 31 36

This is with length checking turned on, RTree (old) vs naïve (new) polygon-line intersection check using the `louisiana.rs` text fixture. We're going to need a bigger polygon.

urschrei:
Same test setup (RTree is old, naïve is new), using `norway_main.rs` (8534 vertices)
Screenshot 2021-12-22 at 13 46 05

urschrei:
Same test setup (RTree is old, naïve is new), using a new `big.rs` (100610 vertices) valid polygon:
Screenshot 2021-12-22 at 15 37 19

If I'm interpreting this correctly:

1. Using a tree for a _single_ Polygon-line intersection check probably isn't worth it for many applications
2. The length checks aren't significant in wall-clock terms
3. The existing logic is very fast in wall-clock terms!

Now, on to Polygon-Polygon intersection checking…

urschrei:
Now we're talking:

Old (red, naïve) vs new (blue, tree) (sorry for reversing the setup this time!)
Polygon-Polygon intersection check using `big.rs` and `norway_main.rs` (100610 and 8534 vertices, respectively)

Screenshot 2021-12-22 at 17 53 34

The naïve version has to potentially do ~**860 million** line-line intersection checks so **165 ms** isn't bad, but the R*-tree mean is only **35.4 ms**…

michaelkirk:
Cool! 

Just as a reminder, @rmanoka did some *somewhat* related intersection perf investigation here: 
https://github.com/georust/geo/issues/649

It's a slightly different use case, but the findings might be relevant. In particular, I sort of expect the performance of R-Tree to be better than the naive approach as a.segments() * b.segements() scales.

So if you *did* want to have  a threshold to toggle between "naive" vs. "rtree" using a multiplicative, rather than an additive might make more sense.

```
if poly_a.num_segments() * poly_b.num_segements() > x {
    use_rtree()
} else {
    use_naive()
}
```

urschrei:
Yeah, the multiplicative is definitely better. Next up, I need to figure out a way of getting the large polygon data into the benchmark; `include!`ing it is taking about 45 mins to compile, and it's only around 2.8 mb.

michaelkirk:
Maybe https://github.com/georust/geo/issues/566?

(lol not to just dump my wishlist on you or anything...)

urschrei:
Hmm as constructed `intersection_candidates_with_other_tree` isn't actually producing any candidates for our (known-to-intersect) test data. I wonder what's going on…

urschrei:
Update: we're checking for line intersections in this case, since we're decomposing the polygons into lines. But that excludes the possibility of `polygon a` completely enclosing `polygon b` (e.g. large square with smaller square inside). 🤔

urschrei:
If / when #351 lands, the easier case (`A` contains `B`) will be solved, but the more tricky case (`A` contains `B` _inside one of its holes_).

geo/src/algorithm/intersects/polygon.rs
```diff
@@ -31,13 +36,42 @@ where
...
     }
 }
+
+impl<T> Intersects<LineString<T>> for Polygon<T>
+where
+    T: GeoNum,
+    geo_types::Line<T>: RTreeObject,
+{
+    fn intersects(&self, linestring: &LineString<T>) -> bool {
+        if (self.exterior().0.len() + self.interiors().iter().map(|ls| ls.0.len()).sum::<usize>())
```
frewsxcv, line 46:
```
        if (self.exterior().0.len() + self.interiors().iter().map(|ls| ls.0.len()).sum::<usize>())
```

https://github.com/georust/geo/issues/706 😄

geo/src/algorithm/intersects/polygon.rs
```diff
@@ -48,13 +82,40 @@ symmetric_intersects_impl!(Rect<T>, Polygon<T>);
...
+                    .interiors()
+                    .iter()
+                    .map(|ls| ls.0.len())
+                    .sum::<usize>())
+            > MAX_NAIVE_SEGMENTS
+        {
+            let lines_a: Vec<_> = self
+                .exterior()
+                .lines()
+                .chain(self.interiors().iter().flat_map(|ls| ls.lines()))
```
frewsxcv, line 101: Should the last coordinate of the exterior ring connect to the first coordinate of the first interior ring?

frewsxcv: Oh wait, that's not what's happening here, nevermind!

frewsxcv: Meep meephttps://github.com/georust/geo/issues/708

michaelkirk:
Now that https://github.com/georust/geo/pull/829 is merged, we might get comparable perf via something like `a.relate(b).is_intersects()`

urschrei:
> a.relate(b).is_intersects()

`relate` requires a `GeoFloat` bound (currently `GeoNum`), and from a first cut, use of `relate` introduces extreme perf regressions (900 %+ in some case) in our nice new boolean ops intersection benchmark suite…

michaelkirk:
>  from a first cut, use of relate introduces extreme perf regressions (900 %+ in some case) in our nice new boolean ops intersection benchmark suite…

Do you still have an example of the data you were using to see this kind of performance? 

I'd expect the R-Tree approach to be a constant factor slower, but better asymptotically — much better for big inputs and a little slower for small inputs. I am curious though as to what range of geometry sizes we can expect to see the tradeoffs.

urschrei:
If you check out this branch and run the boolean ops benchmarks you should see the perf regression in the criterion output. I should also note that this was a quick check that I didn't dig into, so I may have messed something up.

urschrei:
(I rebased against `main` just in case)

urschrei:
We have prepared geometries for this now!

PR #703, opened 2021-12-25
Extract test fixtures into WKT files

### LINKED ISSUE #566: Add WKT to devDependencies for test fixtures

lnicola:
---

Fixes #566

Some notes:

 - we say these are line strings in the WKT; that's not always the case, but it's still closer to the original code
 - I made this into a separate crate, in order to be able to access it from both benches and tests (there's no `#[cfg(bench)]`); it has `publish = false` to prevent accidental publishing
 - the `extremes f64` and `simplify vwp f64` benches were testing an `f32` implementation
 - one quick hull test wasn't using the baseline (which was itself unused)
 - `rust-analyzer` used to balk on the larger `include!`s, it feels much better now
 - the benches seem pretty unstable on my computer


geo/benches/extremes.rs
```diff
@@ -4,21 +4,21 @@ extern crate geo;
...
+        let polygon = Polygon::new(norway, vec![]);
 
         bencher.iter(|| {
             criterion::black_box(criterion::black_box(&polygon).extremes());
         });
     });
 
     c.bench_function("extremes f64", |bencher| {
-        let points = include!("../src/algorithm/test_fixtures/norway_main.rs");
-        let polygon = Polygon::new(LineString::<f32>::from(points), vec![]);
```
lnicola, line 21: Wrong type here ^

geo/benches/simplifyvw.rs
```diff
@@ -39,8 +35,7 @@ fn criterion_benchmark(c: &mut Criterion) {
     });
 
     c.bench_function("simplify vwp f64", |bencher| {
-        let points = include!("../src/algorithm/test_fixtures/louisiana.rs");
-        let ls: LineString<f32> = points.into();
```
lnicola, line 43: Wrong type here, too ^

geo/src/algorithm/convex_hull/qhull.rs
```diff
@@ -206,29 +206,18 @@ mod test {
...
 
     #[test]
     fn quick_hull_test_complex_2() {
-        let coords = include!("../test_fixtures/poly2.rs");
-        let mut v: Vec<_> = coords
-            .iter()
-            .map(|e| Coordinate { x: e.0, y: e.1 })
-            .collect();
-        let res = quick_hull(&mut v);
-        assert!(res.is_strictly_ccw_convex());
```
lnicola, line 231: `poly2_hull` wasn't used

michaelkirk (approved):
Thanks for this @lnicola - LGTM!

I'll wait a couple days before merging in case someone (@urschrei) wants to weigh in.

urschrei:
And me!

Issue #702, opened 2021-12-23
serde is not supported for `Geometry` and `GeometryCollection`


msalib:
Serializing one of those doesn't work because there is no `#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]` for either of those types.

martinfrances107:
+1 I think this need to be made complete.

When I first saw it, a month ago I was in a rush and did not have time to fix it upstream ...here 

The ugly workaround I used was to loop over geometry colections inner elements and serialize/deserialise each element individually ...

I would be happy to help out here...

martinfrances107:
So I have created an initial patch ... testing is on my mind. Geometry primitives like point do not have explicit testing for serde. So I am not sure how to fit in testing?

frewsxcv:
Done in #704

PR #704, opened 2021-12-27
#702 Geometry Collection now supports serde.

martinfrances107:

Initial version ... I want to discuss ways to test this, from what I can see serde support is not tested for geometry primitives like point, or line_string.. so I don't know how to confirm this works for all users.

### DIFF
--- a/geo-types/CHANGES.md
+++ b/geo-types/CHANGES.md
@@ -4,6 +4,8 @@
 
 * Support `rstar` version `0.9` in feature `use-rstar_0_9`
   * <https://github.com/georust/geo/pull/682>
+* Geometry and GeometryCollection now support serde.
+  * <https://github.com/georust/geo/pull/704>
 
 ## 0.7.2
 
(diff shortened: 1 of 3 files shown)


michaelkirk:
I we'd also want to derive serde for `Geometry` - can you add that here as well?

martinfrances107:
> I think we'd also want to derive serde for Geometry - can you add that here as well?

That seems like a good idea... done.

michaelkirk:
bors try

michaelkirk:
LGTM

Personally, I don't think we need additional tests for this particular change, but I'll wait a couple days before merging in case someone else wants to weigh in.

lnicola:
Needs a changelog entry.

martinfrances107:
> I don't think it needs tests, but it's worth adding a changelog entry.

Thanks .. I agree... done.

geo-types/CHANGES.md
```diff
@@ -4,6 +4,8 @@
 
 * Support `rstar` version `0.9` in feature `use-rstar_0_9`
   * <https://github.com/georust/geo/pull/682>
+* Geometry and GeometryCollection now support serde.
```
lnicola, line 7:
```suggestion
* `Geometry` and `GeometryCollection` now support serde.
```

martinfrances107:
@lnicola  That is better .. thank you.

PR #705, opened 2021-12-28
More LineString iterators

urschrei:
---

This PR adds the `iter` and `iter_mut` methods on `LineString` (previously only available via its public `.0` field), yielding `Coordinate`s, and adds an `into_coordinates` method, complementing `into_points`. It also adds `into_iter` for **borrowed** `LineString`s.

Some of these methods overlap with those available in `geo`s [CoordsIter](https://docs.rs/geo/0.18.0/geo/algorithm/coords_iter/trait.CoordsIter.html) trait, but:

- Not all `geo-types` library users are _necessarily_ `geo` users
- The lack of direct `iter` methods is a constant papercut in my opinion.

### DIFF
--- a/geo-types/src/line_string.rs
+++ b/geo-types/src/line_string.rs
@@ -78,7 +78,7 @@ use std::ops::{Index, IndexMut};
 /// let line_string: LineString<f32> = coords_iter.collect();
 /// ```
 ///
-/// You can iterate over the coordinates in the `LineString`:
+/// You can iterate and loop over the coordinates in the `LineString`, yielding [Coordinate]s:
 ///
 /// ```
 /// use geo_types::{Coordinate, LineString};
@@ -87,6 +87,8 @@ use std::ops::{Index, IndexMut};
 ///     Coordinate { x: 0., y: 0. },
 ///     Coordinate { x: 10., y: 0. },
 /// ]);
+///
+/// line_string.iter().for_each(|coord| println!("Coordinate x = {}, y = {}", coord.x, coord.y));
 ///
 /// for coord in line_string {
 ///     println!("Coordinate x = {}, y = {}", coord.x, coord.y);
@@ -132,15 +134,47 @@ impl<'a, T: CoordNum> DoubleEndedIterator for PointsIter<'a, T> {
     }
 }
 
+/// A [Coordinate] iterator returned by the `iter` method over a [LineString]
+#[derive(Debug)]
+pub struct CoordinatesIter<'a, T: CoordNum + 'a>(::std::slice::Iter<'a, Coordinate<T>>);
+
+impl<'a, T: CoordNum> Iterator for CoordinatesIter<'a, T> {
+    type Item = &'a Coordinate<T>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.0.next()
+    }
+}
+
+impl<'a, T: CoordNum> DoubleEndedIterator for CoordinatesIter<'a, T> {
+    fn next_back(&mut self) -> Option<Self::Item> {
+        self.0.next_back()
+    }
+}
+
 impl<T: CoordNum> LineString<T> {
     /// Return an iterator yielding the coordinates of a `LineString` as `Point`s
     pub fn points_iter(&self) -> PointsIter<T> {
         PointsIter(self.0.iter())
     }
 
+    /// Return an iterator yielding the members of a [LineString] as [Coordinate]s
+    pub fn iter(&self) -> CoordinatesIter<T> {
+        CoordinatesIter(self.0.iter())
+    }
+    /// Return an iterator yielding the coordinates of a [LineString] as mutable [Coordinate]s
+    pub fn iter_mut(&mut self) -> CoordinatesIter<T> {
+        CoordinatesIter(self.0.iter())
+    }
+
     /// Return the coordinates of a `LineString` as a `Vec` of `Point`s
     pub fn into_points(self) -> Vec<Point<T>> {
         self.0.into_iter().map(Point).collect()
+    }
+
+    /// Return the coordinates of a [LineString] as a `Vec` of [Coordinate]s
+    pub fn into_coordinates(self) -> Vec<Coordinate<T>> {
+        self.0.into_iter().collect()
     }
 
     /// Return an iterator yielding one `Line` for each line segment
--- a/geo-types/src/line_string.rs
+++ b/geo-types/src/line_string.rs
@@ -163,8 +163,8 @@ impl<T: CoordNum> LineString<T> {
     }
 
     /// Return an iterator yielding the members of a [LineString] as [Coordinate]s
-    pub fn iter(&self) -> CoordinatesIter<T> {
-        CoordinatesIter(self.0.iter())
+    pub fn iter(&self) -> impl Iterator<Item = &Coordinate<T>> {
+        self.0.iter()
     }
 
     /// Return an iterator yielding the coordinates of a [LineString] as mutable [Coordinate]s
--- a/geo-types/src/line_string.rs
+++ b/geo-types/src/line_string.rs
@@ -5,7 +5,7 @@ use crate::{CoordNum, Coordinate, Line, Point, Triangle};
 use std::iter::FromIterator;
 use std::ops::{Index, IndexMut};
 
-/// An ordered collection of two or more [`Coordinate`]s, representing a
+/// An ordered collection of two or more [Coordinate]s, representing a
 /// path between locations.
 ///
 /// # Semantics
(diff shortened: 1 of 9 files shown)


lnicola:
Can `iter` use the same `impl Iterator` trick?

urschrei:
> Can `iter` use the same `impl Iterator` trick?

Done. I don't _think_ there's a way to leverage it for `IntoIter` though, otherwise we could do away with the helper struct.

geo-types/src/line_string.rs
```diff
@@ -132,19 +138,52 @@ impl<'a, T: CoordNum> DoubleEndedIterator for PointsIter<'a, T> {
     }
 }
 
+/// A [Coordinate] iterator used by the `into_iter` method on a [LineString]
+#[derive(Debug)]
+pub struct CoordinatesIter<'a, T: CoordNum + 'a>(::std::slice::Iter<'a, Coordinate<T>>);
```
frewsxcv, line 143: Not blocking, but this could also implement [`ExactSizeIterator`](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html)

geo-types/src/line_string.rs
```diff
@@ -5,7 +5,7 @@ use crate::{CoordNum, Coordinate, Line, Point, Triangle};
 use std::iter::FromIterator;
 use std::ops::{Index, IndexMut};
 
-/// An ordered collection of two or more [`Coordinate`]s, representing a
+/// An ordered collection of two or more [Coordinate]s, representing a
```
frewsxcv, line 8: Nonblocking and subjective, but how do you feel about keeping the inline code block formatting? At least to me, it better indicates that "Coordinate" in this context is referring to a specific geo-types API

```suggestion
/// An ordered collection of two or more [`Coordinate`]s, representing a
```

geo-types/src/line_string.rs
```diff
@@ -132,19 +138,52 @@ impl<'a, T: CoordNum> DoubleEndedIterator for PointsIter<'a, T> {
...
+    /// Return the coordinates of a [LineString] as a `Vec` of [Point]s
     pub fn into_points(self) -> Vec<Point<T>> {
         self.0.into_iter().map(Point).collect()
     }
 
-    /// Return an iterator yielding one `Line` for each line segment
-    /// in the `LineString`.
+    /// Return the coordinates of a [LineString] as a `Vec` of [Coordinate]s
+    pub fn into_coordinates(self) -> Vec<Coordinate<T>> {
+        self.0.into_iter().collect()
```
frewsxcv, line 182: Could this just return `.0`?

```suggestion
        self.0
```

frewsxcv (approved):

geo-types/src/line_string.rs
frewsxcv: I'm fine with the name as-is, but another option is to name it `into_inner`, which returns the data the structure contains, which is a convention in the Rust standard library, and also our `Polygon#into_inner`

urschrei: Fixed `.0`🙈, and renamed.

geo-types/src/line_string.rs
urschrei: Done!

geo-types/src/line_string.rs
urschrei: Done for all types!

urschrei:
Implemented all yr suggestions @frewsxcv.

frewsxcv (approved):

geo-types/src/line_string.rs
```diff
@@ -132,19 +138,58 @@ impl<'a, T: CoordNum> DoubleEndedIterator for PointsIter<'a, T> {
...
 impl<T: CoordNum> LineString<T> {
-    /// Return an iterator yielding the coordinates of a `LineString` as `Point`s
+    /// Return an iterator yielding the coordinates of a [`LineString`] as [`Point`]s
     pub fn points_iter(&self) -> PointsIter<T> {
         PointsIter(self.0.iter())
     }
 
-    /// Return the coordinates of a `LineString` as a `Vec` of `Point`s
+    /// Return an iterator yielding the members of a [`LineString`] as [`Coordinate`]s
+    pub fn iter(&self) -> impl Iterator<Item = &Coordinate<T>> {
```
michaelkirk, line 172: Sorry to bikeshed on naming, but... `iter` strikes me as a little off.

In contrast, these usages of `iter` all make sense to me:
```
multi_point.iter() -> impl Iterator<Item = &Point<T>>
multi_line_string.iter() -> impl Iterator<Item = &LineString<T>>
multi_polygon.iter() -> impl Iterator<Item = &Polygon<T>>
vec_of_strings.iter() => -> impl Iterator<Item = &String>
```

For a collection of `Foo`s, `iter()` will yield `Foo`. But I don't think of a LineString as a collection of Coordinates (it happens to look that way on disk, but so do all of our geometries).

What would you think of `pub fn coords(&self)` or `coords_iter()`?

michaelkirk: `fn coords()` might match nicely with the existing:

```
/// Return an iterator yielding one [Line] for each line segment in the [`LineString`].
pub fn lines(&'_ self) -> impl ExactSizeIterator + Iterator<Item = Line<T>> + '_ { }

/// An iterator which yields the coordinates of a [`LineString`] as [Triangle]s
pub fn triangles(&'_ self) -> impl ExactSizeIterator + Iterator<Item = Triangle<T>> + '_ { }
```

urschrei: Hmmmm. I initially added iter() because _I_ expected LineString to have it, mostly when writing internal code. But you're absolutely correct in that the choice stems purely from my own expectations, which are at least partly rooted in the fact that we had already added IntoIter so we could loop over Coordinates, and I was annoyed that we could loop but not explicitly iter().

I'm open to adding coords() instead, but should we then consider deprecating points_iter() in favour of points() in an effort to be consistent?

If we _do_, I'll probably rewrite the introductory docs a bit to note that we offer four iterators, but note that loops will yield Coordinate (which should be thought of as the default?)

michaelkirk: Oh geeze, software is so hard. 

Honestly, in hindsight, I think it's a little weird that you can do this at all:

```
for x in &line_string { 
    ...
}
```

But anyway, that ship sailed long ago, and maybe consistency is more important at this point. Thanks for bringing up the context, your change seems pretty reasonable in light of that.

Would it be crazy to do all of the following?

- Keep the `iter()` impl as you have it since it complements the existing `into_iter()`,
  - Alternatively, we could remove the into_iter impl from line_string, but I hesitate to break things, and I don't personally think it's doing much harm as is, just that there are clearer ways to write it. 
- add `line_string.coords()` which does the exact same thing as `line_string.iter()` but reads more clearly IMO
- add points() / deprecate `points_iter()` 
  - I think it's pretty idiomatic to omit the `_iter` suffix when returning transformations of the inner data - e.g. `hash_map.values()` and `hash_map.values_mut()` not `hash_map.values_iter()` and `hash_map.values_iter_mut()` same with `string.chars()` not `string.chars_iter()`.

I know it's a real laundry list of things... so if you're in spirit on board, but don't have the time to do it, feel free to punt on some of it.

urschrei:
Pulling https://github.com/georust/geo/pull/705#pullrequestreview-843078728 comments out into the main thread for discussion. @michaelkirk I've diverged slightly from your suggestion: I've removed `iter()` in favour of `coords()`, and re-written the docs to:

- Show all `LineString` creation methods / mechanisms;
- list all five available iterators (`coords[_mut]`, `points`, `lines`, and `triangles`);
- note that our `IntoIterator` impl yields `Coordinates` when looping (impls can't be marked as deprecated, unfortunately, and I agree that it's too late to do anything about its presence now)
- Show `LineString` _decomposition_ methods (returning `Vec`s of `Coordinates` or `Points)`

I think I now prefer the lack of `iter()` combined with clearer docs: it means that you have to be explicit about what you want to iterate over.

michaelkirk:
> I think I now prefer the lack of iter() combined with clearer docs: it means that you have to be explicit about what you want to iterate over.

LGTM! I think this turned out rather nicely! Thanks for allowing the back-and-fourth.

geo-types/src/line_string.rs
```diff
@@ -260,14 +327,14 @@ impl<T: CoordNum> From<Line<T>> for LineString<T> {
...
-/// Turn an iterator of `Point`-like objects into a `LineString`.
+/// Turn an iterator of [`Point`]-like objects into a [`LineString`].
 impl<T: CoordNum, IC: Into<Coordinate<T>>> FromIterator<IC> for LineString<T> {
     fn from_iter<I: IntoIterator<Item = IC>>(iter: I) -> Self {
         LineString(iter.into_iter().map(|c| c.into()).collect())
     }
 }
 
-/// Iterate over all the [Coordinate](struct.Coordinates.html)s in this `LineString`.
+/// Iterate over all the [`Coordinate`]s in this [`LineString`].
```
michaelkirk, line 337: Thanks for cleaning all these up!

urschrei:
> I think this turned out rather nicely!

I think so too! I'm gonna leave it open for a day in case anyone else wants to kick the tires, otherwise I'll merge tomorrow.

urschrei:
bors try

PR #709, opened 2022-01-03
FIX: geo tests not running in CI

michaelkirk:
---

Due to the lack of quoting, the "use-serde" feature was incorrectly
being interpreted as a filter on which tests to run.

e.g. compare

previously working:
https://github.com/georust/geo/runs/3929823924?check_suite_focus=true#step:6:72

broken here:
https://github.com/georust/geo/runs/3929868700?check_suite_focus=true#step:5:184

and now working again:
https://github.com/georust/geo/runs/4694596622?check_suite_focus=true#step:5:145


lnicola:
That's awful.

bors r+

PR #718, opened 2022-01-29
Add new `Transform` algorithm for transforming a geometry with PROJ.

frewsxcv:
  - I will do this if we decide we want this
- [x] Add entry to table of contents in `lib.rs`

### DIFF
--- a/geo/src/algorithm/transform.rs
+++ b/geo/src/algorithm/transform.rs
@@ -1,0 +1,117 @@
+use std::{convert::TryInto, error::Error, fmt};
+
+/// Transform a geometry using PROJ.
+pub trait Transform<T> {
+    type Output;
+
+    /// Transform a geometry.
+    ///
+    /// # Examples
+    ///
+    /// Transform a geometry using a PROJ string definition:
+    ///
+    /// ```
+    /// use geo::{self, prelude::*};
+    ///
+    /// let point = geo::point!(x: -36.508f32, y: -54.2815f32);
+    ///
+    /// assert_eq!(
+    ///     point.transform("+proj=axisswap +order=2,1,3,4").unwrap(),
+    ///     geo::point!(x: -54.2815f32, y: -36.508f32)
+    /// );
+    /// ```
+    ///
+    /// Transform a geometry from one CRS to another CRS:
+    ///
+    /// ```
+    /// use geo::{self, prelude::*};
+    ///
+    /// let point = geo::point!(x: -36.508f32, y: -54.2815f32);
+    ///
+    /// assert_eq!(
+    ///     point.transform(("EPSG:4326", "EPSG:3857")).unwrap(),
+    ///     geo::point!(x: -4064052.0f32, y: -7223650.5f32)
+    /// );
+    /// ```
+    fn transform(
+        &self,
+        proj: impl TryInto<proj::Proj, Error = proj::ProjCreateError>,
+    ) -> Result<Self::Output, TransformError>;
+
+    /// Transform a geometry from one CRS to another CRS.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use geo::{self, prelude::*};
+    ///
+    /// let point: geo::Point<f32> = geo::point!(x: -36.508f32, y: -54.2815f32);
+    ///
+    /// assert_eq!(
+    ///     point.transform_crs_to_crs("EPSG:4326", "EPSG:3857").unwrap(),
+    ///     geo::point!(x: -4064052.0f32, y: -7223650.5f32)
+    /// );
+    /// ```
+    fn transform_crs_to_crs(
+        &self,
+        source_crs: &str,
+        target_crs: &str,
+    ) -> Result<Self::Output, TransformError>;
+}
+
+impl<T, G> Transform<T> for G
+where
+    T: crate::CoordFloat,
+    G: crate::algorithm::map_coords::TryMapCoords<T, T, proj::ProjError>,
+{
+    type Output = G::Output;
+
+    fn transform(
+        &self,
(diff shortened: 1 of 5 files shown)


lnicola:
This is nice to have, but I'm wondering if it's better to take the transform instead of the two CRSes. I don't know how long these take to instantiate in practice, but it seems like you'd want to reuse them when possible.

urschrei:
> the transform instead of the two CRSes
>  it seems like you'd want to reuse them when possible.

I agree, even though the trade-off is a less-nice API. However, without something that takes a `Proj` instance, you can't use a  [`ProjBuilder`](https://docs.rs/proj/0.24.0/proj/struct.ProjBuilder.html#impl-Drop) to enable grid downloads (etc.) requiring network access, which is common for people who need higher-accuracy transforms:

    let pb = ProjBuilder::new();
    pb.enable_network().unwrap();
    let transform = pb.proj_known_crs("EPSG:4326", "EPSG:3857").unwrap(); // Proj instance!

frewsxcv:
What if the argument was `TryInto<Proj>`? So you could pass it a `Proj`, and we could also consider adding a `TryFrom<(&str, &str)> for Proj` implementation in the `proj` crate? So one could do `transform(("EPSG:4326", "EPSG:3857"))` here?

urschrei:
> What if the argument was transform(transformer: TryInto<Proj>)? So you could pass it a Proj directly, and we could also consider adding a TryFrom<(&str, &str)> for Proj implementation in the proj crate, so you could do transform(("EPSG:4326", "EPSG:3857")) here.

I like it.

lnicola:
It's okay, but but really discoverable.

frewsxcv:
I agreed the `TryFrom` and `TryInto` trait implementations are hard to discover, so I would add documentation examples demonstration you can use multiple options here. I added similar docs for [`ToSocketAddr` a while back](https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html), which has multiple implementations.

michaelkirk:
To recap, I agree that accepting a `TryInto<Proj>` param is the most flexible.

And ("EPSG:4326", "EPSG:3857").try_into() is concise, but I don't think that I'd guess for it to exist. The docs you're planning to add will help.

But I usually go to the docs only *after* I fail to intuit the API. If we think that often people will want to use the initially proposed interface (from: &str, to: &str), then I think it makes sense to expose a more discoverable helper method, like this:

```
pub trait Transform<T> {
     type Output;

    /// naming insperation from  https://proj.org/development/reference/functions.html#c.proj_create_crs_to_crs
     fn transform_crs_to_crs(&self, from: &str, to: &str)
         -> Result<Self::Output, TransformError> {
        self.transform((from, to).try_into())
    }

     fn transform(&self, proj: TryInto<Proj>)
         -> Result<Self::Output, TransformError>;
 }
```

WDYT?

If it's controversial, don't worry about it. We could always add it later.

phayes:
I tried this out and it works great!

Question: Does it make sense to use `Proj::convert_array` instead of converting each point individually?   Could be faster to make one call to Proj instead of many, especially on large polygons etc.

michaelkirk:
Great point @phayes!

I *think* it's possible with the current API's we have (poly.interiors_mut(), line_string.0)... and if it's not possible we should consider making it so.

It's a perf improvement that doesn't seem like it would impact the API proposed here, so I'd also be fine with leaving that for a followup PR depending on how much time @frewsxcv wants to sink into this right now.

frewsxcv:
I don't have much time to push this over the finish line over the next few days. If anyone wants to take over and push to my branch, I'd say go for it. I'm also not opposed to merging as-is (ish) and following up with some of these changes.

phayes:
Thinking about the best performance here again, and I think the long-term solution is actually to add a `convert_inter_mut` method to `Proj`.  If you look at the array processing code here (https://github.com/georust/proj/blob/83fca44d25b0def3fe7b313d85cf638aa7fdf493/src/proj.rs#L802-L865), we could avoid some allocations by passing an `Into<Iterator> where Item: &'a mut Coord<F>` or something similar.    It also might  make sense to grab `num_coords` first and do the simple thing if there's a small number of coords. 

Anyways, long-term idea that doesn't affect the API surface here.

frewsxcv:
This is semi-blocked on https://github.com/georust/proj/pull/100 and a `proj` release

phayes:
What do folks think of doing something like this?

https://github.com/georust/proj/pull/105

frewsxcv:
Once https://github.com/georust/proj/pull/103 is reviewed, I will merge it and do a `proj` crate release. Then this PR will be ready for review.

frewsxcv:
This is ready for review!

geo/src/algorithm/transform.rs
```diff
@@ -0,0 +1,117 @@
...
+    fn from(e: proj::ProjCreateError) -> Self {
+        TransformError::ProjCreateError(e)
+    }
+}
+
+impl fmt::Display for TransformError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            TransformError::UnknownCrs => write!(f, "Unknown CRS"),
+            TransformError::ProjCreateError(err) => write!(f, "{}", err),
```
lnicola, line 111: Nit: `err.fmt(f)` might generate less code.

lnicola (approved):

geo/src/algorithm/transform.rs
frewsxcv: Good idea! Addressed in https://github.com/georust/geo/pull/718/commits/323717ab11d244866cc35257eee57e799598ec58

frewsxcv:
bors r=lnicola

PR #725, opened 2022-02-09
Add missing implementations for `Intersects`

msalib:
See #724 

- [X] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/master/CODE_OF_CONDUCT.md).
- [x] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

PR #726, opened 2022-02-11
bump MSRV to 1.55, add rust-1.58

michaelkirk:
---

MSRV for latest 'arbitrary' crate is 1.55

Alternatively, we **could** pin to an older version of `arbitrary` to keep our old MSRV, but in [previous discussion](https://github.com/georust/geo/pull/680#discussion_r753412550) we said we'd do our best to support stable + 3 releases, so I guess "onward" we should go.

FIXES https://github.com/georust/geo/pull/725#issuecomment-1034398133

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,10 +39,10 @@ jobs:
           # giving us about 6 months of coverage.
           #
           # Minimum supported rust version (MSRV)
-          - "georust/geo-ci:rust-1.51"
+          - "georust/geo-ci:rust-1.55"
           # Two most recent releases - we omit older ones for expedient CI
-          - "georust/geo-ci:rust-1.56"
           - "georust/geo-ci:rust-1.57"
+          - "georust/geo-ci:rust-1.58"
     container:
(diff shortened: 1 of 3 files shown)


lnicola:
We could also pin `arbitrary`, but I generally prefer aggressively bumping MSRV.

michaelkirk:
@lnicola  is that r + ? =)

lnicola:
No, it's missing a changelog entry (some people even expect major version bumps when increasing the MSRV, but I don't agree with them).

bors d+

michaelkirk:
bors r+

> No, it's missing a changelog entry (some people even expect major version bumps when increasing the MSRV, but I don't agree with them).

I've added a note to the changelog. It's awkward, because this bump of MSRV doesn't correspond to any changes that *we* made. There's *no* release (not even a minor one) associated with this bump of MSRV.

I think this is an argument for more restrictive versioning in our Cargo.toml - akin to https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277

PR #729, opened 2022-02-14
project tidy

michaelkirk:
- [-] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

An omnibus of project tidying - see individual commits for rationale.

If parts are controversial for some reason I can break this into multiple PR's


lnicola (approved):
I'm not sure the daily build is necessary, but I guess it's not that bad either.

michaelkirk:
> I'm not sure the daily build is necessary, but I guess it's not that bad either.

The idea is to try to preempt and clarify problems like this where a contributor gets a failing CI through no fault of their own code simply because CI hasn't run in a while: https://github.com/georust/geo/pull/725#issuecomment-1034404015

There's a risk that this change will be unhelpfully noisy. If that turns out to be true, I'm ok with removing it.

michaelkirk:
bors r=lnicola

.github/workflows/test.yml
```diff
@@ -1,4 +1,7 @@
-on: push
+on:
+  push:
```
lnicola, line 2: I don't know YAML, should this be `push`?

michaelkirk: I don't know YAML either. But I think it's working as expected. Do you have an example of where it's not?

Issue #555, opened 2020-12-04
Polygon-to-Polygon euclidean_distance hangs forever


leops:
I came across two polygons that do not seem to exhibit any "weird" characteristics (both are convex, have only a few points and do not overlap), but for some reason they hang the `euclidean_distance` function indefinitely.

# Reproduction

```rust
use geo::{polygon, prelude::*};

fn main() {
    let a = polygon!(
        exterior: [
            (x: -1.5350399, y: -0.18596762),
            (x: -0.33726108, y: 0.2206358),
            (x: -0.13836576, y: 0.42173427),
            (x: -0.33946428, y: 0.6206298),
            (x: -1.1405537, y: 0.8162205),
            (x: -1.3405507, y: 0.81511897),
            (x: -1.5372429, y: 0.21402647),
        ],
        interiors: []
    );

    let b = polygon!(
        exterior: [
            (x: 0.6649223, y: -0.17384565),
            (x: 0.8649193, y: -0.1727441),
            (x: 1.0616114, y: 0.4283484),
            (x: 0.8616146, y: 0.4272468),
            (x: 0.66271913, y: 0.22614834),
        ],
        interiors: []
    );

    a.euclidean_distance(&b);
}
```

Perhaps interestingly, calling `b.euclidean_distance(&a)` works correctly.

urschrei:
Hmm. This is almost certainly an issue with the rotating calipers code.

msalib:
Here's a smaller test case consisting of two disjoint rectangles that have been converted into polygons. `euclidean_distance` hangs because `min_poly_dist` hangs.

```rust
        let a = Rect::new((0., 0.), (1., 1.)).to_polygon();
        let b = Rect::new((2., 2.), (3., 3.)).to_polygon();
        min_poly_dist(&a, &b);
```

I have no idea what the fix is, but since this is supposed to be a linear algorithm, perhaps we can find a bound on the expected number of iterations and panic when we exceed that?

urschrei:
That's a good plan, but I'd be equally happy to just disable the fast path until I get around to investigating / debugging, which will probably be at least a couple of months away.

urschrei:
Fixed in #731. Apologies for the delay @leops.

PR #731, opened 2022-02-19
Fix polygon distance fast path

urschrei:
This replaces the previous, buggy iteration logic: we now rely upon the fact that this is a linear algorithm (thank you for the inspiration for the fix @msalib) so the minimum distance can be calculated by spinning the calipers at most n times, where n is the <s>number of edges in the larger of the two polygons</s> total number of vertices in both polygons' exteriors.

Other improvements include re-calculating the slope according to the same logic in the original algorithm, fixing a branching difference in our port, and new tests that cover both the original and minimised errors that were found.

# THAT HAVING BEEN SAID:

This algorithm's implementation is **extremely** magical and if it were proposed for addition today I would be dubious about its inclusion: why does the `100.00` value work in various functions, but `float::max_value()` doesn't? Why is the epsilon (`0.002`) comparatively large, and why does substituting `float::min_value()` not work? Nobody knows.

---

### DIFF
--- a/geo/src/algorithm/euclidean_distance.rs
+++ b/geo/src/algorithm/euclidean_distance.rs
@@ -576,6 +576,7 @@ mod test {
     use crate::algorithm::convex_hull::ConvexHull;
     use crate::algorithm::euclidean_distance::EuclideanDistance;
     use crate::{Line, LineString, MultiLineString, MultiPoint, MultiPolygon, Point, Polygon};
+    use geo_types::Rect;
     use geo_types::{polygon, private_utils::line_segment_distance, Coordinate};
 
     #[test]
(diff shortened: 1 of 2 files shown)


urschrei:
See #555 for previous discussion.

urschrei:
Bors try

urschrei:
Oops, need to fix num iterations to be n + m where n = poly1.exterior().points().len() and m = poly2.exterior().points().len().

See https://escholarship.mcgill.ca/downloads/3x816p60j?locale=en pp30-32

rzru:
@urschrei hi! looks like this is still broken in main branch. I wonder maybe this bugfix was lost during moving from master => main?

urschrei:
No, the bugfix and test that originally reproduced it are in `main`: [`here`](https://github.com/georust/geo/blob/main/geo/src/algorithm/euclidean_distance.rs#L1034-L1068). Do you have a failing test you can show me?

rzru:
@urschrei sure thing 

```rust
    #[test]
    fn test_2_polygons_stucks_forever() {
        let first_polygon = Polygon::new(
            LineString::new(vec![
                Coordinate::from((57.39453770777941, 307.60533608924663)),
                Coordinate::from((67.1800355576469, 309.6654408997451)),
                Coordinate::from((84.89693692793338, 225.5101593908847)),
                Coordinate::from((75.1114390780659, 223.45005458038628)),
                Coordinate::from((57.39453770777941, 307.60533608924663)),
            ]),
            vec![],
        );
        let second_polygon = Polygon::new(
            LineString::new(vec![
                Coordinate::from((138.11769866645008, -45.75134112915392)),
                Coordinate::from((130.50230476949187, -39.270154833870336)),
                Coordinate::from((184.94426964987397, 24.699153900578573)),
                Coordinate::from((192.55966354683218, 18.217967605294987)),
                Coordinate::from((138.11769866645008, -45.75134112915392)),
            ]),
            vec![],
        );

        assert_relative_eq!(first_polygon.euclidean_distance(&second_polygon), 0.0)
    }
```

this test should fail, distance is obvious not 0.0, however it just stucks forever

urschrei:
Reverting to the initial fix causes this test to pass. Urgh.

urschrei:
Hi @rzru, the fix will merge to `main` today – thanks for reporting and providing a test!

rzru:
hey @urschrei, thank you for your fast support and this amazing library!

PR #734, opened 2022-02-20
Use automatic default implementation

nyurik:
Make the code just a bit shorter by using the automatically generated `default` implementation (`#derive(Default)`)

---


frewsxcv (approved):

nyurik:
@frewsxcv sorry, seemed to have missed a T:default requirement. I wasn't sure if `Default` should be added as a requirement for `CoordNum` trait, so instead doing a partial revert but in a way to make it easier to remove manual default impl later.

michaelkirk:
bors retry

PR #735, opened 2022-02-20
Clean up trailing spaces at EOL

nyurik:
This should be a noop

---


nyurik:
This got included into #736

PR #740, opened 2022-02-22
Make clippy happy and add it to CI

### LINKED ISSUE #558: clippy in CI?

nyurik:
* Add clippy to CI pipeline - might as well let CI do its magic :)
* Applied a number of fixes suggested by clippy -- tests seem to still pass, and they do make sense:
  * `&Vec<Coordinate<f64>>` -> `&[Coordinate<f64>]`
  * switch to implicit returns
  * `&String` -> `&str`
  * many boolean `assert_eq!(true, ...)` -> `assert!(...)` -- code much more compact and easy to read
  * `points.iter().copied().collect()` -> `points.to_vec()`
  * a few useless clones on copyable objects
* Allow excessive float precision in tests (complex topic, didn't want it to trigger all the time)
* Bypass a bug in rustc clippy suggestion - that code might be fixed separately if desired
* Made a tiny improvement to `geo/examples/concavehull-usage.rs` by switching `String` to `&str`

- ~I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.~
---

fixes #558


geo/examples/concavehull-usage.rs
```diff
@@ -16,21 +16,21 @@ fn generate_polygon_str(coords: &Vec<Coordinate<f64>>) -> String {
...
         );
     }
-    return circles_str;
+    circles_str
 }
 
-fn produce_file_content(start_str: &String, mid_str: String) -> String {
-    let mut overall_string = start_str.clone();
-    overall_string.push_str(mid_str.as_ref());
+fn produce_file_content(start_str: &str, mid_str: &str) -> String {
```
lnicola, line 29: This could be a `format!`, but it doesn't matter.

lnicola (approved):

nyurik:
@lnicola i'm new to this proj, do i do the bors command myself, or do you (or someone else) usually does it?  Approval seems to not have triggered the auto-merge process

lnicola:
I can approve it, but I'd prefer if someone else could take a second look, just in case I missed something.

nyurik:
@lnicola ah, thx, just wasn't sure if i got the process right (was just reading the contribs doc) :)

geo/src/algorithm/closest_point.rs
```diff
@@ -184,6 +184,7 @@ impl<F: GeoFloat> ClosestPoint<F> for Geometry<F> {
 
 #[cfg(test)]
 mod tests {
+    #![allow(clippy::excessive_precision)]
```
michaelkirk, line 187: Please leave a comment wherever you turn a warning off.

michaelkirk (commented):
Some previous discussion here:
https://github.com/georust/geo/issues/558

The conclusion at the time was that clippy shouldn't fail CI. 

I don't have particularly strong feelings either way.

lnicola:
I tend to agree, in another (larger) project, clippy is pretty much useless because of the false positives, so someone runs it manually from time to time.

frewsxcv (approved):
I feel good trying this out. If it gets in the way and Clippy complains too much about inconsequential things we can always turn it off

geo/src/algorithm/closest_point.rs
nyurik: @michaelkirk what kind of a comment and where would you like me to add? thx :)

michaelkirk: Can you clarify why it is OK to turn this warning off?

nyurik: ah, got it, thx.  Well... actually I do not know this for a fact TBH. This warning is about floating point precision, which is a very tricky subject (might even be CPU dependent?). My thinking is that the floats in the tests were copy-pasted from the failed assertions, so whatever got "printed" during the failed test was expected to be the "right" result. Technically I think we should switch to using [assert_approx_eq!()](https://crates.io/crates/assert_approx_eq), but that would be a separate PR.  I will leave a comment about that.

frewsxcv: If you remove the unnecessary precision, do the tests still pass?

nyurik: @frewsxcv @michaelkirk ok, I reduced precision in the floats and removed the `allow` attributes, the tests seemed to have passed ok.  Feel free to merge if OK

PR #748, opened 2022-02-22
Run CI on pull requests

### LINKED ISSUE #747: CI does not run every PR
Currently new PRs, including those from members, do not automatically trigger a CI build. This means that a PR state is unknown until after someone gives bors the go-ahead, requiring additional cycles of fixes and re-approvals.

nyurik:
This ensures CI automatically runs on PRs, even if submitted from forks.

Doing this avoids a silly case of a PR being ready to merge and having a green "merge" button, only for the bors to start test&submitting it, realizing PR has bugs, and waiting for a fix from a user and another review from the maintainer.

---

fixes #747

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,10 @@
 on:
   push:
+    branches:
+      - main
+      - master
+      - staging
+      - trying
   pull_request:
   schedule: [cron: "45 6 * * *"]
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,6 @@
 on:
   push:
+  pull_request:
   schedule: [cron: "45 6 * * *"]
 
 name: Run tests


.github/workflows/test.yml
```diff
@@ -1,5 +1,6 @@
 on:
   push:
```
lnicola, line 2: We might want to do

```yaml
  push:
    branches:
    - master
    - staging
    - trying
```

here, so CI doesn't run twice for feature branches.

michaelkirk:
I know that this is a pretty popular thing to do on GH, but I've never understand how it reconciles with potential abuse. We don't currently have any access to special secrets in the CI build, so I guess the only thing we really have to worry about is a clogged CI queue.

I'm ok with merging it, but I'm also ok with our current system where contributors can push to the main repo to have CI build their branches on push.

lnicola:
There's a checkbox to prevent workflows from new contributors, we can enable that.

nyurik:
i would even restrict it for all outside collaborators - makes it far easier. And yes, restricting CI to the few key branches should also be done - i will update it

nyurik:
Actually come to think of it - yes, better restrict it just to the first timers until the repo has some real secrets (i.e. a secret to automatically publish to crates.io on new tag)

michaelkirk:
> There's a checkbox to prevent workflows from new contributors, we can enable that.

Oh wow TIL! IIRC this was the only thing that prevented us from using it previously, so I think we can go ahead and merge now, and revisit if necessary.

bors r+

michaelkirk:
Sorry - I see you want to make some changes to the branches. Merge whenever you're ready. 

bors delegate+

nyurik:
bors +r

nyurik:
yep, that's what i meant

bors r+

.github/workflows/test.yml
```diff
@@ -1,5 +1,11 @@
...
+    branches:
+      - main
```
lnicola, line 4: This made me look at the list of existing branches :smile:

PR #751, opened 2022-02-22
Require Default on all num types

nyurik:
This PR requires https://github.com/georust/proj/pull/110 or https://github.com/georust/proj/pull/111 to be released first. See https://github.com/georust/geo/discussions/746 of a potential solution.

* Add `Default` trait constraint to `CoordNum` and `CoordFloat`
* Use `CoordFloat` instead of `::num_traits::Float` for rstar
* A few minor conditional compilation cleanups

---

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -5,6 +5,7 @@
 * Add `LinesIter` algorithm to iterate over the lines in geometries.
   * Very similar to `CoordsIter`, but only implemented where it makes sense (e.g., for `Polygon`, `Rect`, but not `Point`).
   * <https://github.com/georust/geo/pull/757>
+* BREAKING: Add `Default` constraint to all `CoordNum`/`CoordFloat` values.
 
 ## 0.19.0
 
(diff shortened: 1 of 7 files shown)


geo/CHANGES.md
```diff
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-*
+* Add `Default` constraint to all `CoordNum`/`CoordFloat` values.
```
michaelkirk, line 5:
```suggestion
* BREAKING: Add `Default` constraint to all `CoordNum`/`CoordFloat` values.
```

This would be a breaking change for geo-types right? 

I'm not opposed to the new behavior, but it seems like a pretty small QOL improvement. 

As such, I'd advocate for holding off on merging this until we have a batch of other more significant reasons to break geo-types.

What do other people think?

nyurik: @michaelkirk  I'm ok to hold off the default trait. What about other changes - should i split them into a separate PR?

michaelkirk: > What about other changes

I'm not sure what remains. Do you mean this?

```diff
-    T: ::num_traits::Float + ::rstar_0_9::RTreeNum,
+    T: CoordFloat + ::rstar_0_9::RTreeNum,
```

Isn't that also a breaking change?

If there's some non-breaking changes that I'be missed that you'd like merged, I'd recommend splitting them out.

PR #753, opened 2022-02-25
Relax dependency requirements, test min-versions

nyurik:
Use complete patch version per Cargo recommendations such as [this one](https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277).
    
* Add minimum version testing (using nightly) to CI
* Relax geo-types requiremnets to 0.7.0 instead of 0.7.3
* Relax some external deps to their 0th patch version

---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -176,5 +176,5 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
-      - run: rustup toolchian install nightly-x86_64-unknown-linux-gnu
+      - run: rustup toolchain install nightly-x86_64-unknown-linux-gnu
       - run: cargo +nightly test -Z minimal-versions --all-features
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,6 @@ on:
   push:
     branches:
       - main
-      - master
       - staging
       - trying
   pull_request:
@@ -27,6 +26,7 @@ jobs:
       - geo_postgis
       - geo_fuzz
       - bench
+      - min_version
     steps:
       - name: Mark the job as a success
         if: success()
@@ -161,3 +161,19 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v2
       - run: cargo bench --no-run
+
+  min_version:
+    name: min_version
+    runs-on: ubuntu-latest
+    if: "!contains(github.event.head_commit.message, '[skip ci]')"
+    strategy:
+      matrix:
+        container_image:
+          # Only run it with the minimum supported rust version (MSRV)
+          - "georust/geo-ci:rust-1.56"
+    container:
+      image: ${{ matrix.container_image }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - run: cargo +nightly test -Z minimal-versions --all-features
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,6 +2,7 @@ on:
   push:
     branches:
       - main
+      - master
       - staging
       - trying
   pull_request:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -176,4 +176,5 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
+      - run: rustup toolchian install nightly-x86_64-unknown-linux-gnu
       - run: cargo +nightly test -Z minimal-versions --all-features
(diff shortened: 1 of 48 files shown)


nyurik:
> In the past, I sometimes would specific "more general" versions like this, but I think it was a misunderstanding on my part of how cargo resolves dependencies.
> 
> Instead, I think we should specify the patch versions. The reasoning is well articulated here: https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277
> 
> If we want to allow a wider swath of versions, we should specify the ".0" patch version and make sure we're testing that minimum version in CI.

Thanks @michaelkirk !  It was a bit of a surprise to me, didn't know it was actually non-version specific like that. I have relaxed the requirements instead, and added min-version testing CI (using nightly, hopefully will pass the build).

michaelkirk:
@nyurik: To make sure I'm understanding the state of this pr: You've append ".0" to all our dependencies which previously did not specify their patch version.

I think it's unlikely that those versions of our dependencies will build, so this change doesn't seem helpful. In fact, at least with the current state, which also specifies overly general versions, at least it seems clearer to the reader that the patch version is unspecified, rather than specifying a broken one.

Were you planning to add to this PR?

nyurik:
> I think it's unlikely that those versions of our dependencies will build, so this change doesn't seem helpful. In fact, at least with the current state, which also specifies overly general versions, at least it seems clearer to the reader that the patch version is unspecified, rather than specifying a broken one.

@michaelkirk I am not sure I'm following - I thought that's what you suggested in the previous comment here:

> Instead, I think we should specify the patch versions. The reasoning is well articulated here:

The only thing I think we really should implement is the minimum version testing (using nightly cargo feature). The patch versions just make it explicit what we already have there.

michaelkirk:
I was suggesting that these versions should specify whatever minimum versions will actually build. Adding a ".0" doesn't seem to do anybody any good, so I don't understand why we would merge it.

nyurik:
isn't `"0.5"` and `"0.5.0"` identical from the cargo's perspective? If so, we should be explicit that we support .0 version. The real issue is that we do NOT know if it actually builds or not because Cargo automatically picks the latest patch for everything.  I am not too set on this PR - more of a cleanup task to be explicit per that blog post. I can close it if you think its not needed.

frewsxcv:
> isn't `"0.5"` and `"0.5.0"` identical from the cargo's perspective? If so, we should be explicit that we support .0 version

Can you expand on why you think `0.5` implies that it doesn't support `0.5.0`? Regarding the word "explicit", I recommend [this blog post](https://without.boats/blog/things-explicit-is-not/), as what you think is explicit is not necessarily true for others.

nyurik:
> > isn't `"0.5"` and `"0.5.0"` identical from the cargo's perspective? If so, we should be explicit that we support .0 version
> 
> Can you expand on why you think `0.5` implies that it doesn't support `0.5.0`? Regarding the word "explicit", I recommend [this blog post](https://without.boats/blog/things-explicit-is-not/), as what you think is explicit is not necessarily true for others.

Why? Of course `0.5` supports `0.5.0` by definition. Good blog post BTW, thx, it does make a lot of sense.

That said, here I am simply following the recommendations by another [blog post](https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277) suggested by @michaelkirk above. It states that `serde = "1.0.130"  # good!` and `serde = "1"  # bad`.  Do we want to follow these recommendation, or do we not?  If not, then i'll close this PR. If we do, this PR would be a noop that simply follows blog recommendation.  I'm really confused by conflicting recommendations here, and don't have a preference either way.

Plus that blog recommends using `cargo +nightly -Z minimal-versions` as I did in #763, but it clearly fails because our current min versions are incorrect (likely in a sub-dependency).

frewsxcv:
Revisiting this– effectively nothing is currently broken for ourselves, or for our users. So what is a concrete example of a project that this pull request would be helping?

frewsxcv:
Also I'm slightly against testing with the minimum versions cargo flag. That seems like unnecessary extra work for us. We should optimize for easy maintenance for us, rather than optimizing for hypothetical scenarios.

lnicola:
I tend to agree. Testing with `min-versions` is the pedantic thing to do, but I would prefer not having to care about this.

I actually prefer `1.0` style versions, because they make it easier to know when to pay more attention to the new features (in case some of them would be useful). `1.0.209` would serve the same purpose, but I'd have to worry about using `1.1.10` (the current version when I upgrade the dependency) or doing a binary search to see if maybe `1.1.0` would work just as well.

nyurik:
ok, so there seems to be an agreement not to merge this or to add min-version CI test, so closing

PR #754, opened 2022-02-25
More tidying of fast path algorithm

urschrei:
Use built-in cross product instead of "left turn"
Simplify vertex line angle logic
Use accumulated caliper angle as a stopping condition
Use epsilon from num-traits instead of "small number"
Simplify bounds
Add link to original paper

I still don't understand why the magic "100" number works, but at least we're down to a single magic number, and I couldn't trigger a condition that causes it to diverge from the rtree-based result.

If anyone's got an idea for an algorithm that can generate valid polygons, this would be a good candidate for a property-based test:

1. generate two valid polygons 
2. generate their convex hulls 
3. calculate distance using fast path4 
4. compare to rtree-based version

---

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -39,6 +39,8 @@
   * <https://github.com/georust/geo/pull/836>
 * BREAKING: MapCoords/MapCoordsInPlace now map `Coordinate`s rather than `(x,y)` tuples
   * <https://github.com/georust/geo/pull/837>
+* Tidy fast-path distance algorithm
+  * <https://github.com/georust/geo/pull/754>
 
 ## 0.20.1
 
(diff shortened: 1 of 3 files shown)


rmanoka:
I tried to understand the `min_convex_poly_dist` fast-path, but couldn't make much progress on it.  Some food for thought:

1. Can we implement the original rotating calipers algo. on a single convex-polygon to iterate over all antipodal vertices without requiring the scalar `T: Float` ?  My gut feeling is yes.
2. If we could do (1), can the above min. distance between two convex poly be done using it as a sub-routine?
3. Would it help to also expose (1) as another algo. since [wiki page](https://en.wikipedia.org/wiki/Rotating_calipers) says it has many applications.

TL;DR:  the tidying up changes look good to me.

frewsxcv:
What's the status of this PR?

urschrei:
Needs a rebase but otherwise ready to go, @rmanoka's suggestions for future improvements notwithstanding.

PR #741, opened 2022-02-22
Switch to 2021 Rust edition, 1.56 MSRV

nyurik:
Update all crates to 2021 edition, and bump MSRV to 1.56. Note that this will not pass current CI because the `bench` target uses a docker image with Rust 1.53.

- [x] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/master/CODE_OF_CONDUCT.md).
- [x] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

Issue #681, opened 2021-11-20
RFC: Stabilize geo-types to 1.0


rmanoka:
This is a issue to gather discussion (started from #660) around eventually releasing a `geo-types` 1.0.

To streamline the discussions, we could list topics categorized as:

1. Blocker for 1.0
2. Nice to have before 1.0

(Feel free to add more topics)

frewsxcv:
I just went through the issues and labeled all the ones relevant to geo-types: https://github.com/georust/geo/labels/geo-types

It would be good to go through all of these and decide whether they're worth doing or not before a 1.0 release

dmarteau:
Please, could you consider going on  https://github.com/georust/geo/pull/797 before ? It's not a matter of having 3D in geo (which I don't want) but not beeing able to work with interoperable primitives is a really no-go for us to use geo-types (and other geo-types based crates) as first class type in a GIS framework.

Thanks.

PR #757, opened 2022-03-05
Add `LinesIter` algorithm to iterate over lines in a geometry

### LINKED ISSUE #708: Add `lines_iter` or `iter_lines` to `Polygon`
That iterates through all `Line`s in the exterior and interior rings of the `Polygon`

https://github.com/georust/geo/pull/701/files#r776821996

callpraths:
---

Newly introduced `LinesIter` is very similar to `CoordsIter`, but is implemented for only a subset of types (e.g., not for `Point`, and hence not for `Geometry`).

This PR also includes a small tweak to the order in which `Rect::to_lines()` returns the lines -- it now matches the winding order of the polygon created by `Rect::to_polygon()`.

Closes #708

### DIFF
--- a/geo-types/CHANGES.md
+++ b/geo-types/CHANGES.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-* Make `Rect::to_lines` return lines in winding order for `Rect::to_polygon`.
+* BREAKING: Make `Rect::to_lines` return lines in winding order for `Rect::to_polygon`.
   * <https://github.com/georust/geo/pull/757>
 * Macros `coord!`, `point!`, `line_string!`, and `polygon!` now support trailing commas such as `coord! { x: 181.2, y: 51.79, }`
   * <https://github.com/georust/geo/pull/752>
--- a/geo-types/CHANGES.md
+++ b/geo-types/CHANGES.md
@@ -2,8 +2,10 @@
 
 ## Unreleased
 
+* Make `Rect::to_lines` return lines in winding order for `Rect::to_polygon`.
+  * TODO: PR number TBD.
 * Macros `coord!`, `point!`, `line_string!`, and `polygon!` now support trailing commas such as `coord! { x: 181.2, y: 51.79, }`
-  * <https://github.com/georust/geo/pull/752> 
+  * <https://github.com/georust/geo/pull/752>
 
 ## 0.7.3
 
--- a/geo-types/CHANGES.md
+++ b/geo-types/CHANGES.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 * Make `Rect::to_lines` return lines in winding order for `Rect::to_polygon`.
-  * TODO: PR number TBD.
+  * <https://github.com/georust/geo/pull/757>
 * Macros `coord!`, `point!`, `line_string!`, and `polygon!` now support trailing commas such as `coord! { x: 181.2, y: 51.79, }`
   * <https://github.com/georust/geo/pull/752>
 
--- a/geo/src/algorithm/lines_iter.rs
+++ b/geo/src/algorithm/lines_iter.rs
@@ -15,9 +15,9 @@ pub trait LinesIter<'a> {
     /// # Examples
     ///
     /// ```
-    /// use geo::{Coordinate, Line};
     /// use geo::line_string;
     /// use geo::lines_iter::LinesIter;
+    /// use geo::{Coordinate, Line};
     ///
     /// let ls = line_string![
     ///     (x: 1., y: 2.),
@@ -26,16 +26,24 @@ pub trait LinesIter<'a> {
     /// ];
     ///
     /// let mut iter = ls.lines_iter();
-    /// assert_eq!(Some(Line::new(Coordinate{x: 1., y: 2.}, Coordinate{x: 23., y:82.})), iter.next());
-    /// assert_eq!(Some(Line::new(Coordinate{x: 23., y: 82.}, Coordinate{x: -1., y:0.})), iter.next());
+    /// assert_eq!(
+    ///     Some(Line::new(
+    ///         Coordinate { x: 1., y: 2. },
+    ///         Coordinate { x: 23., y: 82. }
+    ///     )),
+    ///     iter.next()
+    /// );
+    /// assert_eq!(
+    ///     Some(Line::new(
+    ///         Coordinate { x: 23., y: 82. },
+    ///         Coordinate { x: -1., y: 0. }
+    ///     )),
+    ///     iter.next()
+    /// );
     /// assert_eq!(None, iter.next());
     /// ```
     fn lines_iter(&'a self) -> Self::Iter;
 }
-
-// ┌────────────────────────────┐
-// │ Implementation for Line    │
-// └────────────────────────────┘
 
 impl<'a, T: CoordNum + 'a> LinesIter<'a> for Line<T> {
     type Scalar = T;
@@ -46,10 +54,6 @@ impl<'a, T: CoordNum + 'a> LinesIter<'a> for Line<T> {
     }
 }
 
-// ┌──────────────────────────────────┐
-// │ Implementation for LineString    │
-// └──────────────────────────────────┘
-
 impl<'a, T: CoordNum + 'a> LinesIter<'a> for LineString<T> {
     type Scalar = T;
     type Iter = LineStringIter<'a, Self::Scalar>;
@@ -59,7 +63,7 @@ impl<'a, T: CoordNum + 'a> LinesIter<'a> for LineString<T> {
     }
 }
 
-#[doc(hidden)]
+/// Iterator over lines in a [LineString].
 #[derive(Debug)]
 pub struct LineStringIter<'a, T: CoordNum>(slice::Windows<'a, Coordinate<T>>);
 
@@ -78,15 +82,11 @@ impl<'a, T: CoordNum> Iterator for LineStringIter<'a, T> {
         //
         // [RFC 2071]: https://rust-lang.github.io/rfcs/2071-impl-trait-existential-types.html
         self.0.next().map(|w| {
-            // slice::windows(2) is guaranteed to yield a slice with exactly 2 elements
+            // SAFETY: slice::windows(2) is guaranteed to yield a slice with exactly 2 elements
             unsafe { Line::new(*w.get_unchecked(0), *w.get_unchecked(1)) }
         })
     }
 }
-
-// ┌───────────────────────────────────────┐
-// │ Implementation for MultiLineString    │
-// └───────────────────────────────────────┘
 
 type MultiLineStringIter<'a, T> =
     iter::Flatten<MapLinesIter<'a, slice::Iter<'a, LineString<T>>, LineString<T>>>;
@@ -99,10 +99,6 @@ impl<'a, T: CoordNum + 'a> LinesIter<'a> for MultiLineString<T> {
         MapLinesIter(self.0.iter()).flatten()
     }
 }
-
-// ┌───────────────────────────────┐
-// │ Implementation for Polygon    │
-// └───────────────────────────────┘
 
 type PolygonIter<'a, T> = iter::Chain<
     LineStringIter<'a, T>,
@@ -120,10 +116,6 @@ impl<'a, T: CoordNum + 'a> LinesIter<'a> for Polygon<T> {
     }
 }
 
-// ┌────────────────────────────────────┐
-// │ Implementation for MultiPolygon    │
-// └────────────────────────────────────┘
-
 type MultiPolygonIter<'a, T> =
     iter::Flatten<MapLinesIter<'a, slice::Iter<'a, Polygon<T>>, Polygon<T>>>;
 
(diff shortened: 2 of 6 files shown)


geo/src/algorithm/lines_iter.rs
```diff
@@ -0,0 +1,328 @@
...
+        })
+    }
+}
+
+// ┌───────────────────────────────────────┐
+// │ Implementation for MultiLineString    │
+// └───────────────────────────────────────┘
+
+type MultiLineStringIter<'a, T> =
+    iter::Flatten<MapLinesIter<'a, slice::Iter<'a, LineString<T>>, LineString<T>>>;
```
callpraths, line 92: I would actually like to introduce newtypes for all these concrete `LinesIter::Iter` types because they currently leak implementation details (about how iterators are chained etc).

I followed the pattern in `CoordsIter` for this PR, but if there agreement, I'm happy to introduce newtypes for both `CoordsIter` and `LinesIter`.

geo/src/algorithm/lines_iter.rs
```diff
@@ -0,0 +1,328 @@
...
+}
+
+impl<'a, T: CoordNum> Iterator for LineStringIter<'a, T> {
+    type Item = Line<T>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // Can't use LineString::lines() because it returns an `impl Trait`
+        // and there is no way to name that type in `LinesIter::Iter` until [RFC 2071] is stabilized.
+        //
+        // [RFC 2071]: https://rust-lang.github.io/rfcs/2071-impl-trait-existential-types.html
```
callpraths, line 79: I am new(ish) to Rust, and I spent some time trying to wrap my head around this.

My current understanding is that as long as RFC-2071 is unstable, crate APIs should likely not use the `impl Trait` feature (in the return position, for existential types). Two reasons that I think I'd avoid this feature:
- Clients have no way to store the returned value in a struct / trait associated type (anywhere they must name the type).
- Conditional implementations (`#[derive(Debug)]` and friends) don't work well). In our case, `impl Iterator<Item=Line<T>` is not `Debug` even if both `T` and `Line` are).

Basing my understanding so far mostly off of https://www.reddit.com/r/rust/comments/8ik620/notes_on_impl_trait/

In this case, I was stuck because of the first issue -- I needed to name the type for `LinesIter::Iter`, but had no way to do it.

callpraths: Did I say I'm new to Rust? Please let me know if I'm understanding this wrong / I'm missing something obvious in how `impl Trait` works.

callpraths: Note to self: Update https://github.com/georust/geo/issues/681 in some form if `impl Trait` should be reconsidered in the public API of `geo-types`.

geo-types/CHANGES.md
```diff
@@ -2,8 +2,10 @@
 
 ## Unreleased
 
+* Make `Rect::to_lines` return lines in winding order for `Rect::to_polygon`.
```
lnicola, line 5:
```suggestion
* BREAKING: Make `Rect::to_lines` return lines in winding order for `Rect::to_polygon`.
```

geo/src/algorithm/lines_iter.rs
```diff
@@ -0,0 +1,328 @@
...
+    ///
+    /// let mut iter = ls.lines_iter();
+    /// assert_eq!(Some(Line::new(Coordinate{x: 1., y: 2.}, Coordinate{x: 23., y:82.})), iter.next());
+    /// assert_eq!(Some(Line::new(Coordinate{x: 23., y: 82.}, Coordinate{x: -1., y:0.})), iter.next());
+    /// assert_eq!(None, iter.next());
+    /// ```
+    fn lines_iter(&'a self) -> Self::Iter;
+}
+
+// ┌────────────────────────────┐
```
lnicola, line 36: Nit: I personally find these more distracting and not very useful, since they cover only one or two items.

geo/src/algorithm/lines_iter.rs
```diff
@@ -0,0 +1,328 @@
...
+impl<'a, T: CoordNum + 'a> LinesIter<'a> for LineString<T> {
+    type Scalar = T;
+    type Iter = LineStringIter<'a, Self::Scalar>;
+
+    fn lines_iter(&'a self) -> Self::Iter {
+        LineStringIter::new(self)
+    }
+}
+
+#[doc(hidden)]
```
lnicola, line 62: I believe `std` iterator structs aren't `#[doc(hidden)]`, and it's strange to use that for a type returned by the public API.

geo/src/algorithm/lines_iter.rs
```diff
@@ -0,0 +1,328 @@
...
+    /// use geo::lines_iter::LinesIter;
+    ///
+    /// let ls = line_string![
+    ///     (x: 1., y: 2.),
+    ///     (x: 23., y: 82.),
+    ///     (x: -1., y: 0.),
+    /// ];
+    ///
+    /// let mut iter = ls.lines_iter();
+    /// assert_eq!(Some(Line::new(Coordinate{x: 1., y: 2.}, Coordinate{x: 23., y:82.})), iter.next());
```
lnicola, line 29: Nit: these aren't formatted, but I don't think there's an easy way to fix that, so never mind.

geo/src/algorithm/lines_iter.rs
```diff
@@ -0,0 +1,328 @@
...
+impl<'a, T: CoordNum> Iterator for LineStringIter<'a, T> {
+    type Item = Line<T>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // Can't use LineString::lines() because it returns an `impl Trait`
+        // and there is no way to name that type in `LinesIter::Iter` until [RFC 2071] is stabilized.
+        //
+        // [RFC 2071]: https://rust-lang.github.io/rfcs/2071-impl-trait-existential-types.html
+        self.0.next().map(|w| {
+            // slice::windows(2) is guaranteed to yield a slice with exactly 2 elements
```
lnicola, line 81:
```suggestion
            // SAFETY: slice::windows(2) is guaranteed to yield a slice with exactly 2 elements
```

geo/src/algorithm/lines_iter.rs
```diff
@@ -0,0 +1,328 @@
...
+// ┌────────────────────────────┐
+// │ Implementation for Rect    │
+// └────────────────────────────┘
+
+impl<'a, T: CoordNum + 'a> LinesIter<'a> for Rect<T> {
+    type Scalar = T;
+    type Iter = <[Line<Self::Scalar>; 4] as IntoIterator>::IntoIter;
+
+    fn lines_iter(&'a self) -> Self::Iter {
+        // Explicitly iterate by value so this works for pre-2021 rust editions.
```
lnicola, line 148: We might switch to 2021 in #741.

geo/src/algorithm/lines_iter.rs
```diff
@@ -0,0 +1,328 @@
...
+// ┌────────────────────────────────┐
+// │ Implementation for Triangle    │
+// └────────────────────────────────┘
+
+impl<'a, T: CoordNum + 'a> LinesIter<'a> for Triangle<T> {
+    type Scalar = T;
+    type Iter = <[Line<Self::Scalar>; 3] as IntoIterator>::IntoIter;
+
+    fn lines_iter(&'a self) -> Self::Iter {
+        // Explicitly iterate by value so this works for pre-2021 rust editions.
```
lnicola, line 163: Ditto.

lnicola (approved):

geo-types/CHANGES.md
callpraths: Done.

geo/src/algorithm/lines_iter.rs
callpraths: I agree they are not useful. I was trying to follow the convetion from `coords_iter`. Removed.

geo/src/algorithm/lines_iter.rs
callpraths: Agreed. I had tried to follow what `coords_iter` does. Added a minimal docstring.

geo/src/algorithm/lines_iter.rs
callpraths: Thanks for calling that out. Formatted.

My hack: Copy into an actual function -> format -> copy back (with a little help from editor to add/remove line comments)

geo/src/algorithm/lines_iter.rs
callpraths: Done.

geo/src/algorithm/lines_iter.rs
callpraths: Should I add a TODO in the comment referencing the PR? I can remove it once 741 is merged.

geo/src/algorithm/lines_iter.rs
```diff
@@ -0,0 +1,307 @@
+use crate::{
+    CoordNum, Coordinate, Line, LineString, MultiLineString, MultiPolygon, Polygon, Rect, Triangle,
+};
+use core::slice;
+use std::fmt::Debug;
+use std::iter;
+
+/// Iterate over lines of a geometry.
+pub trait LinesIter<'a> {
```
frewsxcv, line 9: What do you think about implementing `LinesIter` for `Point` and `MultiPoint` that return empty iterators? Then we could implement `LinesIter` for `GeometryCollection`. I don't feel strongly about it, just an idea.

michaelkirk: It **would** be nice to have LinesIter on GeometryCollection, but I can't decide how surprising it would be for Point/MultiPoint...

My instinct is to hold off on including it for now. It would be a strictly additive change at some point if we wanted to add it.

geo-types/CHANGES.md
```diff
@@ -2,8 +2,10 @@
 
 ## Unreleased
 
+* BREAKING: Make `Rect::to_lines` return lines in winding order for `Rect::to_polygon`.
```
michaelkirk, line 5: This was probably oversight. Good catch! 

I personally don't think this is the kind of change we'd bump semantic versioning for. Does anyone disagree? 

I'm ok to leave "BREAKING" or whatever in the release notes, I'm just loathe to do anything that causes all our downstream crate users to be on an old version.

michaelkirk (approved):
The code all LGTM, but I want to leave it unmerged until someone else weighs in on not bumping semver for this change.

geo/src/algorithm/lines_iter.rs
callpraths: I was also conflicted about `Point` and `MultiPoint`. There is the better ergonomics of being able to iterate over the lines of a `Geometry`, but clients lose some help from compiler in catching cases where they probably don't want to iterate over lines in a `Point`.

I'd vote to not add it right now.

geo/src/algorithm/lines_iter.rs
callpraths: Added TODO

geo/src/algorithm/lines_iter.rs
callpraths: Added TODO.

Will keep an eye on #741 and #758 and update this once those are merged.

callpraths: To double check, we'll still need #741 to be merged, right? Because #758 doesn't actually bump MSRV, so our clients could still be on pre-2021 rustc.

michaelkirk: FYI #741 is building now. I expect it to be merged imminently. 

Sorry for the rigamarole as all the trains come into the station at once...

geo/src/algorithm/lines_iter.rs
frewsxcv: Sounds great, thanks for the feedback!

geo/src/algorithm/lines_iter.rs
callpraths: No worries. I'm happy to have converted a TODO to a to-done ;)

geo-types/CHANGES.md
callpraths: I don't have any context on downstream consumers yet, so I don't have a clear opinion here.

I think this falls in the "Possibly-breaking change" per [cargo book](https://doc.rust-lang.org/cargo/reference/semver.html) and in unstable packages, it's OK to bump just the patch version for these changes. The order of returned lines isn't part of an explicit contract, so someone depending on that order to *not* conform to the winding order is unlikely.

Let me know if I can help make this decision (perhaps start a thread on the channel?)

michaelkirk: In the interest of a certain amount of efficiency, I don't expect to get an explicit "OK" on all things. Typically when I'm about to do something potentially controversial, my approach is to make a proposal in the code review and solicit input. If no one complains (we're all async, so I wait a couple days) I go forward. 

If it were something that seemed potentially *very* controversial, I might wait for longer or for a more explicit OK from a larger group of people.

Since no one has yet replied to the contrary, I estimate that it's reasonable to move forward without bumping major version at this point.

callpraths:
Thank you for a great first PR experience!

PR #760, opened 2022-03-08
Use coord! macro instead of Coordinate ctor

nyurik:
This PR updates the `geo-types` crate. See also #761.

The `coord!` macro is much more flexible way to create a coordinate, and should be used in the same way as all other ones like `line_string` and `polygon`.

This change is a noop for now, but in the future it will make it easier to introduce Z coordinate without substantial code change.

---

### DIFF
--- a/geo-types/src/point.rs
+++ b/geo-types/src/point.rs
@@ -232,10 +232,10 @@ impl<T: CoordNum> Point<T> {
     /// # Examples
     ///
     /// ```
-    /// use geo_types::Point;
+    /// use geo_types::{coord, Point};
     ///
-    /// let point = Point::new(1.5, 0.5);
-    /// let dot = point.dot(Point::new(2.0, 4.5);
+    /// let point = Point(coord! { x: 1.5, y: 0.5 });
+    /// let dot = point.dot(Point(coord! { x: 2.0, y: 4.5 }));
     ///
     /// assert_eq!(dot, 5.25);
     /// ```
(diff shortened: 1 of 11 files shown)


geo-types/src/point.rs
```diff
@@ -232,10 +232,10 @@ impl<T: CoordNum> Point<T> {
     /// # Examples
     ///
     /// ```
-    /// use geo_types::{Coordinate, Point};
+    /// use geo_types::{coord, Point};
     ///
-    /// let point = Point(Coordinate { x: 1.5, y: 0.5 });
-    /// let dot = point.dot(Point(Coordinate { x: 2.0, y: 4.5 }));
+    /// let point = Point(coord! { x: 1.5, y: 0.5 });
```
pka, line 237: I would prefer
```
let point = Point::new(1.5, 0.5);
```
in doc comments.

nyurik: @pka seems like it would create more problems than solve -- https://github.com/georust/geo/runs/5472344274?check_suite_focus=true -- so reverting. Let's keep changes to a minimum per above, and consider other changes separately.

```
error[E0283]: type annotations needed for `geo_types::Point<{float}>`
  --> src/point.rs:237:13
   |
6  | let point = Point::new(1.5, 0.5);
   |     -----   ^^^^^^^^^^ cannot infer type for type `{float}`
   |     |
   |     consider giving `point` the explicit type `geo_types::Point<{float}>`, where the type parameter `{float}` is specified
   |
```

frewsxcv: How about using the `geo::point! { x: , y: }` macro?

frewsxcv: But yeah I'm fine keep it as you have written since you just did a find/replace, and this is unrelated

nyurik: @frewsxcv I agree that we should refactor it further in a subsequent PR, flattening macro usage. Feel free to r plus it when ready :)

frewsxcv (approved):
thank you!

PR #761, opened 2022-03-08
Use coord! macro instead of Coordinate ctor (part 2)

nyurik:
This PR updates `geo` crate.  See also #760.

The `coord!` macro is much more flexible way to create a coordinate, and should be used in the same way as all other ones like `line_string` and `polygon`.

This change is a noop for now, but in the future it will make it easier to introduce Z coordinate without substantial code change.

---

### DIFF
--- a/geo/examples/types.rs
+++ b/geo/examples/types.rs
@@ -1,9 +1,9 @@
 extern crate geo;
 
-use geo::{Coordinate, Point};
+use geo::{coord, Point};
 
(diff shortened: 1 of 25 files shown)


michaelkirk:
Is this PR instead of #760 or in addition to it?

nyurik:
@michaelkirk both are needed, they change different crates

frewsxcv (approved):

PR #763, opened 2022-03-10
Add min_version CI target

nyurik:
This is a test target we should add, but currently it seems like it is not building correctly and needs to be fixed as it is clearly a bug in our code.

---


nyurik:
closing per discussion in #753

PR #764, opened 2022-03-11
 Migrate jts-test-runner code

nyurik:
This is a work-in-progress to see if this approach is possible and workable.

Migrate [jts-test-runner](https://github.com/georust/jts-test-runner) code into this repo, while leaving the XML files in the original place. The xml files are used as a submodule. Eventually we may want to automate XML file download with the `build.rs` script.

The entire history was migrated (without the initial XML check-in). Additional changes were introduced at the end, see https://github.com/georust/geo/pull/764/commits/17b234a09996b402e69ba740130617522cbe5b6d

---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,6 +49,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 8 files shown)


nyurik:
closing in favor of #791

Issue #762, opened 2022-03-10
`CoordinatesIter` does not implement `ExactSizeIterator` correctly, panics


seritools:
This comes from the rust discord server, where someone tried to iterate over all but the last point of a LineString (being the exterior of a Polygon). In the meantime we realized just skipping the first one works just as fine for their usecase, but before that we stumbled upon this panic:

From <https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html>:

> When implementing an ExactSize[Iterator](https://doc.rust-lang.org/std/iter/trait.Iterator.html), you must also implement Iterator. When doing so, the implementation of [Iterator::size_hint](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint) *must* return the exact size of the iterator.

This means that something like

```rs
for coord in line_string.into_iter().rev().skip(1).rev() { ... }
```

panics, failing this assertion: https://doc.rust-lang.org/src/core/iter/traits/exact_size.rs.html#103-108

(It would also be nice for `PointsIter` to implement `ExactSizeIterator` as well I think :P)

I haven't looked at other iterators in the crate, but they might also have this issue.

urschrei:
Hi, could you provide a complete failing test case?

seritools:
Sure thing, here you go:

```rs
use geo::{Coordinate, LineString};

fn main() {
    let ls = LineString::<f64>(vec![
        Coordinate { x: 0., y: 0. },
        Coordinate { x: 10., y: 0. },
    ]);

    // reference to force the `impl IntoIterator for &LineString` impl, giving a `CoordinatesIter`
    for c in (&ls).into_iter().rev().skip(1).rev() {
        println!("{:?}", c);
    }
}
```

urschrei:
Thanks, just wanted to make sure I wasn't misunderstanding.

PR #766, opened 2022-03-12
Add missing size_hint method on LineString iterators

urschrei:
Fixes https://github.com/georust/geo/issues/762

---


michaelkirk:
Thanks for fixing this!

bors r+

PR #771, opened 2022-03-15
Generalize geo-type macros

nyurik:
Make all coordinate-specific logic limited just to the `coord!` macro, and all other macros will pass their parameters as is.

This change is a noop because `coord!` macro still uses the same parameter pattern as before.

This PR removes hardcoded x,y coordinates from the `point!`, `line_string!`, and `polygon!` macros, and instead each macro expects `token: expression` pair(s) that is passed as a parameter to the `coord!` macro.

---

### DIFF
--- a/geo-types/src/macros.rs
+++ b/geo-types/src/macros.rs
@@ -22,8 +22,8 @@
 /// [`Point`]: ./struct.Point.html
 #[macro_export]
 macro_rules! point {
-    (x: $x:expr, y: $y:expr $(,)?) => {
-        $crate::Point::new($x, $y)
+    ( $($tag:tt : $val:expr),* ) => {
+        $crate::Point ( $crate::coord! { $( $tag: $val , )* } )
     };
 }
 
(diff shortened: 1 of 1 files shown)


rmanoka (approved):
Nice; lgtm!

geo-types/src/macros.rs
```diff
@@ -284,3 +284,82 @@ macro_rules! polygon {
...
+            y: 3.4,
+        };
+        assert_eq!(p.x(), 1.2);
+        assert_eq!(p.y(), 3.4);
+    }
+
+    #[test]
+    fn test_line() {
+        let ls = line_string![(x: -1.2f32, y: 3.4f32)];
+        assert_eq!(ls[0], coord! {x: -1.2, y: 3.4});
```
lnicola, line 307: I think these should be formatted as:

```suggestion
        assert_eq!(ls[0], coord! { x: -1.2, y: 3.4 });
```

Regarding your proposal on the GeoRust Discord, this is just one way in which macros are still worse than normal Rust code. I'm strongly against deprecating the other constructors and conversions.

geo-types/src/macros.rs
```diff
@@ -304,7 +304,7 @@ mod test {
     #[test]
     fn test_line() {
         let ls = line_string![(x: -1.2f32, y: 3.4f32)];
-        assert_eq!(ls[0], coord! {x: -1.2, y: 3.4});
+        assert_eq!(ls[0], coord! { x: -1.2, y: 3.4 });
```
lnicola, line 307: There are 14 more instances.

nyurik: ah, thx, didn't look, fixed

frewsxcv (approved):

PR #772, opened 2022-03-15
remove all deprecated features from `geo-types`

nyurik:
This PR includes #798, which should be merged before this one

---

* BREAKING: Remove deprecated functions on the `Geometry<T>`:
  * `into_point` - Switch to `std::convert::TryInto<Point>`
  * `into_line_string` - Switch to `std::convert::TryInto<LineString>`
  * `into_line` - Switch to `std::convert::TryInto<Line>`
  * `into_polygon` - Switch to `std::convert::TryInto<Polygon>`
  * `into_multi_point` - Switch to `std::convert::TryInto<MultiPoint>`
  * `into_multi_line_string` - Switch to `std::convert::TryInto<MultiLineString>`
  * `into_multi_polygon` - Switch to `std::convert::TryInto<MultiPolygon>`
* BREAKING: Remove deprecated `CoordinateType` trait. Use `CoordFloat` or `CoordNum` instead.
* BREAKING: Remove deprecated functions from `LineString<T>`
  * Remove `points_iter()` -- use `points()` instead.
  * Remove `num_coords()` -- use `geo::algorithm::coords_iter::CoordsIter::coords_count` instead.
* BREAKING: Remove deprecated functions from `Point<T>`
  * Remove `lng()` -- use `x()` instead.
  * Remove `set_lng()` -- use `set_x()` instead.
  * Remove `lat()` -- use `y()` instead.
  * Remove `set_lat()` -- use `set_y()` instead.
* BREAKING: Remove deprecated `Polygon<T>::is_convex()` -- use `geo::is_convex` on `poly.exterior()` instead.
* BREAKING: Remove deprecated `Rect<T>::try_new()` -- use `Rect::new` instead, since `Rect::try_new` will never Error. Also removes corresponding `InvalidRectCoordinatesError`.

---


nyurik:
> There is a cost associated with maintaining two branches. I think it would make sense to do something like this once an 0.8 release is in sight, but we don't yet have a single compelling reason to break geo-types, so I'd prefer to hold off for now.

sounds good, I will maintain this PR until more breaking features have accumulated. BTW, note that there is already one breaking change merged to main with the winding order, and another breaking one that we now use Rust 2021 -- https://github.com/georust/geo/blob/master/geo-types/CHANGES.md

lnicola:
I don't think we're really in a position to maintain a separate maintenance release branch. It has its own costs, and looking at the recent changes, most of them have been new features and breaking changes in themselves, as opposed to bug fixes.

urschrei:
>I don't think we're really in a position to maintain a separate maintenance release branch.

Agreed: we do not have the resources to maintain a separate maintenance branch. We can revisit this when 0.8 is in sight – I would personally be fine with removing these features, but others may have a more conservative view.

frewsxcv (approved):
Considering we already have a breaking change in geo-types on `main`, we are already on course for a 0.8.0 release. So this looks good to merge to me. If we're hesitant on merging breaking changes, we should undo the breaking change on `main` so we can aren't blocking ourselves from releasing a new version of geo-types. @michaelkirk @urschrei I wanna make sure I have approval from you both before merging.

michaelkirk:
> Considering we already have a breaking change in geo-types on main

My understanding is that we don't yet have any changes merged that warrant a breaking bump to geo-types yet. 

--- 

> note that there is already one breaking change merged to main with the winding order

Please see the earlier discussion about why I didn't think this was a breaking change: https://github.com/georust/geo/pull/757#discussion_r821124529

The summary that made sense to me was:

> I think this falls in the "Possibly-breaking change" per [cargo book](https://doc.rust-lang.org/cargo/reference/semver.html) and in unstable packages, it's OK to bump just the patch version for these changes. The order of returned lines isn't part of an explicit contract, so someone depending on that order to not conform to the winding order is unlikely.

--- 

> another breaking one that we now use Rust 2021 --

Is upgrading to Rust 2021 a breaking change? To me that seems like taking advantage of any other new feature from an MSRV bump, and like with much of the rust community, we wouldn't consider a breaking change.

urschrei:
I'm OK with the change.

urschrei (approved):

frewsxcv:
@michaelkirk Gotcha, thanks for explaining. In that case, do you (or anyone) have any hesitations about a 0.7.4 release for geo-types?

michaelkirk:
> do you (or anyone) have any hesitations about a 0.7.4 release for geo-types?

SGTM!

frewsxcv:

PR #774, opened 2022-03-16
Add point!(x, y) macro

nyurik:
This follows up from #773, including all of its changes (so better to merge that one first for shorter diff).
    
This PR adds support for the positional `point!(x,y)` macro, making it appear more like a function.

---


nyurik:
abandoning as this doesn't seem as good of a path as using static method

PR #775, opened 2022-03-17
extend point! macro to support more ways to instantiate

nyurik:
To be consistent with other macros, `point!` can now be used as:

```
point! { x: <number>, y: <number> }
point!(<coordinate>)
```

Also replace a few `format!("{}", v)` with `v.to_string()`

See also https://github.com/georust/geo/pull/742#issuecomment-1068731936

---

### DIFF
--- a/geo-types/src/macros.rs
+++ b/geo-types/src/macros.rs
@@ -1,7 +1,6 @@
 /// Creates a [`Point`] from the given coordinates.
 ///
 /// ```txt
-/// point!(<x_number>, <y_number>)
 /// point! { x: <number>, y: <number> }
 /// point!(<coordinate>)
 /// ```
@@ -11,9 +10,14 @@
 /// Creating a [`Point`], supplying x/y values:
 ///
 /// ```
-/// use geo_types::point;
+/// use geo_types::{point, coord};
 ///
 /// let p = point! { x: 181.2, y: 51.79 };
+///
+/// assert_eq!(p.x(), 181.2);
+/// assert_eq!(p.y(), 51.79);
+///
+/// let p = point!(coord! { x: 181.2, y: 51.79 });
 ///
 /// assert_eq!(p.x(), 181.2);
 /// assert_eq!(p.y(), 51.79);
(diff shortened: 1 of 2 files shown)


rmanoka:
I support the improvements to the `point!` macro.   Let's give it a few days to seek more opinions on changing all constructions to use the macro: some users might like the explicit type-based (or `Point::from` which would still work even after aliasing)

nyurik:
@lnicola I made a simpler #777 which should be easier to merge. I will shrink this PR to just add the macro change

geo-types/src/macros.rs
```diff
@@ -1,7 +1,9 @@
 /// Creates a [`Point`] from the given coordinates.
 ///
 /// ```txt
-/// point!(x: <number>, y: <number>)
+/// point!(<x_number>, <y_number>)
```
frewsxcv, line 4: Not a blocker from merging if others are okay with it, but I'm personally in favor of requiring the user to specify `x` and `y` in the constructor instead of them having to memorize lng/lat ordering with these unnamed positional arguments. https://macwright.com/lonlat/

nyurik: @frewsxcv I agree with you here, neither am I, but we have lots of those all over the code in the form of `(x,y).into()` (as well as `[x,y]`, and the `point.x_y()` fn, etc.  I think we should either consistently abandon all positional ones, or make them consistently available.

rmanoka: It's a good point @frewsxcv .  If we removed the support for two positional args ie `point!(x, y)` in this PR, would a lot break?

nyurik: @rmanoka there are no usages of positional `point!(x,y)` because it doesn't exist. We don't have to introduce it, but there is currently several other functions that take positional x,y arguments, and they need to be switched to this macro in order to simplify z&m migration.

rmanoka: So your intention is to add support for point!(x,y,z) and point!(x,y,z,m) as positionals?

An alternative is not have positionals at all, and when migrating to the macros, use named args.  Shall we go with that?

nyurik: I think z and m are far less common, and don't need the short syntaxes. I'm ok to force named args everywhere, but there are quiet a few of the various positional instantions at the moment like .into and ::from. Should we keep those or deprecate them?

rmanoka: Yeah, let's track that in another issue, and take a call.

nyurik: thx, removed the positional one for now

rmanoka:
Ok great, regardless of the Z-M PR, this is an independently good addition, so I'm merging this.

bors r+

PR #779, opened 2022-03-17
Use `LineString::from()` instead of `LineString()`

nyurik:
Make migration simpler by using static function that will still be available if `LineString` becomes a type alias.

Similar to https://github.com/georust/geo/pull/777

---

### DIFF
--- a/geo-types/src/lib.rs
+++ b/geo-types/src/lib.rs
@@ -168,13 +168,13 @@ mod tests {
 
     #[test]
     fn polygon_new_test() {
-        let exterior = LineString(vec![
+        let exterior = LineString::from(vec![
             coord! { x: 0., y: 0. },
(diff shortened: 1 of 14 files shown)


nyurik:
@lnicola I'm beginning to wonder if this PR was a mistake, and not exactly sure how to test it without an extensive performance analysis.

```rust
        let exterior = LineString::from(vec![       coord! { x: 0., y: 0. }, ... ]);
```
calls this trait function

```rust
    fn from(v: Vec<IC>) -> Self {
        Self(v.into_iter().map(|c| c.into()).collect())
    }
```

but it seems compiler might not be able to just make it into a noop, instead creating a new vector and copying things into it?  Compared with the old constructor call like `LineString(vec![...])` which allocated data only once.  Do you have any insight into this?

PR #781, opened 2022-03-18
Use `LineString::new()` instead of `LineString()`

nyurik:
Make migration simpler by using static function that will still be available if `LineString` becomes a type alias.

Similar to https://github.com/georust/geo/pull/777

---

### DIFF
--- a/geo-types/src/line_string.rs
+++ b/geo-types/src/line_string.rs
@@ -35,7 +35,7 @@ use std::ops::{Index, IndexMut};
 /// ```
 /// use geo_types::{coord, LineString};
 ///
-/// let line_string = LineString(vec![
+/// let line_string = LineString::new(vec![
 ///     coord! { x: 0., y: 0. },
 ///     coord! { x: 10., y: 0. },
 /// ]);
@@ -83,7 +83,7 @@ use std::ops::{Index, IndexMut};
 /// ```
 /// use geo_types::{coord, LineString};
 ///
-/// let line_string = LineString(vec![
+/// let line_string = LineString::new(vec![
 ///     coord! { x: 0., y: 0. },
 ///     coord! { x: 10., y: 0. },
 /// ]);
@@ -100,7 +100,7 @@ use std::ops::{Index, IndexMut};
 /// ```
 /// use geo_types::{coord, LineString};
 ///
-/// let line_string = LineString(vec![
+/// let line_string = LineString::new(vec![
 ///     coord! { x: 0., y: 0. },
 ///     coord! { x: 10., y: 0. },
 /// ]);
@@ -120,7 +120,7 @@ use std::ops::{Index, IndexMut};
 /// ```
 /// use geo_types::{coord, LineString, Point};
 ///
-/// let line_string = LineString(vec![
+/// let line_string = LineString::new(vec![
 ///     coord! { x: 0., y: 0. },
 ///     coord! { x: 10., y: 0. },
 /// ]);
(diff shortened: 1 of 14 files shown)


geo-types/src/line_string.rs
```diff
@@ -191,6 +191,11 @@ impl<'a, T: CoordNum> DoubleEndedIterator for CoordinatesIter<'a, T> {
 }
 
 impl<T: CoordNum> LineString<T> {
+    /// Instantiate Self from the raw content value
+    pub fn from_raw(value: Vec<Coordinate<T>>) -> Self {
```
frewsxcv, line 195: I've typically seen `from_raw` used as a name in Rust when using raw pointers. What do you think about just calling the constructor `new`?

frewsxcv: https://doc.rust-lang.org/nightly/std/?search=from_raw

nyurik: @frewsxcv yeah, i'm not too happy about the name either.  The problem with `new` is that we should have a consistently named static function on all geo types for instantiating, and `new` is already taken at least for some of them.  An alternative of course is to switch to a macro

frewsxcv: Can you clarify what you mean by "`new` is already taken at least for some of them"?

lnicola:
I think ideally we'd have:

```rust
// might as well call it new, but new_raw is a pretty bad name
impl<T: CoordNum> From<Vec<Coordinate<T>>> for LineString<T> {
    fn from(v: Vec<Coordinate<T>>) -> Self {
        Self(v)
    }
}

// this is already available
impl<T: CoordNum, IC: Into<Coordinate<T>>> FromIterator<IC> for LineString<T> {
    fn from_iter<I: IntoIterator<Item = IC>>(iter: I) -> Self {
        Self(iter.into_iter().map(|c| c.into()).collect())
    }
```

But I tried this over @nyurik's previous PR and it requires updating some of the tests:

```diff
// notice that the allocation is gone, but we need to specify f32
-        let mut l1: LineString<_> = vec![(0., 0.), (1., 2.)].into();
+        let mut l1: LineString<f32> = LineString::from_iter([(0., 0.), (1., 2.)]);

diff --git i/geo-types/src/polygon.rs w/geo-types/src/polygon.rs
index 104e183..eb57811 100644
--- i/geo-types/src/polygon.rs
+++ w/geo-types/src/polygon.rs
@@ -442,14 +442,13 @@ impl<T: CoordFloat + Signed> Polygon<T> {
 impl<T: CoordNum> From<Rect<T>> for Polygon<T> {
     fn from(r: Rect<T>) -> Self {
        // again, we can avoid an allocation
         Polygon::new(
-            vec![
+            LineString::from_iter([
                 (r.min().x, r.min().y),
                 (r.max().x, r.min().y),
                 (r.max().x, r.max().y),
                 (r.min().x, r.max().y),
                 (r.min().x, r.min().y),
-            ]
-            .into(),
+            ]),
             Vec::new(),
         )
     }
```

geo-types/src/line_string.rs
nyurik: @frewsxcv 3 existing structs already have the new method. For some reason I thought more geotypes had it, so it seems it might be ok to add more.
* `Point::new(x: T, y: T)`
* `Polygon::new(mut exterior: LineString<T>, mut interiors: Vec<LineString<T>>)`
* `GeometryCollection::new()`

nyurik:
Per my above comment - seems like it should be ok to just add more `new(vec)` functions to other geotypes

nyurik:
I updated this PR to use `new()` fn instead. I think this should be a good solution for now, and can be merged. Going forward, it sounds like a good path to expand it to the iter and from coercion.

frewsxcv (approved):

PR #785, opened 2022-03-19
Merge wkt repository with history as /wkt dir

nyurik:
If we agree to merge this, the [wkt repo](https://github.com/georust/wkt) should be archived, and all issues should be moved to this repo. I believe it would be good to merge it here because:
* `geo-test-fixtures` requires `wkt`, and `wkt` requires `geo-types` - creating a circular dependency between two repos, making development more complicated and error-prone
* this is a commonly used geo feature
* this is a pure Rust lib

This merges the entire history of the wkt repository from

https://github.com/georust/wkt/tree/dbd80f22099fc11ad69a1d072c3516f9a1ca4376

as a `/wkt` directory, and adds the CI and adds path links from Crates.toml files:
* wkt/Cargo.toml - link to `/geo-types`
* geo-test-fixtures/Cargo.toml - link to `/wkt`

---
---


urschrei:
Why do you want to do this?

nyurik:
@urschrei I updated the top comment with the explanation -- mostly because there is a circular dependency between geo and wkt repos, and developing them in sync is becoming complicated.  Currently i have to essentially maintain a separate fork just to get the `geo` repo to compile.

frewsxcv:
How does moving the `wkt` crate into this repository solve the circular dependency problem?

I'd consider this to be a rather large change to this repository and the `wkt` repository. In the future, can you open a GitHub Issue or start a GitHub Discussion so we can discuss whether we even want them before reviewing any code?

nyurik:
@frewsxcv sure thing, will do.  I thought that since PR and issues are nearly identical in terms of discussion, and it was fairly simple to create this PR to see the code changes if we wanted to, while we can discuss it as if it was an issue.

Moving WKT does not solve circular dependency. What it does is it brings all components of the circular dependency into a single repo with a single CI. When a breaking change must be made to both modules, it can be done with a single PR.

Having two repos forces us to go through this process:
* create a PR for geo-types
* ignore `geo` CI errors and merge + publish it
* create a PR for `wkt`
* now `wkt` should pass CI since geo-types is already published, so merge+publish
* expect the `geo` to start passing CI again

rmanoka:
@nyurik can you explain this part more?

> ignore geo CI errors and merge + publish it

Why ignore CI?

nyurik:
> @nyurik can you explain this part more?
> > ignore geo CI errors and merge + publish it
> Why ignore CI?

There are two types of changes: those that can be done in steps, e.g. merge and release #786 that introduces `MultiPolygon::new()` and similar. Once published, `wkt` repo can be changed to use the new methods. There are also breaking changes which cannot IMO be done -- CI would not pass if geo repo introduces `PointTZM` - because the matching below wouldn't work

```diff
-/// assert!(matches!(my_type.geometry, Some(Point(_))));
+/// assert!(matches!(my_type.geometry, Some(PointTZM(_))));
```

Granted that there are very few of these, and in most cases they can be worked around, i.e. unit test in the doc can be removed, and a new wkt version would be released without that test, but this is no different than simply disabling certain parts of the CI during the migration, and re-enabling it later.

lnicola:
This isn't a circular dependency. `geo-test-fixtures` is a `dev-dependency` of `geo`, and uses `wkt` without the `geo-types` feature. Unless I'm missing something, we should be able to keep using whatever `wkt` version for reading the test fixtures.

nyurik:
> This isn't a circular dependency. `geo-test-fixtures` is a `dev-dependency` of `geo`, and uses `wkt` without the `geo-types` feature. Unless I'm missing something, we should be able to keep using whatever `wkt` version for reading the test fixtures.

It's a circular dependency in terms of passing ci. See my comment above

rmanoka:
I'm also not convinced yet, that this break CI.  @nyurik  can you point us to where this line you pasted above is:

```
assert!(matches!(my_type.geometry, Some(Point(_))));
```

I'm inclined against this PR, and in general pro-actively merging repos.  If anything, we should see a couple of PRs that do break CI due to circular deps, and try out the alternative / more cumbersome workflow before taking a call on this (imho).

lnicola:
> It's a circular dependency in terms of passing ci. See my comment above

I don't understand -- why does `wkt` care about how to instantiate `MultiPolygon`? It depends on an older version of `geo-types` (without `new`) and that code is feature-gated anyway.

nyurik:
> I'm also not convinced yet, that this break CI. @nyurik can you point us to where this line you pasted above is:
> 
> ```
> assert!(matches!(my_type.geometry, Some(Point(_))));
> ```
> 
> I'm inclined against this PR, and in general pro-actively merging repos. If anything, we should see a couple of PRs that do break CI due to circular deps, and try out the alternative / more cumbersome workflow before taking a call on this (imho).

@rmanoka see https://github.com/georust/wkt/pull/85/files#diff-0fb2eeea3273a4c9b3de69ee949567f546dc8c06b1e190336870d00b54ea0979R140

I created two PRs to demonstrate the difficulty of introducing any breaking changes that affect wkt -- see #790 and https://github.com/georust/wkt/pull/85

nyurik:
closing for now thx for @lnicola proposed workaround - see https://github.com/georust/geo/pull/742/files#diff-d540e35931c0109d7b7d24d53436d1f2e5d883fc264fcfa22d517b8a43ec7cbb

lnicola:
> [lnicola](https://github.com/lnicola) proposed workaround

Just to be clear, that's not the solution I proposed.

nyurik:
@lnicola in that case I must have misunderstood your idea - i thought you meant conversion of old geo-types to the new ones.  Could you sketch a prototype of your idea?

PR #791, opened 2022-03-22
merge jts_test_runner crate into this workspace

michaelkirk:
---

## background

The jts_test_runner crate currently lives at https://github.com/georust/jts-test-runner. It contains a bunch of geometry tests from the JTS project, expressed as xml. It also includes some test-harness code to run those tests.

## motivation

I initially kept jts_test_runner in a separate repository because it's (a little) large. But this separation introduces a circular dependency between geo and jts-test-runner, which makes updates [(a little) awkward](https://github.com/georust/jts-test-runner/pull/12#discussion_r832650534).

Integrating jts-test-runner into this repository avoids that awkwardness. A comparison of the size of a fresh clone of the repo before/after integration:
```
$ du -sh *
4.7M    before
 25M    after
 
# and looking at just the repo size
$ du -sh */.git
3.1M    before/.git
8.0M    after/.git
```

## Here's what changed

First, I imported the external history into the repo by running:

    git remote add jts-test-runner ../jts-test-runner
    git fetch jts-test-runner
    git merge -s ours --no-commit --allow-unrelated-histories jts-test-runner/master
    git read-tree --prefix=jts-test-runner/ -u jts-test-runner/master

The consequence is that we have all the original commits - which makes reviewing this PR a little confusing. So, I'll highlight my manual changes:

- 734e72b303c8a2cf46fcb9c11f027e55cb83f5e5
- ab6acc67

## alternatives considered

1. Make no structural changes. Continue to have updates be a little awkward. IMO having a patches section in jts-test-runner/Cargo.toml isn't a big deal, but it's been a stumbling block for more than one person. I could update the docs to make this clearer. 

2. Use a `git submodule`. This would keep the primary repo size down. But submodules are an unfortunate stumbling block, especially for new contributors who expect to be able to `git clone` and `cargo test` without further ceremony.

---

supersedes #764


nyurik (approved):
thank you!!

pka:
I'm a friend of combining related crates/workpaces in a single repository, but I see an argument, which speeks against it in this case. The JTS test runner could be of use for other crates besides geo for using (a subset of) these test cases. The geo-types specific part is tiny and it would be more natural to add support for other spatial crates in the jts_test_runner repo rather than in the geo repo. But maybe this is too hypothecal and reducing awkwardness is more important.

lnicola:
> The JTS test runner could be of use for other crates besides geo for using (a subset of) these test cases.

But these can still use it even if it's moved to a different repository, so that doesn't seem like such a big issue.

michaelkirk:
>  it would be more natural to add support for other spatial crates in the jts_test_runner repo rather than in the geo repo

That's a fair point. But, as you say, it's pretty hypothetical. Since the proposed layout wouldn't *prevent* that integration, I'm inclined to move forward with the more tangible benefits of merging the repos and revisit if another consumer of jts-test-runner appears.

I'll plan to merge this in a couple days unless I hear otherwise.

jts-test-runner/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+approx = "0.4.0"
+geo = { path = "../geo" }
+include_dir = { version = "0.6", features = ["search"] }
+log = "0.4.14"
+serde =  { version = "1.0.105", features = ["derive"] }
+serde-xml-rs = "*"
```
lnicola, line 12:
```suggestion
serde-xml-rs = "0.5"
```

(or `0.5.0` or `0.5.1`, if you prefer)

lnicola: But we can change this later on, no need to do it in this PR.

lnicola (approved):

jts-test-runner/Cargo.toml
michaelkirk: Good catch! updated.

PR #797, opened 2022-04-06
Implement 3D and Measure support for geo-types only

nyurik:
This PR includes #772

---

This PR focuses on `geo-types` only. This is a part of #742.

This PR changes the underlying geo-type data structures to support 3D data and measurement values (M and Z values). The PR attempts to cause relatively minor disruptions to the existing users (see breaking changes below). My knowledge of the actual geo algorithms is limited, so please ping me for any specific algo change.

## geo-types restructuring
All geo type structs have been renamed from `Foo<T>(...)` to `Foo<T, Z=NoValue, M=NoValue>(...)`, and several type aliases were added:

```rust
// old
pub struct LineString<T: CoordNum>(pub Vec<Coordinate<T>>);

// new
pub struct LineString<T: CoordNum, Z: ZCoord=NoValue, M: Measure=NoValue>(pub Vec<Coordinate<T, Z, M>>);

pub type LineStringM<T, M=T> = LineString<T, NoValue, M>;
pub type LineString3D<T> = LineString<T, T, NoValue>;
pub type LineString3DM<T, M=T> = LineString<T, T, M>;
```

## NoValue magic
`NoValue` is an empty struct that behaves just like a number. It supports all math and comparison operations. This means that a `Z` or `M` value can be manipulated without checking if it is actually there.  This code works for Z and M being either a number or a NoValue:

```rust
pub struct NoValue;

impl<T: CoordNum, Z: ZCoord, M: Measure> Sub for Coordinate<T, Z, M> {
    type Output = Self;

    fn sub(self, rhs: Self) -> Self {
        coord! {
            x: self.x - rhs.x,
            y: self.y - rhs.y,
            z: self.z - rhs.z,
            m: self.m - rhs.m,
        }
    }
}
```

## Variant algorithm implementations
Function implementations can keep just the original 2D `<T>` variant, or add support for 3D `<Z>` and/or the Measure `<M>`. The above example works for all combinations of objects. This function only works for 2D objects with and without the Measure.

```rust
impl<T: CoordNum, M: Measure> Line<T, NoValue, M> {
    /// Calculate the slope (Δy/Δx).
    pub fn slope(&self) -> T {
        self.dy() / self.dx()
    }
}
```

## Breaking changes
* It is no longer possible to create Coordinate with just `x` and `y` values using implicit tuple constructor. Use `coord!` instead.

---


geo-types/src/coordinate.rs
```diff
@@ -25,11 +26,25 @@ use approx::{AbsDiffEq, RelativeEq, UlpsEq};
...
+    pub m: M,
 }
 
+impl<T: CoordNum, Z: ZCoord, M: Measure> CoordTZM<T, Z, M> {
+    pub fn new(x: T, y: T, z: Z, m: M) -> Self {
+        Self { x, y, z, m }
+    }
+}
+
+pub type Coordinate<T> = CoordTZM<T, NoValue, NoValue>;
```
frewsxcv, line 42: One downside of the changes in this pull request is if the user wants to view the documentation for `Coordinate` (or `Coord`, `CoordM`, `CoordZ`, or `CoordZM`), they'll need to click through to `CoordTZM` to actually see it
Screen Shot 2022-04-09 at 12 55 06 AM

Screen Shot 2022-04-09 at 12 55 15 AM

One option here is to move the documentation to a separate Markdown file, and then we can add `#[doc = include_str!("../docs/coord.md")]` above each type alias here. Though the downside there is all the docs will be the same, which we may not want.

@nyurik What do you think?

nyurik: I think I will add proper documentation for each variant, with appropriate examples for each. Variant's description might be shorter than the main one, with a link to the main.

geo-types/src/lib.rs
```diff
@@ -64,61 +64,72 @@ extern crate rstar_0_8;
...
-pub trait CoordNum: CoordinateType + Debug {}
-#[allow(deprecated)]
-impl<T: CoordinateType + Debug> CoordNum for T {}
+pub trait CoordNum: Num + Copy + NumCast + PartialOrd + Debug {}
+impl<T: Num + Copy + NumCast + PartialOrd + Debug> CoordNum for T {}
 
 pub trait CoordFloat: CoordNum + Float {}
 impl<T: CoordNum + Float> CoordFloat for T {}
 
+pub trait ZCoord: NumOps + Zero + Copy + PartialEq + PartialOrd + Debug {}
```
frewsxcv, line 79: We should add docs for the `ZCoord` and `Measure` traits

Also, how come the trait bounds here (`NumOps + Zero + ...`) differ from `CoordNum`?

geo-types/src/novalue.rs
```diff
@@ -0,0 +1,95 @@
+#[cfg(any(feature = "approx", test))]
+use approx::AbsDiffEq;
+use num_traits::{One, Zero};
+use std::fmt::Debug;
+use std::ops::{Add, Div, Mul, Neg, Rem, Sub};
+
+#[derive(Eq, PartialEq, PartialOrd, Clone, Copy, Debug, Hash, Default)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct NoValue;
```
frewsxcv, line 9: Can we add some basic docs here, like:

```
/// Used in geometries where there is an absence of `Z` or `M` parameters.
```

geo-types/src/lib.rs
frewsxcv: > Also, how come the trait bounds here (`NumOps` + `Zero` + ...) differ from `CoordNum`?

Actually, is there any reason why we couldn't make `ZCoord` just extend `CoordNum`?

lnicola (approved):
I didn't see anything obviously wrong here, but it's too complicated for me to properly think about it these days. I'm not convinced this is the best design, but I can't suggest a better one.

nyurik:
Question for the community: for the types with the measure value, should  the `M` type be the same as `T`?  E.g. which of these types is better:

### `LineStringZM<T> = LineStringTZM<T, T, T>`
* PROs: usage simplicity (single generic type), possibly easier mathematics in various algos
* CONs: a bit less flexibility

### `LineStringZM<T, M> = LineStringTZM<T, T, M>`
* PROs: flexibility - measure could be any numeric value different from T
* CONs: always have to specify two generic types

lnicola:
I don't have much experience with these, but it feels that should be different types, even if in practice they might all be floats.

nyurik:
> I don't have much experience with these, but it feels that should be different types, even if in practice they might all be floats.

@lnicola it would still be possible to create values with different M types even if we have a single generic type like `LineStringZM<T>`. If some user wants them different, they can always use the "real" TZM value -- `LineStringTZM<T,Z,M>`, with any combination of the generic types.

This question is more about convenience i guess -- having a single generic type makes the new types easier to declare (you don't need to write `LineStringZM<f64,f64>`) -- but this is only relevant if **typically** M value is OK to be the same type.

lnicola:
I know it's an alias and traditionally they're all floats anyway, but it's not uncommon to define newtypes for different measures (angles, distances etc.). It seems more clear to have different types and show the user that they can put anything there instead of having an easy path of `f32` everywhere.

nyurik:
I looked at all the usages of M I can find - and it seems Postgis is the only major one, which uses identical type for M and others. I will simplify M for the most common usecase for now, unless we have evidence of a common real world use-case where M must be different. This will still allow any user to use TZM variant with any M values they want, even define their own struct that implements numeric operations.

nyurik:
Another thought- actually we can do both -- type aliases are cheap. Just need a good name, like PointMc for custom M

geo-types/src/coordinate.rs
nyurik: @frewsxcv thx for the feedback - I added some documentation to the variants.  I think we may want to do another PR with more doc fixes, or it might be too big and will make reviewing even harder.

geo-types/src/lib.rs
nyurik: In order to support `CoordNum` in `NoValue`, we would need to implement `ToPrimitive`, `NumCast`, and `Num` traits. Possible implementation below.  I am worried though that it will cause more harm than good -- i.e. `Num::from_str_radix()` shouldn't allow all possible values, etc. Also, having fewer traits ensures that algorithms that don't work with NoValues are not accidentally used in that way - e.g. someone wouldn't be able to create a `CoordTZM<NoValue, f64, NoValue>` (only Z value is given) -- we expect coordinate to always have real x,y values.

```rust
impl ToPrimitive for NoValue {
    fn to_i64(&self) -> Option<i64> {
        Some(0)
    }

    fn to_u64(&self) -> Option<u64> {
        Some(0)
    }
}

impl NumCast for NoValue {
    fn from<T: ToPrimitive>(_: T) -> Option<Self> {
        Some(Self::default())
    }
}

impl Num for NoValue {
    type FromStrRadixErr = ();

    fn from_str_radix(str: &str, radix: u32) -> Result<Self, Self::FromStrRadixErr> {
        Ok(Self::default())
    }
}
```

frewsxcv: Thanks for the response, I think I understand. I am going to check out your git branch in the next day or two and experiment with this so I have a better grasp of this

lnicola:
@nyurik why not `pub type LineStringM<T, M = T> = LineStringTZM<T, NoValue, T>;`?

nyurik:
> @nyurik why not `pub type LineStringM<T, M = T> = LineStringTZM<T, NoValue, M>;`?

@lnicola this is a great idea!  Moreover, I think we can get rid of the TZM types all together thanks to this approach -- instead of introducing a new `LineStringTZM`, we can simply define `struct LineString<T, M=NoValue, Z=NoValue> {...}`, plus introduce a few helper type aliases like `type LineStringM<T, M=T> = LineString<T, NoValue, M>;`

nyurik:
Success!  It seems we can introduce this change with far fewer breaking changes than originally anticipated, all thanks to generic parameter defaults.  We can keep the original geo type names as before, and just add two more generic type parameters with `NoValue` as default.

Also, it seems the moment the new `proj` version is released, we can even merge this **without** any `geo-types` unlinking/patching from the `proj` or `wkt`.

New structure (also updated in the description above): 

```rust
// old
pub struct LineString<T: CoordNum>(pub Vec<Coordinate<T>>);

// new
pub struct LineString<T: CoordNum, Z: ZCoord=NoValue, M: Measure=NoValue>(pub Vec<Coordinate<T, Z, M>>);
pub type LineStringM<T> = LineString<T, NoValue, T>;
pub type LineStringZ<T> = LineString<T, T, NoValue>;
pub type LineStringZM<T> = LineString<T, T, T>;
```

geo-types/src/coordinate.rs
```diff
@@ -1,12 +1,14 @@
...
+use std::fmt::Debug;
+use std::ops::{Add, Div, Mul, Neg, Sub};
 
-/// A lightweight struct used to store coordinates on the 2-dimensional
-/// Cartesian plane.
+/// A generic struct used to store a single coordinate with optional
+/// 3D (Z) and measurement support.
 ///
-/// Unlike `Point` (which in the future may contain additional information such
+/// Unlike [Point] (which in the future may contain additional information such
```
frewsxcv, line 11: FYI, we can retain the code formatting _and_ link to the docs here. Which I (subjectively) think is preferable since it indicates that we're talking about the `Point` API, and not just a random abstract notion of a 'point'

```suggestion
/// Unlike [`Point`] (which in the future may contain additional information such
```

geo-types/src/coordinate.rs
```diff
@@ -25,11 +27,37 @@ use approx::{AbsDiffEq, RelativeEq, UlpsEq};
...
-pub struct Coordinate<T: CoordNum> {
+pub struct Coordinate<T: CoordNum, Z: ZCoord = NoValue, M: Measure = NoValue> {
     pub x: T,
     pub y: T,
+    pub z: Z,
+    pub m: M,
+}
+
+impl<T: CoordNum, Z: ZCoord, M: Measure> Coordinate<T, Z, M> {
+    pub fn new(x: T, y: T, z: Z, m: M) -> Self {
```
frewsxcv, line 38: What do you think about adding docs here encouraging the user to utilize the `coord!` macro?

geo-types/src/coordinate.rs
```diff
@@ -365,3 +411,35 @@ where
         }
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_coordinates() {
```
frewsxcv, line 420: Thank you for adding tests!

geo-types/src/lib.rs
```diff
@@ -64,61 +64,83 @@ extern crate rstar_0_8;
...
+impl<T: Num + Copy + NumCast + PartialOrd + Debug> CoordNum for T {}
 
 pub trait CoordFloat: CoordNum + Float {}
 impl<T: CoordNum + Float> CoordFloat for T {}
 
+/// The type of the optional z value of a point/coordinate.
+///
+/// Floats (`f32` and `f64`) and Integers (`u8`, `i32` etc.) implement this.
+/// Also, an empty [`NoValue`] generic type can be used instead of the real value,
+/// allowing * types to avoid having Z value if it is not needed.
```
frewsxcv, line 83: What is the `*` here?

nyurik: good catch, fixed

geo-types/CHANGES.md
```diff
@@ -1,5 +1,28 @@
...
+* BREAKING: Remove deprecated `CoordinateType` trait. Use `CoordFloat` or `CoordNum` instead.
+* BREAKING: Remove deprecated functions from `LineString<T>`
+  * Remove `points_iter()` -- use `points()` instead.
+  * Remove `num_coords()` -- use `geo::algorithm::coords_iter::CoordsIter::coords_count` instead.
+* BREAKING: Remove deprecated functions from `Point<T>`
+  * Remove `lng()` -- use `x()` instead.
+  * Remove `set_lng()` -- use `set_x()` instead.
+  * Remove `lat()` -- use `y()` instead.
+  * Remove `set_lat()` -- use `set_y()` instead.
+* BREAKING: Remove deprecated `Polygon<T>::is_convex()` -- use `geo::is_convex` on `poly.exterior()` instead.
```
frewsxcv, line 22: What is `geo::is_convex`?

frewsxcv (commented):
Code-wise, this is looking great @nyurik! Amazing work 👏🏻 

One thing I'd like to verify before we merge is performance impact. Can you `cargo bench` your branch and `main` and share the results?

geo-types/CHANGES.md
nyurik: @frewsxcv this is part of the deprecation PR #772 -- https://github.com/georust/geo/pull/772/files#diff-3efed880fa0562aa828523c8b8024b977ea2e9a3ea079be1755cc8def1f84108L416-L417 -- i simply copied the deprecation note to the CHANGES.md

nyurik:
> One thing I'd like to verify before we merge is performance impact. Can you `cargo bench` your branch and `main` and share the results?

@frewsxcv theoretically there should be no perf impact, but more practically there are several issues:
* I tried running `cargo bench` on my laptop 3 times just on the main branch itself, and my results were inconsistent up to 10.1%.  That's **without** changing to any other branch, just rerunning on the main branch.
* Running `cargo bench` on this PR makes no sense because the geo-types used by the bench code is not the code that changed (this is the case where patching crate's location actually does make sense, esp because neither wkt nor proj are used).  I will try to run the benchmarks a few times to see what perf i get.

PS I created #809 to fix some annoying warnings during the `cargo bench`. Should be a noop.

nyurik:
OK, seems like the perf issues are gone with a few `#[inline]` attribs.  Ready to proceed, possibly after #812 to measure the impact of this PR more precisely (so far I have observed that some benchmarks actually perform better after this PR).

Also, right before merging i think it should be squashed to simplify history - please poke me to do it before running bors.

nyurik:
Reposting from Discord for more visibility: I just made a change to this PR after posting these thoughts.   Hopefully it will make things even cleaner, plus it allows empty structures for full WKT support. Original post below:

---

I keep thinking about one aspect of the 3D change -- how should we structure the traits for Z and M? Currently, I have two new traits - `ZCoord` and `Measure` . Technically, neither is really needed - they are nearly identical to `CoordNum`. The only reason the whole thing works is because `NoValue` implements those traits. 

```
pub trait CoordNum: Num + Copy + NumCast + PartialOrd + Debug {}
pub trait ZCoord: NumOps + One + Zero + Copy + PartialEq + PartialOrd + Debug {}
pub trait Measure: NumOps + One + Zero + Copy + PartialEq + PartialOrd + Debug {}
```

All structs follow this form:

```
struct Coordinate<T: CoordNum, Z: ZCoord = NoValue, M: Measure = NoValue> { x: T,  y: T,  z: Z,  m: M }
```

The only difference is in implementation - does it want to implement support for ALL generic types (i.e. the implementation supports real values for Z and M), or does it want to implement it only for X,Y, and keep Z/M as NoValue, e.g. 

```
impl<T: CoordNum, Z: ZCoord, M: Measure> Coordinate<T, Z, M> { /// works for real and missing Z/M
```
vs
```
impl<T: CoordNum> Coordinate<T> { /// assumes NoValue for Z/M
```

rmanoka:
@nyurik My vote would be to go for the simplest and extend it later as use cases arise.

michaelkirk (changes requested):
I ran out of time to review everything closely today, I’ll pick it back up tomorrow morning (PDT), but here’s my feedback so far.

overall things seem pretty clean!

rmanoka:
@nyurik can you also share benchmark reports before and  after the ZM changes?  It'll be good to see it's impact on algos.

geo-types/src/geometry.rs
```diff
@@ -207,6 +114,7 @@ macro_rules! try_from_geometry_impl {
     }
 }
 
+// `concat_idents` is not available, so hacking around it
```
michaelkirk, line 117: I don't understand what this comment is about. Are you explaining the trailing comma?

geo-types/src/geometry_collection.rs
```diff
@@ -23,9 +23,9 @@ use std::ops::{Index, IndexMut};
 /// ```
 /// use std::convert::TryFrom;
 /// use geo_types::{Point, point, Geometry, GeometryCollection};
-/// let p = point!(x: 1.0, y: 1.0);
+/// let p = point! { x: 1.0, y: 1.0 };
```
michaelkirk, line 26: Hrmm, why macro delimiter bikeshedding?

geo-types/src/lib.rs
```diff
@@ -64,61 +64,63 @@ extern crate rstar_0_8;
...
-pub trait CoordNum: CoordinateType + Debug {}
-#[allow(deprecated)]
-impl<T: CoordinateType + Debug> CoordNum for T {}
+pub trait CoordNum: Num + Copy + NumCast + PartialOrd + Debug {}
+impl<T: Num + Copy + NumCast + PartialOrd + Debug> CoordNum for T {}
 
 pub trait CoordFloat: CoordNum + Float {}
 impl<T: CoordNum + Float> CoordFloat for T {}
 
+mod novalue;
```
michaelkirk, line 79:
```suggestion
mod no_value;
```

Please rename the module to include a `_` between words

geo-types/src/line.rs
```diff
@@ -1,46 +1,76 @@
-use crate::{CoordNum, Coordinate, Point};
+use crate::{CoordNum, Coordinate, NoValue, Point};
+#[cfg(doc)]
+use crate::{Coordinate3D, Coordinate3DM, CoordinateM};
 #[cfg(any(feature = "approx", test))]
 use approx::{AbsDiffEq, RelativeEq};
 
-/// A line segment made up of exactly two
-/// [`Coordinate`s](struct.Coordinate.html).
+/// A generic line segment made up of exactly two [Coordinate] values.
```
michaelkirk, line 7: It's not clear what "generic" means in this context. It was generic before (<T>), I'd say just omit the word.

(and please restore the backticks around [`Coordinate`])

geo-types/src/line_string.rs
```diff
@@ -1,10 +1,9 @@
-#[cfg(any(feature = "approx", test))]
-use approx::{AbsDiffEq, RelativeEq};
-
-use crate::{CoordNum, Coordinate, Line, Point, Triangle};
+use crate::*;
```
michaelkirk, line 1: Keep the includes explicit please.

geo-types/src/multi_point.rs
```diff
@@ -1,11 +1,11 @@
-use crate::{CoordNum, Point};
-
+use crate::{CoordNum, NoValue, Point};
 #[cfg(any(feature = "approx", test))]
 use approx::{AbsDiffEq, RelativeEq};
-
 use std::iter::FromIterator;
 
-/// A collection of [`Point`s](struct.Point.html). Can
+/// A generic line string with 3D space + Measure value support.
```
michaelkirk, line 6: copy/paste errro - this isn't a line string 

(and same as I mentioned elsewhere, I dont think  the addition of the term "generic" is clear in the context of these docs)

geo-types/src/line_string.rs
```diff
@@ -1,10 +1,9 @@
...
 use std::iter::FromIterator;
 use std::ops::{Index, IndexMut};
 
+/// A generic line string with 3D space + Measure value support.
```
michaelkirk, line 5: Can you rephrase this (and all the similar geometry docs that you're updating) so that they remain geared primarily towards 2d people, with an addendum about 3D/M support.

Something like:
```
/// An ordered collection of two or more [`Coordinate`]s, representing a
/// path between locations.
///
/// LineStrings are 2D by default, but optionally support 3D and Measure parameters.
```

The overwhelming majority of people are going to be doing 2D, so we should keep are docs optimized for them.

michaelkirk (changes requested):
It seems like we might want to wait for https://github.com/georust/geo/issues/816 before merging this, but if you want to keep polishing this in the meanwhile, overall I'm happy to see it (eventually) get merged!

Other than the specific feedback I left, more generally, could you please review all the docs you've updated for the geometry types. It seems like there's a lot of copy and paste that happened and they don't always make sense. 

e.g. the rect docs say it's 3d and then immediately say it's 2d
```
 /// A generic bounded rectangle with 3D space + Measure value support.
 ///
 /// An _axis-aligned_ bounded 2D rectangle whose area is
 /// defined by minimum and maximum `Coordinate`s.
 ///
```

and the 3d triangle docs:
```
 /// A bounded 2D area whose three vertices are defined by
 /// `Coordinate`s. The semantics and validity are that of
 /// the equivalent [Polygon3D]; in addition, the three
 /// vertices must not be collinear and they must be distinct.
 pub type Triangle3D<T> = Triangle<T, T, NoValue>;
```

Please do another once-over on all those yourself.

nyurik:
Thanks for the thorough review! Should the inline pr merge as part of #816? Also, I suspect we may want to sprinkle a few more inlines in other geotypes.

geo-types/src/macros.rs
```diff
@@ -55,8 +55,22 @@ macro_rules! point {
 /// [`Coordinate`]: ./struct.Point.html
 #[macro_export]
 macro_rules! coord {
-    (x: $x:expr, y: $y:expr $(,)?) => {
-        $crate::Coordinate { x: $x, y: $y }
+    (x: $x:expr, y: $y:expr $(,)* ) => {
+        $crate::Coordinate::new__(
+            $x,
+            $y,
+            $crate::NoValue::default(),
```
michaelkirk, line 62:
```suggestion
            $crate::NoValue,
```
(and similar elsewhere)

michaelkirk:
> Thanks for the thorough review! Should the inline pr merge as part of https://github.com/georust/geo/issues/816? Also, I suspect we may want to sprinkle a few more inlines in other geotypes.

I don't think it should be part of #816, but it seems like it'd be good to get it out in the next non-breaking geo-types release if that's what you mean.

nyurik:
Yep, that's what I meant - to merge it for the non breaking release, and possibly do a few more inline prs for that release

geo-types/src/line_string.rs
nyurik: fixed

geo-types/src/macros.rs
nyurik: thx, didn't know empty structs can be instantiated without the default. Fixed.

geo-types/src/multi_point.rs
nyurik: fixed

geo-types/src/multi_polygon.rs
```diff
@@ -1,10 +1,11 @@
...
-
+use crate::{CoordNum, NoValue, Polygon};
 #[cfg(any(feature = "approx", test))]
 use approx::{AbsDiffEq, RelativeEq};
 use std::iter::FromIterator;
 
-/// A collection of [`Polygon`s](struct.Polygon.html). Can
+/// A generic collection of polygons with 3D space + Measure value support.
+///
+/// A collection of [Polygon]s. Can
```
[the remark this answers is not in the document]
nyurik: fixed

geo-types/src/novalue.rs
```diff
@@ -0,0 +1,133 @@
...
+        NoValue::default()
+    }
+
+    #[inline]
+    fn is_zero(&self) -> bool {
+        true
+    }
+}
+
+/// Thhese hacks allows mathematical operations that result in noop due to above ops
```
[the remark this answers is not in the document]
nyurik: fixed

geo-types/src/novalue.rs
```diff
@@ -0,0 +1,133 @@
...
+impl One for NoValue {
+    #[inline]
+    fn one() -> Self {
+        NoValue::default()
+    }
+}
+
+impl ToPrimitive for NoValue {
+    fn to_i64(&self) -> Option<i64> {
+        Some(0)
```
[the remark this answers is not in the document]
nyurik: thx, makes sense - i switched them all to None and Err for parsing.

geo-types/src/point.rs
```diff
@@ -1,16 +1,14 @@
-use crate::{point, CoordFloat, CoordNum, Coordinate};
+use crate::*;
```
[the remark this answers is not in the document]
nyurik: fixed

geo-types/src/point.rs
```diff
@@ -21,18 +19,58 @@ use std::ops::{Add, AddAssign, Div, DivAssign, Mul, MulAssign, Neg, Sub, SubAssi
...
+/// # Examples
+///
+/// ```
+/// use geo_types::{coord, Point3DM};
+/// let p = Point3DM::from(coord! { x: 10., y: 20., z: 30., m: 40. });
+/// ```
+pub type Point3DM<T> = Point<T, T, T>;
+
+impl<T: CoordNum, Z: CoordNum, M: CoordNum> From<Coordinate<T, Z, M>> for Point<T, Z, M> {
+    fn from(x: Coordinate<T, Z, M>) -> Self {
```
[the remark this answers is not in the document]
nyurik: I used `coords` for consistency with all other `from(...)` in the same file. Maybe in a separate cleanup they might be renamed to `coord` as more accurate?

geo-types/src/novalue.rs
```diff
@@ -0,0 +1,133 @@
+#[cfg(any(feature = "approx", test))]
+use approx::AbsDiffEq;
+use num_traits::{Num, NumCast, One, ToPrimitive, Zero};
+use std::fmt::Debug;
+use std::ops::{Add, Div, Mul, Neg, Rem, Sub};
+
+/// An empty placeholder type that can be used instead of the real
+/// numerical value types for 3D (z) and measurement (m) values.
```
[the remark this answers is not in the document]
nyurik: I fixed the comment. Which frewsxcv's suggestion are you referring to?

geo-types/src/point.rs
```diff
@@ -609,9 +647,36 @@ where
 #[cfg(test)]
 mod test {
     use super::*;
-
+    #[cfg(any(feature = "approx", test))]
```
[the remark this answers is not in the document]
nyurik: hmm, i think i saw a warning once in some obscure case. Not sure, deleting.

nyurik (commented):
thanks @michaelkirk for all the comments, addressed and merged from main.

frewsxcv (approved):

geo-types/src/geometry_collection.rs
michaelkirk: Ok, I'm willing concede that curly braces are now the *one true way* we should write these constructor macros.

I guess they're kind of evocative of "direct struct field" syntax, which is arguably useful. 👍

geo-types/src/coordinate.rs
```diff
@@ -232,21 +293,26 @@ use num_traits::Zero;
 /// assert_eq!(p.x, 0.);
 /// assert_eq!(p.y, 0.);
 /// ```
-impl<T: CoordNum> Coordinate<T> {
+impl<T: CoordNum, Z: CoordNum, M: CoordNum> Coordinate<T, Z, M> {
+    #[inline]
```
michaelkirk, line 297: I'm concerned that adding in all these inlines might cover up unrelated perf regressions. 

Can you separate the unrelated inlines into a separate PR?

geo-types/CHANGES.md
```diff
@@ -1,5 +1,30 @@
...
+  * Remove `points_iter()` -- use `points()` instead.
+  * Remove `num_coords()` -- use `geo::algorithm::coords_iter::CoordsIter::coords_count` instead.
+* Remove deprecated functions from `Point<T>`
+  * Remove `lng()` -- use `x()` instead.
+  * Remove `set_lng()` -- use `set_x()` instead.
+  * Remove `lat()` -- use `y()` instead.
+  * Remove `set_lat()` -- use `set_y()` instead.
+* Remove deprecated `Polygon<T>::is_convex()` -- use `geo::is_convex` on `poly.exterior()` instead.
+* Remove deprecated `Rect<T>::try_new()` -- use `Rect::new` instead, since `Rect::try_new` will never Error. Also removes corresponding `InvalidRectCoordinatesError`.
+* Replace deprecated `GeometryCollection::new()` with `GeometryCollection::new(value)`, and remove deprecated `GeometryCollection::new_from(value)`.
```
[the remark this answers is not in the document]
michaelkirk: I deprecated `new_from` in #818 so we should be able to delete it as part of whatever breaking release this goes out in.

geo-types/src/coordinate.rs
```diff
@@ -25,12 +27,45 @@ use approx::{AbsDiffEq, RelativeEq, UlpsEq};
...
+/// See also [`Coordinate`]
+pub type Coordinate3D<T> = Coordinate<T, T, NoValue>;
+
+/// A lightweight struct used to store coordinates on the 3-dimensional
+/// Cartesian plane together with a Measure value of the same type.
+///
+/// See also [`Coordinate`]
+pub type Coordinate3DM<T, M = T> = Coordinate<T, T, M>;
+
 impl<T: CoordNum> From<(T, T)> for Coordinate<T> {
```
michaelkirk, line 67: While working on #816 - I realized it probably doesn't make sense to implement this for all Coordinate<T>, rather only for Coordinate<T, NoValue, NoValue>. And same for `From([T; 2])` below and the similar impls in `Point`. 

We probably should eventually have a way to easily project 3d+ points to 2d, but it seems like it should be more explicit.

WDYT?

michaelkirk, line 67: While working on #816 - I realized it probably doesn't make sense to implement this for all Coordinate<T>, rather only for Coordinate<T, NoValue, NoValue>. And same for `From([T; 2])` below and the similar impls in `Point`. 

We probably should eventually have a way to easily project 3d+ points to 2d, but it seems like it should be more explicit.

WDYT?

michaelkirk:
Another thought, as I've been iterating on this, is it would be nice if we had an explicit use case to evaluate the 3d points against.

You know how sometimes an idea seems good in the abstract, but then once applied you've missed something obvious. 

Does anyone have a good imminent use case for 3d geometries that we could test our ideas against?

nyurik:
I just had a good chat with @komzpa about 3D and M support in PostGIS. He suggested we look at [ST_ClusterKMeans](https://postgis.net/docs/ST_ClusterKMeans.html) function which supports optional 3D and M values -- see [implementation](https://github.com/postgis/postgis/blob/25fe3dd0a651caeceb7aa3b10671d0c56e6dac1a/liblwgeom/lwkmeans.c#L359).  My limited understanding of how it works:
* Convert the input array of geometries/points to an array of 4D points - i.e. each geometry is reduced to a `(x, y, z, m)` tuple, where `m` is 1 by default, and `z` is 0 by default.
* The rest of the algorithm iteratively [works on the 4D point array](https://github.com/postgis/postgis/blob/25fe3dd0a651caeceb7aa3b10671d0c56e6dac1a/liblwgeom/lwkmeans.c#L149), eventually arriving to some result.

From the first glance, I believe we could optimize this algorithm with the NoValue approach, but with some modifications. Instead of a "default to 0" `NoValue`, we should introduce two empty structs:  **"default to 0"** for `z` and **"default to 1 (identity)"** for `m` generic types.  This assumes that all other algorithms treat M value as a weight-like parameter, so in case of a non-M geometries, `1` is a valid default.

If the above assumptions are not correct, or if we choose to prioritize `coord.x` over `coord.x()`, we would have to implement some complex wrapper that converts different types into a common 4D representation (like the linked C code), or we would have to introduce some internal representation that uses internal generic empty structs and copy values into them first.

Komzpa:
> tuple, where m is 1 by default, and z is 0 by default.

er, no. M and Z can be present or absent per-geometry, and Z=0 is not same as absent Z. Box with 0 as Z will only intersect geometries with Z=0, box with absent Z will intersect geometries with any Z values.

nyurik:
@michaelkirk @frewsxcv et al, i'm a bit stuck now.  This PR was approved and ready to merge, and now it is in a bit of a limbo. Since then, there has been a number of changes made to the repo that are fairly difficult to merge with this PR, and I really don't want to restart the merge process until there is an agreement with the direction, or else I'm risking spending many more hours on this only to see it not being merged again.  What are your thoughts?

urschrei:
Something I've been turning over in my mind as a related issue: we currently have algorithms (the vast majority) that operate on Euclidean space, (or more precisely on coordinates in the [real plane](https://en.wikipedia.org/wiki/Two-dimensional_Euclidean_space)). There are also some that operate on spherical coordinates, and I don't want to talk about them for now.

In any case, if we were to merge the minimum viable changes to allow `z` coordinates, we would also have to make it clear that all our existing algorithms ignore z, and potentially add a level to our algorithm module that distinguishes between algorithms that operate on 2D coordinates (**ignoring z values** but maintaining them where they exist), and algorithms that operate on 3D coordinates (and possibly won't accept 2D coordinates), in a more robust way than e.g. our EuclideanDistance / VincentyDistance split.

I haven't thought at all about 3D algorithms and I frankly don't intend to – it's simply not an area I work in – but I'd like to ensure that even if we do manage to make this change in a technically sound way, that we also work out a way that keeps the distinction clear for users of geo, the vast majority of whom are unlikely to need z coordinates.

urschrei:
I have had a go at rebasing this. It is neither correct nor complete, but probably…almost? 

https://github.com/georust/geo/tree/dim-types-rebased

nyurik:
Thanks @urschrei! I keep wondering if @michaelkirk idea might be better (not a hard change really) -- instead of having `Foo<T,Z,M>`, introduce a `Foo<T, C=Coordinate<T>>` -- where the Coordinate is still a struct with `{pub x: T, y: T}`.  This way far less code change, still can use `.x` and `.y` without the parenthesis because all `Foo` will implement the [Deref trait](https://doc.rust-lang.org/std/ops/trait.Deref.html).  This way we can very easily introduce `CoordinateZ` implemented as `{pub x:T, y:T, z:T}` or something with M, or even a custom implementation supplied by the user.

nyurik:
P.S. the only problem with the above is that all existing algorithms will continue to work in 2D space, so if the storage is `Coordinate3D`, which still has `x` and `y`, all algorithms will continue treating them as a 2D object, instead of refusing to compile.  A separate set of functions would be needed that are explicitly 3D focused - i.e. they will all use `.z` in addition to x and y.  Same goes for the functions that support `measure` param.  My initial approach would force each algorithm implementer to explicitly declare which variant they want to handle - i.e. all (ignore z & m), just 3D, just 2D, etc., and if someone tries to use them incorrectly, it would fail to compile.

urschrei:
>  all existing algorithms will continue to work in 2D space

Depending on your point of view, this could actually be a good thing. 2D is the default and – in my opinion, and in keeping with the industry standard, i.e. [JTS](https://locationtech.github.io/jts/jts-faq.html#B4) and GEOS specifically note this (see link) – _privileged_ coordinate space in `geo`. We could have an entirely separate crate for 3D algorithms, while allowing Z and M values to "come along for the ride" in `geo`. That would eliminate a lot of confusion instantly IMO.

nyurik:
@urschrei I think this is slightly more complex. Think of the following cases to compute a distance between two points `a` & `b` using `a.euclidean_distance(&b)`, for different types of `a` and `b`:
* `a` and `b` are in 2D:  all good, we have that now.
* `a` and `b` are in 3D:  silently drops z coordinate, and computes in 2D
* if the type of `a` is different from `b`, even if both have `x` and `y`, it probably won't compile, even though technically they should because euclidean_distance is designed as a 2D only operation, so it doesn't matter what other values are there

Compare this with a more explicit `euclidean_distance` that works differently depending on the coordinate type, and which refuses to compile if the operation is undefined.

I think silent drop of an internal value could lead to surprising bugs

urschrei:
But this is my point: Z values can be supplied, but they aren't used. This is what JTS does [ today](http://javadox.com/com.vividsolutions/jts/1.13/com/vividsolutions/jts/geom/Coordinate.html):

> Coordinates are two-dimensional points, with an additional Z-ordinate. JTS **does not support any operations on the Z-ordinate except the basic accessor functions**. If a Z-ordinate value is not specified or not defined, constructed coordinates have a Z-ordinate of NaN (which is also the value of NULL_ORDINATE). The standard comparison functions ignore the Z-ordinate.

This is what I'm asking: what if we allowed Z values in `Point` and `Coordinate`, kept Z values where possible when processing geometries with algorithms, but didn't _use_ them for _anything_ in `geo`. That's what JTS and GEOS do, and that's what people are expecting. What if 3D algorithms were available in a separate crate that used `geo-types`, but whose algorithms were only type-compatible with 3D coordinates?

Are there use cases that are possible outside the Rust ecosystem today that would then not be feasible in `geo`?

nyurik:
Thanks for the link!  That's exactly what I don't have a good answer to -- is the "silently ignore" a good approach in Rust?  We usually try to make the system as fool proof as possible simply because we have a lot more compilation control, and are able to catch many mistakes at the compile time.

If a person tries to do an `a.euclidean_distance(&b)` on two 3D points, what is the expectation? If the expectation is that it is always in 2D with the z coordinate dropping, we should support the case of `a` being 2D and `b` being 3D, without casting. If the expectation is that it computes the real distance (whatever that is), then a mix of 2D and 3D shouldn't work, but using two 3D points should use a different algorithm internally than for two 2D points.  I honestly don't know what the true expectation is for this and other algorithms we have.

urschrei:
> If a person tries to do an a.euclidean_distance(&b) on two 3D points, what is the expectation? If the expectation is that it is always in 2D with the z coordinate dropping, we should support the case of a being 2D and b being 3D, without casting. 

This is what I am _currently_ in favour of in `geo`. I'm just trying to figure out whether this matches other peoples' expectations, and whether this would actually enable broader compatibility with other software, and eventual adoption of `geo` as a low-level library as an alternative to GEOS.

frewsxcv:
Just like how we have a `# Units` top-level heading for some of our algorithms (like our area algorithms), we can have something like `# Dimensionality` or `# Dimensions` that makes it very clear to readers what dimensions the algorithm operates in. In my opinion that seems like enough clarity to users what they should expect.

nyurik:
If I understand it correctly, given two 3D points and these two choices:
* algorithm not designed to work with 3D silently treats it as 2D
* algorithm not designed to work with 3D fails to compile
we want to use the first approach?  This seems very anti-Rust design.  I think instead we should require an explicit cast to 2D to ensure the user really understands the implications.

frewsxcv:
I don't think it'd be possible to cast a 3D to 2D– you'd have to drop all the 3D coordinate values. Can you go into more detail with what you're thinking with that cast?

urschrei:
I think this what I'm wondering too. Is there an inherent disadvantage to what JTS and GEOS do (treat as 2D)? How would [some other approach] be an improvement?

nyurik:
> I don't think it'd be possible to cast a 3D to 2D– you'd have to drop all the 3D coordinate values. Can you go into more detail with what you're thinking with that cast?

That's exactly what I mean -- if an algorithm is only defined for 2D coordinates, and the object has 3D coordinates, I think there must be a way for the user to explicitly "downcast" it to 2D -- a way for the user to say "yes, I know I have 3D objects, but I want to ignore their z-coordinate when calling this algorithm which only understands 2D".

I am not certain of how we can achieve this -- I'm trying to state (and agree on) the ideal usage first. If not possible to implement, then sure, let's discuss next best thing.

So if we do want users to be explicit (IMO - a much safer approach), we could introduce a "2D object wrapper" that has the following properties:
* works transparently on all 2D objects, without any additional cast, e.g. ` a.euclidean_distance(&b)`
* requires a special casting for 3D objects, e.g. `a.as2d().euclidean_distance(b.as2d())` - where `as2d()` would be a readonly ref, and wouldn't actually allocate any new object.  Essentially this must be a noop when executing.
* The `as_2d()` should be available on all geometry types.

nyurik:
> I think this what I'm wondering too. Is there an inherent disadvantage to what JTS and GEOS do (treat as 2D)? How would [some other approach] be an improvement?

The main disadvantage is a potential for bugs. We should not rely on users reading documentation if we can make compiler do the work. When working with 3D coordinates, dropping to 2D algos should be an explicit desire, not auto-magical effect just because user accidentally forgot to check if the algo is 2d or 3d.  I suspect JTS/GEOS either didn't think it through, or maybe they did, but couldn't implement it in java/c++

bjornharrtell:
Long time user of JTS and it's C# port NTS here with a growing interest in Rust land. The Z handling is definitely an afterthought in JTS and I have a long standing issue that they have actually broken it, from my point of view, some time ago by introducing Z interpolation in many of the (purely 2D algorithms). For my use case ignoring Z with new coords produced by 2D algorithms set Z to undefined was great, it allowed me to post process Z in different ways. So I'm still trying to affect them to fix this situation.

That said I do wish for this PR to go forward. Not having Z in the geo primitives makes it less useful as defacto geometry types in Rust fx. to be an intermediate for other geometry representations.

urschrei:
> That said I do wish for this PR to go forward. Not having Z in the geo primitives makes it less useful as defacto geometry types in Rust fx. to be an intermediate for other geometry representations.

> PS. Note that JTS and its C# port NTS is providing the defacto geometry types in those ecosystems with lots and lots of software benefiting from being able to agree on a common interface.

I think that having Z available in some form is a question of "when", not "if", for exactly the reasons you note: I (and I don't think I'm alone in this, though I don't want to speak for any other core contributors) want to see geo-types and geo used more widely as foundational libraries.

That having been said, let me reiterate **my** opinion that focusing on 3D is a distraction. I want the semantics of interactions between (x, y) and (x, y, z) nailed down and clearly documented for library consumers _as they pertain to planar geometries and algorithms_, and then I want to get on with my life.

I don't particularly want 3D algorithms in `geo` in its current form (They can exist in separate modules or a separate crate – let's have a separate conversation about it, the decision isn't up to me).

frewsxcv:

Issue #721, opened 2022-01-29
Combine `MapCoords` and `MapCoordsInplace` into `MapCoords`


frewsxcv:
Similar to how `translate_inplace` exists on the `Translate` trait

michaelkirk:
I took a pass at this, but it ended up being trickier than I expected. Maybe someone else has some idea:

With this trait definition:
```
pub trait MapCoords<T, NT> {
    type Output;

    fn map_coords(&self, func: impl Fn(&(T, T)) -> (NT, NT) + Copy) -> Self::Output
    where
        T: CoordNum,
        NT: CoordNum;

    fn try_map_coords<E>(
        &self,
        func: impl Fn(&(T, T)) -> Result<(NT, NT), E> + Copy,
    ) -> Result<Self::Output, E>
    where
        T: CoordNum,
        NT: CoordNum;

    fn map_coords_in_place(&mut self, func: impl Fn(&(T, T)) -> (T, T))
        where
            T: CoordNum;

    fn try_map_coords_inplace<E>(
        &mut self,
        func: impl Fn(&(T, T)) -> Result<(T, T), E>,
    ) -> Result<(), E>
        where
            T: CoordNum;
}
```

And this trait impl:
```
impl<T: CoordNum, NT: CoordNum> MapCoords<T, NT> for Point<T> {
    type Output = Point<NT>;

    fn map_coords(&self, func: impl Fn(&(T, T)) -> (NT, NT) + Copy) -> Point<NT>
    {
        let new_point = func(&(self.0.x, self.0.y));
        Point::new(new_point.0, new_point.1)
    }

    fn try_map_coords<E>(
        &self,
        func: impl Fn(&(T, T)) -> Result<(NT, NT), E>,
    ) -> Result<Point<NT>, E>
        where
        NT: CoordNum
    {
        let new_point = func(&(self.0.x, self.0.y))?;
        Ok(Point::new(new_point.0, new_point.1))
    }

    fn map_coords_in_place(&mut self, func: impl Fn(&(T, T)) -> (T, T))
    {
        let new_point = func(&(self.0.x, self.0.y));
        self.0.x = new_point.0;
        self.0.y = new_point.1;
    }

    fn try_map_coords_inplace<E>(
        &mut self,
        func: impl Fn(&(T, T)) -> Result<(T, T), E>,
    ) -> Result<(), E>
        where
            T: CoordNum {
        let new_point = func(&(self.0.x, self.0.y))?;
        self.0.x = new_point.0;
[...]

PR #800, opened 2022-04-11
Introduce `TryMapCoordsInplace` algorithm.

frewsxcv:
---

Fixes https://github.com/georust/geo/issues/720.

This would be helpful in a couple places in [rgis](https://github.com/frewsxcv/rgis)

### DIFF
--- a/geo/src/algorithm/map_coords.rs
+++ b/geo/src/algorithm/map_coords.rs
@@ -194,7 +194,7 @@ pub trait TryMapCoordsInplace<T, E> {
     /// ```
     fn try_map_coords_inplace(
         &mut self,
-        func: impl Fn(&(T, T)) -> Result<(T, T), E> + Copy,
+        func: impl Fn(&(T, T)) -> Result<(T, T), E>,
     ) -> Result<(), E>
     where
         T: CoordNum;
@@ -236,7 +236,7 @@ impl<T: CoordNum> MapCoordsInplace<T> for Point<T> {
 impl<T: CoordNum, E> TryMapCoordsInplace<T, E> for Point<T> {
     fn try_map_coords_inplace(
         &mut self,
-        func: impl Fn(&(T, T)) -> Result<(T, T), E> + Copy,
+        func: impl Fn(&(T, T)) -> Result<(T, T), E>,
     ) -> Result<(), E> {
         let new_point = func(&(self.0.x, self.0.y))?;
         self.0.x = new_point.0;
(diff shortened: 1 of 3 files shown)


geo/src/algorithm/map_coords.rs
```diff
@@ -165,6 +165,45 @@ pub trait MapCoordsInplace<T> {
...
+    ///     ))
+    /// })?;
+    ///
+    /// assert_eq!(
+    ///     p1,
+    ///     geo::point!{x: 1010u32, y: 40u32},
+    /// );
+    /// # Ok::<(), &str>(())
+    /// ```
+    fn try_map_coords_inplace(
```
michaelkirk, line 195:
```suggestion
    fn try_map_coords_in_place(
```

I think "In place" should be spelled as two words, like: https://doc.rust-lang.org/std/ptr/fn.drop_in_place.html?search=in_place (7 matches)

michaelkirk: Wild addendum: We have a pattern in this module of "one method per trait": TryMapCoordsInplace, MapCoords, TryMapCoords, MapCoordsInplace

I think it might be more convenient for users to have all these grouped under a single MapCoords trait. Is there any reason you can think of where it'd be good to keep them separate? What would you think about, instead of introducing this new `TryMapCoordsInplace` trait, instead deprecating `TryMapCoords` and `MapCoordsInplace` and consolidating all their methods (with corrected spelling) on the `MapCoords` trait?
 
Because you're following the existing pattern, I think it's reasonable to merge as is, so consider this non-blocking feedback.

If you'd prefer not to make these changes in this PR, I'll probably follow up with my own PR proposing something like what I outlined above.

geo/src/algorithm/map_coords.rs
```diff
@@ -583,6 +813,27 @@ impl<T: CoordNum> MapCoordsInplace<T> for Triangle<T> {
     }
 }
 
+impl<T: CoordNum, E> TryMapCoordsInplace<T, E> for Triangle<T> {
+    fn try_map_coords_inplace(
+        &mut self,
+        func: impl Fn(&(T, T)) -> Result<(T, T), E>,
```
michaelkirk, line 819: This is another artifact of history - but I'm curious - based on your usage, is it actually convenient to deal with (T, T) tuples here rather than Coordinate<T>?

I think it makes sense to be consistent with all the other ones for now, but maybe something to consider for a future  breaking change or alternate API.

michaelkirk (approved):
LGTM, with some optional feedback.

geo/src/algorithm/map_coords.rs
frewsxcv: Regarding the naming, I noticed that too when I was writing this pull request. I named it this way to follow existing convention, but agree we should rename it. I opened this GitHub Issue to track that: https://github.com/georust/geo/issues/802

And regarding merging them into a single trait, I completely agree. There's already a separate issue for that: https://github.com/georust/geo/issues/721

geo/src/algorithm/map_coords.rs
frewsxcv: My preference is always to prefer working with structures with named fields when working with coordinates, as opposed to unnamed positional arguments. So yes, I do think it would be an improvement to change the tuple here to `Coordinate`. In fact we'll probably need to do that if/when the XYZM coordinate changes come in.

urschrei: I have a funny feeling that I'm responsible for this API, and that it's been somewhat inconvenient since the day it was introduced.

frewsxcv:
bors r=michaelkirk

PR #809, opened 2022-04-16
Hide benchmarking criterion html report warnings

nyurik:
Per criterion instructions, adding html to the crate requirements removes
the future compatibility warning (criterion will remove html by default in the next verion)

---


frewsxcv (approved):

michaelkirk:
That's a clever way to deprecate default functionality into a feature: https://github.com/bheisler/criterion.rs/blob/master/Cargo.toml#L74

Thanks!

bors r+

PR #810, opened 2022-04-16
Remove unneded reference for map coords closure parameter.

frewsxcv:
---


michaelkirk:
Note that this conflicts with #811. I'm happy to see both merged.

This might be a breaking change, right? Is there a way to make old clients work?

michaelkirk (approved):
LGTM

frewsxcv:
bors r=michaelkirk

Issue #802, opened 2022-04-13
Rename 'Inplace' → 'InPlace' and 'inplace' → 'in_place'


frewsxcv:
https://github.com/georust/geo/pull/800#discussion_r849478158

PR #811, opened 2022-04-16
Consolidate traits and rename for consistency with std lib spelling of "in_place"

michaelkirk:
---

Fixes: https://github.com/georust/geo/issues/802 and https://github.com/georust/geo/issues/721

---

Deprecated the old traits.

Note - the diff looks huge, but it's mostly whitespace changes because I put the new implementations into a new module.
Reminder that you can omit whitespace from the diff with https://github.com/georust/geo/pull/811/files?diff=unified&w=1

Without putting the new implementations into a new module, it'd live in the same namespace as the deprecated traits, causing references to `try_map_coords` to be ambiguous, which would require changing a couple dozen lines like: `p.try_map_coords(func)` to `MapCoords::try_map_coords(&p, func)` - I could do that instead if you prefer.

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -7,6 +7,10 @@
   * <https://github.com/georust/geo/pull/810>
 * BREAKING: Bump proj dependency to 0.26 which uses proj lib 9.0
   * <https://github.com/georust/geo/pull/813>
+* MapCoords restrucuring
+  - rename MapCoordsInplace::map_coords_inplace -> MapCoordsInPlace::map_coords_in_place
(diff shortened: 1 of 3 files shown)


frewsxcv (approved):
This all looks good to me! I think my change that I already merged might be a breaking change? If so, we should update the `since` attribute value to `0.21.0`

michaelkirk:
>  we should update the since attribute value to 0.21.0

Good point, done!

frewsxcv (approved):

michaelkirk:
bors r=frewsxcv

(rebased and resolved conflicts in CHANGES.md)

PR #812, opened 2022-04-18
Inline coords to boost perf improvements

nyurik:
* Inline all simple coordinate methods
* use `new()` method in some macros

# Performance improvements

test | improvement
---|---
concave hull f32 | -1.8840%
point outside polygon | -4.9574%
Polygon Euclidean distance rotating calipers f64 | -1.6313%
simplify vwp f32 | -7.0465%

the rest were statistically unchanged

---

### DIFF
--- a/geo-types/src/coordinate.rs
+++ b/geo-types/src/coordinate.rs
@@ -31,6 +31,7 @@ pub struct Coordinate<T: CoordNum> {
 }
 
 impl<T: CoordNum> From<(T, T)> for Coordinate<T> {
+    #[inline]
     fn from(coords: (T, T)) -> Self {
         coord! {
             x: coords.0,
@@ -40,6 +41,7 @@ impl<T: CoordNum> From<(T, T)> for Coordinate<T> {
 }
 
 impl<T: CoordNum> From<[T; 2]> for Coordinate<T> {
+    #[inline]
     fn from(coords: [T; 2]) -> Self {
         coord! {
             x: coords[0],
@@ -49,6 +51,7 @@ impl<T: CoordNum> From<[T; 2]> for Coordinate<T> {
 }
 
 impl<T: CoordNum> From<Point<T>> for Coordinate<T> {
+    #[inline]
     fn from(point: Point<T>) -> Self {
         coord! {
             x: point.x(),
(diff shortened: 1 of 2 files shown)


geo-types/src/coordinate.rs
```diff
@@ -30,7 +30,18 @@ pub struct Coordinate<T: CoordNum> {
     pub y: T,
 }
 
+impl<T: CoordNum> Coordinate<T> {
+    /// Create a new instance of a coordinate.
+    /// **ATTENTION** Use [`coord!`] macro or one of the [`from`] methods instead.
+    #[inline]
+    #[doc(hidden)]
+    pub fn new__(x: T, y: T) -> Self {
```
rmanoka, line 38: The fn new__ is the same as the prev. impl of the macro.  Why is it causing perf. regression?

geo-types/src/coordinate.rs
```diff
@@ -30,7 +30,18 @@ pub struct Coordinate<T: CoordNum> {
...
+        Self { x, y }
+    }
+}
+
 impl<T: CoordNum> From<(T, T)> for Coordinate<T> {
+    #[inline]
```
rmanoka, line 44: These are good inlines IMO!  Thanks!

rmanoka (changes requested):

geo-types/src/coordinate.rs
nyurik: I have no idea, but i did see the numbers change. Regardless, I will move this change to another PR to research separately from this PR.

geo-types/src/coordinate.rs
lnicola: Does `#[inline]` even do much for generic code?

nyurik: @lnicola that's a good question - but would require too substantial of a research to pinpoint why adding these inlines boosts performance. I think it is far easier to just add inline everywhere it seems relevant (tiny code, frequent use).

nyurik:
Note that the change requested by @rmanoka has been done (I simply removed the code in question) - i'm not sure if I should resolve it or if @rmanoka should click "ok"?

rmanoka (approved):

rmanoka:
LGTM.  Let's give it a couple of days for a few more eyes and merge it if no one has further suggestions

frewsxcv (approved):

PR #819, opened 2022-04-25
Add clarification on is_proper flag of LineIntersection

### LINKED ISSUE #664: LineIntersection trait may return incorrect is_proper flag
I'm looking to implement the [Bentley-Ottman] algorithm to fast compute all intersection of a set of lines.  The `LineIntersection` trait is quite relevant as it tries to be very robust while calculating, and neatly flags all the edge-cases.

I notice however that the [`is_proper`] flag may sometimes be set incorrectly.   The problem is that the calculations in [`proper_intersection`] may sometimes overflow, and returns one of the end points of the two lines.  In this case, it shouldn't be flagged as a proper intersection, but the logic seems to always set it to true if `proper_intersection` is used.

In fact, the test `test_central_endpoint_heuristic_failure_1` actually tests this case.  The expected output is one of the end points of the inputs (`line_1.start`), but it expects `is_proper` to be `true` which seems incorrect.  Since this comes from porting JTS code, I don't know if I'm missing something about the semantics.

[`is_proper`]: //github.com/georust/geo/blob/fd302719d381a7541db4fd5c79bec8e94ce5620f/geo/src/algorithm/line_intersection.rs#L14

[`proper_intersection`]: //github.com/georust/geo/blob/fd302719d381a7541db4fd5c79bec8e94ce5620f/geo/src/algorithm/line_intersection.rs#L141

[Bentley-Ottman]: //en.wikipedia.org/wiki/Bentley%E2%80%93Ottmann_algorithm

rmanoka:
- [X] ~I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.~
---

closes #664

### DIFF
--- a/geo/src/algorithm/line_intersection.rs
+++ b/geo/src/algorithm/line_intersection.rs
@@ -17,7 +17,7 @@ pub enum LineIntersection<F: GeoFloat> {
         ///
         /// Due to the limited precision of most float data-types, the
         /// calculated intersection point may be snapped to one of the
-        /// end-points even though the all the end-points of the two
+        /// end-points even though all the end-points of the two
         /// lines are distinct points. In such cases, this field is
         /// still set to `true`. Please refer test_case:
         /// `test_central_endpoint_heuristic_failure_1` for such an
--- a/geo/src/algorithm/line_intersection.rs
+++ b/geo/src/algorithm/line_intersection.rs
@@ -12,6 +12,16 @@ pub enum LineIntersection<F: GeoFloat> {
         /// For Lines which intersect in a single point, that point may be either an endpoint
         /// or in the interior of each Line.
         /// If the point lies in the interior of both Lines, we call it a _proper_ intersection.
+        ///
+        /// # Note
+        ///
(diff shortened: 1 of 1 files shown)


geo/src/algorithm/line_intersection.rs
```diff
@@ -12,6 +12,16 @@ pub enum LineIntersection<F: GeoFloat> {
         /// For Lines which intersect in a single point, that point may be either an endpoint
         /// or in the interior of each Line.
         /// If the point lies in the interior of both Lines, we call it a _proper_ intersection.
+        ///
+        /// # Note
+        ///
+        /// Due to the limited precision of most float data-types, the
+        /// calculated intersection point may be snapped to one of the
+        /// end-points even though the all the end-points of the two
```
michaelkirk, line 20:
```suggestion
        /// end-points even though all the end-points of the two
```

michaelkirk (approved):
A little typo, but otherwise, LGTM. Thanks for documenting this.

rmanoka:
bors r=michaelkirk

PR #826, opened 2022-05-06
Work in progress - remove Copy bound

frewsxcv:
---


geo-types/src/rect.rs
```diff
@@ -170,7 +170,7 @@ impl<T: CoordNum> Rect<T> {
     /// assert_eq!(rect.width(), 10.);
     /// ```
     pub fn width(self) -> T {
-        self.max().x - self.min().x
+        self.clone().max().x - self.min().x
```
michaelkirk, line 173:
```suggestion
        self.max().x.clone() - self.min().x
```

(and similarly throughout this file)

PR #827, opened 2022-05-06
rstar::object for geo_types::Polygon

michaelkirk:
---
superesedes #351

Introduced an rstar::Object implementation for Polygon. This was originally proposed in https://github.com/georust/geo/pull/351,  but that PR has since grown stale because we now support two versions of rstar.

based on #351, this PR:
- merges in main
- update polygon impl to work with the multiple versions of rstar we now support


rmanoka (approved):
LGTM

lnicola (approved):

Issue #822, opened 2022-05-04
Is the performance of boolean operation algorithms good enough?


erikpols:
I've seen a mention of maybe replacing the boolean operation algorithms for example here:
https://github.com/georust/geo/issues/620

What is the general state of the boolean operation algorithms from a performance perspective in this crate?

Reason is I ran a (somewhat simple) comparison test between this crate and npm's [Turf](https://www.npmjs.com/package/@turf/turf) crate. The intersect algorithm on polygons of Turf seems to be magnitudes faster even though it is written in typescript. I had a look at the source, and one reason I saw is the turf algorithm does a bounding box comparison before the actual intersect.  

I would have no problem doing this myself, but it leads me to the question: Are you guys happy with the current algorithms? What can I expect in the future?

Add-on question: Is it possible to retrieve the resulting intersection when doing the intersection between two polygons?

rmanoka:
@erikpols We're happy to improve the performance of our impls.  The bbox. heuristic has been discussed in #649 before;  we could discuss it there.  Reg. retreiving the intersection:  it is not available here, but refer #620 for other options.

erikpols:
ok ty. Those links are super helpful to me, I wasn't aware of geo-clipper and it makes all the difference. I understand its a c++ implementation but for me that's ok.

But just FYI it was a 15-30x difference

rmanoka:
@erikpols Could you share a minimal repo. of your bench-marking code?  Would be very helpful when we address the bbox. issue.

urschrei:
@rmanoka The easiest thing to would be to benchmark and test `geo-crossings` for this, right? Turf uses Martinez-Rueda (IIRC), but Bentley-Ottmann should be comparable in terms of speed.

rmanoka:
Yeah, we have some benchmarks in `geo-crossings` that show r-star heuristics is the best "edge-intersector" used by the `Intersects` and `Relate` impls.  If we can get some good dataset to benchmark against, we can have a before-after comparison and try adding it.

michaelkirk:
I prototyped an integration of rtree into geo here to see what kind of performance wins we might get:
https://github.com/georust/geo/tree/mkirk/rstar-edge-set-intersector

A good benchmark suite to compare against would be great. Lacking good benchmarks, here are some mediocre ones from [benches/relate.rs](https://github.com/georust/geo/blob/mkirk/rstar-edge-set-intersector/geo/benches/relate.rs) :

## relate overlapping 50-point polygons: 
```
[...]

PR #828, opened 2022-05-06
Speed up intersection check by first comparing bounding boxes

michaelkirk:
---

Checking intersections can be quite expensive.

Before doing the expensive intersection check, we now first build up bounding boxes and check if those intersect.

If you want to browse the reports/index.html: [criterion.zip](https://github.com/georust/geo/files/8643913/criterion.zip)

The summary is that most benches are unchanged, none are worse, but the new `intersections` bench improves a few significantly:

### intersection

Screen Shot 2022-05-06 at 4 11 02 PM

`euclidean_distance`, which relies on intersection checks, also improved its benches:

### Polygon Euclidean distance RTree f64

Screen Shot 2022-05-06 at 4 09 45 PM

### Polygon Euclidean distance rotating calipers f64

Screen Shot 2022-05-06 at 4 10 40 PM


geo-test-fixtures/src/lib.rs
```diff
@@ -101,6 +101,24 @@ where
     line_string("shell.wkt")
 }
 
+// From https://geodata.nationaalgeoregister.nl/kadastralekaart/wfs/v4_0?request=GetFeature&service=WFS&srsName=EPSG:4326&typeName=kadastralekaartv4:perceel&version=2.0.0&outputFormat=json&bbox=165593,480993,166125,481552
+pub fn nl_zones<T>() -> MultiPolygon<T>
```
michaelkirk, line 105: These are the polygons from the example in https://github.com/georust/geo/issues/822

geo/src/algorithm/intersects/collections.rs
```diff
@@ -13,6 +14,7 @@ where
     Rect<T>: Intersects<G>,
     Polygon<T>: Intersects<G>,
     MultiPolygon<T>: Intersects<G>,
+    G: BoundingRect<T>,
```
michaelkirk, line 17: Note that I had to add some new bounds to some of our intersection traits. 

This shouldn't affect any of our implementations because BoundingBox was already implemented for all geometry types.

geo/src/algorithm/bounding_rect.rs
```diff
@@ -1,13 +1,13 @@
...
-    MultiPoint, MultiPolygon, Point, Polygon, Rect, Triangle,
+    coord, CoordNum, Coordinate, Geometry, GeometryCollection, GeometryCow, Line, LineString,
+    MultiLineString, MultiPoint, MultiPolygon, Point, Polygon, Rect, Triangle,
 };
 use geo_types::private_utils::{get_bounding_rect, line_string_bounding_rect};
 
 /// Calculation of the bounding rectangle of a geometry.
 pub trait BoundingRect<T: CoordNum> {
-    type Output;
+    type Output: Into<Option<Rect<T>>>;
```
michaelkirk, line 10: This was needed for some of the blanket Intersection implementations which now rely on BoundingRect.

Some geometries bbox is Rect, while others, like an empty Polygon, are Option<Rect>::None.

michaelkirk:
I also threw in 915a0670fbca14ccfe9c950c8dadc5fe3c9750ee - preferring just `intersects` over `contains` for the short-circuit test of zero euclidean_distance. It doesn't measurably affect our benches, but there's no need to check both and intersects is typically a little simpler to check.

rmanoka (approved):
This is very cool @michaelkirk !  LGTM!

geo-test-fixtures/Cargo.toml
```diff
@@ -5,5 +5,5 @@ edition = "2021"
 publish = false
 
 [dependencies]
-wkt = { version = "0.10.0", default-features = false }
+wkt = { version = "0.10.0", features = ["geo-types"], default-features = false }
```
lnicola, line 8: This is a bit unfortunate, since `wkt` being used here without `geo-types` was one of the main arguments against moving `wkt` into the `geo` repository. That's why `line_string` below is used instead of the `wkt` feature.

lnicola:
I'm worried about the fixture changes, but LGTM otherwise.

geo-test-fixtures/Cargo.toml
michaelkirk: I don't think I ever made that argument, but this does introduce a circular dependency, which I think is the concern, right? 

There's been a lot of ink spilled about circular dev-dependencies in georust. I get why it's not conceptually ideal, but in practice I've found it pretty easy to add a temporary entry to `[patch.crates-io]` while I have a change in flight.

@lnicola  - can you expound on what you'd like to see happen here? Is it your preference that I copy and paste the conversion of WKT into a geo-type::Polygon into geo-test-fixtures? 

A bigger change could be to move the integrations into the geo-types crate (behind a `wkt` feature) like we did with `rstar`. What would you think of that?

michaelkirk: > A bigger change could be to move the integrations into the geo-types crate (behind a wkt feature) like we did with rstar. What would you think of that?

To add to this, IIRC we talked about doing something like this in the past, but I was concerned it would mean "breaking" geo-types semver whenever we require a newer breaking version of `wkt`. But we've since learned a nicer way to handle this - as exemplified with our simultaneous rstar-0.8 and rstar-0.9 integrations, so I have less reservations about it than I might have a year ago.

lnicola: > I don't think I ever made that argument, but this does introduce a circular dependency, which I think is the concern, right?

No, I did :sweat_smile:. And yes, it adds an awkward (though not circular) dependency: it pulls in another version of `geo-types`, so the types returned by `wkt` will not always be compatible with those in the current repo.

> Is it your preference that I copy and paste the conversion of WKT into a geo-type::Polygon into geo-test-fixtures?

That's what I did for `line_string`. I think it's not too bad.

> A bigger change could be to move the integrations into the geo-types

That would work. I don't have any strong opinion about which direction it's better to do this in.

michaelkirk: >> Is it your preference that I copy and paste the conversion of WKT into a geo-type::Polygon into geo-test-fixtures?
>
> That's what I did for line_string. I think it's not too bad.

Alright, I've done that for now since it seems easy and localized. I'm not overly worried about hacks in our fixture loading code.

> A bigger change could be to move the integrations into the geo-types

I might follow up one day with this...

lnicola: Thanks, looks good, but please squash at the end to keep only one copy of the WKTs.

michaelkirk: Done - I squashed the fixture changes into the first commit.

michaelkirk:
bors r=lnicola

Issue #623, opened 2021-02-10
Add trait CoordsIter to prelude


Nevsden:
It would be nice if the trait `CoordsIter` is inside prelude to prevent from adding additional dependencies.

frewsxcv:
https://github.com/georust/geo/pull/830

PR #830, opened 2022-05-12
flatten algorithm namespace

### LINKED ISSUE #572: Proposal: Flatten algorithm module layout
# Current layout

Currently, whenever we add an algorithm to `geo`, we create a new module in the `crate::algorithm` module. Within that new module, we have all the necessary structures and traits to utilize the new algorithm. This new module also gets reexported at the top-level of the crate. Visualized, this looks like:

Untitled Diagram(5)

Effectively, our users have two ways to import algorithms in their crate:

```rust
use geo::algorithm::winding_order::Winding;
// or
use geo::winding_order::Winding;
```

Subjectively, my view is this structure is unnecessarily deep, and requires a lot of typing and knowledge about our module layout just to import a single trait.

# History

The `crate::algorithm` module dates back to May 2016 when [`Centroid`, our first algorithm, was implemented in the crate](https://github.com/georust/geo/pull/24). I don't recall us ever having a discussion about module layout, so I'm going to assume there's not much history beyond this!

# Proposed layout

I propose all traits for algorithms live at the root of the crate. Any supporting structures for the trait can live in a submodule specific to that algorithm. Visualized, this looks like:

Untitled Diagram(8)

An example

As a point of comparison, this is similar to how `HashMap` and other collections are laid out in the standard library:

Untitled Diagram(7)

The biggest downside to this is the docs for root of the `geo` crate will get much bigger. I think this can be addressed by having an organized index in the docs of everything at the crate root:

```rust
//! The `geo` crate provides and geospatial primitives and algorithms.
//!
//! # Types
//!
//! - `Coordinate`: ...
//! - `Point`: ...
//!
//! # Algorithms
//!
//! - `BoundingRect`: ...
//!
//! ## Distance
//!
//! - `EuclideanDistance`: ...
```

What does everyone think?

### LINKED ISSUE #793: Add trait Winding to prelude
Might be a good idea to check for other missing traits from the prelude.

michaelkirk:
---

Fixes #572 
Supersedes #576 

Also fixes #793 #623, because the prelude is now just a re-export of `algorithms::*`

### DIFF
--- a/geo/src/algorithm/mod.rs
+++ b/geo/src/algorithm/mod.rs
@@ -1,93 +1,188 @@
 /// Kernels to compute various predicates
 pub mod kernels;
+pub use kernels::{HasKernel, Kernel, Orientation};
 
 /// Calculate the area of the surface of a `Geometry`.
 pub mod area;
+pub use area::Area;
+
 /// Calculate the bearing to another `Point`, in degrees.
 pub mod bearing;
+pub use bearing::Bearing;
+
 /// Calculate the bounding rectangle of a `Geometry`.
 pub mod bounding_rect;
+pub use bounding_rect::BoundingRect;
+
 /// Calculate the centroid of a `Geometry`.
 pub mod centroid;
+pub use centroid::Centroid;
+
 /// Smoothen `LineString`, `Polygon`, `MultiLineString` and `MultiPolygon` using Chaikins algorithm.
 pub mod chaikin_smoothing;
+pub use chaikin_smoothing::ChaikinSmoothing;
+
 /// Calculate the signed approximate geodesic area of a `Geometry`.
 pub mod chamberlain_duquette_area;
+pub use chamberlain_duquette_area::ChamberlainDuquetteArea;
+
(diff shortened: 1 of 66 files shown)


geo/src/algorithm/mod.rs
```diff
@@ -1,93 +1,188 @@
...
 pub mod concave_hull;
+pub use concave_hull::ConcaveHull;
+
 /// Determine whether `Geometry` `A` is completely enclosed by `Geometry` `B`.
 pub mod contains;
+pub use contains::Contains;
+
 /// Calculate the convex hull of a `Geometry`.
 pub mod convex_hull;
+// REVIEW: use qhull, etc?
```
frewsxcv, line 43: My vote is 👎🏻 for all the `REVIEW` comments in this file

frewsxcv (approved):

geo/src/algorithm/mod.rs
michaelkirk: Removed in https://github.com/georust/geo/pull/830/commits/700f008ed4b50040093b13d4cfa2aa7c6bd0aeb8

PR #832, opened 2022-05-13
default impl for CoordNum in Geometries

michaelkirk:
---

Ok firstly, this might be controversial. I spent very little time on it, so I'm happy to let it sit for a while, or just walk away from it.

99% of the time when I'm using geo/geo-types I want an f64. I suspect most downstream users are also generally using f64. It finally occurred to me that we could just make that the default and save some typing. People that want other geometry types can continue to specify `<f32>/<i32>/<usize>` whatever.

The major downside I see is that if people are implementing traits for geo-types, they might inadvertently constrain their impls to f64 when they might actually prefer to make them generic. Personally, that doesn't strike me as a huge concern, but maybe others feel differently.

### DIFF
--- a/geo-postgis/src/from_postgis.rs
+++ b/geo-postgis/src/from_postgis.rs
@@ -15,7 +15,7 @@ pub trait FromPostgis<T> {
     fn from_postgis(_: T) -> Self;
 }
 
-impl<'a, T> FromPostgis<&'a T> for Point<f64>
+impl<'a, T> FromPostgis<&'a T> for Point
 where
     T: postgis::Point,
 {
@@ -23,12 +23,12 @@ where
         Point::new(pt.x(), pt.y())
     }
 }
-impl<'a, T> FromPostgis<&'a T> for LineString<f64>
+impl<'a, T> FromPostgis<&'a T> for LineString
 where
     T: postgis::LineString<'a>,
 {
     fn from_postgis(ls: &'a T) -> Self {
-        let ret: Vec<Point<f64>> = ls.points().map(Point::from_postgis).collect();
+        let ret: Vec<Point> = ls.points().map(Point::from_postgis).collect();
         LineString::from(ret)
     }
 }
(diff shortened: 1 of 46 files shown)


rmanoka:
I quite like this; it's quite ergonomic.  Is there any downside to users who don't want T=f64?  Or does the API remain exactly the same as before for the general case?  May be we could move some of the tests to f32 to explain the usage with a non-default type.

michaelkirk:
> Or does the API remain exactly the same as before for the general case? May be we could move some of the tests to f32 to explain the usage with a non-default type.

The API remains exactly the same as before for the general case.

I've added some docs (and a doc test) to the geo-types/lib.rs - WDYT?

Screen Shot 2022-05-16 at 4 00 52 PM

rmanoka (approved):
LGTM

frewsxcv:
Overall I _think_ I'm in favor. One very small concern I have is this may encourage library authors to write something like:

```rust
struct Shapefile {
    fn from_geo(geom: geo::Geometry) -> Self {
         ...
    }
}
```

When they should really be writing `geo::Geometry<T>` so more types can be supported

michaelkirk:
> they should really be writing geo::Geometry<T> so more types can be supported

I agree - that is a potential downside. One of the reason's I think it's not so bad though, is that, if there's demand for it, those trait implementors could later upgrade to add generic support in a non-breaking way:

```diff
- struct Shapefile {   
-   fn from_geo(geom: geo::Geometry) -> Self {
+ struct Shapefile<T: CoordNum = f64> {
+   fn from_geo(geom: geo::Geometry<T>) -> Self<T> {
          ...
    }
  }
```

frewsxcv:
Could this possibly be considered a breaking change for geo-types? I don't think it would be but I can't help but wonder if there's some edge case we're not considering 🤔

michaelkirk:
This is intended to be non-breaking, and I believe it to be. But it's always possible that I've missing something.

michaelkirk:
Anybody else care to weigh in? 

I've gotten somewhere between one and two positive responses, and (I think) no unaddressed concerns.

Otherwise I'll plan to merge in a couple days.

urschrei:
I like it too, and the doc additions make clear what's happening so I'm not too worried about confusing errors.

PR #836, opened 2022-06-02
Add `{Convert,TryConvert}` traits for coordinate value type conversion.

frewsxcv:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/main/CODE_OF_CONDUCT.md).
- [x] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

- [x] Add table of contents entry

Fixes https//github.com/georust/geo/issues/617

PR #837, opened 2022-06-02
Map coords takes a coordinate

### LINKED ISSUE #804: Replace MapCoords tuple parameter in function argument with a Coodinate parameter
https://github.com/georust/geo/pull/800#discussion_r849490674

michaelkirk:
---

FIXES #804 

Note there will be a little conflict with #836 - whoever merges second will have to make a small update.

### DIFF
--- a/geo/src/algorithm/map_coords.rs
+++ b/geo/src/algorithm/map_coords.rs
@@ -50,18 +50,29 @@ mod modern {
         /// assert_relative_eq!(p2, Point::new(1010., 40.), epsilon = 1e-6);
         /// ```
         ///
-        /// You can convert the coordinate type this way as well
+        /// Note that the input and output numeric types need not match.
+        ///
+        /// For example, consider OpenStreetMap's coordinate encoding scheme, which, to save space,
+        /// encodes latitude/longitude as 32bit signed integers from the floating point values
+        /// to six decimal places (eg. lat/lon * 1000000).
         ///
         /// ```
         /// # use geo::{Coordinate, Point};
         /// # use geo::MapCoords;
         /// # use approx::assert_relative_eq;
         ///
-        /// let p1: Point<f32> = Point::new(10.0f32, 20.0f32);
-        /// let p2: Point<f64> = p1.map_coords(|Coordinate { x, y }| Coordinate { x: x as f64, y: y as f64 });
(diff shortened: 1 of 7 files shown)


geo/src/algorithm/map_coords.rs
```diff
@@ -745,7 +726,10 @@ pub(crate) mod deprecated {
     use super::*;
 
     /// Map a fallible function over all the coordinates in a geometry, returning a Result
-    #[deprecated(since = "0.21.0", note = "use `MapCoords::try_map_coords` instead")]
+    #[deprecated(
+        since = "0.21.0",
+        note = "use `MapCoords::try_map_coords` which takes a `Coordinate` instead of an (x,y) tuple"
+    )]
     pub trait TryMapCoords<T, NT, E> {
```
michaelkirk, line 733: Note, for the deprecated methods, I left the signatures as `(T, T)` tuples, figuring they'll go away soon and this way we can better have the compiler tell the users what to do.

geo/src/algorithm/map_coords.rs
```diff
@@ -41,28 +41,28 @@ mod modern {
...
         /// use approx::assert_relative_eq;
         ///
         /// let p1 = Point::new(10., 20.);
-        /// let p2 = p1.map_coords(|(x, y)| (x + 1000., y * 2.));
+        /// let p2 = p1.map_coords(|Coordinate { x, y }| Coordinate { x: x + 1000., y: y * 2. });
         ///
         /// assert_relative_eq!(p2, Point::new(1010., 40.), epsilon = 1e-6);
         /// ```
         ///
         /// You can convert the coordinate type this way as well
```
frewsxcv, line 53: Unrelated, but we should probs advertise `Convert` here now?

frewsxcv (approved):
lgtm, though you'll need to rebase/merge after my other pr merges

geo/src/algorithm/map_coords.rs
michaelkirk: I've rebased, and pushed up a new commit which references `Convert` for the case of simply changing numeric type , and then updated this example to solve a more complex problem involving both type conversion and a transformation for which `Convert` would be insufficient.

michaelkirk:
bors r=frewsxcv

PR #839, opened 2022-06-11
Reexport everything from `proj`.

frewsxcv:
Seems like if we're going to expose `proj`, we might as well expose
everything. My use case is I need access to `proj::ProjCreateError`,
which wasn't exposed prior to this.

---


michaelkirk (approved):
I think that this makes sense. 

The alternative is that users have to deduce *which* version of proj is being installed by geo, and copy that into their Cargo.toml (and then keep it in sync whenever geo changes). Instead we just "include" all of proj within geo (iff you've enabled the `with-proj` feature).

frewsxcv:
bors r=michaelkirk

PR #841, opened 2022-06-12
Add `cargo publish --dry-run` to CI.

frewsxcv:
---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -103,6 +103,8 @@ jobs:
       - run: cargo check --all-targets --no-default-features
       # we don't want to test `proj-network` because it only enables the `proj` feature
       - run: cargo test --features "use-proj use-serde"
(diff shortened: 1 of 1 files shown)


rmanoka:
Flip side: this will mean that sometimes PRs can't be merged until a geo-types release is made;  not sure if it'll be a blocker for anything though.

michaelkirk:
Would it be helpful if we added an action like this, but it only ran conditionally on pre-release branches?

e.g. by convention, instead of frewsxcv-0.21.0, you could push to something like `release/geo-0.21.0` (or whatever convention) and then we have a condition in our `cargo publish --dry-run` action to run only on `release/*`.

frewsxcv:
I like the sentiment and the thought you're putting into it, but (at least for me) I'm skeptical I'll remember that convention the next go-around

PR #852, opened 2022-06-23
Use robust orientation check in VW and fast path algorithms

urschrei:
---

While `Point` implements a cross product function which can be used to determine whether three lines "turn" clockwise, counter-clockwise, or are collinear, `geo` can do this in either a robust or precise manner for anything implementing the `HasKernel` trait. This PR removes all uses of the naive `cross_prod` in `geo` in favour of the kernel-based `orient2d` function.

The only remaining use of `cross_prod` is now in `geo_types` (in `Polygon`'s `is_convex()` method), because we can't use `robust` there.

Is it worth also adding a note to the `cross_prod` implementation advising that if `geo` is being used, the kernel-based methods are available instead?

### DIFF
--- a/geo/src/algorithm/simplifyvw.rs
+++ b/geo/src/algorithm/simplifyvw.rs
@@ -1,7 +1,7 @@
 use crate::prelude::*;
 use crate::{
-    CoordFloat, Coordinate, Line, LineString, MultiLineString, MultiPolygon, Point, Polygon,
-    Triangle,
+    CoordFloat, Coordinate, HasKernel, Line, LineString, MultiLineString, MultiPolygon, Point,
+    Polygon, Triangle,
 };
 use std::cmp::Ordering;
 use std::collections::BinaryHeap;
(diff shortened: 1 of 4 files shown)


geo/src/algorithm/simplifyvw.rs
```diff
@@ -207,7 +207,7 @@ fn vwp_wrapper<T>(
     epsilon: &T,
 ) -> Vec<Vec<Coordinate<T>>>
 where
-    T: CoordFloat + RTreeNum,
+    T: CoordFloat + RTreeNum + HasKernel,
```
michaelkirk, line 210: I guess this is technically a breaking change for those mythical users hypothetically using non-standard numeric types. I really expect these people don't currently exist, but I'd love to be proven wrong!

FWIW I think it's worth "breaking" to get the precision/robustness of our kernel implementations.

michaelkirk (approved):

michaelkirk:
btw the failed lint should be fixed in master now.

bors try

geo/src/algorithm/simplifyvw.rs
urschrei: Yeah I did wonder about that (show yourselves!), but as you say, I think it's worth it.

Issue #831, opened 2022-05-12
create a `geo::geometries` module that re-exports all the geometry members


michaelkirk:
I often see client code like this:

```
use geo_types::{
    Geometry, GeometryCollection, Line, LineString, MultiLineString, MultiPoint, MultiPolygon, Polygon, Rect, Point, Coord,
};

use geo_types::{coord,  point, line_string, polygon}; 
```

Instead I'd like to:
```
   use geo_types::geometries::*;
   use geo_types::geometries::macros::*;
```

And re-export them as `geo::geometries`/`geo::geometries::macros`.

In general, I'm not a fan of large wild card imports because later in the code it's hard to know where something came from, but for cases like this when it's a small number of well known things, I think it's helpful with little downside.

This would be a non-breaking change - the old paths would remain.

PR #853, opened 2022-06-23
`use geometry::*` to concisely include all the geometry variants

michaelkirk:
---

Fixes https://github.com/georust/geo/issues/831 - an ergonomics issue.

So you can now:
```diff
- use geo_types::{Coordinate, Point, Line, LineString, Polygon, MultiPoint, MultiLineString, MultiPolygon, Rect, Triangle, Geometry, GeometryCollection};
+ use geo_types::geometry::*
```

And similarly with geo:
```diff
- use geo::{Coordinate, Point, Line, LineString, Polygon, MultiPoint, MultiLineString, MultiPolygon, Rect, Triangle, Geometry, GeometryCollection};
+ use geo::geometry::*
```

(quoting from #831):
> In general, I'm not a fan of large wild card imports because later in the code it's hard to know where something came from, but for cases like this when it's a small number of well known things, I think it's helpful with little downside.

Note that I diverged slightly from my original plan in #831.

I initially sought to introduce a new `geometries` module, but then I was afraid it'd be too confusing to have both a `geometry` and a `geometries` module, so opted instead to have a `geometry` module which, as well as defining the `Geometry` enum, also re-exports the inner variants.


geo/src/algorithm/chamberlain_duquette_area.rs
```diff
@@ -49,7 +46,7 @@ use num_traits::Float;
 /// ```
 pub trait ChamberlainDuquetteArea<T>
 where
-    T: Float + CoordNum,
+    T: CoordFloat,
```
michaelkirk, line 49: The CoordFloat additions are an unrelated no-op cleanup. I just prefer to uniformly reference our internal types over the num_types.

frewsxcv (approved):
Sgtm!

michaelkirk:
bors r=frewsxcv

(rebased)

PR #855, opened 2022-06-24
Fixup cargo-doc warnings

michaelkirk:
---

### DIFF
--- a/geo/src/algorithm/mod.rs
+++ b/geo/src/algorithm/mod.rs
@@ -183,7 +183,7 @@ pub use vincenty_length::VincentyLength;
 
 /// Calculate and work with the winding order of `Linestring`s.
 pub mod winding_order;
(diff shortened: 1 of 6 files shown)


geo/src/algorithm/mod.rs
```diff
@@ -183,7 +183,7 @@ pub use vincenty_length::VincentyLength;
 
 /// Calculate and work with the winding order of `Linestring`s.
 pub mod winding_order;
-pub use winding_order::WindingOrder;
+pub use winding_order::Winding;
```
michaelkirk, line 186: I got confused when I re-exported winding_order::WindingOrder from the algorithm module.

Breaking with the convention of our other modules, `winding_order::Winding` is the trait, while `winding_order::WindingOrder` is the type returned by that traits methods.

frewsxcv (approved):

Issue #856, opened 2022-06-24
geo 0.22.0 release


michaelkirk:
We should do another release! 

I'd like to include: 
- https://github.com/georust/geo/pull/850
- https://github.com/georust/geo/pull/855

Note that this geo release will require first releasing (non-breaking) geo-types and then bumping geo's geo-types dependency.

Anything else that we should be waiting for?

frewsxcv:
It'd be nice to have https://github.com/georust/geo/issues/844 in there too, but not a blocker

martinfrances107:
All this depends on how others feel. but I have a set of other dependencies that are outdated
[ by outdated I mean as reported by "cargo outdated"]

So,  I would like to delay the release until that  technical debt is resolved.


All the dependencies I want to fix, would cause breaking changes.

| package                              | dependency  | current  | comment  
--------------------------------------|-------------------|------------|------------------------------------
geo/jts-test-runner                 | include_dir    | 0.6.0      | version 0.7.2 is wanted.
geo/geo                                 | proj                | 0.26      | version 0.27.0 is wanted.
geo/geo-bool-ops-benches   | geojson          | 0.22.3    | version 0.23.0 is wanted.
gep/geo-bool-ops-benches   | gt_prev           | 0.6.2     | version 0.75 is wanted.

1) For proj,   as with bumping approx -- accepting a range of version numbers is the  way forward.
2) For the test runner and benches I was hoping for a simple single bump in number.

Given that I am about to do two PR  - I will link here when I am done.

I appreciated this a annoying in that it causes lots of little minor incompatibilities. I favour  frequent maintenance tasks over  the alternative where things spiral out of control and we end up in logjam which is impossible to resolve.

frewsxcv:
My vote is to not have that block a release, but would like to hear from others. `geo` releases are easy to do so we can just do another after those changes are made. We don't need to cram them in

martinfrances107:
regarding  this https://github.com/georust/geo/issues/856#issuecomment-1165168140
I now have PR for all I want to fix. 

I have no opinion about wheather they should be pre or post release.

michaelkirk:
Thanks for all these @martinfrances107 - I've queued them all for merging, and plan to proceed with the geo 0.22.0 release later today.

PR #857, opened 2022-06-24
geo/geo bump  proj to 0.27.0

martinfrances107:
The is a ongoing discussion in #856 about how to signal to the wider community that our dependencies are changing.

But this is issue follows the pattern developed in #850

### DIFF
--- a/geo/Cargo.toml
+++ b/geo/Cargo.toml
@@ -21,7 +21,7 @@ geo-types = { version = "0.7.5", features = ["approx", "use-rstar_0_9"] }
 geographiclib-rs = "0.2"
 log = "0.4.11"
 num-traits = "0.2"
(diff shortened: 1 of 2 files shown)


michaelkirk:
This LGTM, but I'm *also* not very concerned about publishing a breaking release for geo. I'd also be fine to just strictly bump proj without the range.

My position on this is different with geo-types - which is used across a lot of other crates. I'm much more willing to put effort into *not* breaking geo-types.

I'll wait until tomorrow to merge in case anyone else wants to weigh in.

martinfrances107:
> I'd also be fine to just strictly bump proj without the range.

Great that is my preference..fixed.

michaelkirk:
bors retry

I pushed up a new commit to your branch to adapt to the latest proj.

PR #858, opened 2022-06-24
bump include_dir to 0.7.2 and fix breaking change.

martinfrances107:
The change that include_dir has forced on us is that we must convert a relative path into a absolute path

Here is what the include_dir documentation has to say about it. :-

> When invoking the [include_dir!()](https://docs.rs/include_dir/latest/include_dir/macro.include_dir.html) macro you should try to avoid using relative paths because rustc makes no guarantees about the current directory when it is running a procedural macro

### DIFF
--- a/jts-test-runner/Cargo.toml
+++ b/jts-test-runner/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2021"
 [dependencies]
 approx = "0.4.0"
 geo = { path = "../geo" }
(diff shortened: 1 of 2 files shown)


michaelkirk (approved):
Nice! Thanks for investigating.

bors r+

martinfrances107:
Thanks for looking at this and the prompt response.

michaelkirk++

Issue #728, opened 2022-02-13
Porting scaling and other Affine transformations from JTS


nyurik:
Currently geo implements rotation and translations, but misses others like `scale` (see static methods on [JTS AffineTransformation](https://locationtech.github.io/jts/javadoc/org/locationtech/jts/geom/util/AffineTransformation.html#method.summary)).  Does it make sense to port the entire AffineTransformation as a single matrix-based transformation with similar interface?  For my usecase I only need `scale`, but I'm sure others may encounter other needs. Additionally a matrix based transformation allows multiple transforms to be combined into one - something that doesn't appear to be possible with the rust-geo's implementation.  If ok, I would be happy to try to implement it.

michaelkirk:
Sorry for the slow response.

I for one would love to see a comprehensive affine feature. Do you have a proposal for what the API would look like?

nyurik:
@michaelkirk not really, I am just now getting deeper into the geo-types and various algorithms. I would think it would be similar to JTS -- a builder pattern, combining various transformations into a single matrix, and later a function on that builder to convert a geo-type into a different geo-type (or possibly in-place).  Uncertain yet, suggestions welcomed.

michaelkirk:
> convert a geo-type into a different geo-type

The output type should always be the same as the input. A Linestring will never become a Point, right?

> possibly in-place
*generally* I'd start with a  mutable implementation, because then people have the option of using it without incurring the memory overhead of creating a new object. Building an immutable versions on top of that can be as trivial as`.clone().some_mutable_method()`. Though there *may* be reasons to prefer a different approach depending on the algorithm.

When reasonable, I like to expose both mutable and immutable flavors, like here: https://github.com/georust/proj/blob/main/src/geo_types.rs#L65

My first thought for some api usage would be something like:

(names subject to change, mostly thinking about the way the components interact)

```
use geo_types::{CoordNum, Coordinate, Polygon, Geometry};

pub struct AffineTransform<T: CoordNum>([[T; 3]; 3]);
impl<T: CoordNum> AffineTransform<T> {
    pub fn scale(x: T, y: T) -> Self {
        Self::identity().scaled(x, y)
    }

    pub fn scaled(mut self, x: T, y: T) -> Self {
        self.0[0][0] = self.0[0][0] * x;
        self.0[1][1] = self.0[1][1] * y;
        self
    }

    pub fn identity() -> Self {
        Self([
            [T::one(), T::zero(), T::zero()],
[...]

PR #866, opened 2022-06-25
Add Scale affine transform

urschrei:
---


geo/src/algorithm/shear.rs
```diff
@@ -0,0 +1,46 @@
...
+impl<T, G> Shear<T> for G
+where
+    T: CoordFloat,
+    G: MapCoords<T, T, Output = G>,
+{
+    fn shear(&self, xs: T, ys: T, origin: Point<T>) -> Self {
+        // x offset = -origin.y() * xs.to_radians().tan()
+        // y offset = -origin.x() * ys.to_radians().tan()
+        self.map_coords(|Coordinate { x, y }| Coordinate {
+            x: x + (-origin.y() * xs.to_radians().tan()),
```
urschrei, line 42: Wrong! Somehow.

urschrei:
(Not yet ready to merge because I have somehow messed up the arithmetic to calculate the `Shear` offsets, and I need to figure that out. See above)

michaelkirk:
Some former discussion on what an affine API could look like #728 

(I’m not committed to what was in 728, just FYI)

urschrei:
Oh I didn't see that at all! I agree that going with an interface that allows affine transforms to be composed together is definitely the way to go, mostly because it's a lot more efficient; composition is "just" multiplying e.g. scale, rotation, and shear matrices together (which works because they're all affine matrices, thus are 3 x 3), then applying the result to your input coordinates (a column vector)

The reason I didn't bother with that and just leveraged our blanket `MapCoords` impl in this PR is that I couldn't figure out a straightforward way to perform the matrix multiplications on slices / `Vec`s: because we don't use e.g. `ndarray`, we'll have to roll our own matrix multiplication logic on things-that-arent-matrices, and I don't know whether we'll we able to leverage `MapCoords` for that.

urschrei:
I suppose implementing `compose` is easy because multiplying two `3 x 3` vectors together is relatively straightforward, so we could present nice interfaces for the different transforms, with an internal matrix-like representation (`AffineMatrix`, an array with 9 slots, maybe nested like your `identity` example) which implements a multiplication function.

The trickier part is figuring out whether there's an easy way of implementing multiplication between an `AffineTransform` and a `Vec` or slice of arbitrary length. Maybe I'm overthinking that and it's easy, though.

urschrei:
@michaelkirk @nyurik I've made good progress with this (please note all names etc subject to bikeshedding)!

We now have: an `affineops` module, containing an `AffineTransform` struct, which implements:

- creation of identity matrix
- creation of rotation, scaling, skewing (shear), translation matrices
- composition of those ^ (that was a lot of careful typing 💀)
- application of affine matrices to a `Coordinate`, allowing us to leverage various `MapCoords` flavours

Using `apply` I was able to re-implement all our existing affine ops in terms of `AffineTransform` and add `skew` and `scale`.

**HOWEVER**, I wasn't able to use `CoordNum` for the `AffineTransform` crate, because both rotation and skewing calculations require [`to_radians`](https://docs.rs/num-traits/latest/num_traits/float/trait.Float.html?search=radians), which is only available on types implementing `Float` or `Real`. The bound is now `CoordNum` + [`Real`](https://docs.rs/num-traits/latest/num_traits/real/trait.Real.html), which is both technically a breaking change, and apparently doesn't work with `no_std`:

> This trait is only available with the std feature, or with the libm feature otherwise.

This doesn't seem like an acceptable compromise. Can anyone think of some way we can juggle the bounds to make this work? Maybe the specific affine ops (`skewed`, `rotated` etc) don't have to be implemented as part of the trait? I can't really see the shape of that, though.

urschrei:
OK, I've managed to split common affine ops (scale and translate, will work with `CoordNum` and `CoordFloat`) and those requiring `CoordFloat`: We now have an `AffineOps` trait which has been impld on `AffineTransform`, providing the two common ops, and `AffineRotate` and `AffineSkew` which impl `AffineOps` as a supertrait (i.e. they produce an `AffineTransform`)

This has let me remove `Real` again, and selectively impl `AffineSkew` and `AffineRotate` for our existing `Rotate*` traits and new `Skew` trait.

I've never engaged in this level of trait-wrangling before, but it…it seems to work fine? I've also written docs so if ppl build them they should be able to make sense of the moving parts (and the existing affine trait bounds are unaffected).

geo/src/algorithm/affineops.rs
```diff
@@ -0,0 +1,287 @@
...
+            ],
+            [
+                (self.0[1][0] * other.0[0][0])
+                    + (self.0[1][1] * other.0[1][0])
+                    + (self.0[1][2] * other.0[2][0]),
+                (self.0[1][0] * other.0[0][1])
+                    + (self.0[1][1] * other.0[1][1])
+                    + (self.0[1][2] * other.0[2][1]),
+                (self.0[1][0] * other.0[0][2])
+                    + (self.0[1][1] * other.0[1][1])
```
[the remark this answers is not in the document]
michaelkirk: This is hard to review, but I think I see some errors. Can you double check?

geo/src/algorithm/scale.rs
```diff
@@ -0,0 +1,38 @@
...
+    where
+        T: CoordNum;
+}
+
+impl<T, G> Scale<T> for G
+where
+    T: CoordNum,
+    G: MapCoords<T, T, Output = G> + BoundingRect<T>,
+{
+    fn scale(&self, x: T, y: T, origin: Point<T>) -> Self {
```
michaelkirk, line 34: Bikeshed:  Immutable vs (eventual) Mutable flavors.

Do we want a mutable version of the affine transformations? (I think so, at least eventually!)

Consider further constraining our trait impl from only `MapCoords` to `MapCoords + MapCoordsInPlace` to avoid a future "breakage" when the mutable methods get added.

And then, what should the naming conventions for mutable vs immutable be? Do the existing names here clash with where will end up once there are mutable flavors?

For context, sometimes we use an `_in_place` suffix to denote mutable - as in:
```rust
// Approach A
trait Rotate {
  // immutable, implicitly clones the input.
  fn translate(&self) -> Self;

  // mutable
  fn translate_in_place(&mut self);
}
```

Other places we use verb for mutable and past participle for immutable:
```rust
// Approach B
trait Transform {
  // immutable, implicitly clones the input.
  fn transformed(&self, proj: &Proj) -> Result<Self::Output, ProjError>;

  // mutable
  fn transform(&mut self, proj: &Proj) -> Result<(), ProjError>;
}
```

I generally prefer approach B (and so does the [top comment in some random reddit thread](https://www.reddit.com/r/rust/comments/p0e95q/naming_convention_for_methods_that_mutate_vs/)!), but I'm also ok with the `_in_place` convention, and we're already using it for `translate_in_place`, so there's no clear best option IMO. 🤷

michaelkirk (changes requested):
Some doc fixups and (I think?) some errors in the matrix math. All the rest is just opining that requires no response.

geo/src/algorithm/affineops.rs
```diff
@@ -0,0 +1,287 @@
...
+            .finish()
+    }
+}
+
+/// Affine operations common to all numeric types
+///
+/// These operations are available for both integer and floating point types.
+/// The two Affine transforms that currently only work with floating point types are defined in
+/// [`AffineRotate`](AffineRotate), and [`AffineSkew`](AffineSkew).
+pub trait AffineOps<T: CoordNum> {
```
michaelkirk, line 73: I just realized I was a little confused when I first read this. I had assumed AffineOps would add methods to Geometry types - but I'm realizing it's only implemented for AffineTransform.

Is there a reason that the `AffineOps` trait exists at all at this point, vs having the implementations directly on AffineTransform? 

I'm having a hard time imagining what else would ever implement `AffineOps` besides `AffineTransform`.

urschrei: Possibly because I confused myself when defining the supertrait bounds on `AffineSkew` and `AffineRotate` – maybe there's no need for the supertrait complication if I can define them directly on `AffineTransform`. In my (flimsy) defence my brain had melted after six hours of traits. Let me try it out.

urschrei: All `CoordNum` methods are now directly re-implemented on `AffineTransform`, with the two `CoordFloat` ops as traits. I don't _think_ there's any way to directly impl them?

geo/src/algorithm/scale.rs
urschrei: Added `MapCoordsInPlace` trait bound.

geo/src/algorithm/affineops.rs
urschrei: 👍👍👍 works perfectly

geo/src/algorithm/rotate.rs
```diff
@@ -132,12 +121,11 @@ pub trait RotatePoint<T> {
 impl<T, G> RotatePoint<T> for G
 where
     T: CoordFloat,
-    G: MapCoords<T, T, Output = G>,
+    G: MapCoords<T, T, Output = G> + MapCoordsInPlace<T>,
```
urschrei, line 124: Breaking, but…

geo/src/algorithm/affine_ops.rs
```diff
@@ -0,0 +1,242 @@
...
+        self.0[0][2] = xoff;
+        self.0[1][2] = yoff;
+        self
+    }
+
+    /// Create an affine transform for translation, shifted by offsets along the `x` and `y` dimensions
+    pub fn translated(mut self, xoff: T, yoff: T) -> Self {
+        // first row
+        self.0[0][0] = T::one();
+        self.0[0][1] = T::zero();
```
michaelkirk, line 127: Given the "builder-ish" API from signature, I had assumed I'd be able to chain methods together for composition. 

e.g. this will currently compile (I think):
```
let transform = AffineTransform::default()
    .scaled(10.0, 10.0, Point(Coordinate::zero())
    .translated(2.0, 4.0)
    .rotated(45.0);

polygon.map_coords(|c| transform.apply)
```

I assumed this would give me an output which had been scaled, translated, and rotated.

But I think as phrased it doesn't behave that way - the later calls replace (rather than compose wth) some of the previous values.

I'd say ultimately it'd be great to support this builder pattern style API for AffineTransforms, but alternatively, if you don't want to mess around with that for now, it might be slightly simpler to offer constructors rather than mutable methods. Then at least they wouldn't *look* like they were composing.

e.g. something like this might avoid the above confusion:
```diff
// constructor, not method
/// AffineTransform::translate(x, y).apply(coord)
- pub fn translated(mut self, xoff: T, yoff: T) -> Self {
+ pub fn translate(xoff: T, yoff: T) -> Self {
+     let mut transform = Self::identity();
      // first row
-     self.0[0][0] = T::one();
-     self.0[0][1] = T::zero();
-     self.0[0][2] = xoff;
+     transform.0[0][2] = xoff;
      // second row
-     self.0[1][0] = T::zero();
-     self.0[1][1] = T::one();
-     self.0[1][2] = yoff;
+     transform.0[1][2] = yoff;

-   self
+   transform
}
```

Once methods like that exist, I think creating an intuitive composable builder pattern (possibly in a follow up PR) would be pretty straight forward:
```rust
pub fn translated(mut self, xoff: T, yoff: T) -> Self {
    self.compose(AffineTransform::translate(xoff, yoff))
}
```

I have a similar concern for the other transforms - `scaled`, `rotated`, `skewed`.

geo/src/algorithm/affine_ops.rs
```diff
@@ -0,0 +1,242 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    // given a matrix with the shape
+    // [[a, b, xoff],
+    // [d, e, yoff],
+    // [0, 0, 1]]
+    #[test]
+    fn matrix_multiply() {
```
michaelkirk, line 231: 👍

geo/src/algorithm/affine_ops.rs
michaelkirk: As an extreme example - I think even this should be possible:

```rust
let transform = AffineTransform::identity()
    .scaled(2.0, 2.0, Point(Coordiante::zero()))
    .scaled(2.0, 2.0, Point(Coordiante::zero())).
```

And would expect it to achieve a 4x scaling.  I admit I wouldn't expect someone to write the above code *exactly* but something more reasonable which the currently proposed API doesn't easily support:

```rust
struct MapCanvasState {
    current_transform: AffineTransform
}
impl MapCanvasState {
    fn zoom_in(&mut self, origin: Point<T>) {
        self.current_transform = self.current_transform.scale(2.0, 2.0, origin)
    }
}
```

urschrei: Ohhhh no, your examples are _exactly_ what I would want (I just haven't really messed about with it yet bc I've been too busy assembling it), and that's why I wanted `compose` – let me try it by using `compose` inside the calls.

I wonder whether a `reset` method (returns matrix to the identity) might be useful too…

urschrei: I've now re-written all the mutate-in-place ops to the following format (note assignment of new matrix to `self.0`)

```Rust
pub fn scaled(mut self, xfact: T, yfact: T, origin: Point<T>) -> Self {
    let (x0, y0) = origin.x_y();
    let xoff = x0 - (x0 * xfact);
    let yoff = y0 - (y0 * yfact);
    self.0 = self
        .compose(&AffineTransform::new(
            xfact,
            T::zero(),
            T::zero(),
            yfact,
            xoff,
            yoff,
        ))
        .0;
    self
}
```

and written a test for cumulative ops, which are now behaving as expected. E.g.:

```Rust
let p0 = Point::new(0.0f64, 0.0);
// scale once
let mut scale_a = AffineTransform::default().scaled(2.0, 2.0, p0);
// rotate
scale_a = scale_a.rotated(45.0, p0);
// rotate back
scale_a = scale_a.rotated(-45.0, p0);
// scale up again, doubling
scale_a = scale_a.scaled(2.0, 2.0, p0);
```

Two issues:

1. Do we need the constructors?
2. Do we need to call out the fact that chaining ops on geometries isn't efficient? E.g.

`geom.scale(…).rotate(…)` will produce two `MapCoords` ops. I don't think there's a good way to avoid _that_, but we could a) document the perf pitfall, and b) provide a `Transform` trait for `MapCoords<T, T, Output = G> + MapCoordsInPlace<T>` so a cumulative `AffineTransform` can be built and passed to it. WDYT?

michaelkirk: > I've now re-written all the mutate-in-place ops

Nice!

> Do we need the constructors?

I think `AffineTransform::scale(2.0, 2.0, point)` | `AffineTransform::scale(2.0, 2.0, point).translated(1.0, 1.0)` would be slightly nicer than `AffineTransform::default().scaled(2.0, 2.0, point)` but I don't feel that it's *necessary* and could easily be added later.

>  provide a Transform trait [..] so a cumulative AffineTransform can be built and passed to it. 

I'd like to see that! I'd be OK with it getting added to this PR or in a followup PR. We already have `proj::Transform` in our geo namespace, so it should probably be called something else. I was calling it `trait AffineTransformable` in my [mockup](https://github.com/georust/geo/issues/728#issuecomment-1061078327), but I'm not married to that.

urschrei: Great. I'll open an issue for follow-up work, and broadcast this in case anyone else wants to kick the tires.

geo/src/algorithm/affine_ops.rs
```diff
@@ -0,0 +1,272 @@
...
+/// ```
+/// use geo::{MapCoords, AffineTransform};
+/// use geo::{line_string, BoundingRect, point, LineString, Centroid};
+/// use approx::assert_relative_eq;
+///
+/// let ls: LineString = line_string![
+///     (x: 0.0f64, y: 0.0f64),
+///     (x: 0.0f64, y: 10.0f64),
+/// ];
+/// let origin = ls.bounding_rect().unwrap().centroid();
```
michaelkirk, line 41:
```suggestion
/// let origin = ls.bounding_rect().unwrap().center();
```
(rects have a built in center method, which will be the same as centroid, but we can omit the noise of the Centroid trait in the example)

urschrei: Fixed! But now realise that `center` returns a `Coordinate` and `centroid` returns a `Point`. So displeasing!

michaelkirk (approved):
LGTM!

geo/src/algorithm/affine_ops.rs
```diff
@@ -0,0 +1,272 @@
...
+mod tests {
+    use super::*;
+
+    // given a matrix with the shape
+    // [[a, b, xoff],
+    // [d, e, yoff],
+    // [0, 0, 1]]
+    #[test]
+    fn matrix_multiply() {
+        let a = AffineTransform::new(1, 2, 3, 4, 5, 6);
```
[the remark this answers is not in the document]
urschrei: Yes I stole the ordering (and so much more 🦝) from Shapely. But perhaps we should match the logical layout 🤔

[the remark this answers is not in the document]
urschrei: Updated!

urschrei:
Squashed! Pls merge if / when happy

PR #871, opened 2022-06-30
Add AffineOps trait to transform an entire geometry (plus some tweaks)

michaelkirk:
---

As discussed [in this comment](), adds the ability to compose a transformation and apply the thing to an entire geometry (saving users the hassle of having to map_coords themselves:

```rust
let center = ls.bounding_rect().unwrap().center();
let mut transform = AffineTransform::skew(40.0, 40.0, center).rotated(45.0, center);
ls.affine_transform(&transform);
```

I also included some tweaks based on my experience using `AffineTransform` in another project.

### DIFF
--- a/geo/src/algorithm/affine_ops.rs
+++ b/geo/src/algorithm/affine_ops.rs
@@ -1,5 +1,58 @@
-use crate::{CoordFloat, CoordNum, Coordinate, Point};
+use crate::{CoordFloat, CoordNum, Coordinate, MapCoords, MapCoordsInPlace};
 use std::fmt;
+
+/// Apply an [`AffineTransform`] like [`scale`](AffineTransform::scale),
+/// [`skew`](AffineTransform::skew), or [`rotate`](AffineTransform::rotate) to a
+/// [`Geometry`](crate::geometry::Geometry).
+///
+/// Multiple transformations can be composed in order to be efficiently applied in a single
+/// operation. See [`AffineTransform`] for more on how to build up a transformation.
+///
+/// If you are not composing operations, traits that leverage this same machinery exist which might
+/// be more readable. See: [`Scale`](crate::algorithm::Scale),
+/// [`Translate`](crate::algorithm::Translate), [`Rotate`](crate::algorithm::Rotate),
+/// and [`Skew`](crate::algorithm::Skew).
+///
+/// # Examples
+/// ## Build up transforms by beginning with a constructor, then chaining mutation operations
+/// ```
+/// use geo::{AffineOps, AffineTransform};
+/// use geo::{line_string, BoundingRect, Point, LineString};
+/// use approx::assert_relative_eq;
+///
+/// let ls: LineString = line_string![
+///     (x: 0.0f64, y: 0.0f64),
+///     (x: 0.0f64, y: 10.0f64),
+/// ];
+/// let center = ls.bounding_rect().unwrap().center();
+///
+/// let transform = AffineTransform::skew(40.0, 40.0, center).rotated(45.0, center);
+///
+/// let skewed_rotated = ls.affine_transform(&transform);
+///
+/// assert_relative_eq!(skewed_rotated, line_string![
+///     (x: 0.5688687f64, y: 4.4311312),
+///     (x: -0.5688687, y: 5.5688687)
+/// ], max_relative = 1.0);
+/// ```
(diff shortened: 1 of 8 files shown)


geo/src/algorithm/affine_ops.rs
```diff
@@ -1,6 +1,59 @@
...
+///     (x: -0.5688687, y: 5.5688687)
+/// ], max_relative = 1.0);
+/// ```
+pub trait AffineOps<T: CoordNum> {
+    /// Apply `transform` to mutate `self`.
+    fn affine_transform_mut(&mut self, transform: &AffineTransform<T>);
+
+    /// Apply `transform` immutably, outputting a new geometry.
+    #[must_use]
+    fn affine_transform(&self, transform: &AffineTransform<T>) -> Self;
```
michaelkirk, line 44: Originally I wanted to do:

mutable: `affine_transform` 
immutable: `affine_transformed`

But it just conflicts too much with other implementations. I'm going to follow up with another PR to add mutable flavors with consistent naming for Scale/Skew/Rotate/Translate.

geo/src/algorithm/affine_ops.rs
```diff
@@ -127,6 +183,10 @@ impl<T: CoordNum> AffineTransform<T> {
         )
     }
 
+    pub fn is_identity(&self) -> bool {
```
michaelkirk, line 186: This came in handy while trying to skip application in my code base.

geo/src/algorithm/affine_ops.rs
```diff
@@ -141,8 +201,8 @@ impl<T: CoordNum> AffineTransform<T> {
     /// xoff = origin.x - (origin.x * xfact)
     /// yoff = origin.y - (origin.y * yfact)
     /// ```
-    pub fn scale(xfact: T, yfact: T, origin: Point<T>) -> Self {
-        let (x0, y0) = origin.x_y();
+    pub fn scale(xfact: T, yfact: T, origin: impl Into<Coordinate<T>>) -> Self {
```
michaelkirk, line 204: I'm not sure if this is controversial - but otherwise this is always kind of a papercut for me at the call site.

michaelkirk (commented):
Ok - this is ready for review!

geo/src/algorithm/affine_ops.rs
urschrei: IIRC using `Into` in function signatures _can_ lead to less informative error messages but I can't find the ref now and I would hope that the signature makes it obvious what's going on, so 👍

geo/src/algorithm/affine_ops.rs
urschrei: Could we get a brief example (or explanation as a doc comment)?

geo/src/algorithm/affine_ops.rs
```diff
@@ -1,6 +1,59 @@
...
+///     (x: 0.5688687f64, y: 4.4311312),
+///     (x: -0.5688687, y: 5.5688687)
+/// ], max_relative = 1.0);
+/// ```
+pub trait AffineOps<T: CoordNum> {
+    /// Apply `transform` to mutate `self`.
+    fn affine_transform_mut(&mut self, transform: &AffineTransform<T>);
+
+    /// Apply `transform` immutably, outputting a new geometry.
+    #[must_use]
```
urschrei, line 43: TIL about `must_use`. This is so good.

geo/src/algorithm/affine_ops.rs
```diff
@@ -13,7 +66,7 @@ use std::fmt;
 /// `AffineTransform` is a row-major matrix.
 /// 2D affine transforms require six matrix parameters:
 ///
-/// `[a, b, d, e, xoff, yoff]`
+/// `[a, b, xoff, d, e, yoff]`
```
urschrei, line 69: Ah, yes. I knew I'd forgotten to update something.

urschrei (approved):

urschrei:
Thanks, LGTM! Merge when you're ready.

geo/src/algorithm/affine_ops.rs
michaelkirk: Sometimes I daydream about getting rid of one of Point/Coord... 🔥

michaelkirk:
bors r=urschrei

Thanks for the review!

Issue #287, opened 2018-07-01
Add generic implementation for the Rotate algorithm trait


frewsxcv:
First attempt was in https://github.com/georust/rust-geo/pull/151

Outstanding question: Should `Rotate` in the case of `Polygon`s just use the external line string (not including holes) when determining the centroid? If so, we'll need a trait that does just that. Otherwise, we can just utilize `Centroid`

PR #872, opened 2022-06-30
Add mutable flavors for affine backed traits

michaelkirk:
---

Draft because it's based on #871, so please review that first.

Leverages the new `AffineOps::affine_transform` trait to drive our various affine affiliated traits.

This doubles down on the `do_the_thing() -> Self` | `do_the_thing_mut()` naming convention from #871

Also, in so doing, I was able to re-implement the rotate trait in terms of other traits, which was neat. 

FIXES https://github.com/georust/geo/issues/287 (somewhat incidentally)


geo/src/algorithm/rotate.rs
```diff
@@ -62,60 +40,31 @@ pub trait Rotate<T> {
...
-    where
-        T: CoordFloat;
-
-    #[deprecated(
-        note = "Equivalent to `rotate_around_centroid` except for `Polygon<T>`,
-                    where it is equivalent to rotating around the polygon's outer ring.
-                    Call that instead, or `rotate_around_center` if you'd like to rotate
-                    around the geometry's bounding box center."
-    )]
-    fn rotate(&self, angle: T) -> Self
```
michaelkirk, line 94: Note I'm removing this deprecated method. The deprecation was merged about a year ago in a3470cd3f5b1b658e5ba29c057f947445635095d, but was released about 4 months ago in geo-0.19.0 (releases are more frequent lately 😄).

geo/src/algorithm/rotate.rs
```diff
@@ -1,44 +1,22 @@
...
+/// efficient to compose the transformations and apply them as a single operation using the
+/// [`AffineOps`](crate::AffineOps) trait.
+pub trait Rotate<T: CoordFloat> {
     /// Rotate a geometry around its [centroid](Centroid) by an angle, in degrees
     ///
     /// Positive angles are counter-clockwise, and negative angles are clockwise rotations.
     ///
-    /// # Units
-    ///
-    /// - `angle`: degrees
```
michaelkirk, line 33: After renaming the parameter to `degrees` this documentation felt pretty superfluous.

geo/src/algorithm/scale.rs
```diff
@@ -1,49 +1,150 @@
...
+    ///
+    /// let scaled = ls.scale(2.);
+    ///
+    /// assert_eq!(scaled, line_string![
+    ///     (x: -5., y: -5.),
+    ///     (x: 15., y: 15.)
+    /// ]);
+    /// ```
+    #[must_use]
+    fn scale(&self, scale_factor: T) -> Self;
```
michaelkirk, line 30: This might be controversial, so I wanted to call it out:

I've re-purposed the nice `scale` name for the method I think people will use the most. 

The general purpose method with all the flexible parameters (which I believe will be less commonly used) is now called `scale_around_point`.

WDYT?

geo/src/algorithm/scale.rs
```diff
@@ -1,49 +1,150 @@
...
+        origin: impl Into<Coordinate<T>>,
+    );
 }
 
-impl<T, G> Scale<T> for G
+impl<T, IR, G> Scale<T> for G
 where
-    T: CoordNum,
-    G: MapCoords<T, T, Output = G> + MapCoordsInPlace<T>,
+    T: CoordFloat,
```
michaelkirk, line 99: Another potentially controversial decision: The new scale methods I've added rely on Rect::center which is only defined for Floats, so I opted to make this entire trait for Floats.

The alternative would be to split out the `scale_around_point` method into another trait, but I was afraid the ergonomic damage of having two different Scale traits wouldn't be worth it, and I do suspect that most people will want to do scaling from the center.

People who need to do integer scaling can still use `integer_geom.affine_transform(&AffineTransform::scale(2, 2))`

michaelkirk:
Ok, this is ready for review!

geo/src/algorithm/scale.rs
urschrei: Yep I agree this is an ergonomic improvement!

geo/src/algorithm/scale.rs
```diff
@@ -1,49 +1,150 @@
...
+    fn scale_xy(&self, x_factor: T, y_factor: T) -> Self;
+
+    /// Mutable version of [`scale_xy`](Self::scale_xy).
+    fn scale_xy_mut(&mut self, x_factor: T, y_factor: T);
+
+    /// Scale a geometry around a point of `origin`.
+    ///
+    /// The point of origin is *usually* given as the 2D bounding box centre of the geometry, in
+    /// which case you can just use [`scale`](Self::scale) or [`scale_xy`](Self::scale_xy), but
+    /// this method allows you to specify any point.
```
urschrei, line 63: 👍👍👍

geo/src/algorithm/scale.rs
urschrei: Yep, I think this is fine, but I should also note that I'm biased here because [unintelligible unhelpful comments about integer geom functionality]

geo/src/algorithm/rotate.rs
```diff
@@ -135,215 +84,90 @@ pub trait RotatePoint<T> {
...
-
-    fn rotate(&self, angle: T) -> Self {
-        self.rotate_around_centroid(angle)
+    fn rotate_around_point(&self, degrees: T, point: Point<T>) -> Self {
+        Rotate::rotate_around_point(self, degrees, point)
     }
 }
 
-impl<T> Rotate<T> for Polygon<T>
+impl<G, IP, IR, T> Rotate<T> for G
```
urschrei, line 112: 🤩

urschrei (approved):
This is great – I have only assenting comments / emoji denoting approval here.

michaelkirk:
bors r=urschrei

michaelkirk:
bors retry

(fixed merge conflict in CHANGES.md)

PR #875, opened 2022-07-06
Fix/jts overlay tests

### LINKED ISSUE #845: JTS test suites for boolean ops
Per https://github.com/georust/geo/pull/835#issuecomment-1149399284

We should incorporate (some?) of JTS testing for boolean operations. I had a rough go at it here: https://github.com/georust/geo/commit/a65e9078e913857fbe20e43237a3bd0417aa6d6d

> Some of them are failing. I haven't looked into most of them to determine if they reflect "errors" on our part or just "different assumptions".

rmanoka:
---

fixes #845


rmanoka:
Thanks for the JTS integration code @michaelkirk ;  I took a shot at fixing #845 .  The two primary reasons seem to be:

1. incorrect output topology assumptions:  think I've fixed these in this PR
2. precision model:  we only support vanilla f64, but many of failing test cases (`*Prec*.xml`)  use some sophisticated precision model:
   ``` xml
   <precisionModel scale="1.0" offsetx="0.0" offsety="0.0"/>
   ```
   This apparently means the outputs-coords are required to be f64 but integer coords.  We don't support this, and I've tried to parse and check if a "non-standard" precision model is used during overlay tests.

geo/src/algorithm/relate/mod.rs
```diff
@@ -81,11 +81,11 @@ macro_rules! relate_impl {
 ///
 /// # Examples
 ///
-/// ```no_run
```
rmanoka, line 84: These were causing `cargo test` to fail as `no_run` still tries to compile the code.

geo/tests/jts_tests.rs
```diff
@@ -13,7 +13,7 @@ fn init_logging() {
 fn test_all_general() {
     init_logging();
 
-    let mut runner = TestRunner::new();
+    let mut runner = TestRunner::new().with_precision_floating();
```
rmanoka, line 16: Currently the precision check is done per file (as the precision is available at the top of the xml file).  This may end-up not running some tests that would any way pass.  eg. Relate tests seem to have a non-default precision model, but being a boolean predicate, I think the model doesn't matter?

jts-test-runner/src/input.rs
```diff
@@ -29,10 +30,19 @@ use super::Result;
...
 #[derive(Debug, Deserialize)]
 pub(crate) struct Run {
+    #[serde(rename = "precisionModel", default)]
+    pub precision_model: Option<PrecisionModel>,
+
     #[serde(rename = "case")]
     pub cases: Vec<Case>,
 }
 
+#[derive(Debug, Deserialize)]
```
rmanoka, line 40: This tries to parse: `<precisionModel scale="1.0" offsetx="0.0" offsety="0.0"/>` as `{ ty: "" }`
and `<precisionModel type='FLOATING'/>` as `{ty: "FLOATING"}`

rmanoka:
Rebased the PR to keep the history simpler.

geo/src/algorithm/bool_ops/mod.rs
```diff
@@ -65,19 +64,13 @@ impl<T: GeoFloat> BooleanOps for MultiPolygon<T> {
...
 }
 
 mod op;
 use op::*;
-
-mod rings;
-use rings::{Ring, Rings};
-
-mod laminar;
-use laminar::*;
```
michaelkirk, line 80: The original RFC references this module:

> The region enclosed by the rings form a [Laminar Set]. We re-use the
planar-sweep to again detect the inclusion relations of the rings, and
assemble as polygons.  Ref `geo/src/algorithm/bool_ops/laminar.rs`.

I don't do much work with RFCs. Does it make sense to try to keep the RFC up to date with the current state of things? Or is it intended to be "locked in time" as things were originally implemented?

michaelkirk (commented):
> We don't support this, and I've tried to parse and check if a "non-standard" precision model is used during overlay tests.

Is the fallout of this that we're no longer running some of the tests were previously passing? I'd think we'd want to keep running more and more of the test suite as we add functionality to geo. 

Before:

```
 INFO  jts_test_runner::runner > successes: 227, failures: 0
 INFO  jts_tests               > All 227 cases succeeded in JTS test suite.
```

After:

```
>  INFO  jts_test_runner::runner > successes: 138, failures: 0
 INFO  jts_tests               > All 138 cases succeeded in JTS test suite.
```

If this is related to the precision model stuff - then is it the case that some of these tests just "happened" to be passing (presumably because the actual numbers were compatible with both precision models)? 

It seems kind of shame to be dropping passing tests since they could serve as a regression check, but I'm OK with it. Mostly I just want to make sure this was intentional.

geo/src/algorithm/bool_ops/assembly.rs
```diff
@@ -0,0 +1,335 @@
...
+}
+
+impl<T: GeoFloat> Assembly<T> {
+    pub fn add_edge(&mut self, edge: LineOrPoint<T>) {
+        debug_assert!(edge.is_line());
+        self.segments.push(edge.into());
+    }
+    pub fn finish(self) -> MultiPolygon<T> {
+        let mut iter = CrossingsIter::from_iter(self.segments.iter());
+        let mut snakes = vec![];
```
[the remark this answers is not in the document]
rmanoka: The general problem statement is mentioned in the `Assembly` docs:

> Implements the construction of the final geometry from boundary
> line-segments obtained via the sweep. The line-segments are guaranteed to
> describe a bounded region, do not intersect in their interior, and are not
> degenerate (not a point).

I couldn't find a concrete reference of an algorithm that solves it.  I believe the approach here is along the same lines as in other impls. / boolean op. papers.  The `Snake` specifically is just something that grows by "eating' line-segments until it hits itself reminiscent of the classic family of video games;  these are then broken into the rings that form the boundaries of the output geom.

rmanoka:
> Before:
> 
> ```
>  INFO  jts_test_runner::runner > successes: 227, failures: 0
>  INFO  jts_tests               > All 227 cases succeeded in JTS test suite.
> ```
> 
> After:
> 
> ```
> >  INFO  jts_test_runner::runner > successes: 138, failures: 0
>  INFO  jts_tests               > All 138 cases succeeded in JTS test suite.
> ```
> 
> If this is related to the precision model stuff - then is it the case that some of these tests just "happened" to be passing (presumably because the actual numbers were compatible with both precision models)?
> 
> It seems kind of shame to be dropping passing tests since they could serve as a regression check, but I'm OK with it. Mostly I just want to make sure this was intentional.

So, we should definitely be passing a strict super-set than the current main branch (I assume before is current main).  What command gives the above outputs?  The relate tests are anyway being run separately (in `test_relate` in jts-test-runner crate) without the check for precision-model.  The issue is that in `fn test_all_general` [here](https://github.com/georust/geo/blob/2f05774ce491340ca1d0e7fb5c945970cc175627/geo/tests/jts_tests.rs#L16), we can't run all the tests as the overlays with precision model would fail.  I think a better structuring of which tests to avoid should help us run all the tests we want in one shot, but I don't see how to do that easily.

michaelkirk:
> What command gives the above outputs? 

`RUST_LOG=info cargo test` should do it

---

See https://github.com/georust/geo/pull/879 for a way to check that the number of tests doesn't decrease. If you merge that into this branch, tests will fail with:

> `thread 'tests::test_boolean_ops' panicked at 'We're running 180 less test cases than before. What happened to them?', jts-test-runner/src/lib.rs:99:13`

---

> The issue is that in fn test_all_general [here](https://github.com/georust/geo/blob/2f05774ce491340ca1d0e7fb5c945970cc175627/geo/tests/jts_tests.rs#L16), we can't run all the tests as the overlays with precision model would fail.

Ah right, that's kind of what I figured. Hmmm... 🤔 what if we did some egregious hack like only throwing away precision model tests for the Overlay operations? Something like this: https://github.com/georust/geo/compare/mkirk/fix-jts-overlay-tests?expand=1

(note that branch is based on #875 + #879)

rmanoka:
> Ah right, that's kind of what I figured. Hmmm... thinking what if we did some egregious hack like only throwing away alternate precision model test cases for the `Overlay` operations? Something like this: https://github.com/georust/geo/compare/mkirk/fix-jts-overlay-tests?expand=1

This is awesome;  pulled this in.  Also updated details in the RFC doc.

geo/src/algorithm/bool_ops/assembly.rs
```diff
@@ -0,0 +1,335 @@
...
+}
+
+fn rings_from_snakes<T: GeoFloat>(
+    snakes: &mut [Snake<T>],
+) -> (Vec<Ring<T>>, HashMap<usize, usize>) {
+    let mut snake_idx_map = HashMap::new();
+    let mut rings = vec![];
+    for idx in 0..snakes.len() {
+        if let Some(ls) = Snake::into_ring(snakes, idx, |midx| {
+            snake_idx_map.insert(midx, rings.len());
```
michaelkirk, line 215: I'm having some trouble understanding `rings_from_snakes`/`Snake::into_ring`. 

Starting here, what is `midx`? I'm guessing it stands for "map index" because thats how it's being used here, but where does it come from? Diving into the code deeper (in `Snake::into_ring`)... it looks like it's the snake_idx of the "current" snake we're trying to fit into a ring. So a ring will be made up of potentially many snakes, right?

And `snake_idx_map` is the map of snake_idx to the ring_idx of the ring that the snake was used to build. Is that all correct?

geo/src/algorithm/bool_ops/assembly.rs
```diff
@@ -0,0 +1,335 @@
...
+        }
+    }
+    (rings, snake_idx_map)
+}
+
+#[derive(Debug, Clone)]
+struct Snake<T: GeoFloat> {
+    points: Vec<SweepPoint<T>>,
+    start_pair: usize,
+    end_pair: Option<usize>,
```
michaelkirk, line 227: `start_pair` is the snake_idx of the snake that preceded this snake in the ring construction.
`end_pair` is the snake_idx of the snake that comes *after* this snake in the ring construction.

Is that right?

michaelkirk (commented):
I'm still trying to wrap my head around this, so I had some questions.

geo/src/algorithm/bool_ops/assembly.rs
rmanoka: The construction of rings from segments (via snakes) is essentially the classical eulerian [circuit algorithm](https://en.wikipedia.org/wiki/Eulerian_path#Hierholzer's_algorithm).  The `midx` is an extra requirement here: we maintain some information at the "snake" level (the region above the snake) which is needed to determine which rings are exterior.  The `snakes_idx_map` tells us which snakes were made into a ring, and `midx` is the idx of the snake in the array.

I intend to update [this book](https://rmanoka.github.io/geo-crossings/book/) with an exposition of the ops implementation too, but unsure when.  We could move this into the gh-pages here for reference on the sweep and ops algorithms.

geo/src/algorithm/bool_ops/assembly.rs
rmanoka: The topology guarantees that snakes start out and end in pairs (but they may not be the same pair).  The indices are needed to finally join the snakes into a ring.  At the point of ring construction, we'll have a bunch of points, and snakes connecting different points. This is a graph, and is Eulerian (number of snakes start/ending at each point is even), so we connect them into a set of rings.

There are some additional constraints on the output topology:  for instance two polygons touching at a vertex shouldn't be merged into one; or hole touching the exterior of a polygon shouldn't me merged into the exterior.  This requires additional info. stored in the snake.

PR #879, opened 2022-07-19
check expected number of tests

michaelkirk:
---

Make sure the number of JTS tests we're running doesn't change inadvertently.

This is more of a regression test of the test runner code itself, since it's easy to accidentally change the underlying logic in a way that would run fewer tests.


michaelkirk:
This has been incorporated into #875, which I expect we'll be merging soon, so I'm going to close this PR.

PR #882, opened 2022-07-28
Add the "validation" portion of the jts test suite, fix Contains edge case

michaelkirk:
---

This enables more of the JTS test suite cases. I haven't found authoritative documentation of what the different parts of the test suite are for, but this "validation" section seems to more comprehensively cover edge case geometries than the previous "general" portion we were running prior.

And it uncovered a bug! Or... at least an inconsistency with JTS with respect to degenerate 0-length line strings (like `LINESTRING(1 1,1 1)`

Specifically:
> failed TestRelatePL.xml case "P/L-2: a point and a zero-length line" with error: expected IntersectionMatrix(0FFFFFFF2), actual: IntersectionMatrix(0FFFFF1F2)


rmanoka (approved):
lgtm

michaelkirk:
bors r=rmanoka

PR #880, opened 2022-07-20
impl Contains for all geometries

michaelkirk:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/main/CODE_OF_CONDUCT.md).
- [x] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

Based on the test suite and correctness fixes in #882 so please review that first.

Fixes https://github.com/georust/geo/issues/535

PR #884, opened 2022-07-28
Adds Within trait and fixes more Contains edge cases

michaelkirk:
---

~~Based on #880, so review that first.~~ Merged! This is ready for review!

The new `Within` trait is implemented in terms of `Contains`. 

Running the "within" tests from the JTS suite exposed some further inconsistencies with how we're handling line string edge cases.

### DIFF
--- a/geo/src/algorithm/contains/line_string.rs
+++ b/geo/src/algorithm/contains/line_string.rs
@@ -16,8 +16,8 @@ where
             return false;
         }
 
-        if self.is_closed() && coord == &self.0[0] {
-            return true;
+        if coord == &self.0[0] || coord == self.0.last().unwrap() {
+            return self.is_closed();
         }
 
         self.lines()
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Add `Within` trait to determine if Geometry A is completely within Geometry B
+  * <https://github.com/georust/geo/pull/884>
 * Add `Contains` impl for all remaining geometry types.
   * <https://github.com/georust/geo/pull/880>
 * Add `Scale` affine transform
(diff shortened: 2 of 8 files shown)


geo/src/algorithm/contains/line_string.rs
```diff
@@ -16,8 +16,8 @@ where
             return false;
         }
 
-        if self.is_closed() && coord == &self.0[0] {
-            return true;
+        if coord == &self.0[0] || coord == self.0.last().unwrap() {
+            return self.is_closed();
         }
```
michaelkirk, line 21: Here's an example of the kind of test that was failing before this change:

```
<case>
<desc>P/nsL.1-5-1: a point on a non-simple LineString's start point with crossing line segments [dim(0){A.P.Int = B.nsL.Bdy.SPx}]</desc>
  <a>
    POINT(110 110)
  </a>
  <b>
    LINESTRING(110 110, 220 20, 20 20, 110 110, 220 220)
  </b>
  <test>
    <op name="relate" arg1="A" arg2="B" arg3="F0FFFF102">true</op>
  </test>
  <test><op name="contains"   arg1="A" arg2="B">false</op></test>
  <test><op name="coveredBy"  arg1="A" arg2="B">true</op></test>
  <test><op name="covers"     arg1="A" arg2="B">false</op></test>
  <test><op name="crosses"    arg1="A" arg2="B">false</op></test>
  <test><op name="disjoint"   arg1="A" arg2="B">false</op></test>
  <test><op name="equalsTopo" arg1="A" arg2="B">false</op></test>
  <test><op name="intersects" arg1="A" arg2="B">true</op></test>
  <test><op name="overlaps"   arg1="A" arg2="B">false</op></test>
  <test><op name="touches"    arg1="A" arg2="B">true</op></test>
  <test><op name="within"     arg1="A" arg2="B">false</op></test>
</case>
```

michaelkirk:
Now that https://github.com/georust/geo/pull/880 has been merged, this is ready for review!

michaelkirk:
bors r=frewsxcv

PR #892, opened 2022-08-12
remove old deprecated code from `geo`

michaelkirk:
---

I don't know that we have an official policy on how long to keep deprecated code around, but based on our time frame for MSRV, I've left some things in `geo` whose deprecation was released less than 6 months ago.

Note that this only touches the `geo` crate, not `geo-types`. There's plenty of old deprecated things in `geo-types`, but we haven't warranted a breaking release in that crate yet.


lnicola (approved):

michaelkirk:
bors r=lnicola

michaelkirk:
The idea of a deprecation timeframe was a hedge against our sometimes frequent breaking releases. The benefits of a deprecation cycle are somewhat lost if the deprecation cycle is 2 weeks.  e.g.:

Screenshot 2023-01-13 at 11 16 02 AM

It's true that tracking timeframes is a bit of work. My process was to scan the tags dates to see which release was about 6 months ago and then delete any matching `deprecated(since="...")` tags. It probably added 5 minutes to the release process (which isn't nothing).

I'm not married to "6 months" of deprecation or even any particular timeframe, but I think it's worthwhile to put some effort into letting people update to the latest geo version, while continuing to use code that was only deprecated a couple weeks prior. I'm comfortable leaving it as a judgement call.

PR #893, opened 2022-08-16
Set MSRV for geo to 1.58 in cargo.toml

debugger22:
---

`geo` crate fails to compile on Rust versions lower than `1.58`.


lnicola:
Thanks!

bors r+

debugger22:
Thanks! This was my first open source contribution in Rust! 😄

PR #904, opened 2022-09-10
Detect outliers using LOF

urschrei:
---

Still draft due to:
- <s>perf: two of the LRD passes look very similar. I need to figure out whether this can be tidied up, or whether it's ultimately not going to affect perf much</s> Can't think of a particularly more elegant way to do this
- Tests, tests, tests
- Other geometries / collections?

### DIFF
--- a/geo/src/algorithm/outlier_detection.rs
+++ b/geo/src/algorithm/outlier_detection.rs
@@ -1,0 +1,446 @@
+use std::iter::Sum;
+use std::ops::RangeInclusive;
+
+use crate::{GeoFloat, MultiPoint, Point};
+
+use rstar::primitives::GeomWithData;
+use rstar::RTree;
+
+/// Calculate the [Local Outlier Factor](https://en.wikipedia.org/wiki/Local_outlier_factor) of a set of points
+///
+/// Based on: Breunig, M., Kriegel, H., Ng, R., and Sander, J. (2000). *LOF: identifying density-based local
+/// outliers.* In ACM Int. Conf. on Management of Data, pages 93-104. doi: [10.1145/335191.335388](https://doi.org/10.1145/335191.335388)
+///
+/// LOF is an unsupervised algorithm that uses local data for anomaly detection.
+///
+/// Outlier vs inlier classification is **highly dependent** on the shape of the data. LOF values <= 1
+/// can generally be considered inliers, but e.g. a highly concentrated, uniform dataset could result in
+/// points with a LOF of 1.05 being outliers.
+/// LOF scores should thus be evaluated in the context of the dataset as a whole in order to classify outliers.
+///
+/// If you wish to run multiple outlier detection processes with differing neighbour counts in order
+/// to build up data for more robust detection (see p. 100-1 above), you can use the [`OutlierDetection::prepared_detector`] method, which retains
+/// the spatial index and point set between runs for greater efficiency. The [`OutlierDetection::generate_ensemble`] method
+/// will efficiently run the LOF algorithm over a contiguous range of neighbour inputs,
+/// allowing aggregations to be carried out over the resulting data.
+pub trait OutlierDetection<T>
+where
+    T: GeoFloat,
+{
+    /// The LOF algorithm. `k_neighbours` specifies the number of neighbours to use for local outlier
+    /// classification. The paper linked above (see p. 100) suggests a `k_neighbours` value of 10 - 20
+    /// as a lower bound for "real-world"
+    /// data.
+    ///
+    /// # Note on Erroneous Input
+    /// If `k_neighbours` >= points in the set, or `k_neighbours` < 1, all input points will be returned with an LOF score of 1.
+    /// If there are at least `k_neighbours` duplicate points of an input point, LOF scores can be `∞` or `NaN`.
+    /// It is thus advisable to **deduplicate** or otherwise ensure the uniqueness of the input points.
+    ///
+    /// # Note on Returned Points
+    /// Outlier scores are always returned corresponding to input point order
+    ///
+    /// # Examples
+    ///
+    /// ## MultiPoint
+    ///
+    /// ```
+    /// use approx::assert_relative_eq;
+    /// use geo::OutlierDetection;
+    /// use geo::{point, MultiPoint};
+    ///
+    /// let v = vec![
+    ///     point!(x: 0.0, y: 0.0),
+    ///     point!(x: 0.0, y: 1.0),
+    ///     point!(x: 3.0, y: 0.0),
+    ///     point!(x: 1.0, y: 1.0),
+    /// ];
+    ///
+    /// let lofscores = v.outliers(2);
+    /// // The third point is an outlier, resulting in a large LOF score
+    /// assert_relative_eq!(lofscores[2], 3.0);
+    /// // The last point is an inlier, resulting in a small LOF score
+    /// assert_relative_eq!(lofscores[3], 1.0);
+    /// ```
+    ///
+    /// ## Computing indices, sorting by LOF score
+    ///```
+    /// use geo::OutlierDetection;
+    /// use geo::{point, MultiPoint};
+    ///
+    /// // these points contain 4 strong outliers
+    /// let v = vec![
+    ///     point!(x: 0.16, y: 0.14),
+    ///     point!(x: 0.15, y: 0.33),
+    ///     point!(x: 0.37, y: 0.25),
+    ///     point!(x: 0.3 , y: 0.4),
+    ///     point!(x: 0.3 , y: 0.1),
+    ///     point!(x: 0.3 , y: 0.2),
+    ///     point!(x: 1.3 , y: 2.3),
+    ///     point!(x: 1.7 , y: 0.2),
+    ///     point!(x: 0.7 , y: -0.9),
+    ///     point!(x: 0.21, y: 2.45),
+    ///     point!(x: 0.8 , y: 0.7),
+    ///     point!(x: 0.9 , y: 0.7),
+    ///     point!(x: 0.8 , y: 0.6),
+    ///     point!(x: 0.73, y: 0.65),
+    ///     point!(x: 0.9 , y: 0.6),
+    ///     point!(x: 1.0, y: 0.6),
+    ///     point!(x: 1.0, y: 0.7),
+    ///     point!(x: 0.25, y: 0.29),
+    ///     point!(x: 0.2 , y: 0.2),
+    /// ];
+    /// let lofs = &mut v.outliers(3);
+    /// let mut idx_lofs: Vec<(usize, f64)> = lofs
+    ///     .iter()
+    ///     .enumerate()
+    ///     .map(|(idx, score)| (idx, *score))
+    ///     .collect();
+    /// // sort by LOF score
+    /// idx_lofs.sort_by(|a, b| a.1.partial_cmp(&b.1).unwrap());
+    /// // most likely outliers first
+    /// idx_lofs.reverse();
+    /// // four outliers, LOF scores way above 10
+    /// idx_lofs
+    ///     .iter()
+    ///     .take(4)
+    ///     .for_each(|score| assert!(score.1 > 10.0));
+    ///```
+    fn outliers(&self, k_neighbours: usize) -> Vec<T>;
+
+    /// Create a prepared outlier detector allowing multiple runs to retain the spatial index in use.
+    /// A [`PreparedDetector`] can efficiently recompute outliers with different `k_neigbhours` values.
+    fn prepared_detector(&self) -> PreparedDetector<T>;
+
+    /// Perform successive runs with `k_neighbours` values between `bounds`,
+    /// generating an ensemble of LOF scores, which may be aggregated using e.g. min, max, or mean
+    ///
+    /// # Examples
+    ///```
+    /// use geo::OutlierDetection;
+    /// use geo::{point, Point, MultiPoint};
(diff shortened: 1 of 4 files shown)


geo/src/algorithm/outlier_detection.rs
```diff
@@ -0,0 +1,302 @@
...
+            // for each point's neighbour set, calculate kth distance
+            let kth_dist = neighbours
+                .iter()
+                .map(|(_, distance)| distance)
+                .last()
+                .unwrap();
+            // sum neighbour set LRD scores
+            let lrd_scores: T = neighbours
+                .iter()
+                .map(|(neighbour, _)| local_reachability_densities[neighbour.data])
```
urschrei, line 135: I suspect there's a way to calculate LRDs for a subset ( the point's neighbours) as we go, but I can't get it right so we're stuck with `local_reachability_densities` and a redundant `kth_dist` calculation for now. I think it would ultimately end up being two passes over the data anyway, but it might use less space.

urschrei:
The paper notes that in order to calculate outliers in a robust fashion it might be necessary to calculate an "ensemble" using a range of nearest-neighour values. Since the r*-tree is expensive to fill and can be re-used, it would make sense to have a "prepared outlier calculator" API that holds on to the tree, allowing you to pass in a new `k_neighbours` value and getting a re-calculated dataset back out – a dataset is guaranteed to have the same dimensions (one `LofScore` entry per input point), so it will be easy to either aggregate scores once you've run it over a range of input neighbour sizes, or calculate them as you go in order to save space if you know what aggregate measure will give you good results and you want to save space.

All suggestions for an API gratefully received.

urschrei:
> The paper notes that in order to calculate outliers in a robust fashion it might be necessary to calculate an "ensemble" using a range of nearest-neighour values. Since the r*-tree is expensive to fill and can be re-used, it would make sense to have a "prepared outlier calculator" API that holds on to the tree, allowing you to pass in a new `k_neighbours` value and getting a re-calculated dataset back out – a dataset is guaranteed to have the same dimensions (one `LofScore` entry per input point), so it will be easy to either aggregate scores once you've run it over a range of input neighbour sizes

This is done!

geo/src/algorithm/outlier_detection.rs
```diff
@@ -0,0 +1,417 @@
...
+}
+
+fn lof<T>(
+    points: &[Point<T>],
+    tree: &RTree<GeomWithData<Point<T>, usize>>,
+    kneighbours: usize,
+) -> Vec<LofScore<T>>
+where
+    T: GeoFloat + Sum,
+{
```
michaelkirk, line 203:
```suggestion
{
    assert!(kneighbours > 0);
```

Would give a better panic message than the later `unwrap` panic.

Also, I'm not sure the meaning of this algo when kneighbors == 1, but maybe that's fine and not worth panicking over.

geo/src/algorithm/outlier_detection.rs
```diff
@@ -0,0 +1,417 @@
...
+        .iter()
+        .map(|neighbours| {
+            // for each point's neighbour set, calculate kth distance
+            let kth_dist = neighbours
+                .iter()
+                .map(|(_, distance)| distance)
+                .last()
+                .unwrap();
+            T::from(neighbours.len()).unwrap()
+                / neighbours
```
michaelkirk, line 234: It seems like we'd get a div-by-0 here when there are sufficient repeated points. Does the algorithm provide sensible results in that case?

geo/src/algorithm/outlier_detection.rs
```diff
@@ -0,0 +1,417 @@
...
+    /// let mp = MultiPoint(v);
+    ///
+    /// let lofscores = mp.outliers(3);
+    ///
+    /// assert_relative_eq!(lofscores[2].score, 3.3333333333333335);
+    /// ```
+    fn outliers(&self, k_neighbours: usize) -> Vec<LofScore<T>>;
+
+    /// Create a prepared outlier detector for ensemble detection operations
+    fn prepared_detector(&self) -> PreparedDetector<T>;
```
michaelkirk, line 66: Nice! Maybe mention in the docs that you'd want to do this for efficiency sake. It's implied by use of the word "prepared" if you know what that means, but is that kind of obscure?

Maybe something that mentions "efficiency" like:
>  /// A PreparedDetector can efficiently recompute outliers with different k_neigbhors value.

geo/src/algorithm/outlier_detection.rs
urschrei: It returns either infinity or `NaN` at the moment (which is documented). I'm not sure how we can avoid that without affecting performance, though…

geo/src/algorithm/outlier_detection.rs
urschrei: 1 will just give lower-quality results. Note that passing `kneighbours = 0` won't cause a panic down the line - we immediately return all the LOF scores set to 1, which is essentially saying "we couldn't calculate them". I've added a `debug_assert` – dunno if that's good practice though.

michaelkirk: Hmm, when I run this:
```
        let v = [
            Point::new(0.0, 1.0),
            Point::new(0.0, 2.0),
            Point::new(3.0, 1.0),
            Point::new(1.0, 1.0),
        ];

        let scores = v.outliers(0);
```

I see this panic:
```
thread 'algorithm::outlier_detection::tests::test_same_points' panicked at 'called `Option::unwrap()` on a `None` value', geo/src/algorithm/outlier_detection.rs:232:18
stack backtrace:
   0: rust_begin_unwind
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/panicking.rs:142:14
   2: core::panicking::panic
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/panicking.rs:48:5
   3: core::option::Option<T>::unwrap
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/option.rs:775:21
   4: geo::algorithm::outlier_detection::lof::{{closure}}
             at ./src/algorithm/outlier_detection.rs:228:28
   5: core::iter::adapters::map::map_fold::{{closure}}
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/iter/adapters/map.rs:84:28
   6: core::iter::traits::iterator::Iterator::fold
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/iter/traits/iterator.rs:2370:21
   7: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/iter/adapters/map.rs:124:9
   8: core::iter::traits::iterator::Iterator::for_each
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/iter/traits/iterator.rs:787:9
   9: <alloc::vec::Vec<T,A> as alloc::vec::spec_extend::SpecExtend<T,I>>::spec_extend
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/alloc/src/vec/spec_extend.rs:40:17
  10: <alloc::vec::Vec<T> as alloc::vec::spec_from_iter_nested::SpecFromIterNested<T,I>>::from_iter
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/alloc/src/vec/spec_from_iter_nested.rs:62:9
  11: <alloc::vec::Vec<T> as alloc::vec::spec_from_iter::SpecFromIter<T,I>>::from_iter
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/alloc/src/vec/spec_from_iter.rs:33:9
  12: <alloc::vec::Vec<T> as core::iter::traits::collect::FromIterator<T>>::from_iter
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/alloc/src/vec/mod.rs:2645:9
  13: core::iter::traits::iterator::Iterator::collect
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/iter/traits/iterator.rs:1792:9
  14: geo::algorithm::outlier_detection::lof
             at ./src/algorithm/outlier_detection.rs:224:48
  15: geo::algorithm::outlier_detection::PreparedDetector<T>::outliers
             at ./src/algorithm/outlier_detection.rs:192:9
  16: <[geo_types::geometry::point::Point<T>] as geo::algorithm::outlier_detection::OutlierDetection<T>>::outliers
             at ./src/algorithm/outlier_detection.rs:299:9
  17: geo::algorithm::outlier_detection::tests::test_same_points
             at ./src/algorithm/outlier_detection.rs:429:22
  18: geo::algorithm::outlier_detection::tests::test_same_points::{{closure}}
             at ./src/algorithm/outlier_detection.rs:420:5
  19: core::ops::function::FnOnce::call_once
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/ops/function.rs:248:5
  20: core::ops::function::FnOnce::call_once
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/ops/function.rs:248:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

geo/src/algorithm/outlier_detection.rs
```diff
@@ -0,0 +1,417 @@
...
+    /// Perform successive runs with `k_neighbour` values between `ubound` and `lbound` (inclusive),
+    /// generating an ensemble of LOF scores, which may be aggregated using e.g. min, max, or mean
+    fn generate_ensemble(&self, lbound: usize, ubound: usize) -> Vec<Vec<LofScore<T>>>;
+}
+
+/// This struct holds the original index of the input point, and the calculated LOF score
+#[derive(Copy, Clone, Debug)]
+pub struct LofScore<T> {
+    /// The original index of the input [`Point`]
+    pub index: usize,
```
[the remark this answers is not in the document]
urschrei: Just to clarify:
Are you suggesting that instead of an outlier score (and attendant custom type), by default we just return a `Vec<T>` and if someone wants the indices they can do `self.outliers(neighbor_count).iter().enumerated({ |(score, idx)|  ...  })`?

geo/src/algorithm/outlier_detection.rs
urschrei: Ugh yes my fault: we only check that neighbours < points. But we could just extend that check to neighbours > 0 instead of just panicking? Which is better: indeterminate result, or a panic?

geo/src/algorithm/outlier_detection.rs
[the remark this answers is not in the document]
michaelkirk: Sorry, I see now that my example code had a bug - it should have been `Vec<T>`, not just a single `T`.

So, yes! You've understood what I was suggesting. Do you think it's an improvement?

geo/src/algorithm/outlier_detection.rs
michaelkirk: Neighbors less than 0 seems almost certainly to be a programmer error, so I'm very biased for panicing in debug builds.

Doing our best to avoid panics in release builds seems nice though, so I have a slight bias in favor of "blundering on" there.

urschrei: We now panic in release, and return 1 in release if `neighbours > points` OR `neighbours < 1`.

geo/src/algorithm/outlier_detection.rs
urschrei: I've made this possible result extra-clear in the docs, and because this input can't produce a panic I'm OK with it as is.

geo/src/algorithm/outlier_detection.rs
michaelkirk: > We now panic in release,

We now panic in ~~release~~ debug,

I think this was just a typo in your github comment. The code looks right -thank you for making this change!

geo/src/algorithm/outlier_detection.rs
```diff
@@ -0,0 +1,446 @@
...
+    /// idx_lofs.sort_by(|a, b| a.1.partial_cmp(&b.1).unwrap());
+    /// // most likely outliers first
+    /// idx_lofs.reverse();
+    /// // four outliers, LOF scores way above 10
+    /// idx_lofs
+    ///     .iter()
+    ///     .take(4)
+    ///     .for_each(|score| assert!(score.1 > 10.0));
+    ///```
+    fn outliers(&self, k_neighbours: usize) -> Vec<T>;
```
michaelkirk, line 110: You changed this to `outlier_scores` in response to my feedback, and then changed it back to `outliers`. I just wanted to make sure that was intentional.

urschrei: Yep! Sorry.

urschrei:
All review comments addressed, I think.

michaelkirk (approved):
LGTM! Not sure if you were looking for more reviews, so I'll leave it to you if you want to merge it.

urschrei:
Thanks for the review! I think it's in good shape now but I'll leave it open for a couple of days / until @frewsxcv has kicked the tires on it.

PR #907, opened 2022-09-13
Fix trivial_hull implementation for collinear points

### LINKED ISSUE #905: Convex hull of three colinear points contains all three points
When calculating the convex hull of `(0, 0)`, `(1, 1)` and `(2, 2)`, the result contains `(1, 1)` even though it should not.

This is also inconsistent with computing the convex hull for `n > 3` colinear points, which only contains the two endpoints, which can be tested with the following code:
```rust
use geo::geometry::*;
use geo::{point, ConvexHull};

fn hull_size(n: usize) -> usize {
    let points: Vec<_> = (0..n as isize).map(|i| point!(x: i, y: i)).collect();
    let mp = MultiPoint::new(points);

    let hull = mp.convex_hull();
    hull.exterior().0.len()
}

fn main() {
    for i in 0..10 {
        println!("{}: {}", i, hull_size(i));
    }
}
```

When running this with `geo` v0.23.0 the output is:
```
0: 0
1: 2
2: 3
3: 4
4: 3
5: 3
6: 3
7: 3
8: 3
9: 3
```

`3: 3` is the expected output instead of `3: 4`.

Tyilo:
---
Fixes #905

### DIFF
--- a/geo/src/algorithm/convex_hull/mod.rs
+++ b/geo/src/algorithm/convex_hull/mod.rs
@@ -1,5 +1,3 @@
-use std::cmp::Ordering;
-
 use crate::geometry::{Coordinate, LineString, Polygon};
 use crate::kernels::*;
 use crate::GeoNum;
--- a/geo/src/algorithm/convex_hull/mod.rs
+++ b/geo/src/algorithm/convex_hull/mod.rs
@@ -1,4 +1,7 @@
+use std::cmp::Ordering;
+
 use crate::geometry::{Coordinate, LineString, Polygon};
+use crate::kernels::*;
 use crate::GeoNum;
 
 /// Returns the convex hull of a Polygon. The hull is always oriented counter-clockwise.
@@ -77,7 +80,21 @@ where
     // are to be included.
     let mut ls: Vec<Coordinate<T>> = points.to_vec();
     if !include_on_hull {
+        fn coord_compare<T: GeoNum>(c1: &Coordinate<T>, c2: &Coordinate<T>) -> Ordering {
+            match c1.x.partial_cmp(&c2.x) {
+                None | Some(Ordering::Equal) => match c1.y.partial_cmp(&c2.y) {
+                    Some(ordering) => ordering,
+                    None => Ordering::Equal,
+                },
+                Some(ordering) => ordering,
+            }
+        }
+
+        ls.sort_by(coord_compare);
         ls.dedup();
+        if ls.len() == 3 && T::Ker::orient2d(ls[0], ls[1], ls[2]) == Orientation::Collinear {
+            ls.remove(1);
+        }
     }
 
     // A linestring with a single point is invalid.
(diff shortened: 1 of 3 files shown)


urschrei:
Thanks! This looks good to me, but it would be great if you could add a brief explanation of the fix to your comment above, for those of us who can't easily visualise what's going on.

Tyilo:
If `include_hull` is `false`, then given the input `[(0, 0), (1, 1), (2,2)]` (in any order) the output should be `[(0, 0), (2, 2), (0, 0)]` (or `[(2, 2), (0, 0), (2, 2)]`).

Before this fix, with input `[(0, 0), (1, 1), (2, 2)]` the output is `[(0, 0), (1, 1), (2, 2), (0, 0)]`.

To fix it, we need to remove the "middle" point if the three points are colinear.
To do this, we first sort the three points (first by `x`, then by `y`) and then remove the point at index 1, if they are colinear.

The sorting is needed to give the correct output for any permutation of `[(0, 0), (1, 1), (2, 2)]`, for instance `[(2, 2), (0, 0), (1, 1)]`.

geo/src/algorithm/convex_hull/mod.rs
```diff
@@ -77,7 +80,21 @@ where
     // are to be included.
     let mut ls: Vec<Coordinate<T>> = points.to_vec();
     if !include_on_hull {
+        fn coord_compare<T: GeoNum>(c1: &Coordinate<T>, c2: &Coordinate<T>) -> Ordering {
```
rmanoka, line 83: Could you use `utils::lex_cmp` for this?  It compares by `x` and then by `y` but I don't think that should affect this logic.

geo/src/algorithm/convex_hull/mod.rs
```diff
@@ -77,7 +80,21 @@ where
     // are to be included.
     let mut ls: Vec<Coordinate<T>> = points.to_vec();
```
rmanoka, line 81: Is it possible to avoid this extra `Vec` allocation?  Can you sort the input mut-slice directly?

rmanoka (approved):
Thanks for the PR @Tyilo  !   Just have a few minor suggestions, but the overall logic looks good to me.

geo/src/algorithm/convex_hull/mod.rs
Tyilo: We can sort the mut-slice directly, however we can't do `ls.remove(1);` on a slice, so I don't see how we can do it without creating a new `Vec`.

@rmanoka I find it weird that the convex hull algorithms (`trivial_hull`, `graham_hull` and `quick_hull`) takes `points: &mut [Coordinate<T>]` as input.
We don't guarantee anything about how `points` has been mutated by the algorithm, right?

So wouldn't it be better to take either `&[Coordinate<T>]` or `Vec<Coordinate<T>>` as input?

urschrei: I think we need mutable points for the [`partition_slice`](https://github.com/georust/geo/blob/main/geo/src/utils.rs#L11-L32) function which reorders coordinates based on a predicate?

Tyilo: > I think we need mutable points for the [`partition_slice`](https://github.com/georust/geo/blob/main/geo/src/utils.rs#L11-L32) function which reorders coordinates based on a predicate?

That would also work if we took input as `Vec<Coordinate<T>>`.

urschrei: > > I think we need mutable points for the [`partition_slice`](https://github.com/georust/geo/blob/main/geo/src/utils.rs#L11-L32) function which reorders coordinates based on a predicate?
> 
> That would also work if we took input as `Vec<Coordinate<T>>`.

I am confused – could you say more about what you have in mind here?

Tyilo: > > > I think we need mutable points for the [`partition_slice`](https://github.com/georust/geo/blob/main/geo/src/utils.rs#L11-L32) function which reorders coordinates based on a predicate?
> > 
> > 
> > That would also work if we took input as `Vec<Coordinate<T>>`.
> 
> I am confused – could you say more about what you have in mind here?

So there are basically three options for taking slices as input in rust:
- Owned: `Box<[Coordinate<T>]>` or `Vec<Coordinate<T>>`. Makes sense if we need to consume the input to do the computation.
- Borrowed: `&[Coordinate<T>]`. Makes sense if we only need to read the input and consuming the input (as in "Owned") doesn't speed up the computation.
- Mutably borrowed: `&mut [Coordinate<T>]`. Makes sense if the computation needs to change the input, but also the user of the function should (at least in some situations) be interested in the mutated input. If mutated input is never useful for the user, we should prefer to use an owned input.

I'm arguing that the mutated `points` input is never useful for the user and thus to prevent the user from accidentally using the now garbage list of points and to make the api more clear, we should choose the "Owned" option.

rmanoka: Yeah, the function signature was primarily whatever works for the implementation.  Note that the usage is anyway via the trait.  If it simplifies code without extra allocations, we can move to something more intuitive.

Tyilo: > Yeah, the function signature was primarily whatever works for the implementation. Note that the usage is anyway via the trait. If it simplifies code without extra allocations, we can move to something more intuitive.

Makes sense. I don't think that change is relevant to this pull request.
Do I need to fix anything else before this can get merged?

rmanoka:
@Tyilo pls. pull changes from main to fix the other CI issues. I think it's ready to merge once CI passes.

Tyilo:
> @Tyilo pls. pull changes from main to fix the other CI issues. I think it's ready to merge once CI passes.

Done.

urschrei:
bors r=rmanoka

PR #908, opened 2022-09-14
Simplify `CoordsIter` `trait` with GATs by removing an unneeded trait lifetime

frewsxcv:
---

https://github.com/rust-lang/rust/issues/44265


frewsxcv:
GAT's are now in stable Rust. So this can be ready whenever we want to update our MSRV

https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html

michaelkirk (approved):
1.65 was released about 9 months ago. Given our `MSRV  >= CURRENT - 3` (about 6 months), I'd say we should go for it and bump to 1.65.

IIRC there can be some nice downstream cleanups for traits implemented in terms of CoordsIter like ConvexHull.

michaelkirk (approved):

PR #899, opened 2022-08-29
point.x_mut() and point.y_mut()

Junkyards3:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/main/CODE_OF_CONDUCT.md).
- [x] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---
I added methods point.x_mut() and point.y_mut() as described in the issue #898. This is my first PR, I read the code of conduct and contribution guidelines but may have missed something, I'd be happy to hear about what I can do better

PR #909, opened 2022-09-14
fixup! added point.x_mut() and point.y_mut()

michaelkirk:
---

The changelog was correct, but it didn't coincide with the actual method implementations which were not named according to convention.

This hasn't been released yet, so it's not considered a breaking change.

Followup to #899

### DIFF
--- a/geo-types/src/geometry/point.rs
+++ b/geo-types/src/geometry/point.rs
@@ -117,11 +117,11 @@ impl<T: CoordNum> Point<T> {
     /// use approx::assert_relative_eq;
     /// use geo_types::Point;
     /// let mut p = Point::new(1.234, 2.345);
-    /// let mut p_x = p.mut_x();
+    /// let mut p_x = p.x_mut();
     /// *p_x += 1.0;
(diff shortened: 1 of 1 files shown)


rmanoka:
Wouldn't it be nicer to instead of Deref and DerefMut to Coordinate instead?  We might need a repr transparent on `Point`, but seems fair to add it.

rmanoka:
The CI failure seems to be from arbitrary crate's min-rust requirement I think.

frewsxcv:
My vote would be to merge this, and then have a separate discussion about adding a `Deref` `impl` for `Point` to `Coordinate`. There are some people in the Rust community who think `Deref` should be reserved for pointer types, so it could warrant a little discussion.

lnicola (approved):

michaelkirk:
bors r=lnicola

I'm gonna go ahead and merge this (I think CI will succeed now that #910). 

I'm happy to talk about Deref, though I gotta tell you, lately I've been thinking we should just resurrect #15.

Issue #732, opened 2022-02-20
Add the new Transform trait to the root docs index


frewsxcv:
Add it here: https://github.com/georust/geo/blob/1d3b32970a49746de6ce2400305725edd21f7c15/geo/src/lib.rs#L124-L130

PR #915, opened 2022-10-01
Add docline for Transform trait to root.

mishakuzma:
---

This PR fixes Issue #732 by adding a doc line about the Transform trait to the root index in the recommended place.


michaelkirk:
bors r+

thanks!

PR #916, opened 2022-10-05
Handle NaN gracefully when closing LineString

frewsxcv:
---

Prior to this commit:

```rust
let start = coord! { x: 0., y: f32::NAN };
let end = coord! { x: 0., y: f32::NAN };
let mut line_string = LineString::new(vec![start, end]);
assert_eq!(2, line_string.0.len());
line_string.close();
assert_eq!(3, line_string.0.len());
```

After this commit:

```rust
let start = coord! { x: 0., y: f32::NAN };
let end = coord! { x: 0., y: f32::NAN };
let mut line_string = LineString::new(vec![start, end]);
assert_eq!(2, line_string.0.len());
line_string.close();
assert_eq!(2, line_string.0.len());
```

This is because `NaN != NaN`

### DIFF
--- a/geo-types/src/geometry/coordinate.rs
+++ b/geo-types/src/geometry/coordinate.rs
@@ -95,6 +95,41 @@ impl<T: CoordNum> Coordinate<T> {
     pub fn x_y(&self) -> (T, T) {
         (self.x, self.y)
     }
+
+    /// # Examples
+    ///
+    /// ```ignore
+    /// use geo_types::coord;
+    ///
+    /// let c1 = coord! {
+    ///     x: 40.02f64,
+    ///     y: f64::NAN,
+    /// };
+    /// let c2 = coord! {
+    ///     x: 40.02f64,
+    ///     y: f64::NAN,
+    /// };
+    ///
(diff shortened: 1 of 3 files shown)


michaelkirk:
I'm curious how you came to this. Was this encountered in real world data, and does this change make things work better for you?

I'm a little bit worried to try to put in the effort of making sense of NaN like this. The definition of closed means the first and last point are the same - but NaN can't be the same as anything else, even another Nan (hence `NaN != NaN`).

So, while I agree that the current behavior is weird, this change makes things a little more complicated and I'm not sure it's actually meaningfully improving the chaos that is NaN. I think it would be equally reasonable to throw up our hands and say "The behavior of `close` is not defined for linestrings with NaN coordinates". 

If we did want a more coherent way to handle "closed" WRT NaN, the only thing I can think of would be to introduce a LinearRing type and `LineString::close -> LinearRing`, but that'd be a bigger change. And ultimately if your geometry contains NaN, I'd expect a lot of downstream algorithms will also fail you.

lnicola:
Is there some prior art we can look at? How does JTS handle NaNs?

Issue #917, opened 2022-10-06
polygon_a..euclidean_distance(&polygon_b) result is different than what I expected


drbartling:
I expected the distance between two polygons to be equal to the smallest distance between the two polygons.

In each of these cases, I have a triangle just above a rectangle.

```
__
\/
 __
|__|
```

Am I expecting the wrong result?  If so, could we add some examples for polygon to polygon in the docs?
Otherwise

```
    use float_cmp::assert_approx_eq;
    use geo::{
        orient::Direction, polygon, Coordinate, EuclideanDistance, LineString,
        Orient, Polygon,
    };
    use pretty_assertions::assert_eq;
    use rstest::*;

   fn simple_check_local_spacing_test(
        removed_geometry: &[Polygon],
        local_geometry: &[Polygon],
        region_minimum_spacing: f64,
        design_minimum_spacing: f64,
        expected_minimum_spacing: f64,
    ) -> Result<()> {
        let output_minimum_spacing = check_local_spacing(
            removed_geometry,
            local_geometry,
            design_minimum_spacing,
            region_minimum_spacing,
        )?;
        assert_eq!(expected_minimum_spacing, output_minimum_spacing);
        Ok(())
    }

    // Proof of differences in distance methods
    #[rstest]
    #[case(
        polygon!(
            (x: 0_f64, y: 0_f64),
            (x: 100_f64, y: 0_f64),
            (x: 100_f64, y: 100_f64),
            (x: 0_f64, y: 100_f64),
        ),
        polygon!(
            (x: 50_f64, y: 150_f64),
            (x: 100_f64, y: 200_f64),
            (x: 0_f64, y: 200_f64),
        ),
        50.0, 158.114,
    )]
    #[case(
        polygon!(
            (x: 0_f64, y: 0_f64),
            (x: 300_f64, y: 0_f64),
            (x: 300_f64, y: 100_f64),
            (x: 0_f64, y: 100_f64),
        ),
        polygon!(
            (x: 50_f64, y: 150_f64),
            (x: 100_f64, y: 200_f64),
            (x: 0_f64, y: 200_f64),
        ),
        50.0, 200.000,
    )]
    #[case(
        polygon!(
            (x: 0_f64, y: 0_f64),
            (x: 300_f64, y: 0_f64),
            (x: 300_f64, y: 100_f64),
            (x: 0_f64, y: 100_f64),
        ),
        polygon!(
            (x: 100_f64, y: 150_f64),
[...]

PR #920, opened 2022-10-11
Remove rotating calipers algorithm

urschrei:
---

The `Polygon-Polygon` distance rotating calipers algorithm for convex polygons has been a source of bugs for a while, and I've exhausted my ability to fix the latest one (#917), so, as its author, I'm proposing the algorithm's removal.

<s>This PR currently makes one other small change: it removes the `Polygon-Polygon` intersection check as it's relatively expensive and our iterative NN algorithm performs the check anyway as part of the point-line distance check.</s>


michaelkirk (approved):
I'm sad to see this implementation go, but it's really important we have correct implementations. 

I hope we can bring it back in the future, but I agree it makes sense to disable it for now.

frewsxcv:
Should we add a regression test after this is removed so we don't accidentally reintroduce the bug?

Issue #921, opened 2022-10-12
Calling `contains` with large coordinates causes panic due to overflow


Snoeprol:
**updated description by @michaelkirk:**

Calling `contains` with large coordinates causes panic due to overflow.

```
let world: Polygon<i32> = polygon![
    (x: 43348848, y: 114780793),
    (x: 93348848, y: 214865793),
    (x: 13518848, y: 714848793),
    (x: 23518848, y: 814780793),
    (x: 63348848, y: 914780793),
];

let point: Point<i32> = Point::new(53348922, 514780838);
```

crash output:
```
attempt to multiply with overflow
thread 'algorithm::contains::polygon::tests::gh_issue_921' panicked at 'attempt to multiply with overflow', /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library/core/src/ops/arith.rs:369:1
stack backtrace:
   0: rust_begin_unwind
             at /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library/core/src/panicking.rs:142:14
   2: core::panicking::panic
             at /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library/core/src/panicking.rs:48:5
   3: <i32 as core::ops::arith::Mul>::mul
             at /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library/core/src/ops/arith.rs:362:45
   4: geo::algorithm::kernels::Kernel::orient2d
             at ./src/algorithm/kernels/mod.rs:31:19
   5: geo::algorithm::intersects::line::<impl geo::algorithm::intersects::Intersects<geo_types::geometry::coordinate::Coordinate<T>> for geo_types::geometry::line::Line<T>>::intersects
             at ./src/algorithm/intersects/line.rs:11:9
   6: geo::algorithm::coordinate_position::coord_pos_relative_to_ring
             at ./src/algorithm/coordinate_position.rs:376:12
   7: <geo_types::geometry::polygon::Polygon<T> as geo::algorithm::coordinate_position::CoordinatePosition>::calculate_coordinate_position
             at ./src/algorithm/coordinate_position.rs:245:15
   8: geo::algorithm::coordinate_position::CoordinatePosition::coordinate_position
[...]

PR #922, opened 2022-10-12
Turn bug report from #921 into an executable test

michaelkirk:
#921 reported a crash when running something like the following code.

Strangely this isn't crashing for me (aarch64). Let's see what CI says though.


Snoeprol:
It has to do with the area not being a square, I will get back to it later.

michaelkirk:
Ah, ok - thanks for clarifying the behavior. I've updated your original issue with a more thorough description.

PR #924, opened 2022-10-21
Introduce `Coord`, deprecate `Coordinate`

frewsxcv:
---

When we're ready :)

Fixes https://github.com/georust/geo/issues/540

### DIFF
--- a/geo-types/src/geometry/coordinate.rs
+++ b/geo-types/src/geometry/coordinate.rs
@@ -8,7 +8,7 @@ use approx::{AbsDiffEq, RelativeEq, UlpsEq};
 ///
 /// Unlike `Point` (which in the future may contain additional information such
 /// as an envelope, a precision model, and spatial reference system
-/// information), a `Coordinate` only contains ordinate values and accessor
+/// information), a `Coord` only contains ordinate values and accessor
 /// methods.
 ///
 /// This type implements the [vector space] operations:
--- a/geo-types/src/geometry/coordinate.rs
+++ b/geo-types/src/geometry/coordinate.rs
@@ -25,10 +25,13 @@ use approx::{AbsDiffEq, RelativeEq, UlpsEq};
 /// [vector space]: //en.wikipedia.org/wiki/Vector_space
 #[derive(Eq, PartialEq, Clone, Copy, Debug, Hash, Default)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub struct Coordinate<T: CoordNum = f64> {
+pub struct Coord<T: CoordNum = f64> {
     pub x: T,
     pub y: T,
 }
+
+#[deprecated(note = "Renamed to `geo_types::Coord` (or `geo::Coord`)")]
+pub type Coordinate<T: CoordNum = f64> = Coord<T>;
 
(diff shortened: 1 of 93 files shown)


michaelkirk:
As written, doesn't this cause a bajillion deprecation warnings within our own crate?

michaelkirk:
Ah, and this too:
```
error: bounds on generic parameters are not enforced in type aliases
[1360](https://github.com/georust/geo/actions/runs/3297886652/jobs/5439279686#step:6:1361)
  --> geo-types/src/geometry/coordinate.rs:34:24
[1361](https://github.com/georust/geo/actions/runs/3297886652/jobs/5439279686#step:6:1362)
   |
[1362](https://github.com/georust/geo/actions/runs/3297886652/jobs/5439279686#step:6:1363)
34 | pub type Coordinate<T: CoordNum = f64> = Coord<T>;
```

> When we're ready :)

FWIW - I'm ready!

frewsxcv:
> As written, doesn't this cause a bajillion deprecation warnings within our own crate?

Yep! I was figuring we can `#![allow(deprecation_warnings)]` temporarily and incrementally work through them instead of making a big PR that will cause a bunch of merge of conflicts

michaelkirk:
It's true that one big PR is at risk of merge conflicts.

But, I wouldn't want to release anything with deprecation warnings, and I don't like having main in a state that's unreleasable, 
 so I'm personally on team "one big PR", but that's because I'm assuming this is a pretty quick search-and-replace job. Is there a reason it's more complicated than that?

frewsxcv:
@michaelkirk This is ready for review!

geo-types/src/geometry/coordinate.rs
```diff
@@ -25,12 +25,15 @@ use approx::{AbsDiffEq, RelativeEq, UlpsEq};
...
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub struct Coordinate<T: CoordNum = f64> {
+pub struct Coord<T: CoordNum = f64> {
     pub x: T,
     pub y: T,
 }
 
-impl<T: CoordNum> From<(T, T)> for Coordinate<T> {
+#[deprecated(note = "Renamed to `geo_types::Coord` (or `geo::Coord`)")]
+pub type Coordinate<T> = Coord<T>;
```
michaelkirk, line 34:
```suggestion
pub type Coordinate<T = f64> = Coord<T>;
```

Without this suggested change, when building a client application (abstreet), I hit this error:

```
error[E0107]: missing generics for type alias `geo_types::Coordinate`
  --> /Users/mkirk/.cargo/registry/src/github.com-1ecc6299db9ec823/contour-0.7.0/src/contour.rs:8:26
   |
8  | pub type Pt = geo_types::Coordinate;
   |                          ^^^^^^^^^^ expected 1 generic argument
   |
```


I (similarly to you I think), assumed the default trait implementation would translate to the type alias, but it looks not to be the case.

michaelkirk (approved):
One request re: the default trait impl, but the other one is "take it or leave it"./

Other than that, LGTM.

frewsxcv:
bors r=michaelkirk

michaelkirk:
🎉

PR #931, opened 2022-11-09
Remove repeated points

mthh:
---
Implement a `RemoveRepeatedPoint` trait that removes repeated points from a `MultiPoint` and repeated consecutive coordinates from `LineString`, `Polygon`, `MultiLineString` and `MultiPolygon`.

I chose to make an implementation that simply returns a clone for the geometry types that were not candidates for this removal (`Point`, `Line`, `Rect` and `Triangle`), so that it could then be implemented for `Geometry`.

When I wanted to implement it for `Geometry` I tried to use `geometry_delegate_impl` but it was not possible because `impl<T: CoordNum> From<GeometryCollection<T>> for Geometry<T>` is not implemented yet, hence the manual implementation.

About the naming, I thought "remove repeated coords" or "remove duplicate" could have been considered too, but I chose the same naming as in PostGIS.

### DIFF
--- a/geo/src/algorithm/remove_repeated_points.rs
+++ b/geo/src/algorithm/remove_repeated_points.rs
@@ -18,7 +18,7 @@ where
 {
     /// Create a new geometry with (consecutive) repeated points removed.
     fn remove_repeated_points(&self) -> Self;
-    /// Returns the geometry with (consecutive) repeated points removed inplace.
+    /// Remove (consecutive) repeated points inplace.
     fn remove_repeated_points_mut(&mut self);
 }
 
--- a/geo/src/algorithm/remove_repeated_points.rs
+++ b/geo/src/algorithm/remove_repeated_points.rs
@@ -7,12 +7,16 @@ use num_traits::FromPrimitive;
 
 /// Remove repeated points from a `MultiPoint` and repeated consecutive coordinates
 /// from `LineString`, `Polygon`, `MultiLineString` and `MultiPolygon`.
+///
+/// For `GeometryCollection` it individually removes the repeated points
+/// of each geometry in the collection.
+///
+/// For `Point`, `Line`, `Rect` and `Triangle` it returns a clone of the geometry.
 pub trait RemoveRepeatedPoints<T>
 where
     T: CoordNum + FromPrimitive,
 {
-    /// Create a new geometry with (consecutive) repeated points removed
-    /// (only applies to `LineString`, `Polygon`, `MultiLineString` and `MultiPolygon`).
+    /// Create a new geometry with (consecutive) repeated points removed.
     fn remove_repeated_points(&self) -> Self;
 }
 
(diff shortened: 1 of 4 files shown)


geo/src/algorithm/remove_repeated_points.rs
```diff
@@ -0,0 +1,395 @@
...
+}
+
+impl<T> RemoveRepeatedPoints<T> for LineString<T>
+where
+    T: CoordNum + FromPrimitive,
+{
+    /// Create a LineString with consecutive repeated points removed.
+    fn remove_repeated_points(&self) -> Self {
+        let mut coords = vec![];
+        for c in self.0.iter() {
```
michaelkirk, line 46: This is the same as `LineString(self.0.dedup().collect())` right?  If so, can you use that instead?

michaelkirk (changes requested):

geo/src/algorithm/remove_repeated_points.rs
mthh: Oh for sure, I had completely missed the `dedup` method of Vec !

mthh: Done.

But because `dedup` acts in-place I couldn't implement it in one line as proposed and I had to clone the coordinate vector first.

However, this raises the question of whether `RemoveRepeatedPoints` should create a new geometry (which is proposed here) or act inplace (which would not be a problem from a conceptual point of view in my opinion, and which should not be much more complicated to implement).

michaelkirk: >  because dedup acts in-place I couldn't implement it in one line as proposed and I had to clone the coordinate vector first.

Oh sorry, my mistake.

> whether RemoveRepeatedPoints should create a new geometry 

I think ideally we would have both! This is a common pattern in geo, though we're a bit inconsistent in how naming goes. I'd suggest:

```
remove_repeated_points
remove_repeated_points_mut
```

mthh:
I added a `remove_repeated_points_mut` method as you suggested.

geo/src/algorithm/remove_repeated_points.rs
```diff
@@ -0,0 +1,502 @@
...
+        mpoly.remove_repeated_points_mut();
+        let expected = MultiPolygon(vec![make_result_poly1(), make_result_poly2()]);
+
+        assert_eq!(mpoly, expected);
+    }
+
+    #[test]
+    fn test_remove_repeated_points_mut_geometrycollection() {
+        let mut gc = GeometryCollection::new_from(vec![
+            make_result_mp1().into(),
```
michaelkirk, line 488:
```suggestion
            make_test_mp1().into(),
```

geo/src/algorithm/remove_repeated_points.rs
```diff
@@ -0,0 +1,502 @@
...
+
+        let expected = MultiPolygon(vec![make_result_poly1(), make_result_poly2()]);
+
+        assert_eq!(mpoly.remove_repeated_points(), expected);
+    }
+
+    #[test]
+    fn test_remove_repeated_points_geometrycollection() {
+        let gc = GeometryCollection::new_from(vec![
+            make_result_mp1().into(),
```
michaelkirk, line 417:
```suggestion
            make_test_mp1().into(),
```

michaelkirk (approved):
Two little bugs in the tests, but otherwise LGTM! 

I'll wait a couple days before merging in case anyone else wants to weight in.

mthh:
> Two little bugs in the tests, but otherwise LGTM!

I fixed the 2 tests. Thanks for the review! 

> I'll wait a couple days before merging in case anyone else wants to weight in.

Sure, there's no rush to merge it.

Issue #883, opened 2022-07-28
Support (Convex) Minkowski Sum construction


rmanoka:
Minkowski sum is a sort of generalization of buffering.  The buffering of a geometry G upto radius r is same as minkowski sum G + B, where B is a circular area centered at the origin, and has radius r.   This sum can be calculated via boolean ops when B is a convex polygon.

This should address #641 to some extent:  taking minkowski sum of G with appropriate approximation of a circular region is decent approximation to buffering.  I don't know if the output is aesthetically nice (because of the approximation, there may be some sharp edges), but it satisfies the functional requirement of buffering:  any point within distance r from the original geometry is contained in the output.

Issue #641, opened 2021-04-02
Geometry buffering


urschrei:
# Overview
The ability to buffer geometries is fundamental to a high-quality geometry library. Thus far, we lack that ability. This is a kind of tracking issue for the implementation of this functionality in geo. [This](https://geogra.uah.es/patxi/gisweb/BuffersModule/Buff_poly.htm) is the most simple explanation I could find.

The more technical term for this functionality is inward or outward **polygon offsetting**, though for the purposes of this issue we include 1D geometries such as `Point` and `LineString` as entities that we wish to buffer.

# Background
[This page](http://fcacciola.50webs.com/Offseting%20Methods.htm) provides a good overview of the problem, and some strategies for implementing the functionality, as well as references in the literature. In particular, the ability to quickly and robustly compute a **straight skeleton** is the fundamental first step to producing offset curves, and many straight-skeleton algorithms require a fast, robust sweep-line algorithm. [This paper](https://mcmains.me.berkeley.edu/pubs/DAC05OffsetPolygon.pdf) by Chen and McMains (2005) provides a more in-depth overview of existing methodologies before presenting a new (in 2005) approach to computing raw offset curves.

This more [recent paper by Palfrader and Held](https://www.tandfonline.com/doi/pdf/10.1080/16864360.2014.997637) (2015) presents a Voronoi-based straight skeleton approach to computing offsets.

The even [more recent post](https://www.sthu.org/blog/14-skeleton-offset-topology/index.html) provides an up-to-date overview of the relationship between the straight skeleton and the offset curve.

# Implementations
## CGAL
As ever, [CGAL](https://doc.cgal.org/Manual/3.2/doc_html/cgal_manual/Straight_skeleton_2/Chapter_main.html) is a useful source of data on how one might approach the [design of the API](https://doc.cgal.org/Manual/3.2/doc_html/cgal_manual/Straight_skeleton_2/Chapter_main.html#Section_16.3). [Here's the overview](http://www.dma.fi.upm.es/personal/mabellanas/tfcs/skeleton/html/documentacion/A%20CGAL%20implementation%20of%20the%20Straight%20Skeleton%20of%20a%20Simple%202D%20Polygon%20with%20Holes.pdf) of the CGAL approach.

[...]

PR #934, opened 2022-11-12
Propose simple offset algorithim

thehappycheese:
---

Hi there,

May I please have some feedback on this pull request. The work is not complete, but since I am not very experienced at contributing to rust projects I figured it might be better to get early feedback.

Is the project interested in adding this kind of simple offset algorithm? 
 - Issue #641 suggested a more advanced algorithim for buffering
 - Issue #883 suggests a Minkowski Sum

The algorithm I am trying to implement here (described in `rfcs/2022-11-11-offset.md`) is much simpler and more prone to creating self-overlapping output than the above two proposals. But for my use cases it is good enough and perhaps others would find is useful also.

Have I started things off in the right direction?

I am still getting to grips with the num_traits crate which I am finding very difficult to navigate. I am not always clear on how to properly use the CoordFloat / CoordNum types etc. I think I also need to change some bits of my code to use the Kernel / RobustKernel where possible.

Many thanks for your time


thehappycheese:
ok so first thing I learned was to remember to run rust format before pull requests. oops

PR #935, opened 2022-11-18
Implement Offset Algorithm

thehappycheese:
---

Hi there,

This is a draft PR to implement an `Offset` trait

The work is not complete, but since I am not very experienced at contributing to rust projects I figured it might be better to make a draft PR while I am still working on it and get feedback as early as possible.

## Choice of Algorithm

 - Issue #641 suggests a more advanced algorithm (Voronoi-based straight skeleton approach to computing offsets)
 - Issue #883 suggests a Minkowski Sum

I am not sure if the algorithm I am trying to implement here is any better or worst than those, but it is one I have got working in the past in python so I figure I have the best chance of success with it. It loosely follows the algorithm described by
[Xu-Zheng Liu, Jun-Hai Yong, Guo-Qin Zheng, Jia-Guang Sun. An offset algorithm for polyline curves. Computers in Industry, Elsevier, 2007, 15p. inria-00518005](https://hal.inria.fr/inria-00518005/document)

## Trait Implementations

- [X] `Line<impl CoordFloat>`
- [X] `LineString<impl CoordFloat>`
- [X] `MultiLineString<impl CoordFloat>`
- [ ] `Triangle<impl CoordFloat>`
- [ ] `Rectangle<impl CoordFloat>`
- [ ] `Polygon`
- [ ] `MultiPolygon`
- [ ] `Geometry` & `GeometryCollection
  
## Issues to Fix:

- [ ] Currently does not make proper use of the Kernels
- [ ] Not sure if I can make it work for CoordNum, everything seems to work with CoordFloat
- [ ] No check is implemented to prevent execution if the specified offset  distance is zero
- [ ] No Mitre-limit is implemented; A LineString which doubles back on itself will produce an elbow at infinity
- [ ] I have not implemented the full clipping algorithm yet; Non-adjacent  line segments in the output may be self-intersecting.
- [ ] Does not handle closed shapes yet

Hopefully I have started off my contribution in the right direction? :)

Many thanks for your time


geo/src/algorithm/offset/line_intersection.rs
```diff
@@ -0,0 +1,321 @@
...
+    T: CoordFloat,
+{
+    let ab = *b - *a;
+    let cd = *d - *c;
+    let ac = *c - *a;
+
+    // TODO: I'm still confused about how to use Kernel / RobustKernel;
+    //       the following did not work. I need to read more code
+    //       from the rest of this repo to understand.
+    // if Kernel::orient2d(*a, *b, *d) == Orientation::Collinear {
```
urschrei, line 153: `Orientation` is an enum, so you have to use a [`match`](https://doc.rust-lang.org/book/ch06-02-match.html) for control flow.

something like

    match Kernel::orient2d(*a, *b, *c) {
        Orientation::Collinear => { // keep going }
        _ => { // not collinear, bail out or test different combination?? }
    }

As you note, the use of arbitrary "magic" threshold values without considerable justification and accompanying comments explaining the reasoning behind it for future maintainers is unlikely to be accepted (I've just finished removing an algorithm containing a lot of these that I had the misfortune of writing myself…), so if we can use existing mechanisms that would be much better.

geo/src/algorithm/offset/cross_product.rs
```diff
@@ -0,0 +1,85 @@
+use crate::CoordFloat;
```
urschrei, line 1: I'm a bit confused about this. Would the [dot product sign](https://docs.rs/geo/latest/geo/algorithm/kernels/trait.Kernel.html#method.dot_product_sign), i.e. the [Orientation](https://docs.rs/geo/latest/geo/algorithm/kernels/enum.Orientation.html) suffice? It looks like you're just checking whether the result is `-1`, `1`, or `0`…

geo/CHANGES.md
```diff
@@ -11,6 +11,7 @@
   * <https://github.com/georust/geo/pull/928>
 * Added `RemoveRepeatedPoints` trait allowing the removal of (consecutive)
   repeated points.
+* Added very basic `Offset` trait and implemented it for `Line` and `LineString`
```
urschrei, line 14: A minor nit: there's no need to qualify this as "very basic", since the docs will make the limitations clear.

geo/src/algorithm/offset/offset_trait.rs
```diff
@@ -0,0 +1,336 @@
...
+    LineString,
+    MultiLineString,
+    // Polygon,
+};
+
+/// # Offset Trait
+///
+/// Signed offset of Geometry assuming cartesian coordinate system.
+///
+/// This is a cheap offset algorithm that is suitable for flat coordinate systems
```
urschrei, line 28: There's no need to specify this as `geo` operates only on coordinates in the real plane (with the exception of some explicitly documented convenience algorithms for spherical coordinates). There is an explicit assumption that only projected / cartesian coordinates are used. That this algorithm may be suitable for coordinates near the equator is a coincidence as far as the library is concerned.

geo/src/algorithm/offset/offset_trait.rs
```diff
@@ -0,0 +1,336 @@
...
+/// - [X] Line<impl CoordFloat>
+/// - [X] LineString<impl CoordFloat>
+/// - [X] MultiLineString<impl CoordFloat>
+/// - ... maybe some closed shapes like triangle, polygon?
+///
+/// The following are a list of known limitations,
+/// some may be removed during development,
+/// others are very hard to fix.
+///
+/// - No checking for zero length input.
```
urschrei, line 41: Is there a specific difficulty you've encountered while trying to implement a check against this?

geo/src/algorithm/offset/offset_trait.rs
```diff
@@ -0,0 +1,336 @@
...
+///   Invalid results may be caused by division by zero.
```
urschrei, line 42: Where could this occur?

geo/src/algorithm/offset/offset_trait.rs
```diff
@@ -0,0 +1,336 @@
...
+/// - No check is implemented to prevent execution if the specified offset
```
urschrei, line 43: We will need a check for this

geo/src/algorithm/offset/offset_trait.rs
```diff
@@ -0,0 +1,336 @@
...
+///   distance is zero.
+/// - Only local cropping where the output is self-intersecting.
+///   Non-adjacent line segments in the output may be self-intersecting.
+/// - There is no mitre-limit; A LineString which
```
urschrei, line 47: This is a fairly significant limitation, but if the algorithm can't handle it, then we should consider a check for self-crossing LineStrings (I believe we already have one) and bail out early if this is encountered.

geo/src/algorithm/offset/offset_trait.rs
```diff
@@ -0,0 +1,336 @@
...
+/// ## Algorithm
+/// Loosely follows the algorithm described by
+/// [Xu-Zheng Liu, Jun-Hai Yong, Guo-Qin Zheng, Jia-Guang Sun. An offset algorithm for polyline curves. Computers in Industry, Elsevier, 2007, 15p. inria-00518005]
+/// (https://hal.inria.fr/inria-00518005/document)
+/// This was the first google result for 'line offset algorithm'
+impl<T> Offset<T> for LineString<T>
+where
+    T: CoordFloat,
+{
+    fn offset(&self, distance: T) -> Self {
```
urschrei, line 105: My expectation is that a positive (or outward) offset on a LineString produces a `Polygon` (as per JTS, GEOS, and derived libraries). Could you explain the reasoning here?

urschrei (changes requested):
Hi @thehappycheese, welcome and thanks for the contribution! I've left a load of review comments – please don't feel that you need to address them all at once. In particular, it looks like you're having some trouble using the `Kernel` trait, and this might be a good place to start. I and other contributors are happy to help here, and we can do it inline in the PR, and / or on Discord: https://discord.gg/Fp2aape

geo/src/algorithm/offset/offset_trait.rs
dabreegster: I have no context on how JTS/GEOS/etc does things, but offsetting a LineString and getting a LineString has many use cases. If you start with something representing the center of a road and want the left or right edge, or the center of a lane, for example. To buffer around a LineString and get a polygon, you could offset inwards (the hole) and outwards (the exterior), right?

urschrei: @dabreegster I'm struggling to visualise that. Do you have an example? The standard implementations of offset (in plain language, **buffer**) operations always produce a polygon from a Line or LineString _as far as I know_. The Shapely docs are representative here: https://shapely.readthedocs.io/en/stable/manual.html#object.buffer

michaelkirk: Disclaimer that I haven't looked at this implementation, but just a note on terminology...

JTS has an "offset curve" which is a line string to line string transformation. 

> a line lying on one side of another line at a given offset distance

http://lin-ear-th-inking.blogspot.com/2022/01/jts-offset-curves.html?m=1

thehappycheese: Hi @urschrei thank you so much for all your feedback I really appreciate it, I will do my best to address each one :)

On this @dabreegster has already explained my reasoning really well; To me a buffer operation different to an offset; I am interested in offsetting road center lines, not producing enclosed polygons.

screenshot

I recognize that a buffer operation is more generally useful, but I also just need this signed offset function. This offset trait can be extended by a a few clipping and stitching steps to a full buffer operation, this is the end goal of the paper I am following.

dabreegster: > I'm struggling to visualise that. Do you have an example?

https://a-b-street.github.io/docs/tech/map/geometry/index.html#projecting-a-polyline, and for the inspiration behind it, https://wwwtyro.net/2019/11/18/instanced-lines.html. The JTS offset curve looks like the same thing. I don't know what this linestring -> linestring operation should be called, but I have plenty of use cases for it, as well as buffering (to get a polygon)

urschrei: Thanks all, it looks like it's just confusion about terminology, mostly on my part. It might be worth discussing whether the current trait name conveys the functionality clearly enough, though maybe it's just a question of ensuring that the module-level docstring does this.

frewsxcv: @thehappycheese By the way, your screenshot link doesn't appear to work: https://github.com/thehappycheese/powerbi-visual-geojson-map-1/raw/main/git_docs/hero_image.jpg (404)

geo/src/algorithm/offset/cross_product.rs
thehappycheese: Apologies, looks like my inept git-fu has disconnected all your comments from the code 😕 I think maybe re-basing the branch on my fork was the wrong thing to do.

`cross_product` simplifies `line_segment_intersection_with_parameters()`: While it is used for a collinear check at the start of that function (which could be replaced with `Kernel::dot_product_sign` or `Kernel::orient2d`) it is used twice more further along to calculate the parameters `t_ab` and `t_cd` so I do not believe it can be substituted for an orientation check.

```rust
let t_ab = cross_product_2d(ac, cd) / ab_cross_cd;
let t_cd = -cross_product_2d(ab, ac) / ab_cross_cd;
```

`line_segment_intersection_with_parameters()` is currently only used via a wrapper function `line_segment_intersection_with_relationships()`; in that function it checks each `t_ab` and `t_cd` to find which of three ranges they lie in:

- less than zero
- between zero and one
- greater than one

It's possible there is some tricky way to express this check in terms of `orient2d()`, I'll have a think about it.

It does feel weird to have `cross_product()` out in it's own module. In the past I have worked with libraries that have a `Vector2` type where this kind of function would be implemented. It is simple enough to be inline perhaps, but I wrote up all the tests and documentation because in the past I have made many mistakes with putting negatives in the wrong place and if you make this kind of mistake an even number of times you might not realize since it cancels out 😋

geo/CHANGES.md
thehappycheese: 👍 I will try to reword it to be more specific, or be more clear about any limitations in the rest of the docs

geo/src/algorithm/offset/offset_trait.rs
thehappycheese: Have added this check to LineString but it probably isn't worth it for Line. I have added comments to the relevant code in my next commit.

geo/src/algorithm/offset/offset_trait.rs
thehappycheese: Keen to align terminology with suggestions / other libraries :)

`offset_curve` is nice because it makes it more obvious why this type of signed offset cannot be implemented for a `Point` object. `Point` objects can only have a "buffer" or "outset" applied, not an 'offset_curve'.

Ah apologies for the broken link I picked it from one of my public repos but didn't realize the image link is to a private repo; this one should work?

Screenshot of road data custom PowerBI visualization

This is from a custom PowerBI visual I created. I will eventually make that code public when I get a chance to clean up /republish repo. I want the PowerBI visual to dynamically connect to a rust server for a specialized kind of chainage-based  geo-coding.

thehappycheese:
I have deployed an interactive tester thing here <http://52.65.41.2/>
(I haven't bought a domain name for it so that link looks very sketchy.)
https://github.com/thehappycheese/geo-offset-test-server

Screenshot of interactive tester

Hoping it will help me find edge cases as I start to implement the miter limit an clipping step of the algorithm. To be honest though after playing with it for a bit, I am a little worried that some edge cases are going to be pretty challenging.

michaelkirk:
As an aside, Wow, that's a really nice demo!  

We've talked in the past about having interactive tools like this as a kind of documentation for some of our algorithms, but I think think this is the first actual working example. 👏 👏👏

michaelkirk:
Hi @thehappycheese - this is an exciting new feature that I think a lot of people would like to use. Are you still interested in working on it? Is there anything I can do to help move it along?

thehappycheese:
> Hi @thehappycheese - this is an exciting new feature that I think a lot of people would like to use. Are you still interested in working on it? Is there anything I can do to help move it along?

Hi @michaelkirk 😃

I have to admit I stalled out on this project; There are a few contributing factors

1. While the basic offset algorithm is done, the clipping of the output part is not started.
   - Clipping the output geometry is pretty tricky, I'm feeling a bit apprehensive about tackling it
   - And also, my main motivating use-case for this feature does not actually require this additional clipping step
   - Perhaps this could just be an `offset_curve_raw` and in the future there could be an `offset_curve_clipped`?
3. I have introduced several new data structures and traits which overlap/conflict with existing code elsewhere in the project... On the one hand it makes sense to push forward and get the `offset_curve` function working then resolve these later. On the other hand, perhaps I should be trying to build up the required infrastructure with smaller separate pull requests first.
    - eg [struct LineMeasured](https://github.com/thehappycheese/geo/blob/356dd400e43d475c547cf384424fdc444826e5fe/geo/src/algorithm/offset_curve/line_measured.rs#L27)
    - eg [fn line_segment_intersection_with_parameters](https://github.com/thehappycheese/geo/blob/356dd400e43d475c547cf384424fdc444826e5fe/geo/src/algorithm/offset_curve/line_intersection.rs#L144)
    - eg [trait VectorExtensions](https://github.com/thehappycheese/geo/blob/356dd400e43d475c547cf384424fdc444826e5fe/geo/src/algorithm/offset_curve/vector_extensions.rs#L18)
      - This one is particularly problematic... I think I should be trying to get these functions into `Kernel` or something 
      - But I rather like the ergonomics of `impl VectorExtensions for Coord` but it is a really different approach
4. There are now two implementations of [`trait OffsetCurveOld`](https://github.com/thehappycheese/geo/blob/356dd400e43d475c547cf384424fdc444826e5fe/geo/src/algorithm/offset_curve/offset_curve_trait_old.rs#L91) (old) and [`trait OffsetCurve`](https://github.com/thehappycheese/geo/blob/356dd400e43d475c547cf384424fdc444826e5fe/geo/src/algorithm/offset_curve/offset_curve_trait.rs#L33) (new)
    - The new one uses [`trait LineStringOffsetSegmentPairs`](https://github.com/thehappycheese/geo/blob/356dd400e43d475c547cf384424fdc444826e5fe/geo/src/algorithm/offset_curve/offset_segments_iterator.rs#L32) which was intended to make things tidy inside the `trait OffsetCurve` which I think it does... but it came at the cost of much headaches for me working out how to make iterators work.
    - Anyway I succeeded and now I think I will delete `trait OffsetCurveOld`
    - BUT the iterator `trait LineStringOffsetSegmentPairs` needs one more feature;
      - support wrapping over to the first segment at the end to work on closed shapes
      - I stopped before implementing that and took a break over Christmas and have not made it back here yet in the new year :/

I am still keen to get back into this but it might take me a few more weeks to get going again. In the meantime, if anyone is keen to jump in I would be happy to collaborate in any way possible :)

thehappycheese:
Trying to figure out if any of my code can be broken off into more bite-sized contributions.
I'll probably edit this comment later or move it into 

## Required for Simple Offset

- I use `impl VectorExtensions for Coord` which I like, but `VectorExtensions` doesn't gel with the rest of the crate.
  To move forward I need to find replacements for the following functions

  - [ ] Explicit 90 degree rotation around origin for `Coord`
    - My current implementation `::<VectorExtensions>::left()` and `::<VectorExtensions>::right()`
    - could add associated functions for `Rotate` trait `::left()` and `::right()`
    - could be added as special constructors for `struct AffineTransform` `::left()` and `::right()`
    - abandon explicit / specialized functions and just use `::<Rotate>::rotate(std::f64::consts::PI/2.0)`

  - [ ] 2D Wedge Product (aka exterior product, 2d cross product) (See [docs](https://github.com/thehappycheese/geo/blob/d95420f1f8d6b0d2dd1e6d803fe3d3e2a2b3dd13/geo/src/algorithm/line_intersection.rs#L73))
    - Simplifies line intersection calculations
    - Currently implemented directly on `geotypes::Point` but its is not right it takes three arguments which is not what I need
    - Related to `Kernel::orient2d` but serves a different purpose
 
  - [ ] Dot Product
    - Currently implemented directly on `geotypes::Point`
    - Feels like it should live together with some of these other ops... should they all be in the `impl Point` or should it be in a trait?

- [ ] Specialized `Line` segment intersection
  (capable of reporting virtual intersections + which side.
   See [offset_curve/line_intersection.rs](https://github.com/thehappycheese/geo/blob/d95420f1f8d6b0d2dd1e6d803fe3d3e2a2b3dd13/geo/src/algorithm/offset_curve/line_intersection.rs#LL1C1-L75C2) )
  - Continue with specialized function [algorithim/offset_curve/line_segment_intersection_with_relationships](https://github.com/thehappycheese/geo/blob/d95420f1f8d6b0d2dd1e6d803fe3d3e2a2b3dd13/geo/src/algorithm/offset_curve/line_intersection.rs#LL210C1-L240C2)
  - augment existing [algorithm/line_intersection.rs](https://github.com/thehappycheese/geo/blob/d95420f1f8d6b0d2dd1e6d803fe3d3e2a2b3dd13/geo/src/algorithm/line_intersection.rs#L73) or add alternative implementation to that module

## Required for Clipped Offset

- [ ] Availability of #986 (LineString Slicing)
- [ ] Need a '`CookieCut`' operation; separate `LineString`, `Polygon` etc by intersection with a circle

frewsxcv:
@thehappycheese Thank you for all the effort you put into this pull request! We landed geometry buffering in geo in the last release, via https://github.com/georust/geo/pull/1365. We really appreciate you leading the way here!

thehappycheese:
> @thehappycheese Thank you for all the effort you put into this pull request! We landed geometry buffering in geo in the last release, via https://github.com/georust/geo/pull/1365. We really appreciate you leading the way here!

Thats super exciting thanks so much to everyone that got it there and much appreciated that you remembered my small initial efforts <3

PR #426, opened 2020-03-15
No std

brenzi:
fixes #422 

builds with rust stable.
must patch rstar dependency until upstream PR gets merged: https://github.com/Stoeoef/rstar/pull/27

PR #936, opened 2022-11-19
`no_std` support for `geo-types`

### LINKED ISSUE #422: support no_std
I'd like to use geo in a `no_std` build. Any chance this will be supported soon? If not on you roadmap, would you accept a PR?

wetheredge:
---

This is my take on `no_std` support. This adds a new `std` that behave similarly to `serde`'s.

Rough overview of my changes:
- For the most part, simply swapping `alloc` or `core` in for `std`.
- Also added a second `doc(hidden)` public module (`_alloc`) to give macros access to `Vec` and `Box` even if the crate they are used from does not have `extern crate alloc`. This could be merged into the existing `private_utils`. It could also be dropped entirely, but that would require modifying the macros to reference `std` when that feature is enabled and the cleanest way I found to do that is still somewhat ugly: match on `()` with `cfg`'d branches.

I've looked some at #426 and the main difference I see is that it adds `core-error` for the `Error` trait, while I've just put that behind the `std` feature. If that is preferred, I can add the dependency.

---

I have also started working on `no_std` support in `geo`, but it's not quite as straight-forward, so I was keeping it for a second PR.

fixes #422, partially supersedes #426

### DIFF
--- a/geo-types/src/lib.rs
+++ b/geo-types/src/lib.rs
@@ -1,3 +1,4 @@
+#![cfg_attr(not(feature = "std"), no_std)]
 #![warn(missing_debug_implementations)]
 #![doc(html_logo_url = "https://raw.githubusercontent.com/georust/meta/master/logo/logo.png")]
 //! The `geo-types` library defines geometric types for the [GeoRust] ecosystem.
@@ -75,9 +76,11 @@
 //! [OGC-SFA]: https://www.ogc.org/standards/sfa
 //! [rstar]: https://github.com/Stoeoef/rstar
 //! [Serde]: https://serde.rs/
+extern crate alloc;
 extern crate num_traits;
+
+use core::fmt::Debug;
 use num_traits::{Float, Num, NumCast};
-use std::fmt::Debug;
 
 #[cfg(feature = "serde")]
 #[macro_use]
@@ -131,10 +134,22 @@ mod arbitrary;
 #[doc(hidden)]
 pub mod private_utils;
 
+#[doc(hidden)]
+pub mod _alloc {
+    //! Needed to access these types from `alloc` in macros when the std feature is
+    //! disabled and the calling context is missing `extern crate alloc`. These are
+    //! _not_ meant for public use.
+
+    pub use ::alloc::boxed::Box;
+    pub use ::alloc::vec;
+}
+
 #[cfg(test)]
 mod tests {
+    use alloc::vec;
+
     use super::*;
-    use std::convert::TryFrom;
+    use core::convert::TryFrom;
 
     #[test]
     fn type_test() {
--- a/geo-types/src/lib.rs
+++ b/geo-types/src/lib.rs
@@ -69,8 +69,8 @@
 //! - `use-rstar_0_9`: Allows geometry types to be inserted into [rstar] R*-trees (`rstar v0.9`)
 //!
 //! This library can be used in `#![no_std]` environments if the default `std` feature is disabled. At
-//! the moment, the `arbitrary` and `use-rstar_0_8` features cannot be used without `std`. This may
-//! change in a future release.
+//! the moment, the `arbitrary` and `use-rstar_0_8` features require `std`. This may change in a
+//! future release.
 //!
 //! [approx]: https://github.com/brendanzab/approx
 //! [arbitrary]: https://github.com/rust-fuzz/arbitrary
--- a/geo-types/src/lib.rs
+++ b/geo-types/src/lib.rs
@@ -61,11 +61,16 @@
 //!
 //! The following optional [Cargo features] are available:
 //!
+//! - `std`: Enables use of the full `std` library. Enabled by default.
 //! - `approx`: Allows geometry types to be checked for approximate equality with [approx]
 //! - `arbitrary`: Allows geometry types to be created from unstructured input with [arbitrary]
 //! - `serde`: Allows geometry types to be serialized and deserialized with [Serde]
 //! - `use-rstar_0_8`: Allows geometry types to be inserted into [rstar] R*-trees (`rstar v0.8`)
 //! - `use-rstar_0_9`: Allows geometry types to be inserted into [rstar] R*-trees (`rstar v0.9`)
+//!
+//! This library can be used in `#![no_std]` environments if the default `std` feature is disabled. At
+//! the moment, the `arbitrary` and `use-rstar_0_8` features cannot be used without `std`. This may
+//! change in a future release.
 //!
 //! [approx]: https://github.com/brendanzab/approx
 //! [arbitrary]: https://github.com/rust-fuzz/arbitrary
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -76,7 +76,7 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
-      - run: rustup target add thumbv6m-none-eabi
+      - run: rustup target add thumbv7em-none-eabihf
       - run: cargo check --all-targets --no-default-features
       - run: cargo check --lib --target thumbv7em-none-eabihf --no-default-features -F use-rstar_0_9,serde
       - run: cargo test --all-features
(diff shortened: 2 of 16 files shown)


michaelkirk:
Hello! Thanks for taking a pass at this. 

Firstly, I'm not sure if we should be trying to support `no_std` (that is, _maybe_ we should, I'm just not _sure_.)

The `std`->`core` stuff is easy enough and I don't personally think it's controversial. 

But I'm a little hesitant about the `alloc` feature. It seems like trying to support a no-alloc build is more disruptive and also to me it seems unlikely to be useful. Do you actually have this use case? Can I ask what you're using it for?

What do people think about going the route that rstar went, which is no_std but requiring alloc outright?

Also, I think to have any hope of not breaking this in the future we'd need to be sure the no_std build is added to CI.

frewsxcv:
Agreed we should start out by having `alloc` be required, since it seems to disrupt the code quite a bit, unless there's a _really_ compelling use-case. I think `no_std` for `geo_types` is reasonable, but `geo` might warrant a discussion since that would likely be difficult.

urschrei:
> Agreed we should start out by having `alloc` be required, since it seems to disrupt the code quite a bit, unless there's a _really_ compelling use-case. I think `no_std` for `geo_types` is reasonable, but `geo` might warrant a discussion since that would likely be difficult.

Agreed w/both.

wetheredge:
I added the `alloc` feature because since that was mentioned as a possibility in the last PR (https://github.com/georust/geo/pull/426#issuecomment-1021990969). I'm fine with requiring `alloc` for `no_std`. The crate I need this for can't go no-`alloc` anyway.

I'll go ahead and remove the `alloc` feature, unless you want to leave it in unused (and documented as such) for the possibility of adding support for no-`alloc` in a non-breaking release in the future.

> Also, I think to have any hope of not breaking this in the future we'd need to be sure the no_std build is added to CI.

I'll add a build on a bare metal ARM target that can't provide `std` to verify that it works, based on advice from [this article](https://blog.dbrgn.ch/2019/12/24/testing-for-no-std-compatibility/).

My other question is whether to add [`core-error`](https://lib.rs/crates/core-error) as a dependency so that `no_std` has an `Error` impl, since `core` doesn't yet provide it natively. The last PR did, but I don't know if it's worth another dependency.

> `geo` might warrant a discussion since that would likely be difficult.

Agreed, it's not as simple. I'll take another look at it, perhaps after this is complete, and open a PR for that discussion only if it's not too bad.

michaelkirk:
> My other question is whether to add [core-error](https://lib.rs/crates/core-error) as a dependency so that no_std has an Error impl, since core doesn't yet provide it natively. The last PR did, but I don't know if it's worth another dependency.

I've done very little with no_std. What is common across the no_std ecosystem? What's the cost? What's the benefit?

wetheredge:
> What's the cost?

- Longer compile time (likely minimal tbf)
- Another source for bugs
- Looks to be [somewhat dead](https://github.com/core-error/core-error/issues/15)

> What's the benefit?

- Removes the difference between the `std` api & the `no_std`+`alloc` api

I don't have all that much experience with `no_std` either, but based on [lib.rs](https://lib.rs/crates/core-error/rev)/[crates.io](https://crates.io/crates/core-error/reverse_dependencies) & [GitHub](https://github.com/core-error/core-error/network/dependents), it looks like the only major crate that uses it is `rhai`, and only as an optional dependency. Could add it as an optional dep, but that adds yet another feature…. I lean toward not including it at all.

frewsxcv:
@wetheredge Are you still working on this? I think the next step is to require alloc for no_std, instead of making it optional

wetheredge:
Looks like the current CI does run with `--no-default-features`, but that isn't 100% accurate for testing `no_std` compatibility: https://blog.dbrgn.ch/2019/12/24/testing-for-no-std-compatibility/. Should I add one of those workarounds? I'm using the build for ARM method in my `no_std` crate.

frewsxcv:
I think it makes sense to add a `--no-default-features` build to our CI, yeah

frewsxcv:
Oh disregard my last message, I misread yours. Let me think about this

michaelkirk:
https://github.com/hobofan/cargo-nono looks nice. It hasn't been updated very recently, but maybe that's a sign of maturity vs. abandonment.

frewsxcv:
My vote would be to try adding a new no-std target (e.g. `thumbv6m-none-eabi` from the blog post) since that simulates the closest thing to what our goal is with no-std. But if that's not feasible then fallback to cargo-nono.

wetheredge:
I've used `thumbv7em-none-eabihf` since that's what I've got installed and what's in [my CI](https://github.com/wetheredge/blackbox/blob/a1b6ec47ec8eee460731823f8d581b471d4fbf5b/.github/workflows/ci.yaml#L22), but I can change it to `thumbv6m-none-eabi` to match the blog post if wanted. Any of the [Tier 2 targets](https://doc.rust-lang.org/rustc/platform-support.html#tier-2) with an asterisk in the middle column should do.

Most of the features are `no_std` compatible, except for `rstar` v0.8 & `arbitrary`. Looks like there's [an open PR](https://github.com/rust-fuzz/arbitrary/pull/74) to add `no_std` support to `arbitrary`. Could wait for that, or just document it as incompatible with `no_std` until that's merged and updated. `no_std` in `rstar` v0.8 would likely be pretty simple, tho I haven't checked. `rstar` v0.9 does work.

frewsxcv:
Let's disable `rstar` and `arbitrary` support for now, and if it's something you think you'll need, let's open a follow-up GitHub issue tracking those. No need for this PR to get held up by those dependencies. How does that sound?

wetheredge:
Unless I've missed something, this should be ready.

frewsxcv (approved):
lgtm! thank you! will leave it open for another day or two to see if anyone else has feedback

wetheredge:
Re semver:

The [cargo docs](https://doc.rust-lang.org/cargo/reference/semver.html#cargo-feature-add) aren't entirely clear on adding a new default feature that controls existing functionality. I'd be inclined to say it's technically a major change, but since there weren't any default features before, *perhaps* you could get away with treating it as minor.

Following [the same guidelines](https://doc.rust-lang.org/cargo/reference/semver.html#minor-changing-dependency-features), disabling features on dependencies only needs to be a minor change since, as [RFC 1105 says](https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md#minor-change-altering-the-use-of-cargo-features), any crate relying on those features without enabling them themselves is a bug.

wetheredge:
Is this ready to merge?

frewsxcv:
Thanks for your patience!

bors r+

geo-types/src/lib.rs
```diff
@@ -60,12 +61,17 @@
...
 //!
+//! - `std`: Enables use of the full `std` library. Enabled by default.
 //! - `approx`: Allows geometry types to be checked for approximate equality with [approx]
 //! - `arbitrary`: Allows geometry types to be created from unstructured input with [arbitrary]
 //! - `serde`: Allows geometry types to be serialized and deserialized with [Serde]
 //! - `use-rstar_0_8`: Allows geometry types to be inserted into [rstar] R*-trees (`rstar v0.8`)
 //! - `use-rstar_0_9`: Allows geometry types to be inserted into [rstar] R*-trees (`rstar v0.9`)
 //!
+//! This library can be used in `#![no_std]` environments if the default `std` feature is disabled. At
+//! the moment, the `arbitrary` and `use-rstar_0_8` features cannot be used without `std`. This may
```
frewsxcv, line 72:
```suggestion
//! the moment, the `arbitrary` and `use-rstar_0_8` features require `std`. This may
```

frewsxcv:
It looks like `byteorder` can't be built on our new target? 🤔

geo-types/src/lib.rs
wetheredge: Applied in [5f0bd98](https://github.com/georust/geo/pull/936/commits/5f0bd98339452ec5f2e9210ce3e33060f11c9b66)

wetheredge:
Oops was installing thumbv6 but building with v7…. Looks like it's fixed.

michaelkirk:
Thanks for seeing this through @wetheredge!

PR #937, opened 2022-11-23
prepare for geo-0.23.1 non-breaking release

michaelkirk:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/main/CODE_OF_CONDUCT.md).
- [x] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

Anything else we should be waiting for before release?

Diff: https://github.com/georust/geo/compare/geo-0.23.0...release/geo-0.23.1

PR #941, opened 2022-11-29
Fix RDP recursion bug

### LINKED ISSUE #940: BUG: Infinite recursion in `compute_rdp` with easy fix
While trying to figure out why calling `LineString::simplify` caused a stack overflow with a short line, I figured out that in cases where `euclidean_distance` returns 0.0 for all intermediate coords (so `farthest_distance` is 0.0), `farthest_index` will be 0 even though it should never be 0, which leads to infinite recursion!
The fix is to change the condition `distance > farthest_distance` to `distance >= farthest_distance` here:
https://github.com/georust/geo/blob/3b0d5738f54bd8964f7d1f573bd63dc114587dc4/geo/src/algorithm/simplify.rs#L79

So that `farthest_index` will never be 0 after `fold` returns.

urschrei:
Fixes #940

---

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -13,6 +13,8 @@
   repeated points.
 * Remove polygon-polygon fast path due to ongoing lack of reliability
   * <https://github.com/georust/geo/pull/920>
+* Fix RDP recursion bug
+  * <https://github.com/georust/geo/pull/941>
 
 
 ## 0.23.0
--- a/geo/src/algorithm/simplify.rs
+++ b/geo/src/algorithm/simplify.rs
@@ -83,6 +83,7 @@ where
                 }
             },
         );
+    debug_assert_ne!(farthest_index, 0);
 
     if farthest_distance > *epsilon {
         // The farthest index was larger than epsilon, so we will recursively simplify subsegments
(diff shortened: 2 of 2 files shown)


michaelkirk:
LGTM - feel free to merge with or without the unit test.

I'll plan to include it in #937.

frewsxcv:
Thoughts on adding a `debug_assert_ne!(farthest_distance, 0)` after line 85? To let the reader know we never expect it to be 0, and to prevent stack overflows from happening, and to indicate users to report a bug if this is encountered.

urschrei:
My debug assertion is now causing test failures. Sigh.

frewsxcv:
> My debug assertion is now causing test failures. Sigh.

I was wondering if this would be the case 😅 Did you have time to look into why it's 0?

urschrei:
Not yet. Tomorrow!

michaelkirk:
>  Did you have time to look into why it's 0?

I took a stab at this in #950

Seems like the issue was just the degenerate case when we recurse to simplify a two coordinate line segment.

frewsxcv:
Fixed in https://github.com/georust/geo/pull/950 Thank you @urschrei and @michaelkirk !

PR #943, opened 2022-11-30
`Simplify` now generates `Polygon` rings with at least four coordiantes

### LINKED ISSUE #142: Using Simplification algorithm on a polygon can result in an invalid polygon
https://github.com/georust/rust-geo/pull/135#issuecomment-320788702

Maybe the algorithm should sometimes return an `Option<T>` since it can sometimes fail.

frewsxcv:
---

Prior to this pull request, running `Simplify` on `Polygon` could generate rings with two coordinates, which would make it no longer a valid ring. This pull request updates `Simplify` to ensure resulting `Polygon` rings always have at least four coordinates.

Fixes #142

### DIFF
--- a/geo/src/algorithm/simplify.rs
+++ b/geo/src/algorithm/simplify.rs
@@ -299,7 +299,7 @@ mod test {
             coord! { x: 12.0, y: 100.0 },
         ];
         let compare = vec![coord! {x: 8.0, y: 100.0}, coord! {x: 12.0, y: 100.0}];
-        let simplified = rdp(vec.into_iter(), &1.0);
+        let simplified = rdp::<_, _, 2>(vec.into_iter(), &1.0);
         assert_eq!(simplified, compare);
     }
 
--- a/geo/src/algorithm/simplify.rs
+++ b/geo/src/algorithm/simplify.rs
@@ -301,21 +301,21 @@ mod test {
             coord! { x: 11.0, y: 5.5 },
             coord! { x: 27.8, y: 0.1 },
         ];
-        let simplified = rdp::<_, 2>(vec.into_iter(), &1.0);
+        let simplified = rdp::<_, _, 2>(vec.into_iter(), &1.0);
         assert_eq!(simplified, compare);
     }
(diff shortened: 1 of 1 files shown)


geo/src/algorithm/simplify.rs
```diff
@@ -87,15 +98,34 @@ where
...
+    // The farthest index was less than or equal to epsilon, so we will retain only the first
+    // and last indices, resulting in the indices inbetween getting culled.
+
+    // Update `simplified_len` to reflect the new number of indices by subtracting the number
+    // of indices we're culling.
+    *simplified_len -= rdp_indices.len() - 2;
+
+    // If `simplified_len` is now lower than the minimum number of indices needed, then don't
+    // perform the culling and return the original input.
+    if *simplified_len <= INITIAL_MIN {
```
michaelkirk, line 123: Can you imagine a situation where this returns a line much larger than INITIAL_MIN?

I'm not very familiar with this algorithm, but it seems like given a pathological 1000 coord polygon could be simplified in 1 step to 2 points, and then because that's below INITIAL_MIN, we'll instead return the original 1000 coord polygon. But I think I'd expect to get a 4-pt polygon in that case.

Or maybe that's not possible and I am misunderstanding.

frewsxcv: Agreed we would ideally return a 4 coordinate polygon in that case, but I'm not sure how to modify this algorithm to make that happen, and which is maybe worthy of a separate issue. Unless you know a good solution for this case

frewsxcv: @michaelkirk How do you feel about opening up a follow-up issue for optimizing in that scenario?

michaelkirk: I actually don't feel that strongly about it, it's just what occurred to me. I'm happy to see it merged if you think it's an improvement.

michaelkirk:
bors d=frewsxcv

frewsxcv:
bors r=michaelkirk

PR #950, opened 2022-12-06
Fix RDP recursion bug #2

michaelkirk:
---

Supersedes https://github.com/georust/geo/pull/941

I rebased #940 against main to include #942 and added a guard to avoid the spurious assertion failure introduced in https://github.com/georust/geo/commit/c61abfa1e758675cc6aa9925e3e6468cf104eaa4


michaelkirk:
bors r=urschrei

Issue #575, opened 2020-12-27
Make capitalization consistent for for Visvalingam-Whyatt simplification algorithms


frewsxcv:
`SimplifyVW`, `SimplifyVWPreserve`, `SimplifyVwIdx`

https://docs.rs/geo/0.16.0/geo/algorithm/simplifyvw/index.html

We should change these all to `SimplifyVw*`

PR #957, opened 2022-12-18
Make `SimplifyVw` algorithm naming consistent.

frewsxcv:
---

Fixes: https://github.com/georust/geo/issues/575

Changed:

* `SimplifyVW` -> `SimplifyVw`
* `SimplifyVWPreserve` -> `SimplifyVwPreserve`
* `simplifyvw` -> `simplify_vw`

Unchanged:

* `SimplifyVwIdx`

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -2,8 +2,10 @@
 
 ## unreleased
 
+* BREAKING: Make `SimplifyVw` naming consistent
+  * <https://github.com/georust/geo/pull/957>
 * BREAKING: Update to float_next_after-1.0.0
-  <https://github.com/georust/geo/pull/952>
+  * <https://github.com/georust/geo/pull/952>
 
 ## 0.23.1
 
--- a/geo/benches/simplifyvw.rs
+++ b/geo/benches/simplifyvw.rs
@@ -4,14 +4,14 @@ extern crate geo;
 
 use criterion::Criterion;
 use geo::prelude::*;
-use geo::simplifyvw::SimplifyVWPreserve;
+use geo::simplify_vw::SimplifyVwPreserve;
 
 fn criterion_benchmark(c: &mut Criterion) {
     c.bench_function("simplify vw simple f32", |bencher| {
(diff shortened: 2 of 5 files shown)


michaelkirk (commented):
Sorry that this (and all your other simple PR's) sat here for so long. I guess I lost track of them!

I agree that these new names are more idiomatic. 

It'd be nicer to deprecate this in a non-breaking release. Do you have time to make that change?

frewsxcv:
> Sorry that this (and all your other simple PR's) sat here for so long. I guess I lost track of them!
> 
> I agree that these new names are more idiomatic.
> 
> It'd be nicer to deprecate this in a non-breaking release. Do you have time to make that change?

I do, but I don't think it's worth the maintenance burden for us. We have deprecated features lingering in the code and it's extra effort for us to remember to remove them, and this is a trivial one or two character change for end users. I don't feel strongly about this though.

michaelkirk:
I think it's nicer for our users to prefer deprecation unless there is some abnormal circumstances which would make it unrealistic to apply them (granted that sometime's it's just not practical).

It's unfortunate that tracking deprecations would be an inordinate burden. Granted it's always going to be more than zero work - they are something a maintainer can _do_ (hence, a burden), for the sake of a hopefully nicer experience for users.

I typically scan the deprecations before doing a release, and delete any "old" ones: https://github.com/georust/geo/pull/892/files 

Maybe there's a better system by which we could track them?

I agree it's a small thing in this particular case. Actually, I think the compiler will already suggest the appropriate rename for the trait (but seemingly not the method, surprisingly). So I'm fine with merging this as is, but let me know if you have any thoughts on a better way to track deprecations for the future.

frewsxcv:
> So I'm fine with merging this as is, but let me know if you have any thoughts on a better way to track deprecations for the future.

It's too bad there's not an easy way to set reminders in Discord like there is in Slack. So at least we'd get some sort of notification in the future that we can remove deprecated things.

frewsxcv:
bors r=michaelkirk

PR #958, opened 2023-01-02
Implemented "Closest Point" from a `Point` on a `Geometry` using spherical geometry.

bbetov:
---
Essentially implementing `ClosestPoint` trait for spherical geometry. I named the new trait `HaversineClosestPoint`.

There is a slight (intentional) difference in behavior in the case when a segment (arc)'s length is very close to zero, the spherical algorithm will return `Closest::SinglePoint` instead of `Closest::Indeterminate`. I think even if the segment has degenerated to a point, there is still a closest point.


geo/src/algorithm/haversine_closest_point.rs
```diff
@@ -0,0 +1,579 @@
...
+        if let Closest::SinglePoint(pt) = line.haversine_closest_point(&p_from) {
+            assert!(pt.approx_eq(&Point::new(5.481_094_923_165_54, 82.998_280_987_615_33)));
+        } else {
+            panic!("Did not get Closest::SinglePoint!");
+        }
+    }
+
+    #[test]
+    fn point_to_linestring() {
+        let wkt = "LineString (3.86503906250000284 11.71231367187503736, 9.48691406250000568 17.3341886718750402, \
```
michaelkirk, line 435: This is not a big deal, so feel free to leave it as is, but since WKT disregards extra whitespace, you can omit the `\` and just have a multiline string.

This makes it slightly easier on yourself to write and for anyone wanting to copy/paste this WKT into another GIS engine

geo/src/algorithm/haversine_closest_point.rs
```diff
@@ -0,0 +1,579 @@
...
+    Coord, Geometry, GeometryCollection, Line, LineString, MultiLineString, MultiPoint,
+    MultiPolygon, Polygon, Rect, Triangle,
+};
+
+use num_traits::FromPrimitive;
+
+/// Closest point on a geometry from a point using great circles.
+///
+/// https://edwilliams.org/avform147.htm#XTE
+///
```
michaelkirk, line 14: Maybe it's worth mentioning that this trait only makes sense for coordinates in a spherical coordinate systems, like WGS84?

bbetov: Done.

geo/src/algorithm/haversine_closest_point.rs
```diff
@@ -0,0 +1,579 @@
...
+{
+    fn haversine_closest_point(&self, from: &Point<T>) -> Closest<T> {
+        multi_geometry_nearest(self, from)
+    }
+}
+
+#[cfg(test)]
+mod test {
+    trait ApproxEq<T> {
+        fn approx_eq(&self, other: &Self) -> bool;
```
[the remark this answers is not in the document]
bbetov: Done. I should have noticed.

geo/src/algorithm/haversine_closest_point.rs
```diff
@@ -0,0 +1,579 @@
...
+
+    use wkt::TryFromWkt;
+
+    use super::*;
+
+    #[test]
+    fn point_to_point() {
+        let p_1 = Point::new(-84.74905, 32.61454);
+        let p_2 = Point::new(-85.93942, 32.11055);
+        assert_eq!(p_1.haversine_closest_point(&p_2), Closest::SinglePoint(p_2));
```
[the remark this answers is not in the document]
bbetov: My bad. Fixed now. I switched the points.

geo/src/algorithm/haversine_closest_point.rs
bbetov: Done.

frewsxcv (approved):
Thank you for your patience @bbetov!

geo/src/algorithm/haversine_closest_point.rs
```diff
@@ -0,0 +1,579 @@
...
+}
+
+impl<T> HaversineClosestPoint<T> for Line<T>
+where
+    T: GeoFloat + FromPrimitive,
+{
+    fn haversine_closest_point(&self, from: &Point<T>) -> Closest<T> {
+        // There is a question how small this distance should be. This is really tiny (~10e-7 meters or smaller),
+        // when working on earth, but it is arbitrary.
+        const DISTANCE_EPSILON: f64 = 1e-12;
```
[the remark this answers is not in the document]
michaelkirk: Can add to the documentation that non-intersecting geometries may be considered to intersect if they are very close within `T::EPSILON`.

I am paranoid that this might cause someone some agony, but I don't have a better solution and would love to see this get merged!

michaelkirk:
bors r=frewsxcv,michaelkirk

> Can add to the documentation that non-intersecting geometries may be considered to intersect if they are very close within T::EPSILON.

I'll follow up with an update to the docs - no need to make this PR wait even longer on account of me.

michaelkirk:
There were some clippy errors too, due to a deprecation/rename that's happened in the meanwhile. Fixed in #1026.

PR #959, opened 2023-01-03
add minimum bounding box func of polygon

hzw456:
---
add minimun bounding box of polygon(mbr)

### DIFF
--- a/geo/src/algorithm/minimum_rotated_rect.rs
+++ b/geo/src/algorithm/minimum_rotated_rect.rs
@@ -1,17 +1,23 @@
+use num_traits::Float;
+
 use crate::{
     algorithm::{centroid::Centroid, rotate::Rotate, BoundingRect, CoordsIter},
-    Area, ConvexHull, CoordFloat, GeoFloat, GeoNum, Point, Polygon,
+    Area, ConvexHull, CoordFloat, GeoFloat, GeoNum, LinesIter, Polygon,
 };
-/// Return the minimun bounding rectangle of polygon
+/// Return the minimum bounding rectangle(MBR) of geometry
+/// reference: https://en.wikipedia.org/wiki/Minimum_bounding_box
+/// minimum rotated rect is the rectangle that can enclose all points given
+/// and have smallest area of all enclosing rectangles
+/// the rect can be any-oriented, not only axis-aligned.
 ///
 /// # Examples
 ///
 /// ```
 /// use geo_types::{line_string, polygon, LineString, Polygon};
-/// use geo::MinimunRotatedRect;
+/// use geo::MinimumRotatedRect;
 /// fn returns_polygon_mbr() {
 ///     let poly: Polygon<f64> = polygon![(x: 3.3, y: 30.4), (x: 1.7, y: 24.6), (x: 13.4, y: 25.1), (x: 14.4, y: 31.0),(x:3.3,y:30.4)];
-///     let mbr = MinimunRotatedRect::minimun_rotated_rect(&poly).unwrap();
+///     let mbr = MinimumRotatedRect::minimum_rotated_rect(&poly).unwrap();
 ///     assert_eq!(
 ///         mbr.exterior(),
 ///         &LineString::from(vec![
@@ -24,37 +30,34 @@ use crate::{
 ///     );
 /// }
 /// ```
-pub trait MinimunRotatedRect<'a, T> {
+pub trait MinimumRotatedRect<'a, T> {
     type Scalar: GeoNum;
-    fn minimun_rotated_rect(&'a self) -> Option<Polygon<Self::Scalar>>;
+    fn minimum_rotated_rect(&'a self) -> Option<Polygon<Self::Scalar>>;
 }
 
-impl<'a, T, G> MinimunRotatedRect<'a, T> for G
+impl<'a, T, G> MinimumRotatedRect<'a, T> for G
 where
     T: CoordFloat + GeoFloat + GeoNum,
     G: CoordsIter<'a, Scalar = T>,
-    f64: From<T>,
 {
     type Scalar = T;
-    fn minimun_rotated_rect(&'a self) -> Option<Polygon<Self::Scalar>> {
+
+    fn minimum_rotated_rect(&'a self) -> Option<Polygon<Self::Scalar>> {
         let convex_poly = ConvexHull::convex_hull(self);
-        let mut min_area = f64::MAX;
+        let mut min_area: T = Float::max_value();
         let mut min_angle: T = T::zero();
-        let points = convex_poly.exterior().clone().into_points();
         let mut rect_poly: Option<Polygon<T>> = None;
-        let mut rotate_point: Option<Point<T>> = None;
-        for i in 0..points.len() - 1 {
-            let ci = points[i];
-            let cii = points[i + 1];
+        let rotate_point = convex_poly.centroid();
+        for line in convex_poly.exterior().lines_iter() {
+            let (ci, cii) = line.points();
             let angle = (cii.y() - ci.y()).atan2(cii.x() - ci.x()).to_degrees();
-            let rotated_poly = Rotate::rotate_around_centroid(&convex_poly, -angle);
+            let rotated_poly = Rotate::rotate_around_point(&convex_poly, -angle, rotate_point?);
             let tmp_poly = rotated_poly.bounding_rect()?.to_polygon();
-            let area = f64::from(tmp_poly.unsigned_area());
+            let area = tmp_poly.unsigned_area();
             if area < min_area {
                 min_area = area;
                 min_angle = angle;
                 rect_poly = Some(tmp_poly);
-                rotate_point = convex_poly.centroid();
             }
         }
         Some(rect_poly?.rotate_around_point(min_angle, rotate_point?))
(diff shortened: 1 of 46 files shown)


rmanoka:
Thanks for the excellent contribution @hzw456 .  A couple of suggestions:

The bounding box in the name is ambiguous as it typically used for axis aligned ones.  Shapely calls this functionality as "minimum_rotated_rectangle".

Since the approach uses convex hull, would it generalize to all geoms that support conv hull?

geo/src/algorithm/minimun_rotated_rect.rs
```diff
@@ -0,0 +1,106 @@
+use crate::{
```
rmanoka, line 1: Please fix all clippy warnings to fix the ci failure

geo/src/algorithm/minimun_rotated_rect.rs
```diff
@@ -0,0 +1,106 @@
...
+    algorithm::{centroid::Centroid, rotate::Rotate, BoundingRect, CoordsIter},
+    coords_iter, Area, ConvexHull, CoordFloat, GeoFloat, GeoNum, Point, Polygon,
+};
+/// Returns the minimun bounding rect of polygon.
+///
+
+pub trait MinimunRotatedRect<'a, T> {
```
rmanoka, line 8: It looks like `T` is unused in the definition.  Remove?

hzw456: i found `ConvexHull` function has same usage for `T`，not sure about its real meaning.

hzw456: if i delete `T`,it will have lifetime error

rmanoka: Oh lol, let me create a separate issue to track removing it there.  Let's leave this as-is for now.

geo/src/algorithm/minimun_rotated_rect.rs
```diff
@@ -0,0 +1,100 @@
...
+impl<'a, T, G> MinimunRotatedRect<'a, T> for G
+where
+    T: CoordFloat + GeoFloat + GeoNum,
+    G: CoordsIter<'a, Scalar = T>,
+    f64: From<T>,
+{
+    type Scalar = T;
+    fn minimun_rotated_rect(&'a self) -> Option<Polygon<Self::Scalar>> {
+        let convex_poly = ConvexHull::convex_hull(self);
+        let mut min_area = f64::MAX;
```
rmanoka, line 41: Can we use the generic [`T::max_value`](https://docs.rs/num-traits/0.2.15/num_traits/float/trait.Float.html#tymethod.max_value) instead of `f64` here?

geo/src/algorithm/minimun_rotated_rect.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+impl<'a, T, G> MinimunRotatedRect<'a, T> for G
+where
+    T: CoordFloat + GeoFloat + GeoNum,
+    G: CoordsIter<'a, Scalar = T>,
+    f64: From<T>,
+{
+    type Scalar = T;
+    fn minimun_rotated_rect(&'a self) -> Option<Polygon<Self::Scalar>> {
```
rmanoka, line 39: Could you explain the outline of the algo.  and any citations if available for future reference?

geo/src/algorithm/minimun_rotated_rect.rs
```diff
@@ -0,0 +1,100 @@
...
+            let cii = points[i + 1];
+            let angle = (cii.y() - ci.y()).atan2(cii.x() - ci.x()).to_degrees();
+            let rotated_poly = Rotate::rotate_around_centroid(&convex_poly, -angle);
+            let tmp_poly = rotated_poly.bounding_rect()?.to_polygon();
+            let area = f64::from(tmp_poly.unsigned_area());
+            if area < min_area {
+                min_area = area;
+                min_angle = angle;
+                rect_poly = Some(tmp_poly);
+                rotate_point = convex_poly.centroid();
```
rmanoka, line 57: Looks to be the same value being assigned every time.  Just initialize it directly?

rmanoka (approved):
I've added a few minor code suggestions, but otherwise lgtm!  

To summarize for other reviewers:  this PR adds support for the `minimum_rotated_rect` algorithm for all our geometry types.  This is the smallest area rectangle that contains the given geometry.  The algorithm involves two steps:

1. compute convex hull of geom:  since rectangle is convex, any covering rect. should cover the convex hull of the geom.
2. for any convex shape, it is known that the min rectangle overlaps with some side of the polygon (citation needed).  This implementation iterates over all sides and computes the min. area rectangle.  The current implementation constructs rotated polygons, to compute the bounding rect thus giving a quadratic total run-time for this step.  This may be optimized in the future via the rotating calipers technique.

@urschrei @frewsxcv @michaelkirk  Any one else have time for a quick skim?  We could merge it in a week if all looks good.

geo/src/algorithm/minimun_rotated_rect.rs
```diff
@@ -0,0 +1,100 @@
...
+///             (1.7000000000000006, 24.6),
+///             (1.4501458363715918, 30.446587428904767),
+///             (14.4, 31.0),
+///             (14.649854163628408, 25.153412571095235),
+///             (1.7000000000000006, 24.6),
+///         ])
+///     );
+/// }
+/// ```
+pub trait MinimunRotatedRect<'a, T> {
```
frewsxcv, line 27:
```suggestion
pub trait MinimumRotatedRect<'a, T> {
```

geo/src/algorithm/minimun_rotated_rect.rs
```diff
@@ -0,0 +1,100 @@
...
+    type Scalar: GeoNum;
+    fn minimun_rotated_rect(&'a self) -> Option<Polygon<Self::Scalar>>;
```
frewsxcv, line 29:
```suggestion
    fn minimum_rotated_rect(&'a self) -> Option<Polygon<Self::Scalar>>;
```

geo/src/algorithm/minimun_rotated_rect.rs
```diff
@@ -0,0 +1,100 @@
...
+    fn minimun_rotated_rect(&'a self) -> Option<Polygon<Self::Scalar>> {
+        let convex_poly = ConvexHull::convex_hull(self);
+        let mut min_area = f64::MAX;
+        let mut min_angle: T = T::zero();
+        let points = convex_poly.exterior().clone().into_points();
+        let mut rect_poly: Option<Polygon<T>> = None;
+        let mut rotate_point: Option<Point<T>> = None;
+        for i in 0..points.len() - 1 {
+            let ci = points[i];
+            let cii = points[i + 1];
```
frewsxcv, line 48: If you want, I think you should be able to do something like:

```rust
use geo::algorithm::lines_iter::LinesIter;

for line in convex_poly.exterior().lines_iter() {
    let (ci, cii) = line.points();
    // ...
}
```

geo/src/algorithm/minimun_rotated_rect.rs
hzw456: any process with this?

rmanoka:
Looks like all reviewer suggestions/comments have been addressed.  Will merge it in a day.

frewsxcv (approved):

geo/src/algorithm/minimum_rotated_rect.rs
```diff
@@ -0,0 +1,103 @@
+use num_traits::Float;
+
+use crate::{
+    algorithm::{centroid::Centroid, rotate::Rotate, BoundingRect, CoordsIter},
+    Area, ConvexHull, CoordFloat, GeoFloat, GeoNum, LinesIter, Polygon,
+};
+/// Return the minimum bounding rectangle(MBR) of geometry
+/// reference: https://en.wikipedia.org/wiki/Minimum_bounding_box
+/// minimum rotated rect is the rectangle that can enclose all points given
```
michaelkirk, line 9: The grammer / punctuation here seems a little off.

geo/src/algorithm/minimum_rotated_rect.rs
```diff
@@ -0,0 +1,103 @@
...
+/// and have smallest area of all enclosing rectangles
+/// the rect can be any-oriented, not only axis-aligned.
+///
+/// # Examples
+///
+/// ```
+/// use geo_types::{line_string, polygon, LineString, Polygon};
+/// use geo::MinimumRotatedRect;
+/// fn returns_polygon_mbr() {
```
michaelkirk, line 18: This defines a doc-test, but the test doesn't actually run anything, it only defines the function. 

if you remove the function declaration, and just have the inner code, a `main` will be synthesized and the test will run.

michaelkirk (changes requested):
Please add an entry for this new algorithm in our top level docs: file:///Users/mkirk/src/georust/geo/target/doc/geo/index.html#boundary

geo/src/algorithm/minimum_rotated_rect.rs
hzw456: fixed

frewsxcv:
@hzw456 One more thing– can you add this to `CHANGES.md`?

frewsxcv (approved):
thank you so much!

Issue #834, opened 2022-05-18
Implement `HaversineDistance` for all combinations of geometry types


frewsxcv:

michaelkirk:
Does anyone have any thoughts on how to go about this? Or have any examples of other implementations?

michaelkirk:
from http://www.faqs.org/faqs/graphics/algorithms-faq/
> Subject 1.02: How do I find the distance from a point to a line?

Has a good euclidean answer (JTS uses this under the hood for `geom_a.distance(geom_b)`).

We have implemented this algorithm as [`geo_types::private_utils::point_line_euclidean_distance`](https://github.com/georust/geo/blob/main/geo-types/src/private_utils.rs#L106)

I don't see how it could work in non-euclidean space though.

urschrei:
Oh, I wonder whether we could use ClosestPoint to get the destination point…

urschrei:
I _think_ this is a correct way to work it out:

```rust
impl<T> HaversineDistance<T, Point<T>> for Polygon<T>
where
    T: GeoFloat + FromPrimitive,
{
    fn haversine_distance(&self, rhs: &Point<T>) -> T {
        match self.closest_point(rhs) {
            crate::Closest::Intersection(_) => todo!(),
            crate::Closest::SinglePoint(sp) => rhs.haversine_distance(&sp),
            crate::Closest::Indeterminate => todo!(),
        }
    }
}
```

But we can't do this in practice because the `HaversineDistance` trait doesn't return a `Result` but we have to account for the possibility of the `closest_point` result being an intersection or indeterminate. Ugh.

michaelkirk:
I think this only kicks the can down the road a bit, since `closest_point` is itself computed with assumptions on euclidean space. 

We'd get the haversine meter distance to the point that is closest in terms of degrees, but may not be closest in terms of meters - right?

michaelkirk:
I'm thinking we'd need something specially formulated for the spherish model of the earth. Maybe I finally need to read the chamberlain duquette paper?

https://trs.jpl.nasa.gov/bitstream/handle/2014/41271/07-0286.pdf?sequence=1&isAllowed=y

michaelkirk:
I wasn't familiar with the term before, but "cross track distance" is the distance between a point and a great circle.

http://www.movable-type.co.uk/scripts/latlong.html#cross-track
https://gis.stackexchange.com/questions/209540/projecting-cross-track-distance-on-great-circle

Maybe that's useful.

[...]

PR #961, opened 2023-01-07
Add CrossTrackDistance calculation, should be independently useful, but is also a starting point for the implementation of #834.

mbcltd:
---

### DIFF
--- a/geo/src/algorithm/cross_track_distance.rs
+++ b/geo/src/algorithm/cross_track_distance.rs
@@ -1,0 +1,128 @@
+use num_traits::FromPrimitive;
+use geo_types::{CoordFloat, Point};
+use crate::{Bearing, HaversineDistance, MEAN_EARTH_RADIUS};
+
+/// Determine the cross track distance (also known as the cross track error) which is the shortest
+/// distance between a point and a continuous line.
+pub trait CrossTrackDistance<T, Rhs = Self> {
+    /// Determine the cross track distance between this point and a line
+    /// which passes through line_point_a and line_point_b
+    ///
+    /// # Units
+    ///
+    /// - return value: meters
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// use geo::prelude::*;
+    /// use geo::point;
+    ///
+    /// // New York City
+    /// let p1 = point!(x: -74.006f64, y: 40.7128f64);
+    ///
+    /// // Miami
+    /// let line_point_a = point!(x: -80.1918f64, y: 25.7617f64);
+    ///
+    /// // Washington
+    /// let line_point_b = point!(x: -120.7401, y: 47.7511f64);
+    ///
+    /// let distance = p1.cross_track_distance(&line_point_a, &line_point_b);
+    ///
+    /// assert_eq!(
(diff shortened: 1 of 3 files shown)


frewsxcv:
Hmm I'm wondering if instead of creating a new algorithm, we could add a new implementation for Haversine distance. e.g. `impl HaversineDistance<Point> for Line`. What do you think?

mbcltd:
That was my thought initially too, but this algorithm is the distance from point a to the great circle that passes through b and c, not the shortest distance to the line bounded by the point b and c. I made the function be from a point to `line_point_a` and `line_point_b` because I felt that most people would assume that `impl HaversineDistance<Point> for Line` or vice-versa would be the minimum distance to the line itself.

ctd_v_hd

I created a gist to give an example of how to use this to calculate the distance between a point and a line here: https://gist.github.com/mbcltd/09b62b706dd1bd4d623f3167aaf5c53f

geo/src/algorithm/cross_track_distance.rs
```diff
@@ -0,0 +1,128 @@
...
+}
+
+#[cfg(test)]
+mod test {
+    use crate::CrossTrackDistance;
+    use crate::HaversineDistance;
+    use crate::Point;
+
+    #[test]
+    fn should_equal_haversine_distance_for_short_lines() {
```
michaelkirk, line 61: Can you explain what this test is about? 

The current test name seems misleading - the fact that they are equal is not because the line is short, but because the point p is "approximately" perpendicular to the line, so that the haversine distance is "almost" the same as the cross track distance.

Said differently, this test would also pass if we made the line very long:

```
let line_point_a = Point::new(1., 100.);
let line_point_b = Point::new(1., -100.);
```

And conversely, a very short line, depending on where it's placed, could fail this test:
```
let line_point_a = Point::new(10., 10.);
let line_point_b = Point::new(9.99, 9.99);
```

Is that right, or have I misunderstood something? 

If so it seems redundant with  `cross_track_distance_to_line_orthogonal_to_point` below. I'd say delete it or at least give it a clearer name like `cross_track_distance_to_line_orthogonal_to_point_with_short_line` or whatever.

michaelkirk (approved):
Sorry that this sat her for so long. I have a nit-picky clarification about one of the tests but it otherwise looks good.

I'll give you a little time to respond and/or rebase if you're around, otherwise since it's just a small clarification, I'll plan to update, rebase, and merge it myself soon.

geo/src/algorithm/cross_track_distance.rs
mbcltd: Hi @michaelkirk you're absolutely right, this is effectively a pointless variation of `cross_track_distance_to_line_orthogonal_to_point`, I think it's best to remove it.

michaelkirk:
While trying to resolve a merge conflict in the change log I accidentally merged directly to `main` without bors (🤦) and some things are failing the CI lint. https://github.com/georust/geo/actions/runs/4996284063/jobs/8949273865

I've just opened a PR for the next geo PR where I've addressed the CI failures: https://github.com/georust/geo/pull/1014

PR #966, opened 2023-01-13
bump MSRV to 1.63

### LINKED ISSUE #964: Consider bumping MSRV
We've settled on supporting 3 releases in https://github.com/georust/geo/blob/main/.github/workflows/test.yml#L66, but we're testing against 1.58, 1.62 and 1.63. The latest version is 1.66, with 1.67 coming out soon. This prevents up from upgrading the `arbitrary` dependency.

CC #963, @martinfrances107

michaelkirk:
---

Fixes #964

### DIFF
--- a/geo/src/algorithm/bool_ops/assembly.rs
+++ b/geo/src/algorithm/bool_ops/assembly.rs
@@ -57,6 +57,7 @@ impl<T: GeoFloat> RegionAssembly<T> {
             };
 
             // Connect consecutive segments
+            #[allow(clippy::bool_to_int_with_if)]
             let mut idx = if prev_region { 1 } else { 0 };
 
             while idx < iter.intersections().len() {
@@ -149,7 +150,7 @@ impl<T: GeoFloat> RegionAssembly<T> {
                     }
                     parent_snake_idx = parent.parent_snake_idx;
                 }
-                let this_children = children.entry(parent_ring_idx).or_insert(vec![]);
+                let this_children: &mut Vec<_> = children.entry(parent_ring_idx).or_default();
                 this_children.push(ring_idx);
             } else {
                 continue;
--- a/geo/src/algorithm/convex_hull/qhull.rs
+++ b/geo/src/algorithm/convex_hull/qhull.rs
@@ -42,9 +42,8 @@ where
         // (2) max_idx = min_idx: then any point could be
         // chosen as max. But from case (1), it could now be
         // 0, and we should not decrement it.
-        if max_idx > 0 {
-            max_idx -= 1;
-        }
+        max_idx = max_idx.saturating_sub(1);
+
         let max = swap_remove_to_first(&mut points, max_idx);
         (min, max)
     };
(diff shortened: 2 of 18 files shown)


lnicola (approved):

geo/src/algorithm/bool_ops/assembly.rs
```diff
@@ -149,7 +150,7 @@ impl<T: GeoFloat> RegionAssembly<T> {
                     }
                     parent_snake_idx = parent.parent_snake_idx;
                 }
-                let this_children = children.entry(parent_ring_idx).or_insert(vec![]);
+                let this_children: &mut Vec<_> = children.entry(parent_ring_idx).or_default();
```
michaelkirk, line 153: Without the `or_insert` argument, type inference doesn't work, so we have to add a more explicit type annotation.

geo/src/algorithm/convex_hull/qhull.rs
```diff
@@ -42,9 +42,8 @@ where
         // (2) max_idx = min_idx: then any point could be
         // chosen as max. But from case (1), it could now be
         // 0, and we should not decrement it.
-        if max_idx > 0 {
-            max_idx -= 1;
-        }
+        max_idx = max_idx.saturating_sub(1);
```
michaelkirk, line 45: Unclear if this is an improvement... 😒📎

geo/src/algorithm/relate/geomgraph/geometry_graph.rs
```diff
@@ -336,6 +336,7 @@ where
 
         // determine the current location for the point (if any)
         let boundary_count = {
+            #[allow(clippy::bool_to_int_with_if)]
             let prev_boundary_count =
                 if Some(CoordPos::OnBoundary) == label.position(arg_index, Direction::On) {
```
michaelkirk, line 341: The clippy suggestion was to do something like: 

`let prev_boundary_count = usize::from(Some(CoordPos::OnBoundary) == label.position(arg_index, Direction::On))`

Which, in my opinion, is just too dense. 

Pulling out a temp could address that:
```
let is_on_boundary = Some(CoordPos::OnBoundary) == label.position(arg_index, Direction::On);
let prev_boundary_count = usize::from(is_on_boundary);
```

But, I still prefer what's already there. I guess my (perhaps controversial) opinion is that booleans shouldn't be treated as numbers and that clippy is wrong to encourage this. 🤷

geo/src/algorithm/convex_hull/qhull.rs
urschrei: I like it better, if only because you have immediate access to what saturating_sub does if you're using an IDE / LSP

lnicola: But.. you don't need to read do docs to understand -= 1 either 😀.

michaelkirk: This entire commit is a bikeshed honeypot / nerd snipe. Everybody be on your best behavior.

geo/src/algorithm/relate/geomgraph/geometry_graph.rs
lnicola: I'd write it as `matches!(label.position(arg_index, Direction::On), Some(CoordPos::OnBoundary)) as usize`, but yeah, whatever.

But if this comes up again, I'd be for globally suppressing this lint.

lnicola (approved):

lnicola:
bors d+

Issue #965, opened 2023-01-13
MSRV policy 


martinfrances107:
PR #963 strove to fix all the notifications reported by 
```
 cargo outdated
```
We finished by reducing the patch to ONLY updating the packages that could be fixed without a formal policy decision.

I think we have a 'living' policy, something not written down. 

> "All our dependencies should be bump, to the latest, and as a fallback be bumped to the a version that is AT LEAST supported."

Formalising this will be disruptive, so this issue is a request for feedback..

1)  Write a PR that fixes any problems that were causing build failures
2) pull in a MSRV policy from an existing project 

I am going to work in (1) NOW

[All credit to lnicola. This issue is largely a write up of his reviews in that PR]

Today after #963 was merged below  is the "outdated" report.

My immediate want is de-duplicate this list to only include packages once 
( generic_array, has32 etc. ) 
```
Name                               Project  Compat   Latest   Kind    Platform
----                               -------  ------   ------   ----    --------
arbitrary                          1.1.3    ---      1.2.2    Normal  ---
as-slice->generic-array            0.12.4   0.13.3   Removed  Normal  ---
as-slice->generic-array            0.13.3   ---      Removed  Normal  ---
as-slice->generic-array            0.14.6   0.13.3   Removed  Normal  ---
as-slice->stable_deref_trait       1.2.0    ---      Removed  Normal  ---
atomic-polyfill->critical-section  1.1.1    Removed  ---      Normal  ---
generic-array->typenum             1.16.0   ---      Removed  Normal  ---
generic-array->version_check       0.9.4    ---      Removed  Build   ---
generic-array->version_check       0.9.4    Removed  Removed  Build   ---
heapless->as-slice                 0.1.5    ---      Removed  Normal  ---
heapless->atomic-polyfill          0.1.11   Removed  ---      Normal  cfg(target_arch = "avr")
heapless->generic-array            0.14.6   ---      Removed  Normal  ---
heapless->hash32                   0.1.1    ---      0.2.1    Normal  ---
heapless->hash32                   0.2.1    0.1.1    ---      Normal  ---
heapless->rustc_version            0.4.0    Removed  ---      Build   ---
heapless->spin                     0.9.4    Removed  ---      Normal  cfg(target_arch = "x86_64")
lock_api->autocfg                  1.1.0    Removed  ---      Build   ---
lock_api->scopeguard               1.1.0    Removed  ---      Normal  ---
rstar                              0.8.4    ---      0.9.3    Normal  ---
[...]

PR #968, opened 2023-01-14
Breaking change, rust v1.59 is not longer supported, simplify

martinfrances107:
By Simplify I mean replace  the "use_rstat_0_8" and "use_rstat_0_9" with a single  "use_rstar".  The namedspace-features issue in no longer relevant.

### DIFF
--- a/geo-types/CHANGES.md
+++ b/geo-types/CHANGES.md
@@ -5,13 +5,16 @@
 * Return `DoubleEndedIterator` from `LineString::points` and `LineString::points_mut`
   * <https://github.com/georust/geo/pull/951>
 * POSSIBLY BREAKING: Minimum supported version of Rust (MSRV) is now 1.63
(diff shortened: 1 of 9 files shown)


martinfrances107:
Opps I have just seen a problem, I need to update the change log

my bad. one sec.

lnicola:
As mentioned in https://github.com/georust/geo/issues/965#issuecomment-1382159152, this involves a breaking change to `geo-types`. We'll do it at some point (when?), but I think we'll want to keep supporting the one old version of `rstar` (0.9 when 0.10 comes out) until it reaches 1.0. This is in line with other popular crates like https://docs.rs/postgres/latest/postgres/#features.

martinfrances107:
understood, I will make changes tomorrow.

martinfrances107:
This issue has been fixed elsewhere

PR #969, opened 2023-01-14
doc(centroid): document each impl and include example

RobWalt:
---

Hey everyone!

This PR just includes documentation and some examples of the implementations of the `Centroid` trait for all the `geo-types`.

I also moved some existing docs to the impl of the `centroid` function and I moved the impl of the trait for `Triangle` further up below the impl of the trait for `Rect` since I think this grouping makes more sense.

### DIFF
--- a/geo/src/algorithm/centroid.rs
+++ b/geo/src/algorithm/centroid.rs
@@ -63,6 +63,24 @@ where
 {
     type Output = Point<T>;
 
+    /// The Centroid of a [`Line`] is it's middle point
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use geo::Centroid;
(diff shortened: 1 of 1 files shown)


geo/src/algorithm/centroid.rs
```diff
@@ -63,6 +63,24 @@ where
 {
     type Output = Point<T>;
 
+    /// The Centroid of a [`Line`] is it's middle point
```
urschrei, line 66: Should be "its"!

RobWalt: Fixing this up everywhere in the file, had multiple occurences

frewsxcv (approved):
thank you sooo much! these docs will be so helpful to users 🎉

Issue #350, opened 2019-03-29
Polygon::{exterior,interiors}_mut can be replaced with methods that pass *owned* data


frewsxcv:
Instead of passing a mutable reference of the exterior `LineString` to the `exterior_mut` closure, we can pass the actual underlying `LineString` temporarily to the closure, as long as it gets returned from the closure:

```rust
    // (I'm not thrilled about the exterior_owned name)
    pub fn exterior_owned<F>(&mut self, mut f: F)
    where
        F: FnMut(LineString<T>) -> LineString<T>,
    {
        let mut ls = LineString(vec![]);
        ::std::mem::swap(&mut ls, &mut self.exterior);
        self.exterior = f(ls);
        self.exterior.close();
    }
```

This allows for more flexibility than passing the mutable reference to the underlying data.

The same can be done for `interiors_mut`, and we might be able to remove our `interiors_push` method since you'd then be able to push onto the `Vec` in the closure.

michaelkirk:
I want to make sure I understand this proposal. Sorry for not responding sooner - it's been here a while!

> we might be able to remove our interiors_push method since you'd then be able to push onto the Vec in the closure.

So currently we can write this:
```
polygon.push_interior(line_string)
```

With this proposal, we could write:
```
polygon.interiors( |interiors| interiors.push(line_string); interiors )
```

I think even if we could write the latter, I'd still prefer to keep the former. Is that what you meant, or did I misunderstand?

> This allows for more flexibility than passing the mutable reference to the underlying data.

I'm trying and failing to think of examples here. Can you contrive some code where we'd want owned data for these methods? It seems like it'd be equivalent for most usages, except you now have to add a return statement at the end.

mutable reference:
```
polygon.interiors_mut_ref( |interiors| {
    for interior in interiors {
        interiors.translate_mut(foo); 
    }  
});
```

owned:
```
polygon.interiors_owned( |interiors| {
    for interior in interiors {
        interiors.translate_mut(foo); 
    }
    return interiors;
})
```

A downside with this `owned` approach is that it'd involve introducing two moves, right?

frewsxcv:
@michaelkirk After taking some time to digest your comment, and rethink this API change, I think you're right. Unless we know of a concrete situation where someone needs an owned `LineString` temporarily, this API change seems premature. Thanks for writing up your comment.

PR #971, opened 2023-01-21
Rename `LineString::exterior_mut` (and `interiors`) to `exterior_mut_ref`. Deprecate old fn.

frewsxcv:
---
In preparation for https://github.com/georust/geo/issues/350.

Eventually we can have these deprecated functions become non-deprecated and be passed owned data.

### DIFF
--- a/geo-types/src/geometry/polygon.rs
+++ b/geo-types/src/geometry/polygon.rs
@@ -236,7 +236,63 @@ impl<T: CoordNum> Polygon<T> {
     /// ```
     ///
     /// [will be closed]: #linestring-closing-operation
+    #[deprecated(note = "Renamed to `exterior_mut_ref`. In a future release, `exterior_mut` will be passed an owned `LineString`.")]
     pub fn exterior_mut<F>(&mut self, f: F)
+    where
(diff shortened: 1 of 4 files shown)


michaelkirk:
I left some comments at https://github.com/georust/geo/issues/350#issuecomment-1399314655, since I have some questions with the underlying proposal, before getting into implementation specifics.

geo-types/src/geometry/polygon.rs
```diff
@@ -244,6 +247,61 @@ impl<T: CoordNum> Polygon<T> {
...
     }
 
+    /// Execute the provided closure `f`, which is provided with a mutable
+    /// reference to the exterior `LineString` ring.
+    ///
+    /// After the closure executes, the exterior `LineString` [will be closed].
+    ///
+    /// # Examples
+    ///
+    /// ```
```
metasim, line 257: I've been typically adding `rust, no_run`, because docs.rs builder doesn't have GDAL pre-installed in the environment.

metasim:
(just deleted a comment because I thought this was the `gdal` repo 🤦)

PR #978, opened 2023-01-26
update geo docs for MapCoords/InPlace

### LINKED ISSUE #977: `MapCoordsInPlace` not listed under `Iteration` in `lib.rs`
`MapCoordsInPlace` should be listed under `Iteration` in  `lib.rs`, but right now it's not mentioned. Instead, the deprecated `MapCoordsInplace` is mentioned. I can imagine how this can confuse users of this library.

https://github.com/georust/geo/blob/main/geo/src/lib.rs#L111

`TryMapCoords` and `TryMapCoordsInplace` should probably be addressed somehow as well, since they are deprecated too. I suggest removing them since `MapCoords` and (soon) `MapCoordsInPlace`  are listed there, and contain the non-deprecated versions of these.

michaelkirk:
- [x]  I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---
I missed this in #811
Fixes #977

### DIFF
--- a/geo/src/lib.rs
+++ b/geo/src/lib.rs
@@ -105,16 +105,12 @@
 //!
 //! ## Iteration
 //!
-//! - **[`CoordsIter`](CoordsIter)**: Iterate over the coordinates of a geometry
-//! - **[`MapCoords`](MapCoords)**: Map a function over all the coordinates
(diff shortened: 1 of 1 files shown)


michaelkirk:
bors cancel

Noticed a typo 😬

PR #987, opened 2023-02-23
Add new feature to support rstar_0_10

gliderkite:
---

Add a new feature `use-rstar_0_10` in `geo-types` and enabled it in `geo`.

### DIFF
--- a/geo/Cargo.toml
+++ b/geo/Cargo.toml
@@ -18,13 +18,13 @@ use-serde = ["serde", "geo-types/serde"]
 
 [dependencies]
 float_next_after = "1.0.0"
-geo-types = { version = "0.7.8", features = ["approx", "use-rstar_0_9"] }
+geo-types = { version = "0.7.8", features = ["approx", "use-rstar_0_10"] }
 geographiclib-rs = "0.2"
 log = "0.4.11"
 num-traits = "0.2"
 proj = { version = "0.27.0", optional = true }
 robust = "0.2.2"
-rstar = "0.9.3"
+rstar = "0.10.0"
 serde = { version = "1.0", optional = true, features = ["derive"] }
 
 [dev-dependencies]
--- a/geo-types/CHANGES.md
+++ b/geo-types/CHANGES.md
@@ -7,6 +7,7 @@
 * POSSIBLY BREAKING: Minimum supported version of Rust (MSRV) is now 1.63
 * Add `no_std` compatibility when the new default `std` feature is disabled
   * <https://github.com/georust/geo/pull/936>
+* Support `rstar` version `0.10` in feature `use-rstar_0_10`.
 
 ## 0.7.8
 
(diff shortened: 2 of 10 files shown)


geo/Cargo.toml
```diff
@@ -18,13 +18,13 @@ use-serde = ["serde", "geo-types/serde"]
...
 float_next_after = "1.0.0"
-geo-types = { version = "0.7.8", features = ["approx", "use-rstar_0_9"] }
+geo-types = { version = "0.7.8", features = ["approx", "use-rstar_0_10"] }
 geographiclib-rs = "0.2"
 log = "0.4.11"
 num-traits = "0.2"
 proj = { version = "0.27.0", optional = true }
 robust = "0.2.2"
-rstar = "0.9.3"
+rstar = "0.10.0"
```
lnicola, line 27: I think this is a breaking change in `geo` because `rstar` is exposed in some type bounds (you might want to double-check, though), we should at least add it to the changelog and maybe bump the crate version.

gliderkite: What's the process I should follow to upgrade the `geo-types` version (`0.8.0`) before changing the `geo` version? I see the `wkt` may also be linked to this.

lnicola: My understanding is that we want to avoid breaking chages in `geo-types`, but they're fine to do in `geo`. I don't think you've made any of those, so it should be fine?

Regarding `wkt`, it doesn't use `geo`, and it can stay on the current version of `geo-types`. `cargo` will pick up the new one automatically, since they're compatible.

gliderkite: > My understanding is that we want to avoid breaking chages in geo-types

Yes, that's also my understanding, but this introduces a new dependencies and I was thinking this should be reflected in the version number (I should have written `0.7.9`).

If it's ok to just bump the `geo` version to `0.24.0` I'll be happy to do so, otherwise I think we would need to publish first the `geo-types` version (and possibly the `wkt` since it depends on a pinned version of `geo-types`).

lnicola: Yeah, it should be a patch bump on geo-types and it's disabled by default, so it doesn't matter. Neither does it matter for wkt, since cargo will happily use the new version. https://github.com/georust/wkt/blob/main/Cargo.toml#L13 isn't a pin, and it will pick 0.7.5, 0.7.6, or any other compatible version.

I just don't know if we want to bump the geo version now, or just before the release.

gliderkite: Right, so checking other [unreleased breaking changes](https://github.com/georust/geo/blob/main/geo/CHANGES.md#unreleased) the `geo` version was not bumped.

For now I'll add a new entry in the CHANGES.md.

lnicola (approved):

geo/Cargo.toml
michaelkirk: Perfectly handled, thanks!

michaelkirk:
I fixed a conflict in the change log due to another PR being merged in the meanwhile. 

Otherwise, everything LGTM - thanks!

bors r+

PR #988, opened 2023-02-27
Adding support geodesic area and perimeter calculations from geographlib-rs

phayes:
---

This is a PR for adding a `GeodesicArea` trait for highly accurate geodesic area and perimeter calculations. 

See https://github.com/georust/geographiclib-rs/pull/49


geo/src/algorithm/geodesic_area.rs
```diff
@@ -0,0 +1,504 @@
...
+    /// (x: 0.00388383, y: 51.501574),
+    /// (x: 0.00538587, y: 51.502278),
+    /// (x: 0.00553607, y: 51.503299),
+    /// (x: 0.00467777, y: 51.504181),
+    /// (x: 0.00327229, y: 51.504435),
+    /// (x: 0.00187754, y: 51.504168),
+    /// (x: 0.00087976, y: 51.503380),
+    /// (x: 0.00107288, y: 51.502324),
+    /// (x: 0.00185608, y: 51.501770),
+    /// (x: 0.00388383, y: 51.501574),
```
frewsxcv, line 49:
```suggestion
    ///     (x: 0.00388383, y: 51.501574),
    ///     (x: 0.00538587, y: 51.502278),
    ///     (x: 0.00553607, y: 51.503299),
    ///     (x: 0.00467777, y: 51.504181),
    ///     (x: 0.00327229, y: 51.504435),
    ///     (x: 0.00187754, y: 51.504168),
    ///     (x: 0.00087976, y: 51.503380),
    ///     (x: 0.00107288, y: 51.502324),
    ///     (x: 0.00185608, y: 51.501770),
    ///     (x: 0.00388383, y: 51.501574),
```

phayes:
This is now ready for merging / review. 

The only open question is wether we like what I've done with having a bunch of different methods for getting area, perimeter or both.   Maybe we want to ignore perimeter and only focus on area?  Maybe we always want to get both together since they're computed together?

geo/src/algorithm/geodesic_area.rs
```diff
@@ -0,0 +1,504 @@
...
+    fn geodesic_perimeter_area_signed(&self) -> (f64, f64) {
+        geodesic_area(self, true, false, false)
+    }
+
+    fn geodesic_perimeter_area_unsigned(&self) -> (f64, f64) {
+        geodesic_area(self, false, false, false)
+    }
+}
+
+fn geodesic_area(poly: &Polygon, sign: bool, reverse: bool, exterior_only: bool) -> (f64, f64) {
```
michaelkirk, line 189: `reverse` and `exterior_only` seems to always be false. Do you want to get rid of them? Or if you want to keep them, can you comment as to their future purpose?

geo/src/algorithm/geodesic_area.rs
```diff
@@ -0,0 +1,504 @@
...
+                pa.add_point(p.y(), p.x());
+            });
+            let (perimeter, area, _) = pa.compute(sign);
+            inner_area += area.abs();
+            inner_perimeter += perimeter;
+        });
+        (inner_perimeter, inner_area)
+    };
+
+    if outer_area < 0.0 && inner_area > 0.0 {
```
michaelkirk, line 226: Why this check? It seems like you might be trying to "fix" someone's winding for them?

geo/src/algorithm/geodesic_area.rs
```diff
@@ -0,0 +1,504 @@
...
+            (x: 125., y: -15.),
+            (x: 144., y: -15.),
+            (x: 154., y: -27.),
+            (x: 148., y: -39.),
+            (x: 130., y: -33.),
+            (x: 117., y: -37.),
+            (x: 113., y: -22.),
+            (x: 125., y: -15.),
+        ];
+        assert_relative_eq!(-7786102826806.07, polygon.geodesic_area_signed());
```
michaelkirk, line 370: This one fails for me locally.

```
---- algorithm::geodesic_area::test::test_negative stdout ----
thread 'algorithm::geodesic_area::test::test_negative' panicked at 'assert_relative_eq!(-7786102826806.07, polygon.geodesic_area_signed())

    left  = -7786102826806.07
    right = -7786102826806.072
```

michaelkirk (commented):
Thanks for another great PR! 

The overall direction I feel good about, but I had some questions/comments.

geo/src/algorithm/geodesic_area.rs
phayes: Originally I was planning on more features, but decided to reign it in a bit. So I guess some open questions:

1. Do we want to *fully* support clockwise-wounded geometries?  The Documentation (https://docs.rs/geo/latest/geo/#semantics) states "The geospatial types provided here aim to adhere to the [OpenGIS Simple feature access](https://www.ogc.org/standards/sfa) standards."  This suggests to me that we don't really need to *actually* support reverse-wound geometries other than on a minimal "best effort" basis.  If this is true, I'll remove the "reverse" option here to only fully support counterclockwise-wound geometries as per the Simple Feature spec. 

2. Is it useful to have an option to only get the perimeter of the exterior?  Originally I thought this might be useful, but in retrospect just calling `geodesic_length()` on the exterior ring does the same thing.   The only trickly thing here is that sometimes I've seen polygons constructed where the first point of the exterior ring and the last point of the exterior ring are not the same (there's an *implicit* closure of the ring, instead of an explicit closure by duplicating the origin point).   If this case calling `geodesic_length()` on the exterior-ring would result in an incorrect result, but we could correct for that here knowing that the exterior-rings and interior-rings should always close, even if only implicitly.

geo/src/algorithm/geodesic_area.rs
phayes: This is a best-effort fix for badly wound polygons.  I found this looking at the test suite for `ChamberlainDuquetteArea`. The test-suite contains a badly-wound polygon here (https://github.com/georust/geo/blob/main/geo/src/algorithm/chamberlain_duquette_area.rs#L228).  Both the exerior-ring and interior-rings are wound counter-clockwise, which is incorrect as per Simple Features spec.   However, we can easily detect this and correct for it, which is what you see here.

phayes: It might make sense to add a bunch more test to ensure that this logic isn't having any unintended consequences.  I'd be happy to do so if you think it useful.

phayes:
I've added a bunch more tests to confirm that my "winding correction logic" is good and adjusted the epsilons.  I would appreciate another test run on your local to confirm that my epsilons are good (I still think it's really weird that there's a difference in behavior between our machines!).

michaelkirk:
> I've added a bunch more tests to confirm that my "winding correction logic" is good and adjusted the epsilons. 

Without the "winding correction logic" you added, the mis-wound holes would *add to* rather than *subtract from* the area, right?

That logic seems like it would be helpful for most people, and the tests give me confidence that you've implemented it correctly.

As devil's advocate, I'm trying to imagine if there's a world where people would intentionally wind their polygons this way, where the existing computation would be meaningful, and now we've inadvertently broken some special use case that these people might have... but I'm having a hard time imagining such a thing. 

On balance, I think what you've done makes sense, and is a nice guard rail, so we should merge it.

> I would appreciate another test run on your local to confirm that my epsilons are good 

Passing now — Thanks!

> (I still think it's really weird that there's a difference in behavior between our machines!).

I'm no expert on floating point implementations, but it's my understanding that there are small discrepancies across architectures and compilers. Some operations are encompassed by the IEEE754 spec, and should perform uniformly, but some operations (e.g. trigonometric functions) are not and might diverge.

I don't know if that's what's going on here, but it's common enough that if the divergence is very small, I've chosen to add a small epsilon for testing and spend my time elsewhere.

---

Overall, this LGTM, but I'll wait a couple days before merging in case anyone else wants to chime in.

(Though I think it needs a `cargo fmt` before we can merge)

bors try

phayes:
> Without the "winding correction logic" you added, the mis-wound holes would _add to_ rather than _subtract from_ the area, right?

Correct


> As devil's advocate, I'm trying to imagine if there's a world where people would intentionally wind their polygons this way, where the existing computation would be meaningful, and now we've inadvertently broken some special use case that these people might have... but I'm having a hard time imagining such a thing.

There *is* prior art for always considering the interior rings to be negative.   I've been using QGIS to double-check my numbers in my tests, and QGIS always calculates interior-rings as reducing the magnitude of the area. 

> On balance, I think what you've done makes sense, and is a nice guard rail, so we should merge it.
Yay!

I've now run `cargo fmt`

bors try

michaelkirk:
bors try

geo/src/algorithm/geodesic_area.rs
```diff
@@ -0,0 +1,667 @@
...
+    /// use geo::polygon;
+    /// use geo::Polygon;
+    ///
+    /// // Describe a polygon that covers all of the earth EXCEPT this small square.
+    /// // The outside of the polygon is in this square, the inside of the polygon is the rest of the earth.
+    /// let mut polygon: Polygon<f64> = polygon![
+    /// (x: 0.0, y: 0.0),
+    /// (x: 0.0, y: 1.0),
+    /// (x: 1.0, y: 1.0),
+    /// (x: 1.0, y: 0.0),
```
frewsxcv, line 88:
```suggestion
    ///     (x: 0.0, y: 0.0),
    ///     (x: 0.0, y: 1.0),
    ///     (x: 1.0, y: 1.0),
    ///     (x: 1.0, y: 0.0),
```

frewsxcv (approved):
lgtm! thank you for adding this 🌻

will leave this open in case michael kirk or someone else has comments

urschrei:
I think the area calculation you've got now is the correct approach – I don't think we have to cater to what would be a fundamental divergence from the spec _and_ convention on the part of a user.

urschrei (approved):

PR #989, opened 2023-03-06
prepare for geo-types-0.7.9 release

michaelkirk:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/main/CODE_OF_CONDUCT.md).
- [x] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

Anything we should wait for before releasing 0.7.9?

PR #990, opened 2023-03-06
prepare for geo-0.24.0 release

michaelkirk:
---

~~Note: this depends on #989 being released first~~ Update: released!

Anything else we should be waiting on before a geo-0.24.0 release?

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -1,6 +1,6 @@
 # Changes
 
-## unreleased
+## 0.24.0
 
 - BREAKING: Make `SimplifyVw` naming consistent
   - <https://github.com/georust/geo/pull/957>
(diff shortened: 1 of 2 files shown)


geo-types/CHANGES.md
```diff
@@ -1,6 +1,6 @@
 # Changes
 
-## unreleased
```
frewsxcv, line 3: this is the wrong changelog change

frewsxcv (approved):

geo-types/CHANGES.md
michaelkirk: Good catch! I've updated the geo CHANGES file.

(note that this PR still contains updates to the geo-types CHANGES file by virtue of it being based on #989)

michaelkirk:
I'll plan to wait on #991  (which seems straight forward) before releasing geo 0.24.0

michaelkirk:
published: https://crates.io/crates/geo/0.24.0
tagged: https://github.com/georust/geo/releases/tag/geo-0.24.0

PR #991, opened 2023-03-06
Adds `GeodesicDestination` and `GeodesicBearing`

phayes:
---

This adds the following:

1. `GeodesicDestination` trait to complement the `HaversineDestination` trait
2. `GeodesicBearing` trait to complement the `Bearing` trait
3. Adds missing documentation for the  `GeodesicIntermediate` trait

### DIFF
--- a/geo/src/algorithm/geodesic_destination.rs
+++ b/geo/src/algorithm/geodesic_destination.rs
@@ -50,7 +50,11 @@ mod test {
     fn returns_a_new_point() {
         let p_1 = Point::new(9.177789688110352, 48.776781529534965);
         let p_2 = p_1.geodesic_destination(45., 10000.);
-        assert_eq!(p_2, Point::new(9.27411867078536, 48.8403266058781));
+
+        let expected = Point::new(9.27411867078536, 48.8403266058781);
+        assert_relative_eq!(p_2.x(), expected.x(), epsilon = 1.0e-6);
+        assert_relative_eq!(p_2.y(), expected.y(), epsilon = 1.0e-6);
+
         let distance = p_1.geodesic_distance(&p_2);
         assert_relative_eq!(distance, 10000., epsilon = 1.0e-6)
     }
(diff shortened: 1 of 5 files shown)


geo/src/algorithm/mod.rs
```diff
@@ -86,6 +86,14 @@ pub use extremes::Extremes;
 pub mod frechet_distance;
 pub use frechet_distance::FrechetDistance;
 
+/// Calculate the Geodesic distance between two `Point`s.
```
michaelkirk, line 89: Is this a copy/paste misfire?

michaelkirk:
LGTM - but I'll leave it open a bit in case anyone else wants to comment.

> `GeodesicBearing` trait to complement the `Bearing` trait

Not in this pull request, but I wonder if in the future we should rename the `Bearing` to something unambiguous like `HaversineBearing`

geo/src/algorithm/mod.rs
phayes: Oops!  Fixed.

geo/src/algorithm/geodesic_destination.rs
```diff
@@ -50,7 +50,11 @@ mod test {
     fn returns_a_new_point() {
         let p_1 = Point::new(9.177789688110352, 48.776781529534965);
         let p_2 = p_1.geodesic_destination(45., 10000.);
-        assert_eq!(p_2, Point::new(9.27411867078536, 48.8403266058781));
+
+        let expected = Point::new(9.27411867078536, 48.8403266058781);
+        assert_relative_eq!(p_2.x(), expected.x(), epsilon = 1.0e-6);
```
michaelkirk, line 55: Since we've implemented some approx traits for Point, you should be able to just `assert_relative_eq(p2, expected, epsilon = 1.e-6)`.

phayes: Oh I didn't know that!  Good catch. Fixed

michaelkirk:
bors r+

I'm going to merge this, and plan to include it in geo-0.24.0, which I'll plan to publish tomorrow unless I hear otherwise.

PR #996, opened 2023-03-15
Fix #995: off by one on validity short-circuit

### LINKED ISSUE #995: Change in behavior of LineString::simplify from 0.23.1 to 0.24.0
The results of calling `simplify` on a linestring has changed from version 0.23.1 to 0.24.0 - at least sometimes, it appears to be doing nothing at all.

I have a reproduction here: https://github.com/rjmac/simplify - it just does

```rust
let unsimplified = line_string![
    (x: 0.0, y: 0.0),
    (x: 5.0, y: 4.0),
    (x: 11.0, y: 5.5),
    (x: 17.3, y: 3.2),
    (x: 27.8, y: 0.1)
];
let simplified = unsimplified.simplify(&30.0);
println!("Before: {unsimplified:?}");
println!("After: {simplified:?}");
 ```
With geo-23 it prints
```
Before: LineString([Coord { x: 0.0, y: 0.0 }, Coord { x: 5.0, y: 4.0 }, Coord { x: 11.0, y: 5.5 }, Coord { x: 17.3, y: 3.2 }, Coord { x: 27.8, y: 0.1 }])
After: LineString([Coord { x: 0.0, y: 0.0 }, Coord { x: 27.8, y: 0.1 }])
```
and with 24,
```
Before: LineString([Coord { x: 0.0, y: 0.0 }, Coord { x: 5.0, y: 4.0 }, Coord { x: 11.0, y: 5.5 }, Coord { x: 17.3, y: 3.2 }, Coord { x: 27.8, y: 0.1 }])
After: LineString([Coord { x: 0.0, y: 0.0 }, Coord { x: 5.0, y: 4.0 }, Coord { x: 11.0, y: 5.5 }, Coord { x: 17.3, y: 3.2 }, Coord { x: 27.8, y: 0.1 }])
```

This is probably related to #943 - I haven't verified it yet, but that's the only thing between the two versions that changed simplify code that I've noticed.

### PR DESCRIPTION
**Pull request by michaelkirk**
---

FIXES #995

/cc @rjmac

### DIFF
--- a/geo/src/algorithm/simplify.rs
+++ b/geo/src/algorithm/simplify.rs
@@ -32,10 +32,14 @@ where
         .enumerate()
         .map(|(idx, coord)| RdpIndex { index: idx, coord })
         .collect::<Vec<RdpIndex<T>>>();
-    compute_rdp::<T, INITIAL_MIN>(rdp_indices, &mut rdp_indices.len(), epsilon)
-        .into_iter()
-        .map(|rdpindex| rdpindex.coord)
-        .collect()
+    let mut simplified_len = rdp_indices.len();
+    let simplified_coords: Vec<_> =
+        compute_rdp::<T, INITIAL_MIN>(rdp_indices, &mut simplified_len, epsilon)
+            .into_iter()
+            .map(|rdpindex| rdpindex.coord)
+            .collect();
+    debug_assert_eq!(simplified_coords.len(), simplified_len);
+    return simplified_coords;
 }
 
 // Wrapper for the RDP algorithm, returning simplified point indices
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: geo/src/algorithm/simplify.rs [hunk 1]
```diff
@@ -293,13 +304,14 @@ mod test {
...
-        let vec = [
+        let input = [
             coord! { x: 8.0, y: 100.0 },
             coord! { x: 9.0, y: 100.0 },
             coord! { x: 12.0, y: 100.0 },
         ];
-        let simplified = rdp::<_, _, 2>(vec.into_iter(), &1.0);
-        assert_eq!(simplified, vec);
+        let actual = rdp::<_, _, 2>(input.into_iter(), &1.0);
+        let expected = [coord! { x: 8.0, y: 100.0 }, coord! { x: 12.0, y: 100.0 }];
```
michaelkirk, line 313: This test was originally added in c61abfa1e758675cc6aa9925e3e6468cf104eaa4, and changed (I think incorrectly) in 9d81bf49ed31fdfeb76728563c8236879c2e2473 as part of #943

This reverts it to it's previous behavior.

urschrei (approved):

urschrei:
Good catch @michaelkirk!

michaelkirk:
Did you want to take a look at this too @frewsxcv?

frewsxcv (approved):

frewsxcv:
Thoughts on pushing out a new bugfix release?

PR #972, opened 2023-01-23
Reimplement coord_pos_relative_to_ring using wn algo

bjornharrtell:
Doing this mostly for fun but AFAIK the winding number algo is superior to the ray-intersection algo.

PR #1004, opened 2023-04-01
Performance fix for point in polygon

b4l:
---

### DIFF
--- a/geo/src/algorithm/coordinate_position.rs
+++ b/geo/src/algorithm/coordinate_position.rs
@@ -1,5 +1,5 @@
 use crate::geometry::*;
-use crate::intersects::point_in_rect;
+use crate::intersects::value_in_between;
 use crate::kernels::*;
 use crate::{BoundingRect, HasDimensions, Intersects};
 use crate::{GeoNum, GeometryCow};
@@ -239,11 +239,6 @@ where
             return;
         }
 
-        // Ok to `unwrap` since we know `self` is non-empty, so bounding-rect is non-null
-        if !self.bounding_rect().unwrap().intersects(coord) {
-            return;
-        }
-
         match coord_pos_relative_to_ring(*coord, self.exterior()) {
             CoordPos::Outside => {}
             CoordPos::OnBoundary => {
@@ -371,21 +366,30 @@ where
     // See: https://en.wikipedia.org/wiki/Point_in_polygon#Winding_number_algorithm
     let mut winding_number = 0;
     for line in linestring.lines() {
+        // Edge Crossing Rules:
+        //   1. an upward edge includes its starting endpoint, and excludes its final endpoint;
+        //   2. a downward edge excludes its starting endpoint, and includes its final endpoint;
+        //   3. horizontal edges are excluded
+        //   4. the edge-ray intersection point must be strictly right of the coord.
         if line.start.y <= coord.y {
-            let o = T::Ker::orient2d(line.start, line.end, coord);
-            if o == Orientation::Collinear && point_in_rect(coord, line.start, line.end) {
-                return CoordPos::OnBoundary;
-            }
-            if line.end.y > coord.y && o == Orientation::CounterClockwise {
-                winding_number += 1;
-            }
+            if line.end.y >= coord.y {
+                let o = T::Ker::orient2d(line.start, line.end, coord);
+                if o == Orientation::CounterClockwise && line.end.y != coord.y {
+                    winding_number += 1
+                } else if o == Orientation::Collinear
+                    && value_in_between(coord.x, line.start.x, line.end.x)
+                {
+                    return CoordPos::OnBoundary;
+                }
+            };
         } else if line.end.y <= coord.y {
             let o = T::Ker::orient2d(line.start, line.end, coord);
-            if o == Orientation::Collinear && point_in_rect(coord, line.start, line.end) {
-                return CoordPos::OnBoundary;
-            }
             if o == Orientation::Clockwise {
-                winding_number -= 1;
+                winding_number -= 1
+            } else if o == Orientation::Collinear
+                && value_in_between(coord.x, line.start.x, line.end.x)
+            {
+                return CoordPos::OnBoundary;
             }
         }
     }
--- a/geo/benches/contains.rs
+++ b/geo/benches/contains.rs
@@ -43,14 +43,56 @@ fn criterion_benchmark(c: &mut Criterion) {
         });
     });
 
-    c.bench_function("point outside complex polygon", |bencher| {
+    c.bench_function(
+        "point outside, but within bbox, of complex polygon",
+        |bencher| {
+            let polygon = Polygon::<f64>::new(geo_test_fixtures::louisiana(), vec![]);
+            // lake borgne - near and mostly surrounded by, but not inside of, Louisiana
+            let point = point!(x: -89.641854, y: 30.026283);
+            bencher.iter(|| {
+                assert!(!criterion::black_box(&polygon).contains(criterion::black_box(&point)));
+            });
+        },
+    );
+
+    c.bench_function("point outside bbox of complex polygon", |bencher| {
         let polygon = Polygon::<f64>::new(geo_test_fixtures::louisiana(), vec![]);
-        // lake borgne - near and mostly surrounded by, but not inside of, Louisiana
-        let point = point!(x: -89.641854, y: 30.026283);
+        let point = point!(x: 2.3522, y: 48.8566);
         bencher.iter(|| {
             assert!(!criterion::black_box(&polygon).contains(criterion::black_box(&point)));
         });
     });
+
+    c.bench_function(
+        "point horizontal to comb teeth aka bart's haircut",
+        |bencher| {
+            // Testing a pathological case where the point is horizontal to lots of edges.
+            //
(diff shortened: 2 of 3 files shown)


geo/src/algorithm/coordinate_position.rs
```diff
@@ -239,11 +239,6 @@ where
             return;
         }
 
-        // Ok to `unwrap` since we know `self` is non-empty, so bounding-rect is non-null
```
michaelkirk, line 242: Since bbox checks are typically pretty quick, this was a short-circuit for when the point is "far" outside the ring. 

I honestly expected to see regressions for that case, so I just now wrote an additional bench that tests points far from their ring:
https://github.com/georust/geo/commit/efac869b7b7430bb27bb91922bacdf4bcfaa1e21

...but I was wrong! You've sped up the winding number logic so much that even computing a bbox is slower.

geo/src/algorithm/coordinate_position.rs
```diff
@@ -371,21 +366,30 @@ where
     // See: https://en.wikipedia.org/wiki/Point_in_polygon#Winding_number_algorithm
     let mut winding_number = 0;
     for line in linestring.lines() {
+        // Edge Crossing Rules:
+        //   1. an upward edge includes its starting endpoint, and excludes its final endpoint;
+        //   2. a downward edge excludes its starting endpoint, and includes its final endpoint;
+        //   3. horizontal edges are excluded
+        //   4. the edge-ray intersection point must be strictly right of the point P.
```
michaelkirk, line 373: What is "the point P" referring to in these notes?

geo/src/algorithm/coordinate_position.rs
```diff
@@ -371,21 +366,30 @@ where
...
         if line.start.y <= coord.y {
-            let o = T::Ker::orient2d(line.start, line.end, coord);
-            if o == Orientation::Collinear && point_in_rect(coord, line.start, line.end) {
-                return CoordPos::OnBoundary;
-            }
-            if line.end.y > coord.y && o == Orientation::CounterClockwise {
-                winding_number += 1;
-            }
+            if line.end.y >= coord.y {
```
michaelkirk, line 375: I'm trying to understand why these changes make things faster. 

With this new check `line.end.y >= coord.y`, we ensure we only call orient2d if a horizontal ray passing through `coord` would also pass through `line`. Checking `line.end.y >= coord.y` is much cheaper than orient2d, so we gain speed there.

But then, shouldn't we have a similar check (`if line.start.y >= coord.y {`) for the downward edge case below?

I confess that I tried it, but saw somewhere between small and no gains. I don't understand why though.

michaelkirk (approved):
Our `contains` benchmarks show some really great improvements from this!

```
point in simple polygon time:   [7.0496 ns 7.0730 ns 7.1007 ns]                                                                                                                            
                        change: [-55.196% -55.112% -55.020%] (p = 0.00 < 0.05)                                                                                                             
                        Performance has improved.                                                                                                                                          

point outside simple polygon
                        time:   [4.3600 ns 4.3647 ns 4.3701 ns]
                        change: [-23.790% -23.631% -23.451%] (p = 0.00 < 0.05)

point inside complex polygon
                        time:   [870.00 ns 871.77 ns 873.88 ns]
                        change: [-86.956% -86.892% -86.833%] (p = 0.00 < 0.05)
                        Performance has improved.

point outside complex polygon
                        time:   [876.98 ns 878.42 ns 880.08 ns]
                        change: [-85.146% -85.114% -85.082%] (p = 0.00 < 0.05)
                        Performance has improved.

line across complex polygon
                        time:   [331.40 µs 332.98 µs 334.75 µs]
                        change: [-4.7109% -3.7784% -2.8551%] (p = 0.00 < 0.05)
                        Performance has improved.

complex polygon contains polygon
                        time:   [509.71 µs 511.27 µs 512.83 µs]
                        change: [-4.8577% -4.3909% -3.9272%] (p = 0.00 < 0.05)
                        Performance has improved.
```

I confess I'm not super confident in my review of the algorithm. Some testing around edge cases, like if the ray is passing through an internal or external vertex would be nice, but I'm biased towards merging it even without that.

I'll leave it open another day or two in case anyone else wants to take a peak.

Thanks for the pr!

michaelkirk:
> Some testing around edge cases, like if the ray is passing through an internal or external vertex would be nice, but I'm biased towards merging it even without that.

Actually! It looks like we already this: see `point_in_polygon_with_ray_passing_through_a_vertex_test` and neighbors.

urschrei:
> > Some testing around edge cases, like if the ray is passing through an internal or external vertex would be nice, but I'm biased towards merging it even without that.
> 
> 
> 
> Actually! It looks like we already this: see `point_in_polygon_with_ray_passing_through_a_vertex_test` and neighbors.
> 
> 
> 
> 

I love past, diligent [whoever contrived that test]

michaelkirk:
Doing a little git archeology... @bjornharrtell is responsible for most of the current implementation, introduced in https://github.com/georust/geo/pull/972.

Maybe @bjornharrtell would like to review?

And the good tests were from @amatissart as part of https://github.com/georust/geo/pull/526

bjornharrtell:
Nice work 🙂

geo/src/algorithm/coordinate_position.rs
b4l: The point to test for intersection/location/inclusion.

bjornharrtell:
Had a second check and nice catch about the orientation checks. I don't remember why I ended up using those as they were not in the original code that I ported...

geo/src/algorithm/coordinate_position.rs
b4l: Its basically this, that for all edges with starting y coord below the ray, the orientation was calculated before checking for the line ending above the line.

> But then, shouldn't we have a similar check (`if line.start.y >= coord.y {`) for the downward edge case below?

This is implicitly given by the test in the if clause `line.start.y <= coord.y` which gives us `line.start.y > coord.y` in the else clause what matches with the edgecrossing rule 2. a downward edge excludes its starting endpoint.

b4l:
Thanks for the reviews!

Honestly the tests were what gave me confidence to open the pr in the first place.

I think the only open issue is whether to put back in the proceeding bounding rect intersection test. No preference here.

michaelkirk:
> I think the only open issue is whether to put back in the proceeding bounding rect intersection test. No preference here.

My best guess is to leave it out, since it seemed to only slow things down in all of our benchmarks. 

How'd you feel about including this bench though? I think it's an interesting case if we revisit this code in the future: https://github.com/georust/geo/commit/efac869b7b7430bb27bb91922bacdf4bcfaa1e21

geo/src/algorithm/coordinate_position.rs
michaelkirk: That seems well reasoned to me. Thank you for explaining.

michaelkirk: Just to make sure I'm understanding... this is the kind of case where we might expect a regression right?

Where the point (green) is horizontal (blue) to an inordinate number of edges in the polygon (red).

image

(forgive my graphic art skills)

geo/src/algorithm/coordinate_position.rs
michaelkirk: So can we call it "`coord`" to match with the code it's documenting - we don't use `P` anywhere else in the docs or code.

geo/src/algorithm/coordinate_position.rs
b4l: yes, exactly! we could prevent this with a bounding check, though not when the point is inside the bounds. for the latter some randomized ray direction shenanigans could help.

geo/src/algorithm/coordinate_position.rs
b4l: yes, done!

b4l:
> How'd you feel about including this bench though? I think it's an interesting case if we revisit this code in the future: [efac869](https://github.com/georust/geo/commit/efac869b7b7430bb27bb91922bacdf4bcfaa1e21)

Sure, if it helps. Performance wise your nice comb would be even better including a version with the point inside the bounds.

geo/src/algorithm/coordinate_position.rs
michaelkirk: oh duh, `line.start.y >= coord.y` is always true at that point. Thank you. =)

michaelkirk:
> Sure, if it helps. Performance wise your nice comb would be even better including a version with the point inside the bounds.

Sure, I'll throw one together right now, standby.

michaelkirk:
Ok done! https://github.com/georust/geo/compare/coord-pos...mkirk/coord-pos?expand=1

If it looks good to you, merge it into this branch, and then we can merge the whole shebang into main.

michaelkirk:
btw, here's the bench diff for that new case:

```
point horizontal to comb teeth aka bart's haircut
                        time:   [16.153 ns 16.169 ns 16.189 ns]
                        change: [+156.71% +157.26% +157.85%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

I think that's fine given the pathological nature of the input.

b4l:
@michaelkirk thanks! I hope I did this right with the merge.

michaelkirk (commented):
bors r+

Issue #1002, opened 2023-03-29
Offer `Triangulate` algorithm trait


frewsxcv:
Options:

- delaunator
- spade
- earcutr

urschrei:
Yeah: earcut is faster / offers higher-quality triangles on average (i.e. less chance of degenerate triangles), but some algorithms rely on a Delaunay triangulation, and it gives you a Voronoi tesselation "for free".

urschrei:
> Also consider this will likely add another dependency (or more), we should consider putting this behind a Cargo feature flag.

Worth noting that Delaunator-rs wouldn't add any new deps (it only requires robust), and Spade would add two (optional and smallvec).

Earcutr only adds itertools.

We can make a call, but we aren't exactly blowing up our dependency graph with any two of these.

frewsxcv:
I opened a draft PR for earcut triangulation: https://github.com/georust/geo/pull/1007

RobWalt:
I'm working on integrating [`spade`](https://docs.rs/spade/latest/spade/index.html) now in https://github.com/georust/geo/pull/1083

doc-E-brown:
If it is helpful I'm preparing a delaunay triangulation trait in https://github.com/franklin-ai/geo/blob/8bae5523f7f7aeba0e8b8d11d9159a6a3c6bf0c3/geo/src/algorithm/triangulate_delaunay.rs which I'll be preparing for an upstream PR soon.  We've had an internet need for this trait.

RobWalt:
@doc-E-brown looks cool! I'll give it a look. Note that we just integrated `spade` into master which also offers that functionality as well as constrained delaunay triangulations. Nevertheless your code looks promising as it seemingly doesn't need external dependencies and is also rather concise. Do you plan on working on more in that direction? If not I might take over the work of extending this at some point as I think it's quiet interesting and important! I'm looking forward to the PR! ❤️

frewsxcv:
Both are implementations of Delauney triangulation right? I think it's worth doing a perf comparison of the implementations

doc-E-brown:
[...]

PR #1007, opened 2023-04-03
Add `TriangulateEarcut` algorithm.

frewsxcv:
---

https://github.com/georust/geo/issues/1002

### DIFF
--- a/geo/src/algorithm/triangulate_earcut.rs
+++ b/geo/src/algorithm/triangulate_earcut.rs
@@ -133,7 +133,7 @@ impl<T: CoordFloat> TriangulateEarcut<T> for Polygon<T> {
 /// The raw result of triangulating a polygon from `earcutr`.
 #[derive(Debug, PartialEq, Clone)]
 pub struct Raw<T: CoordFloat> {
-    /// One-dimensional vector of polygon vertices (in XY order).
+    /// Flattened one-dimensional vector of polygon vertices (in XY order).
     pub vertices: Vec<T>,
 
     /// Indices of the triangles within the vertices vector.
--- a/geo/src/algorithm/triangulate_earcut.rs
+++ b/geo/src/algorithm/triangulate_earcut.rs
@@ -1,0 +1,259 @@
+use crate::{coord, CoordFloat, CoordsIter, Polygon, Triangle};
+
+/// Triangulate polygons using an [ear-cutting algorithm](https://www.geometrictools.com/Documentation/TriangulationByEarClipping.pdf).
+pub trait TriangulateEarcut<T: CoordFloat> {
+    /// # Examples
+    ///
+    /// ```
+    /// use geo::{coord, polygon, Triangle, TriangulateEarcut};
+    ///
+    /// let square_polygon = polygon![
+    ///     (x: 0., y: 0.), // SW
+    ///     (x: 10., y: 0.), // SE
+    ///     (x: 10., y: 10.), // NE
+    ///     (x: 0., y: 10.), // NW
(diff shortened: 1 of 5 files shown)


geo/src/algorithm/triangulate_earcut.rs
```diff
@@ -0,0 +1,259 @@
...
+        Raw {
+            vertices: input.vertexes,
+            triangle_indices,
+        }
+    }
+}
+
+/// The raw result of triangulating a polygon from `earcutr`.
+#[derive(Debug, PartialEq, Clone)]
+pub struct Raw<T: CoordFloat> {
```
michaelkirk, line 135: Since this is public, I feel like a more descriptive name is in order. Maybe `Triangulation` ?

michaelkirk (approved):
This looks great. 

I left a bunch of small suggestions that I hoped would make it a little more intuitive to use, but I don't feel strongly about any of it, so please only apply what resonates with you.

michaelkirk (approved):
This looks great. 

I left a bunch of small suggestions that I hoped would make it a little more intuitive to use, but I don't feel strongly about any of it, so please only apply what resonates with you.

geo/src/algorithm/triangulate_earcut.rs
urschrei: Perhaps worth bearing in mind that if we went with Triangulation it would also have to be suitable for any future triangulations (e.g. Delaunay) we might add.

michaelkirk: This is name spaced to the `triangulate_earcut::Triangulation` module (v.s. the hypothetical `delaunay::Triangulation`), so I think it'd be fine.

That said, I'm also fine with the clarity/verboseness tradeoff of `triangulate_earcut::EarcutTriangulation`

I just think `Raw` is a bit inexplicable.

frewsxcv: There is a case to be made for `triangulate_earcut::Triangulation`, but I wonder if we should have a larger discussion about naming and namespacing to be consistent with our other traits. For example we also don't do `geo::euclidean_distance::Distance` nor `geo::geodeisc_length::Length`.

michaelkirk: > geo::euclidean_distance::Distance nor geo::geodeisc_length::Length

I don't understand the comparison to EuclideanDistance and GeodesicLength which both return scalars.  We don't currently have bespoke types for those operations. 

We *could* introduce something like:
```diff
 mod euclidean_distance {
+    struct Distance<T: CoordNum>(T);
     trait EuclideanDistance<T> {
-         fn euclidean_distance(&self) -> T;
+         fn euclidean_distance(&self) -> Distance<T>;
     }
 }
```
Is that what you mean? I don't see the value in it, but maybe I'm missing something.

In the case of this triangulation module, we already *have* this bespoke type: `Raw`, I'm saying we should give it a more descriptive name than "Raw".

frewsxcv: Disregard my previous comment! I interpreted your message:

> This is namespaced to the module, so `triangulate_earcut::Triangulation` ...

to mean that you were suggesting renaming the `TriangulateEarcut` trait to `Triangulation` and forgot this was a discussion about the struct that is being returned.

michaelkirk: Ah, right! Sorry for not being clearer from the get-go.  I probably should have just led with:

```suggestion
pub struct Triangulation<T: CoordFloat> {
```

frewsxcv:
bors r=michaelkirk

Issue #67, opened 2016-09-18
Introduce trait for accessing geometries


pka:
In the course of [refactoring rust-postgis](https://github.com/andelf/rust-postgis/pull/9) to use rust-geo geometries, I'm looking at needed conversions between different georust geometries. One goal would be to store a rust-gdal geometry with rust-postgis as efficient as possible.

One possible solution would be using iterators returning a Point trait instead of accessing the structs directly. If rust-gdal would implement this trait, storing a geometry with rust-postgis could be done without any conversion. And if the geo algorithms would use this trait, they could be applied directly on geometries implementing this trait. It should also solve the ownership problem in #21.

A simplified trait could be

```
pub trait PointType {
    fn x(&self) -> f64;
    fn y(&self) -> f64;
    //...
}
```

When this trait is implemented for `geo::Point` and `geo::LineString` implements `IntoIterator` with `Item=&PointType`, you can iterate like

```
let xsum = linestring.into_iter().fold(0., |sum, p| sum + p.x());
```

frewsxcv:
Sorry, didn't see this until today. I'll need a day or two to think this through but I think this is a good idea. I'm not renting sure how this might work for something like `LineStringType`. I'll try you respond again tomorrow after I've played around with this idea.

pka:
My current experiments look like this:

```rust
pub trait PointType {
    fn x(&self) -> f64;
    fn y(&self) -> f64;
    //...
}

pub trait LineStringType<'a> {
    fn points(self) -> PointRefIter<'a>;
}

pub trait MultiLineStringType<'a> {
    fn lines(self) -> LineStringRefIter<'a>;
}
```

I got the point iterator working, but not yet the line iterator.

```rust
let sumx = linestring.points().fold(0., |sum, p| sum + p.x());
```

frewsxcv:
I played around with this idea today: https://is.gd/tqkzBp

I *think* this is a good idea. If you have any feedback for my link above, let me know. Commented out section was something I tried, but I couldn't get the typechecker to behave. Will have to do some more thinking.

pka:
I made some progress implementing the following traits:

```rust
pub trait Point {
    fn x(&self) -> f64;
    fn y(&self) -> f64;
    //...
}

/// Iterator for points of a line
pub struct Points<'a, T: 'a + Point>
{
    pub iter: Iter<'a, T>
}

pub trait LineString<'a> {
    type PointType: Point;

[...]

PR #1011, opened 2023-04-20
Initial geo-traits implementation

kylebarron:
---

As discussed on the discord, in https://github.com/georust/geo/discussions/838, and originally in https://github.com/georust/geo/issues/67, this is an initial bare-bones implementation of geometry access traits for discussion. As @frewsxcv suggested, for now this is a separate crate in the workspace to let us iterate on some ideas.

#### Minimize cloning in the geo-types implementation

I got lost in lifetime errors so to get things to compile I added cloning to the geo-types implementations. I.e. for `MultiPoint` each point is cloned when accessed.


geo-traits/src/coord.rs
```diff
@@ -0,0 +1,78 @@
...
+    type T: CoordNum + Send + Sync;
+
+    /// x component of this coord
+    fn x(&self) -> Self::T;
+
+    /// y component of this coord
+    fn y(&self) -> Self::T;
+
+    /// Returns a tuple that contains the x/horizontal & y/vertical component of the coord.
+    fn x_y(&self) -> (Self::T, Self::T);
```
frewsxcv, line 13: Do you think it's reasonable to have a default implementation for this so implementers don't need to implement it?

geo-traits/src/line_string.rs
```diff
@@ -0,0 +1,54 @@
...
+use super::point::PointTrait;
+use std::iter::Cloned;
+use std::slice::Iter;
+
+pub trait LineStringTrait<'a>: Send + Sync {
+    type ItemType: 'a + PointTrait;
+    type Iter: Iterator<Item = Self::ItemType>;
+
+    /// An iterator over the points in this LineString
+    fn points(&'a self) -> Self::Iter;
```
frewsxcv, line 12: Thoughts on calling this (and the other methods) `coords`? Since, at least in the `geo` world, a `LineString` is made up of "coordinates" and not "points".

kylebarron: Makes sense. I originally didn't have a `CoordTrait` and so had these named `point*` and having the item be an instance of `PointTrait` instead of `CoordTrait`

geo-traits/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "geo-traits"
+version = "0.1.0"
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/georust/geo"
+documentation = "https://docs.rs/geo-traits/"
+readme = "../README.md"
+keywords = ["gis", "geo", "geography", "geospatial"]
+description = "Geospatial traits"
+rust-version = "1.65"
```
kylebarron, line 10: Bumped this to 1.65 for this crate because the traits use GATs, which came in 1.65

michaelkirk:
Thank you for moving the traits discussion forward with some concrete code! I think these proposed trait implementations looks reasonable in the abstract. I'd love to see how some algorithms can be implemented in terms of these traits, and what kind of changes/concessions/benefits arise when rubber meets the road.

I think seeing how they can be used will be a necessary measure of how "good" these trait definitions are.

geo-traits/src/geometry.rs
```diff
@@ -0,0 +1,87 @@
...
+    MY: 'a + MultiPolygonTrait<'a>,
+    GC: 'a + GeometryCollectionTrait<'a>,
+{
+    Point(&'a P),
+    LineString(&'a L),
+    Polygon(&'a Y),
+    MultiPoint(&'a MP),
+    MultiLineString(&'a ML),
+    MultiPolygon(&'a MY),
+    GeometryCollection(&'a GC),
```
frewsxcv, line 52: Do you have any thoughts on adding the other types offered in `geo-types`? Namely `Triangle`, `Line`, and `Rect`

geo-traits/src/line_string.rs
```diff
@@ -0,0 +1,54 @@
+use geo_types::{Coord, CoordNum, LineString};
+
+use super::CoordTrait;
+use std::iter::Cloned;
+use std::slice::Iter;
+
+pub trait LineStringTrait<'a>: Send + Sync {
+    type ItemType: 'a + CoordTrait;
+    type Iter: Iterator<Item = Self::ItemType>;
```
frewsxcv, line 9:
```suggestion
    type Iter: ExactSizeIterator<Item = Self::ItemType>;
```

Since we already require the exact count, I wonder if we should just do `ExactSizeIterator` here

geo-traits/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+edition = "2021"
+
+[features]
+
+[dependencies]
+geo-types = "0.7.9"
```
frewsxcv, line 16: What do you think about making `geo-types` optional? In theory if someone wanted to just implement `geo-traits` for their own geometry types, they shouldn't need access to `geo-types`, I think

frewsxcv: Hmm actually this depends on `CoordTrait`. Maybe that should live in this crate someday 🤔

geo-traits/src/polygon.rs
```diff
@@ -0,0 +1,64 @@
+use super::line_string::LineStringTrait;
+use geo_types::{CoordNum, LineString, Polygon};
+use std::iter::Cloned;
+use std::slice::Iter;
+
+pub trait PolygonTrait<'a>: Send + Sync {
+    type ItemType: 'a + LineStringTrait<'a>;
```
frewsxcv, line 7: Maybe `RingType` would be more descriptive? 🤷🏻

geo-traits/src/polygon.rs
```diff
@@ -0,0 +1,64 @@
+use super::line_string::LineStringTrait;
+use geo_types::{CoordNum, LineString, Polygon};
+use std::iter::Cloned;
+use std::slice::Iter;
+
+pub trait PolygonTrait<'a>: Send + Sync {
```
frewsxcv, line 6: I wonder if we'll want to specify guarantees around winding order for implementors.

geo-traits/src/line_string.rs
frewsxcv: In fact if we did this, we probably wouldn't need a `num_coords` method

geo-traits/src/geometry.rs
kylebarron: Definitely something to discuss. I didn't implement them yet because I wasn't sure the pros/cons of how hard that might be for some data structures to implement vs the usefulness in adapting for geo algorithms. If a data structure doesn't natively have a `Rect` type, should it allow a polygon to fill in as a `Rect`?

geo-traits/src/line_string.rs
kylebarron: I updated to use `ExactSizeIterator` but didn't yet remove `num_coords`

geo-traits/Cargo.toml
kylebarron: Yeah I think making `geo-types` optional definitely makes sense but in this initial implementation I made a hard dependency for simplicity

geo-traits/src/polygon.rs
kylebarron: Should we change the name `ItemType` in each trait then? I.e. `CoordType` in `LineStringTrait`?

geo-traits/src/polygon.rs
kylebarron: I don't know... some stores might not have that information? I think GEOS objects don't have a known winding order?

kylebarron:
> I'd love to see how some algorithms can be implemented in terms of these traits, and what kind of changes/concessions/benefits arise when rubber meets the road.

I agree things will become clearer with code implementation. Should I make a new stacked PR to iterate or add edits to this PR?

frewsxcv:
My vote would be to merge as-is, and then iterate on it with subsequent PRs. Then it's easier for others to collaborate too

kylebarron:
Any changes I should make in this PR?

frewsxcv:
Let's wait to hear what @michaelkirk thinks! If he approves of the plan then yeah we can merge it

michaelkirk:
My concern is that we may be putting the cart before the horse a bit. My understanding of this work is that ultimately the purpose of it is to be able to define algorithms in terms of these traits. geo-types would be one implementor, but people might want to bring their own types which are better suited for their use cases for whatever reasons, and write their own `impl geo_traits::* for MyGeometryType` - like an Arrow backed geometry. Right?

I don't think we need to come up with a *perfect and complete* design before merging. I'm saying "show me that the code you want to merge works at all". Yes it compiles, but if the point of these traits is to be able to implement algorithms, then lets see what it looks like to implement an algorithm with them.

My hope is that if this is a reasonably good design for geometry traits, that this will not be an unreasonably hard task. If it turns out to be hard to do, then maybe this is not a good design for a geometry traits and we all will learn something.

geo-traits/src/coord.rs
```diff
@@ -0,0 +1,64 @@
+use geo_types::{Coord, CoordNum, Point};
```
michaelkirk, line 1: 🧑‍🚒 hot take: It was a mistake to have `geo_types::Coordinate` separate from `geo_types::Point` they are redundant, and I'm guessing this only happened because of ideas copied over from Java where a "light weight" representation required splitting the two concepts.

Rust doesn't have the overhead of "primitive type" vs "object". Regardless of how many methods or whatever we add to it, `geo-types::Coordinate<f64>` will always be the same size as `[f64; 2]`.

The cost of our current approach is frequent confusion about whether a method can/should accept a Coordinate, a Point, or both.

Maybe geo-traits could be an opportunity to fix that wart. We could delete `CoordTrait`, and have only `PointTrait`. *Both* `geo-types::Coordinate` *and* `geo_types::Point` would implement `PointTrait`, allowing any methods defined in terms of `PointTrait` to accept both, and finally resolving the question "Should this be a Point or a Coord?"

This is a half baked armchair suggestion, so no changes demanded, but think about it!

In fact, this might be a relatively nice way to start showing the value of a trait system proposal - by simplifying some of our existing APIs.

michaelkirk: Also, no judgement on the original `geo_types::Coordinate` vs `geo_types::Point` decision. It's gotten us pretty far, and it seemed totally reasonable to me at first too, until years of stubbing my toe on it.

frewsxcv:
@michaelkirk For the purposes of a proof-of-concept, do you see the implementations you desire living in `geo` or `geo-traits`?

michaelkirk:
I was assuming our existing algorithms in `geo` would be modified to work in terms of geo-traits where possible.

Did you have something else in mind?

michaelkirk:
> For the purposes of a proof-of-concept

I guess I'm a little uncertain about the usage of "proof of concept". I think that if there is code in georust/geo we should expect that people will use it. 

If we want a place to experiment with crazy stuff that we don't want people to use I don't think it should live in this repository.

frewsxcv:
> I think that if there is code in georust/geo we should expect that people will use it.

Curious where this concern comes from! And how them using unpublished code would affect the lives of us as maintainers. As long as we are unpaid volunteers supporting their efforts, we don't owe them any guarantees about how we run our work. Unlike many technical decisions, I feel pretty strongly about unpaid labor haha

frewsxcv:
> I was assuming our existing algorithms in `geo` would be modified to work in terms of geo-traits where possible.
> 
> Did you have something else in mind?

I ask because if they live in `geo` then that muddies the water in terms of separating our draft code and keeping everything in the `geo` directory deployable.

geo-traits/src/coord.rs
kylebarron: I'd be fully supportive of making `Coord` and `Point` both implement a shared `PointTrait`! I admit I do get confused when to use coord vs point.

michaelkirk:
I agree that we should keep `geo/` deployable at all times (modulo a version bump in Cargo.toml or whatever).

Can you elaborate on how you'd like to see geo-traits development proceed out @frewsxcv? All I can tell is that you seem to not be quite on board with my vision for it. 😅

frewsxcv:
I see us iterating on it in a separate crate directory akin to the other crate directories. If a pull request change to `geo-traits` is an improvement, we approve it and merge it in. There's no need to increase the barrier to contribution with a separate repository or ensuring the full vision is there. We can do that when we're ready to publish.

b4l:
Alternatively, we could agree to work on a `traits` branch analogous to a development branch for a while until we figure out if it is the way to go.  There we would also have the liberty to directly refactor the existing algorithms and put the code in the right place and have a nice diff.
Mostly working with 3d data nowadays and the stalled effort to introduce another coordinate dimension in `geo-types` prevents me from using it which I find quite unfortunate.
Just providing options here and hope we can find space to be bold at times and traits may as well offer some possibility to support somehow arbitrary dimension support.

michaelkirk:
Another concern I have is less objective, but subjectively, I suspect my position as a maintainer gets easier (and thus more delightful) when there are clear delineations between "this code is thought to work" vs. "this code is a prototype, and I'm not sure to what extent it is intended to work". I'm willing to accept that this is an idiosyncrasy founded on my own personal trauma doing open source development on The Internet  (🤣😭) and may not be relevant to other people, but it's still my preference. 

> I see us iterating on it in a separate crate directory akin to the other crate directories.

Ok, I'm with you here, but more specifically, if we merged these trait definitions, what's next? I'm assuming implementing some algorithms with these traits would be the next step (right?).

So how would we go about that? My best guess is you mean we'd be duplicating the existing algorithms from geo into geo-traits and editing that copy to work with geo-traits instead of geo-types? Or did you intend something else?

> we could agree to work on a `traits` branch analogous to a development branch for a while until we figure out if it is the way to go. 

(Thanks for chiming in @b4l!)

I think long running branches can be fraught, typically because they can get out of sync, but I do think it's preferable to duplicating/re-implementing the algorithms in geo-traits. At least with a long running branch, version control can give us some hints as to where conflicts arise (and hopefully it won't be *that* long running anyway, right? 😬). The alternative of duplicating the implementations means we are relying on our own fragile minds to reason about/remember when conflicts occur.

> There we would also have the liberty to directly refactor the existing algorithms and put the code in the right place and have a nice diff.

👍👍 Having a diff against the current geo algorithm implementation is the primary thing I'm looking for when trying to evaluate how "good" these trait implementations are. 

---

What do other people think about merging traits work into a `traits` branch until we're comfortable releasing something usable and relevant to geo? 

I'd especially like to hear from @kylebarron, who seems most affected by this decision in the near term.

kylebarron:
I'm happy to have this merge onto a `traits` branch, and ongoing PRs can have that as the base branch. We could also have an ongoing open draft PR from the `traits` branch to the `main` branch to make it easy for people to see and comment on the state of the two branches at large

rmanoka:
This an excellent contribution; thanks @kylebarron !   I'm also quite in favour of exploring the trait-based approach for `geo`.   Specifically w.r. to this PR, I would also prefer we merge this as-is/quickly into a separate branch until we identify a plan to switch to it (either incrementally or in one release).  This would help have smaller PRs to help push this out.

Particularly we should try to understand:

1.  Port effort.   Is the changes needed per algo. straight-forward or needs knowledge of algo?  Do we need to port all algos in one release or can it be done incrementally?
2. Dependencies.  geo-traits abstracts over geo-types, and hence the new dependency chain is: geo -> geo-traits -> geo-types.  In the long run, geo won't even directly depend on geo-types.  But, are there any dependency issues to understand?  Will every breaking bump of geo-traits need bump of geo?  This is fine, but would we then have to impl. the algos for different versions of geo-traits via complicated macros?
3. Performance.  Would be good to benchmark some algos on this approach to ensure the compiler understands that we're just wrapping around [f64; 2] or similar.

Pls. add more, and we could track this in the PR that merges the `traits` branch to `main`.  Reg. @michaelkirk valid concern on the feature branch diverging:  since the traits are just wrapping types, at least until we finalize the details of the traits, only changes in geo-types should affect the difference between the branches.  So, may be worth the risk?

urschrei:
(apologies for the drive-by comment)
I also prefer the idea of a separate branch for now. Yes, there's some overhead to keeping it in sync, but as has been pointed out it hopefully won't be too long-running.

frewsxcv (approved):

frewsxcv:
I just updated the base branch of this pull request to be a new `traits` branch. Anyone else have blockers before merging?

kylebarron:
Seems fine on my side

frewsxcv:
bors r+

🎉 🎉 🎉 🎉 🎉

michaelkirk:
🥳
Thanks for your flexibility @frewsxcv.

PR #1015, opened 2023-05-19
Implement haversine distance from a point to a bounded line, further step to implementing #834

mbcltd:
---

### DIFF
--- a/geo/src/algorithm/haversine_distance_to_line.rs
+++ b/geo/src/algorithm/haversine_distance_to_line.rs
@@ -13,9 +13,9 @@ impl<T: CoordFloat + FromPrimitive> HaversineDistanceToLine<T, Line<T>> for Poin
         let lines_start = Point::from(line.start);
         let lines_end = Point::from(line.end);
 
-        if is_inner_angle_between_points_obtuse(&self, &lines_start, &lines_end) {
+        if is_inner_angle_between_points_obtuse(self, &lines_start, &lines_end) {
             self.haversine_distance(&lines_start)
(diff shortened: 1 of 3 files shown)


geo/src/algorithm/mod.rs
```diff
@@ -60,10 +60,14 @@ pub use convert::{Convert, TryConvert};
 pub mod convex_hull;
 pub use convex_hull::ConvexHull;
 
-/// Cross track distance
+/// Determine the shortest distance between this point and a line which passes through line_point_a and line_point_b
```
michaelkirk, line 63:
```suggestion
/// Determine the shortest distance between this point and a great circle which passes through line_point_a and line_point_b
```

clearer?

geo/src/algorithm/haversine_distance_to_line.rs
```diff
@@ -0,0 +1,122 @@
+use num_traits::FromPrimitive;
+
+use geo_types::{CoordFloat, Line, Point};
+
+use crate::{CrossTrackDistance, HaversineBearing, HaversineDistance};
+
+pub trait HaversineDistanceToLine<T, Rhs> {
```
michaelkirk, line 7: My initial impulse would be to hang this implementation on the existing `HaversineDistance` trait instead of introducing a new trait. Is there a reason it's better to introduce a new trait like this?

frewsxcv:
@mbcltd Thank you for this pull request! This does get us one step closer to implementing haversine for all geometry types which is exciting. Whenever you have a chance could you reply to @michaelkirk's comments above?

frewsxcv:
If you have capacity to pick this up again feel free to reopen

PR #1019, opened 2023-06-21
Add new `geo-traits` crate

frewsxcv:
---

See https://github.com/georust/geo/pull/1011. We can use this draft pull request to view the changes from the work-in-progress `traits` branch.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,6 +26,7 @@ jobs:
       - geo
       - geo_postgis
       - geo_fuzz
+      - geo_traits
       - bench
     steps:
       - name: Mark the job as a success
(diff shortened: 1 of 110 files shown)


RobWalt:
Just a comment: I think the name of the crate is a bit confusing. Since `geo` relies on the `num` crate quiet heavily I thought the new `geo-traits` crate was similar to the `num-traits` crate which separates all the traits one can find in `num` nicely so we don't have to pull in all of `num` everywhere.

However, it seems that `geo-traits` implements entirely new traits instead.

I was just looking for a lightweight way to get the `HasKernel` trait which is required for a data type I'm defining in a separate crate for use everywhere else

frewsxcv:
Closing in favor of https://github.com/georust/geo/pull/1157

PR #1020, opened 2023-06-21
Implement `From<&Line>` for `LineString`

gibbz00:
---

Manual implementation tends to be implemented inline (e.g https://github.com/frewsxcv/geo-bevy/blob/348085c43b58da8f321e04e77fa9bf7d624e60b4/src/lib.rs#L147) as `From<Line>` is possibly more restrictive than necessary. Not sure whether `From<Line>` uses copy over move even if it's an owned parameter.

### DIFF
--- a/geo-types/CHANGES.md
+++ b/geo-types/CHANGES.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Add new changes here.
+* Implement `From<&Line>` for `LineString`
 
 ## 0.7.9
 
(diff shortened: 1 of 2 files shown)


michaelkirk (approved):
Thanks! 

> Not sure whether From<Line> uses copy over move even if it's an owned parameter.

My guess is that they are equivalent in this case. A move *is* a copy after all, but we expect the compiler to optimize it out in most cases.

This LGTM, but I'll leave it open for a couple days in case anyone else wants to weigh in.

frewsxcv (approved):

lnicola (approved):

PR #1021, opened 2023-06-23
WIP Implement area functions on geometry traits

kylebarron:
### Change list

- Remove `Send + Sync` restriction on `T` on the trait. I don't know whether this restriction makes sense here. I originally added it because rust-postgis [included it](https://github.com/andelf/rust-postgis/blob/f25925fae51050000df8eb66af3ba554949f94b5/src/types.rs). But it appeared that `CoordFloat` doesn't include those trait bounds so it was easier in this context to remove them.
- Add fully parametrized `T: CoordNum` on each trait, and define other item types in terms of `T`. I needed to do this so that some algorithms could enforce `CoordFloat` and not just `CoordNum`. 
    I'm not sure if `type ItemType = LineString<Self::T>;` on the `impl` is necessary instead of `type ItemType = LineString<T>;`.
- Implement pure functions for most combinations of geometry type + signed/unsigned area.
- Drop 1.63 CI from this branch

### Questions:

- Is there a simple way with lifetimes to fix this `temporary value dropped while borrowed` error? It's on [this line](https://github.com/kylebarron/geo/blob/fd62519c82551e408c0b654408d17ccb64b7d1a0/geo/src/algorithm/area.rs#L226). I'm not sure if this is an issue with the trait design or implementation or with my code

### DIFF
--- a/geo-traits/src/geometry.rs
+++ b/geo-traits/src/geometry.rs
@@ -9,7 +9,7 @@ use super::{
 };
 
 #[allow(clippy::type_complexity)]
-pub trait GeometryTrait<'a>: Send + Sync {
+pub trait GeometryTrait<'a> {
     type Point: 'a + PointTrait;
     type LineString: 'a + LineStringTrait<'a>;
     type Polygon: 'a + PolygonTrait<'a>;
@@ -52,7 +52,7 @@ where
     GeometryCollection(&'a GC),
 }
 
-impl<'a, T: CoordNum + Send + Sync + 'a> GeometryTrait<'a> for Geometry<T> {
+impl<'a, T: CoordNum + 'a> GeometryTrait<'a> for Geometry<T> {
     type Point = Point<T>;
     type LineString = LineString<T>;
     type Polygon = Polygon<T>;
--- a/geo-traits/src/geometry.rs
+++ b/geo-traits/src/geometry.rs
@@ -1,5 +1,3 @@
-use std::ops::SubAssign;
-
 use geo_types::{
     CoordNum, Geometry, GeometryCollection, LineString, MultiLineString, MultiPoint, MultiPolygon,
     Point, Polygon,
@@ -54,7 +52,7 @@ where
     GeometryCollection(&'a GC),
 }
 
-impl<'a, T: CoordNum + Send + Sync + SubAssign + 'a> GeometryTrait<'a> for Geometry<T> {
+impl<'a, T: CoordNum + Send + Sync + 'a> GeometryTrait<'a> for Geometry<T> {
     type Point = Point<T>;
     type LineString = LineString<T>;
     type Polygon = Polygon<T>;
--- a/geo-traits/src/geometry.rs
+++ b/geo-traits/src/geometry.rs
@@ -1,3 +1,5 @@
+use std::ops::SubAssign;
+
 use geo_types::{
     CoordNum, Geometry, GeometryCollection, LineString, MultiLineString, MultiPoint, MultiPolygon,
     Point, Polygon,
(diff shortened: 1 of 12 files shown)


frewsxcv:
> Is there a simple way with lifetimes to fix this temporary value dropped while borrowed error? It's on [this line](https://github.com/kylebarron/geo/blob/fd62519c82551e408c0b654408d17ccb64b7d1a0/geo/src/algorithm/area.rs#L226). I'm not sure if this is an issue with the trait design or implementation or with my code

When I run into situations like this, I typically start incrementally loosening lifetime bounds that don't need to be explicitly specified. With these changes, your code compiles: https://gist.github.com/frewsxcv/75e719227b889ee2930ccb59930b26de

geo/src/algorithm/area.rs
```diff
@@ -27,12 +35,25 @@ where
...
     let mut tmp = T::zero();
-    for line in linestring.lines() {
-        use crate::MapCoords;
-        let line = line.map_coords(|c| c - shift);
+    for i in 0..linestring.num_coords() - 1 {
+        let mut c1 = Coord {
+            x: linestring.coord(i).unwrap().x(),
+            y: linestring.coord(i).unwrap().y(),
+        };
+        c1.x = c1.x - shift.x();
```
kylebarron, line 46: I switched these to not use `-=` to avoid having to require the `SubAssign` type on `T`. Maybe requiring `SubAssign` and similar is such a low bar that we should require it?

kylebarron:
> When I run into situations like this, I typically start incrementally loosening lifetime bounds that don't need to be explicitly specified. With these changes, your code compiles: [gist.github.com/frewsxcv/75e719227b889ee2930ccb59930b26de](https://gist.github.com/frewsxcv/75e719227b889ee2930ccb59930b26de)

Thanks! That was helpful! I was able to make some progress. But it seems like the trait methods that really need `&'a self` are the iterators. 

So I can make progress by removing `'a` everywhere but that forces me to write ugly code and avoid using the iterators, doing stuff like

```rs
for polygon_idx in 0..geom.num_polygons() {
```

Especially with the `GeometryCollection` and `Geometry` traits, it gets complex and it's not clear to me we can fully remove the lifetimes.

I tried to implement `unsigned_area_geometry` which takes anything that implements `GeometryTrait`, but there `as_type` requires `'a`.

geo-traits/src/geometry.rs
```diff
@@ -9,14 +9,15 @@ use super::{
...
-    type LineString: 'a + LineStringTrait<'a>;
-    type Polygon: 'a + PolygonTrait<'a>;
-    type MultiPoint: 'a + MultiPointTrait<'a>;
-    type MultiLineString: 'a + MultiLineStringTrait<'a>;
-    type MultiPolygon: 'a + MultiPolygonTrait<'a>;
-    type GeometryCollection: 'a + GeometryCollectionTrait<'a>;
+pub trait GeometryTrait<'a> {
+    type T: CoordNum;
+    type Point: 'a + PointTrait<T = Self::T>;
+    type LineString: 'a + LineStringTrait<'a, T = Self::T>;
```
frewsxcv, line 15: This can now be:

```rust
type LineString<'a>: 'a + LineStringTrait<'a, T = Self::T>;
```

Or something like that

kylebarron: What's the benefit to adding `<'a>` onto the type name as well?

frewsxcv: You can remove the lifetime from the trait itself, which means any references to the trait don't need to specify that lifetime. Similar to https://github.com/georust/geo/pull/908/files

kylebarron: Oh wow that is super nice!

kylebarron: It ended up being a lot more work than expected but I implemented that change in https://github.com/geoarrow/geoarrow-rs/pull/253 (I'm incubating and testing out trait impls there); I'll make a PR to the traits branch with those updated changed

frewsxcv:
From what I understand, https://github.com/georust/geo/pull/1115 is a newer version of this so I'm going to close this one. Feel free to reopen if this is wrong

PR #1024, opened 2023-06-26
geo-types: Weaken `serde/std` and `approx/std` features

w-flo:
Previously, if I wanted only the `std` feature of geo_types, this would enable the `serde` and `approx` features, too. Change these features to be "weak dependency features", so they're only enabled (and their crates pulled in) if the corresponding optional crate is already pulled in for other reasons.

---
---
I'm always eager to optimize my build process, and this would help with that :-)

But I'm not sure if this could potentially be a breaking change? Ideally, it would be a purely internal change, but I'm not sure if it could cause issues for downstream crates / users. And I'm not sure if a CHANGES.md entry would be needed / what it should say?

### DIFF
--- a/geo-types/Cargo.toml
+++ b/geo-types/Cargo.toml
@@ -12,7 +12,7 @@ edition = "2021"
 
 [features]
 default = ["std"]
-std = ["approx/std", "num-traits/std", "serde/std"]
+std = ["approx?/std", "num-traits/std", "serde?/std"]
 # Prefer `use-rstar` feature rather than enabling rstar directly.
 # rstar integration relies on the optional approx crate, but implicit features cannot yet enable other features.
 # See: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#namespaced-features


michaelkirk:
Good catch! 

[We added the "std" feature](https://github.com/georust/geo/pull/936) in an attempt to support no-std builds, and in the process it looks like we inadvertently pulled some existing disabled-by-default features into the default build. Is that right?

I fear we might have made the same mistake with some of our other no_std crates.

w-flo:
Oh, yes! I hadn't seen that commit/PR. I believe you're right, that's where the `approx` and `serde` crates started to be pulled in by default because their corresponding features were not marked as "weak" using the "?" syntax. Now you can only drop them by disabling the `std` feature. The `std` feature itself is a great addition though.

Note that the ["?" cargo syntax](https://doc.rust-lang.org/nightly/cargo/reference/features.html#dependency-features) was stabilized in Rust 1.60, so it should be OK for `geo_types` (MSRV 1.63), but maybe some other georust crates use MSRV < 1.60?

I only use `geo_types` and `gpx`, and `gpx` does not appear to have this issue (but two other issues https://github.com/georust/gpx/pull/91 )

frewsxcv (approved):

lnicola:
bors r=frewsxcv,lnicola

PR #1025, opened 2023-06-27
Vector2DOps Trait - Proposal

thehappycheese:
---

Proposal to define `Vector2DOps` trait and implement it for the `Coord` struct.
Intention is to facilitate general purpose linear algebra operations such as 'dot product' etc all in one place.

Most of these functions are are already implemented in various places throughout the crate or sometimes hard-coded in place. This would bring them all together.

For example

- `wedge_product`( aka cross/perp/exterior product) `geotypes::Line::determinant`,
  - `geotypes::Point::cross_prod` (which takes three arguments for some reason) etc.
  - Also used in the default implementation of Kernel orient2d, although this serves a different purpose as it returns an enum
- `dot_product` `geotypes::Point::dot`,
- `magnitude` in `Line::line_euclidean_length`
- `magnitude_squared` hard coded ( [here](https://github.com/thehappycheese/geo/blob/9b477fc7b24fcb8731bd5e01d559b613e79ada0c/geo/src/algorithm/line_locate_point.rs#L61C9-L61C29))
- `normalize` ... interestingly it is hard to find an example of this operation in the existing crate. Possibly because it is generally best avoided due to floating point instability.

For more discussion about motivation for this proposal, see [this comment](https://github.com/georust/geo/pull/935#issuecomment-1600917519) on PR #935 . 

Possibly the timing of this proposal is weird, I am sure that the `geotraits::CoordTrait` will affect how this should be implemented. I suspect it may allow the `Vector2DOps` to define default implementations?

Note:

 - Currently `Vector2DOps` isn't used anywhere so it will trigger dead code warnings.
 - For an example that illustrates how I imagine it would be used in the OffsetCurve trait, see [this snippet](https://github.com/thehappycheese/geo/blob/d95420f1f8d6b0d2dd1e6d803fe3d3e2a2b3dd13/geo/src/algorithm/offset_curve/offset_line_raw.rs#L46C1-L51C1) from this PR #935 
   - Note that PR #935 is a separate, older branch, in that branch I was calling it `VectorExtensions`. In this PR I renamed it to `Vector2DOps` so that it is more similar to the existing `AffineOps`

Many thanks for your time and any feedback you can provide

### DIFF
--- a/geo/src/algorithm/mod.rs
+++ b/geo/src/algorithm/mod.rs
@@ -232,6 +232,11 @@ pub mod triangulate_earcut;
 #[cfg(feature = "earcutr")]
 pub use triangulate_earcut::TriangulateEarcut;
 
+/// # Vector Operations for 2D coordinates
+/// 
+mod vector_ops;
+pub(crate) use vector_ops::Vector2DOps;
+
 /// Calculate the Vincenty distance between two `Point`s.
 pub mod vincenty_distance;
 pub use vincenty_distance::VincentyDistance;
--- a/geo/src/algorithm/mod.rs
+++ b/geo/src/algorithm/mod.rs
@@ -232,10 +232,9 @@ pub mod triangulate_earcut;
 #[cfg(feature = "earcutr")]
 pub use triangulate_earcut::TriangulateEarcut;
 
-/// # Vector Operations for 2D coordinates
-/// 
+/// Vector Operations for 2D coordinates
 mod vector_ops;
-pub(crate) use vector_ops::Vector2DOps;
+pub use vector_ops::Vector2DOps;
 
 /// Calculate the Vincenty distance between two `Point`s.
 pub mod vincenty_distance;
--- a/geo/src/algorithm/vector_ops.rs
+++ b/geo/src/algorithm/vector_ops.rs
@@ -111,6 +111,9 @@ where
     fn try_normalize(self) -> Option<Self>;
 
     /// Returns true if both the x and y components are finite
+    // Annotation to disable bad clippy lint; It is not good to use
+    // `&self` as clippy suggests since Coord is Copy
+    #[allow(clippy::wrong_self_convention)]
     fn is_finite(self) -> bool;
 }
 
--- a/geo/src/algorithm/vector_ops.rs
+++ b/geo/src/algorithm/vector_ops.rs
@@ -1,35 +1,5 @@
 //! This module defines the [Vector2DOps] trait and implements it for the
 //! [Coord] struct.
-//!
-//! In the future [Vector2DOps] might be implemented on other types:
-//!
-//! - Based on community discussions it seems like the existing struct
-//!   [crate::Coord] is just one of many future data structures which may hold
-//!   coordinate information. For example future data structures may also hold Z
-//!   and/or M ordinates, or other arbitrary data.
-//! - [geotraits::CoordTrait] is a future trait defining accessor methods
-//!   `.x()` and `.y()` which will facilitate these generic data structures.
-//!
-//! > Note: [Vector2DOps] is not implemented for [crate::Point] because users
-//! > probably expect to see more familiar geospatial functions like
-//! > `a.euclidean_distance(b)` at that level and generally not linear algebra
-//! > like `a.dot(b)`.
-//!
-//! For now, it is assumed that users of [Vector2DOps] are satisfied to upcast
-//! everything to [CoordFloat]. In future it might make sense to split this
-//! trait into 3 flavors supporting progressively more functions:
-//!
-//! - `trait Vector2DOpsCoordNum` for [CoordNum]
-//!     - Supports magnitude_squared, dot
-//! - `trait Vector2DOpsCoordNumSigned:Vector2DOpsCoordNum` for [CoordNum] + [Signed]
-//!     - Supports left, right, wedge_product
-//! - `trait Vector2DOpsCoordFloat:Vector2DOpsCoordNumSigned` for [CoordFloat]
-//!     - Supports magnitude, normalize, etc
-//!
-//! Maybe if these traits were restricted to the future [geotraits::CoordTrait]
-//! then they could provide default implementations using the accessors `.x()`
-//! and `.y()`?
-//!
 
 use crate::{Coord, CoordFloat, CoordNum};
 
@@ -142,16 +112,6 @@ where
     /// These values should avoid overflowing to Infinity for coordinate values
     /// in the range typically relevant for spatial data (+-40e6) which is the
     /// approximate length of the earth's great circle in metres.
-    ///
-    /// > Note to Reviewer: it may be annoying to have to provide a value for
-    /// > minimum_magnitude, but that seems to be what `nalgebra` does
-    /// > (See https://docs.rs/nalgebra/latest/src/nalgebra/base/norm.rs.html#301-307).
-    /// > Some other parts of the api do not require the user to specify a
-    /// > value, but I haven't yet figured out how those work because it is
-    /// > wrapped up in the simba SIMD crate in complicated macros.
-    /// >
-    /// > Open to suggestions about how this can be better handled, or the
-    /// > try_normalize function can just be removed for now.
     fn try_normalize(self, minimum_magnitude: Self::Scalar) -> Option<Self>;
 }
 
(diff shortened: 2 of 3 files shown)


geo/src/algorithm/mod.rs
```diff
@@ -232,6 +232,11 @@ pub mod triangulate_earcut;
 #[cfg(feature = "earcutr")]
 pub use triangulate_earcut::TriangulateEarcut;
 
+/// # Vector Operations for 2D coordinates
+/// 
+mod vector_ops;
+pub(crate) use vector_ops::Vector2DOps;
```
michaelkirk, line 238: Is there a reason for this not to be public?

michaelkirk (commented):
Sorry for the delay - I thought I submitted a review a long time ago, but it seems like it didn't go through (or I'm hallucinating).

geo/src/algorithm/mod.rs
thehappycheese: Hi @michaelkirk no stress about the delay, thankyou for your time :)

No problems making it public. I was thinking of this as an internal tool, but some crate users may find it useful. 

One benefit of making it private is to reduce the complexity of the API exposed to the user. For example, a user might want to find the length of a `Line`, but miss the `EuclideanLength` trait in the docs, and mistakenly assume they should use `Vector2DOps::magnitude` on the `start` and `end`.

On the other hand making it private means 'cargo doc' may hide it away and make it harder for other crate contributors to use it. The user also misses out on some useful functionality.

I think changing it to be public is probably the way to go, and follows the established pattern in `algorithm/mod.rs`?

geo/src/algorithm/vector_ops.rs
```diff
@@ -0,0 +1,347 @@
...
+/// bring together the various hand-crafted linear algebra operations used
+/// throughout other algorithms and attached to various structs.
+///
+///
+
+pub trait Vector2DOps<Rhs = Self>
+where
+    Self: Sized,
+{
+    type NumericType: CoordNum + Send + Sync;
```
michaelkirk, line 48:
```suggestion
    type Scalar: CoordNum + Send + Sync;
```

We have, I think, similar associated types elsewhere in the codebase which we call `Scalar` - can we use that naming here too to stay conventional, or is that not accurate in this case?

geo/src/algorithm/mod.rs
michaelkirk: :+1: I'm in favor of making them public and treating them like all our other algorithm modules. They seem like generally useful methods.

michaelkirk (commented):
Is there a reason that this is a draft, or would you be happy to move forward with merging it?

kylebarron:
> Possibly the timing of this proposal is weird, I am sure that the `geotraits::CoordTrait` will affect how this should be implemented. 

I think coordinate traits will take a while to get stabilized, so I wouldn't put too much thought in that work for your PR

geo/src/algorithm/vector_ops.rs
thehappycheese: Changed to `Scalar`, that is the correct terminology 👍

geo/src/algorithm/mod.rs
thehappycheese: Ok sounds like a plan to me :)

michaelkirk (approved):
Other than some final nits, this LGTM.

geo/src/algorithm/vector_ops.rs
```diff
@@ -0,0 +1,347 @@
...
+            y: f64_minimum_threshold / 2.0
+        };
+        assert_eq!(a.try_normalize(f64_minimum_threshold), None);
+
+        // Zero vector - Fails because below threshold
+        let a = coord! { x: 0.0, y: 0.0 };
+        assert_eq!(a.try_normalize(f64_minimum_threshold), None);
+
+        // Large vector Pass;
+        // Note: this fails because the magnitude calculation overflows to infinity
```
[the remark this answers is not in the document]
thehappycheese: ah yes correct that one is meant to pass... looks like I botched a copy paste

geo/src/algorithm/vector_ops.rs
```diff
@@ -0,0 +1,347 @@
...
+        //       this could be avoided my modifying the implementation to use scaling be avoided using scaling
+        let a = coord! {
+            x: f64::sqrt(f64::MAX/2.0),
+            y: f64::sqrt(f64::MAX/2.0)
+        };
+        assert!(a.try_normalize(f64_minimum_threshold).is_some());
+
+        // Large Vector Fail;
+        // Note: This test uses the unstable float_next_up_down feature
+        // let a = coord! { x: f64::sqrt(f64::MAX/2.0), y: f64::next_up(f64::sqrt(f64::MAX/2.0)) };
```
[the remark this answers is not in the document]
michaelkirk: Thanks for reading through my unreasonable suggestion and coming up with something better!

michaelkirk (approved):
LGTM, but I'll give it a couple more days before merging in case anyone else wanted to weigh in.

michaelkirk:
Can you please run `cargo fmt` and push back up? It doesn't like your CRLF line terminators and a couple other things.

thehappycheese:
> Can you please run `cargo fmt` and push back up? It doesn't like your CRLF line terminators and a couple other things.

Hi @michaelkirk, appologies for that, I ran cargo fmt again and I think it is fixed now in [12155aa](https://github.com/georust/geo/pull/1025/commits/12155aaea512b5ea91a00059942a17c1d63f928f)

I'm pretty sure it has been committed using LF and not CRLF.
The git docs say 'You can tell Git to convert CRLF to LF on commit but not the other way around by setting core.autocrlf to `input`' so i have done that. I am surprised github has no built-in way to verify if it worked.

urschrei:
Bors try

urschrei:
This is a spurious Clippy lint failure imo, because the suggested change:

```diff
--- a/geo/src/algorithm/vector_ops.rs
+++ b/geo/src/algorithm/vector_ops.rs
@@ -111,7 +111,7 @@ where
     fn try_normalize(self) -> Option<Self>;
 
     /// Returns true if both the x and y components are finite
-    fn is_finite(self) -> bool;
+    fn is_finite(&self) -> bool;
 }
 
 impl<T> Vector2DOps for Coord<T>
@@ -165,7 +165,7 @@ where
         }
     }
 
-    fn is_finite(self) -> bool {
+    fn is_finite(self: &geo_types::Coord<T>) -> bool {
         self.x.is_finite() && self.y.is_finite()
     }
 }
```

Introduces an unnecessary borrow of `Coord` (a `Copy` type).

If you annotate the trait fn instead:

```diff
--- a/geo/src/algorithm/vector_ops.rs
+++ b/geo/src/algorithm/vector_ops.rs
@@ -111,6 +111,7 @@ where
     fn try_normalize(self) -> Option<Self>;
 
     /// Returns true if both the x and y components are finite
+    #[allow(clippy::wrong_self_convention)]
     fn is_finite(self) -> bool;
 }
```

All should be well.

urschrei:
bors try

thehappycheese:
I am confused... There should be a 12th commit where I added the annotation, but it is not showing up in this PR. Is it because it happened while bors was doing something?

urschrei:
I'm confused too. I don't _think_ it was bors, and it's…somewhere:

https://github.com/georust/geo/commit/006a48c1be8430a74ad916674790cb9cc8ae2f1d

urschrei:
Can you pull your remote branch and then push again, just to be sure? I can see the commit on your remote, but it's now not in the PR for some reason.

thehappycheese:
I cant find a way to trigger it manually. I tried the open pull request button but it just navigates to this existing PR. I can try commit some minor change to a comment or something and see if that triggers it to get included? Maybe I need to rebase it or something? It still says it can be merged tho.

urschrei:
bors r=michaelkirk

urschrei:
Now it's timing out. This is hilarious.

urschrei:
bors retry

urschrei:
I though we turned off the merge queue, but apparently not, and it's now conflicting with bors. The PR passes so I'm merging using the queue, then going for a little walk.

thehappycheese:
Hooray, thanks for all the help to get it merged, I'm looking forward to working on the next part of the other offset curve PR 😄

michaelkirk:
I hadn't turned it off - I just never got it to work.

I did, just now, turn it off though. For the record it's under Settings > Branches (branch protection rules) > Edit

Screenshot 2023-07-31 at 10 57 11

PR #1026, opened 2023-06-27
haversine closest point fixups

michaelkirk:
---

Addendum to #958


michaelkirk:
bors r+

I'm going to be bold and self-merge this since the diff from #958 is very modest.

Issue #979, opened 2023-01-26
Implement EuclideanDistance for all combinations of geometry types


michaelkirk:
Currently we can compute EucldieanDistance for lots of things, but not _everything_.

Ultimately, we want to add a 
```
impl EuclideanDistance for Geometry<T> {
    fn euclidean_distance(&self, other: &Geometry<T>) -> T
}
```

Implementing that should lead you down a path of compiler errors leading to any other missing combinations.

JosiahParry:
This would be _very_ helpful! Right now I'm trying to write a wrapper for calculating a distance matrix between two vectors of Geometrys. But the branching of types  is more than my smooth brain can handle!

urschrei:
Closed by #1029. Thanks @mcassels!

PR #1029, opened 2023-07-09
implement euclidean distance for all geometry types

mcassels:
- [ x] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/main/CODE_OF_CONDUCT.md).
---

This change implements https://github.com/georust/geo/issues/979

### DIFF
--- a/geo/src/algorithm/euclidean_distance.rs
+++ b/geo/src/algorithm/euclidean_distance.rs
@@ -402,8 +402,7 @@ where
 // └─────────────────────────────┘
 
 /// Implements EuclideanDistance between Geometry<T> and specific geometry types
-/// by converting the specific types to Geometry<T> and using the Geometry<T> implementation.
-/// Note that currently this macro cannot be used for GeometryCollection because Geometry::from(GeometryCollection) is currently disabled.
+/// by using the Geometry's inner type's implementation.
 macro_rules! impl_euclidean_distance_to_geometry {
   ([$($for:ty),*]) => {
       $(
@@ -412,8 +411,18 @@ macro_rules! impl_euclidean_distance_to_geometry {
               T: GeoFloat + FloatConst + RTreeNum + FloatConst,
           {
               fn euclidean_distance(&self, geom: &Geometry<T>) -> T {
-                  let self_as_geom = Geometry::from(self.clone());
-                  self_as_geom.euclidean_distance(geom)
+                  match geom {
+                      Geometry::Point(p) => self.euclidean_distance(p),
+                      Geometry::Line(l) => self.euclidean_distance(l),
+                      Geometry::LineString(ls) => self.euclidean_distance(ls),
+                      Geometry::Polygon(p) => self.euclidean_distance(p),
+                      Geometry::MultiPoint(mp) => self.euclidean_distance(mp),
+                      Geometry::MultiLineString(mls) => self.euclidean_distance(mls),
+                      Geometry::MultiPolygon(mp) => self.euclidean_distance(mp),
+                      Geometry::GeometryCollection(gc) => self.euclidean_distance(gc),
+                      Geometry::Rect(r) => self.euclidean_distance(r),
+                      Geometry::Triangle(t) => self.euclidean_distance(t),
+                  }
               }
           }
 
@@ -422,15 +431,25 @@ macro_rules! impl_euclidean_distance_to_geometry {
               T: GeoFloat + FloatConst + RTreeNum + FloatConst,
           {
               fn euclidean_distance(&self, other: &$for) -> T {
-                  let other_as_geom = Geometry::from(other.clone());
-                  other_as_geom.euclidean_distance(self)
+                  match self {
+                      Geometry::Point(p) => p.euclidean_distance(other),
+                      Geometry::Line(l) => l.euclidean_distance(other),
+                      Geometry::LineString(ls) => ls.euclidean_distance(other),
+                      Geometry::Polygon(p) => p.euclidean_distance(other),
+                      Geometry::MultiPoint(mp) => mp.euclidean_distance(other),
+                      Geometry::MultiLineString(mls) => mls.euclidean_distance(other),
+                      Geometry::MultiPolygon(mp) => mp.euclidean_distance(other),
+                      Geometry::GeometryCollection(gc) => gc.euclidean_distance(other),
+                      Geometry::Rect(r) => r.euclidean_distance(other),
+                      Geometry::Triangle(t) => t.euclidean_distance(other),
+                  }
               }
           }
       )*
   };
 }
 
-impl_euclidean_distance_to_geometry!([Point<T>, Line<T>, LineString<T>, Polygon<T>, MultiPolygon<T>, Triangle<T>, Rect<T>]);
+impl_euclidean_distance_to_geometry!([Point<T>, Line<T>, LineString<T>, Polygon<T>, MultiPolygon<T>, Triangle<T>, Rect<T>, GeometryCollection<T>]);
 
 /// Euclidean distance implementation for multi geometry types.
 macro_rules! impl_euclidean_distance_for_iter_geometry {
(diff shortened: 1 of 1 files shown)


michaelkirk (commented):
Hi @mcassels - thanks for the PR!

I haven't reviewed very thoroughly yet, but I had a question about the usage of clone - see my inline commentary.

geo/src/algorithm/euclidean_distance.rs
```diff
@@ -480,47 +397,229 @@ where
...
+/// Note that currently this macro cannot be used for GeometryCollection because Geometry::from(GeometryCollection) is currently disabled.
+macro_rules! impl_euclidean_distance_to_geometry {
+  ([$($for:ty),*]) => {
+      $(
+          impl<T> EuclideanDistance<T, Geometry<T>> for $for
+          where
+              T: GeoFloat + FloatConst + RTreeNum + FloatConst,
+          {
+              fn euclidean_distance(&self, geom: &Geometry<T>) -> T {
+                  let self_as_geom = Geometry::from(self.clone());
```
michaelkirk, line 415: This usage of clone is unfortunate, since it potentially doubles our memory usage. Did you explore alternatives that don't require a `clone` here? Maybe it'd be a ton of code?

mcassels: Thanks for the review! I've updated this to avoid the use of clone

JosiahParry: I'm quite personally interested in this :) Could the use of cloning here be avoided by creating the `impl`s manually?

mcassels: @JosiahParry the most recent commit removes this use of clone :)

geo/src/algorithm/euclidean_distance.rs
```diff
@@ -480,47 +397,248 @@ where
...
-impl<T> EuclideanDistance<T, Line<T>> for MultiPolygon<T>
-where
-    T: GeoFloat + FloatConst + Signed + RTreeNum,
-{
-    fn euclidean_distance(&self, other: &Line<T>) -> T {
-        other.euclidean_distance(self)
+impl_euclidean_distance_for_iter_geometry!(MultiPoint<T>, [Point<T>, Line<T>, LineString<T>, MultiLineString<T>, Polygon<T>, MultiPolygon<T>, GeometryCollection<T>]);
+impl_euclidean_distance_for_iter_geometry!(MultiLineString<T>, [Point<T>, Line<T>, LineString<T>, Polygon<T>, MultiPolygon<T>, GeometryCollection<T>]);
+impl_euclidean_distance_for_iter_geometry!(MultiPolygon<T>, [Point<T>, Line<T>, LineString<T>, Polygon<T>, GeometryCollection<T>]);
+impl_euclidean_distance_for_iter_geometry!(GeometryCollection<T>, [Point<T>, Line<T>, LineString<T>, Polygon<T>]);
```
michaelkirk, line 498: Each or these macros is taking a smaller list of input for the second parameter. I assume that's because the implementations have already been covered by a previous macro. 

The trouble is, it's hard to review this and be confident that we didn't miss one.

For example, while `geometry.euclidean_distance(&multi_polygon)` is implemented, we are missing `geometry.euclidean_distance(&multi_point)`.

One idea I had is, even though it might be more typing, to not have the macro implementation be reflexive.
e.g. only implement `impl<T> EuclideanDistance<T, $target> for $for`, but not the inverse `impl<T> EuclideanDistance<T, $for> for $target`.

I think that might make it easier to scan the source code for missing combinations. e.g. something like

```
impl_euclidean_distance_for_iter_geometry!(MultiPoint<T>,         [Point<T>, MultiPoint<T>, Line<T>, LineString<T>, MultiLineString<T>, Polygon<T>, MultiPolygon<T>, GeometryCollection<T>, Geometry<T>]);
impl_euclidean_distance_for_iter_geometry!(MultiLineString<T>,    [Point<T>, MultiPoint<T>, Line<T>, LineString<T>, MultiLineString<T>, Polygon<T>, MultiPolygon<T>, GeometryCollection<T>, Geometry<T>]);
impl_euclidean_distance_for_iter_geometry!(MultiPolygon<T>,       [Point<T>, MultiPoint<T>, Line<T>, LineString<T>, MultiLineString<T>, Polygon<T>, MultiPolygon<T>, GeometryCollection<T>, Geometry<T>]);
impl_euclidean_distance_for_iter_geometry!(GeometryCollection<T>, [Point<T>, MultiPoint<T>, Line<T>, LineString<T>, MultiLineString<T>, Polygon<T>, MultiPolygon<T>, GeometryCollection<T>, Geometry<T>]);
```

(Note I included `Geometry` in the second parameter list)

Possibly that second parameter could be baked into a macro itself so you don't have to type it all out, but I don't feel that strongly about that. 

Similarly with the other macros.

geo/src/algorithm/euclidean_distance.rs
```diff
@@ -480,47 +397,248 @@ where
...
-impl<T> EuclideanDistance<T, Point<T>> for Triangle<T>
+impl<T> EuclideanDistance<T> for Geometry<T>
 where
-    T: GeoFloat,
+    T: GeoFloat + FloatConst,
 {
-    fn euclidean_distance(&self, point: &Point<T>) -> T {
-        if self.intersects(point) {
-            return T::zero();
+    fn euclidean_distance(&self, other: &Geometry<T>) -> T {
```
michaelkirk, line 552: My condolences to your wrists. 😉 

I'd expect this to be implementable as
```
impl<T> EuclideanDistance<T> for Geometry<T>
 where T: GeoFloat + FloatConst 
{
    crate::geometry_delegate_impl! {
        fn euclidean_distance(&self, other: &Geometry<T>) -> T;
    }
}
```

Please revisit this method after addressing the missing implementations mentioned above.

michaelkirk (changes requested):
Thanks for getting rid of that clone - this is getting closer!

michaelkirk:
Hi @mcassels, are you still interested in working on this? Let me know if you have any questions. 

If you don't think you'll have time, I can try to finish it up.

mcassels:
@michaelkirk thanks for the review! I'm still planning to work on this, sorry for the delay -- been travelling with limited internet access. I'm hoping to get back to this within the next couple days

geo/src/algorithm/euclidean_distance.rs
mcassels: Thanks, I agree that it's hard to see if any combinations are missing. I've now updated this to make the macro implementations not reflexive, and also re-organized the macros a bit. I hope that it's more readable and clear now.

One note regarding this point:

> geometry.euclidean_distance(&multi_polygon) is implemented, we are missing geometry.euclidean_distance(&multi_point)

In the new version, all the implementations of euclidean distance between `Geometry<T>` and specific geometry types are handled by 2 separate macros, so `Geometry<T>` is not included in the parameter lists for the multi-geometry macros. This seemed cleaner, since we need to have `impl<T> EuclideanDistance<T, Geometry<T>>` for all geometry types.

geo/src/algorithm/euclidean_distance.rs
mcassels: thanks! done

geo/src/algorithm/euclidean_distance.rs
```diff
@@ -480,47 +397,248 @@ where
...
+                  }
+              }
+          }
+
+          impl<T> EuclideanDistance<T, $for> for Geometry<T>
+          where
+              T: GeoFloat + FloatConst + RTreeNum + FloatConst,
+          {
+              fn euclidean_distance(&self, other: &$for) -> T {
+                  match self {
```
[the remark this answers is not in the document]
mcassels: thanks! done

michaelkirk (approved):
LGTM - thanks!

bors r+

JosiahParry:
Amazing!

PR #1031, opened 2023-07-09
Update CI with libproj 9.2.1 containers

urschrei:
---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,12 +43,12 @@ jobs:
       matrix:
         container_image:
           # Use the latest stable version. No need for older versions.
-          - "georust/geo-ci:rust-1.66"
+          - "georust/geo-ci:proj-9.2.1-rust-1.70"
     container:
       image: ${{ matrix.container_image }}
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - run: rustup component add rustfmt clippy
       - run: cargo fmt --all -- --check
       - run: cargo clippy --all-features --all-targets -- -Dwarnings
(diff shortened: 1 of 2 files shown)


urschrei:
@michaelkirk would you prefer to merge https://github.com/georust/docker-images/pull/25 prior to regenerating 1.63? (I thought it would include libproj 9.2.1, but I may be confused, as can so often be the way when Docker is involved)

.github/workflows/test.yml
```diff
@@ -69,8 +69,8 @@ jobs:
           # Minimum supported rust version (MSRV)
           - "georust/geo-ci:rust-1.63"
```
michaelkirk, line 70:
```suggestion
          - "georust/geo-ci:proj-9.2.1-rust-1.63"
```

.github/workflows/test.yml
```diff
@@ -97,8 +97,8 @@ jobs:
...
```
michaelkirk, line 98:
```suggestion
          - "georust/geo-ci:proj-9.2.1-rust-1.63"
```

michaelkirk (approved):
I didn't clobber the existing tags with the new image. Maybe I should have?

In any case, I think if you reference the tags as in my suggestion things will "just work".

urschrei:
> I didn't clobber the existing tags with the new image. Maybe I should have?

I'm not sure! It's a chicken and egg situation I think.

.github/workflows/test.yml
```diff
@@ -124,13 +124,13 @@ jobs:
...
```
michaelkirk, line 125:
```suggestion
          - "georust/geo-ci:proj-9.2.1-rust-1.63"
```

michaelkirk: Sorry, missed one.

urschrei:
There is something weird going on here. Gonna close this PR and try again.

Issue #976, opened 2023-01-24
Panic in BooleanOps::difference for Multipolygon


cassiepearson-deca:
The BooleanOps::difference function panics about half the time when running the included test case due to a failure in the sweep algorithm. Similar datasets succeed without issue. The provided test case was reduced from a larger test case. I attempted to find individual polygons that failed but was unable to find a smaller input Multipolygon.

Test case: https://github.com/Deca-Technologies/geo_boolean_difference_panic

I expected to see the structures successfully difference'd returning the entire subject multipolygon. Instead, geo panics during the difference operation about 50% of the time. The code includes a for loop to run until the failure is hit.

Issue https://github.com/georust/geo/issues/913 may be related but I am not certain as it uses a different version of geo.

Backtrace:
```
thread 'main' panicked at 'segment not found in active-vec-set: 8', /Users/cassie/.cargo/registry/src/github.com-1ecc6299db9ec823/geo-0.23.1/src/algorithm/sweep/vec_set.rs:22:14
stack backtrace:
   0: rust_begin_unwind
             at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
             at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/panicking.rs:65:14
   2: core::result::unwrap_failed
             at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/result.rs:1791:5
   3: core::result::Result<T,E>::expect
             at /rustc/69f9c33d71c871fc16ac445211281c6e7a340943/library/core/src/result.rs:1070:23
   4: geo::algorithm::sweep::vec_set::VecSet<geo::algorithm::sweep::active::Active<T>>::index_of
             at /Users/cassie/.cargo/registry/src/github.com-1ecc6299db9ec823/geo-0.23.1/src/algorithm/sweep/vec_set.rs:20:9
   5: geo::algorithm::sweep::proc::Sweep<C>::handle_event
             at /Users/cassie/.cargo/registry/src/github.com-1ecc6299db9ec823/geo-0.23.1/src/algorithm/sweep/proc.rs:232:30
   6: geo::algorithm::sweep::proc::Sweep<C>::next_event::{{closure}}
             at /Users/cassie/.cargo/registry/src/github.com-1ecc6299db9ec823/geo-0.23.1/src/algorithm/sweep/proc.rs:46:13
[...]

PR #1032, opened 2023-07-10
Falliable Boolean Operations

RobWalt:
---

The idea comes from [here](https://github.com/georust/geo/issues/976#issuecomment-1623394857).

tldr: I think it would be best to resultify every panic that could happen in the `BooleanOps` trait as a short term solution. This gives `geo` users the chance to catch and deal with these scenarios (e.g. adjust the geometry slightly and try again) or to ignore them if the results aren't too important.

---

# Some things to discuss

The current state of this PR is merely a PoC for now. There are some things I would like to discuss first

## Error handling

Most of the errors are just placeholders for the first iteration of this PR. We can use `thiserror` to make error handling more robust but I didn't want to go that route yet without any second opinion.

## Breaking changes

I tried to avoid breaking changes as much as possible for now. Technically I probably did create some breaking changes which could be ironed out easiliy. I just didn't want to create a copy of all of the existing code parts yet. If you want a really clean solution, we can just create the falliable version of everything completely detached in separate functions from the current state of `geo`

--- 

# Todos

- [ ] un-break intersections iterator
- [ ] re-export the error type to let users handle them correctly


geo/Cargo.toml
```diff
@@ -40,6 +40,7 @@ rand_distr = "0.4.3"
 
 ### boolean-ops test deps
 wkt = "0.10.1"
+serde_json = "1.0"
```
michaelkirk, line 43: Can you use WKT for your test fixtures? We already have the dependency and IMO it's easier to read as a human.

geo/src/types.rs
```diff
@@ -37,6 +37,16 @@ impl<F: GeoFloat> Closest<F> {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum GeoError {
```
michaelkirk, line 41: Calling this `GeoError` is a bit aspirational since they are only for boolean ops. Is it actually useful to distinguish between all these cases? (genuine question)

geo/src/types.rs
```diff
@@ -37,6 +37,16 @@ impl<F: GeoFloat> Closest<F> {
...
+    MissingOverlap,
+    SegmentNotFoundInActiveSet(usize),
+    SegmentAlreadyFoundInActiveSet(usize),
+    ExpectedNonemptyEvents,
+    NotEulierian(usize),
+    Unreachable(&'static str),
```
michaelkirk, line 47: Just to put my cards on the table, I think writing panics into our code is reasonable if the programmer truly expects the code to be unreachable, and that trying to turn every single panic into a runtime error is probably an anti-pattern.

My preference is that this be only the set of run time errors we're actually encountering. Is it the case that you're hitting all of these? 

If we are actually encountering one of the "Unreachable" cases, it's not unreachable. Can you think of a better name that might help people understand what's going on?

michaelkirk (commented):
I'm on board with the idea of returning runtime errors where users are hitting panics, but I'd like to make sure we're not overly convoluting the code. See my comments on the GeoError enum.

geo/Cargo.toml
RobWalt: Yeah that makes total sense. I'll change that :+1:

geo/src/types.rs
RobWalt: yeah that isn't final yet 😅 I just figured it would be a good start to test it all out. I'm totally on board with you in creating a more specialized error enum there. The distinction between all the cases was made so that we would basically still have some rough information in the normal boolops code. Since we're just unwrapping there we will at least see the error enum variant which was chosen to mirror the previous error as close as possible.

geo/src/types.rs
RobWalt: Yeah, unfortunately I was able to trigger the unreachable code (see the 3rd case in the spoiler of [this comment](https://github.com/georust/geo/issues/976#issuecomment-1615055706)).

Again, I agree here as well. We should probably pick a better name here. I'll look into that. I wasn't diving too deep into that code yet since I just wanted to have some kind of parity to the existing panicing implementation, but I guess now's a good time for that.

geo/src/types.rs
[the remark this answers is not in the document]
RobWalt: I'm not really getting what you meant here. Is it showing the right code here?

RobWalt:
A friend of mine found another major point of failure of the normal `BooleanOps` algo that I wasn't able to patch yet. In the implementation of `Ord` for `Active<T>` there is a potential panic which does occur in the wild.

I'm not really sure how to fix that. We can potentially just make it not panic in the `None` case there by defaulting to `Ordering::Less` in this case but I'm not sure what the implications of this are tbh.

geo/src/types.rs
michaelkirk, line 41: > I'd guess we want this to be non-exhaustive

I mean like this:

```suggestion
#[non_exhaustive]
pub enum GeoError {
```

That way, if we need to add new error cases in a subsequent release, we can do so without breaking existing user's code.

PR #1033, opened 2023-07-10
update CI containers to use libproj 9.2.1

urschrei:
---


urschrei:
bors r=michaelkirk

PR #1041, opened 2023-07-26
Add Hausdorff Distance trait 

JosiahParry:

---

This PR adds a new trait `HausdorffDistance` which is implemented for all geometries. 

This is my first Rust library PR so I'll need some guidance getting this through the finish line! 

### Outstanding questions for maintainers: 

- This PR is lacking tests. Should there be a test for each combo of geometry?
- `CHANGES.md` does not have a development version listed, how should it be modified?
- what amount of documentation is needed / suggested?

### DIFF
--- a/geo/src/algorithm/hausdorff_distance.rs
+++ b/geo/src/algorithm/hausdorff_distance.rs
@@ -1,32 +1,32 @@
-use crate::GeoFloat;
-use crate::CoordsIter;
 use crate::algorithm::EuclideanDistance;
-use num_traits::Bounded; // used to have T as generic type in folding
+use crate::CoordsIter;
+use crate::GeoFloat;
 use geo_types::*;
+use num_traits::Bounded; // used to have T as generic type in folding
 
 /// Determine the distance between two geometries using the [Hausdorff distance formula].
 ///
-/// Hausdorff distance is used to compare two point sets. It measures the maximum euclidean 
+/// Hausdorff distance is used to compare two point sets. It measures the maximum euclidean
 /// distance of a point in one set to the nearest point in another set. Hausdorff distance
-/// is often used to measure the amount of mismatch between two sets. 
-/// 
+/// is often used to measure the amount of mismatch between two sets.
+///
 /// [Hausdorff distance formula]: https://en.wikipedia.org/wiki/Hausdorff_distance
 pub trait HausdorffDistance<T, Rhs = Self> {
     fn hausdorff_distance(&self, rhs: &Rhs) -> T;
 }
 
 // We can take advantage of the coords_iter() method for all geometries
-// to iterate across all combos of coords (infinum). Simplifies the 
+// to iterate across all combos of coords (infinum). Simplifies the
 // implementations for all geometries.
 macro_rules! impl_hausdorff_distance_coord_iter {
     ($from:ident, [$($to:ident),*]) => {
         $(
             impl<T> HausdorffDistance<T, $to<T>> for $from<T>
             where
-                T: GeoFloat 
+                T: GeoFloat
             {
                 fn hausdorff_distance(&self, geom: &$to<T>) -> T {
-                    // calculate from A -> B 
+                    // calculate from A -> B
                     let hd1 = self
                         .coords_iter()
                         .map(|c| {
@@ -38,7 +38,7 @@ macro_rules! impl_hausdorff_distance_coord_iter {
                             .fold(<T as Bounded>::max_value(), |accum, val| accum.min(val))
                         })
                         .fold(<T as Bounded>::min_value(), |accum, val| accum.max(val));
-            
+
                     // Calculate from B -> A
                     let hd2 = geom
                         .coords_iter()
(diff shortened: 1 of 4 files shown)


geo/src/algorithm/hausdorff_distance.rs
```diff
@@ -0,0 +1,194 @@
+use crate::GeoFloat;
+use crate::CoordsIter;
+use crate::algorithm::EuclideanDistance;
+use num_traits::Bounded; // used to have T as generic type in folding
+use geo_types::*;
+
+/// Determine the distance between two geometries using the [Hausdorff distance formula].
+///
+/// [haversine formula]: https://en.wikipedia.org/wiki/Hausdorff_distance
```
frewsxcv, line 9: > `[haversine formula]`

This should be updated

geo/src/algorithm/hausdorff_distance.rs
```diff
@@ -0,0 +1,194 @@
...
+            {
+                fn hausdorff_distance(&self, geom: &$to<T>) -> T {
+                    // calculate from A -> B 
+                    let hd1 = self
+                        .coords_iter()
+                        .map(|c| {
+                            geom
+                            .coords_iter()
+                            .map(|c2| {
+                                c.euclidean_distance(&c2)
```
frewsxcv, line 32: Can you clarify in the docs that this algorithm utilizes euclidean distance under the hood? In case someone might be expecting it to use haversine or something else

JosiahParry:
@frewsxcv thanks! I've added a note in the documentation that it uses euclidean distance. Is this explicit enough?

https://github.com/JosiahParry/geo/blob/1c44e3b5578b4e2af24bf3755827ca2b87c7e2eb/geo/src/algorithm/hausdorff_distance.rs#L9-L10

I've also added a handful of tests as well. Please let me know if this is sufficient or if there ought to be exhaustive tests (every combo of geometry types).

geo/src/algorithm/hausdorff_distance.rs
```diff
@@ -0,0 +1,278 @@
...
+                            })
+                            .fold(<T as Bounded>::max_value(), |accum, val| accum.min(val))
```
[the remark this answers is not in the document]
JosiahParry: Erm, I don't know? Since the map returns `T` not `Option<T>` does `.unwrap_or()` have a use here beyond calling the or portion? Additionally, I was unaware you can call `min()` on an iterator!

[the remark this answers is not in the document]
JosiahParry: I've tried replacing this with `.min()` and received the following compiler error. It appears that since `Ord` trait is not implemented for `T` `min()` cannot be used. 
image

lnicola (approved):

JosiahParry:
I've pushed a commit that addresses formatting using `cargo fmt`.

geo/src/algorithm/hausdorff_distance.rs
```diff
@@ -0,0 +1,261 @@
...
+
+                    // The max of the two
+                    hd1.max(hd2)
+                }
+            }
+        )*
+    };
+}
+
+impl_hausdorff_distance_coord_iter! {
```
michaelkirk, line 63: Thanks for getting comprehensive implementations with the macros. 

I think you could delete a lot of this code if you implemented generically, based on the `CoordsIter` trait.

The impl's would look something like this:

```rust
impl<'a, 'b, T, G1, G2> HausdorffDistance<'a, 'b, T, G2> for G1
where
    T: GeoFloat,
    G1: CoordsIter<'a, Scalar = T>,
    G2: CoordsIter<'b, Scalar = T>,
{
    fn hausdorff_distance(&'a self, rhs: &'b G2) -> T {
       todo!()
    }
}

impl<'a, 'b, T, G> HausdorffDistance<'a, 'b, T, G> for Coord<T>
where
    T: GeoFloat,
    G: CoordsIter<'b, Scalar = T>
{
    fn hausdorff_distance(&'a self, geom: &'b G) -> T {
        todo()!
    }
}
```

... and you'd have to add the lifetimes to the trait declaration. We've done similar things for ConvexHull.

What do you think?

geo/src/algorithm/mod.rs
```diff
@@ -120,6 +120,10 @@ pub use geodesic_intermediate::GeodesicIntermediate;
 pub mod geodesic_length;
 pub use geodesic_length::GeodesicLength;
 
+/// Calculate the Hausdorff distance between two geometries.
```
michaelkirk, line 123: Thank you for maintaining alphabetical order in this file. ❤️

geo/src/algorithm/hausdorff_distance.rs
JosiahParry: I've been trying to get this sorted out for a little while this morning but I'm finding myself in lifetime hell. It seems like there might be a mismatch between lifetimes between `CoordsIter` and `T`. 

I'll check the implementation for ConvexHull and see if i can get it sorted out.

JosiahParry: Resolved in newest commit.

JosiahParry:
I've added a note to the changelog. Please let me know what else is needed to close this PR. Thanks! :)

michaelkirk:
bors retry

Whoops. My bad I think. Let's try again now that I have (for real this time) disabled the merge queue.

michaelkirk:
Thanks!

JosiahParry:
Thank you!

PR #1048, opened 2023-08-02
chore: remove unused impl of private trait ActiveSet

RobWalt:
- [x] ~I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.~
---

The motivation for removing this is that it is a bit misleading and confusing when starting to look deeper into the code. I spent quiet some time trying to wrap my head around why we need this but it was simply unused.

This seem to be leftover parts of an earlier implementation. The trait is private, so there is no chance that this will break user code. I also made sure to check that the trait impl for `BTreeSet` isn't used anymore. It has been completely replaced by `VecSet` everywhere.

### DIFF
--- a/geo/src/algorithm/sweep/active.rs
+++ b/geo/src/algorithm/sweep/active.rs
@@ -1,10 +1,4 @@
-use std::{
-    borrow::Borrow,
-    cmp::Ordering,
-    collections::BTreeSet,
-    fmt::Debug,
-    ops::{Bound, Deref},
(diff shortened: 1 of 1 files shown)


rmanoka:
Yes, it's a left-over from a previous implementation but I wasn't sure if we'll switch back to `BTreeSet` so left it in there.  We can remove this if it helps with some linting.

rmanoka (approved):

PR #1051, opened 2023-08-08
Vector2DOps fixes

thehappycheese:
- [x] ran `cargo fmt` then `cargo fmt --all -- --check`
- [x] ran `cargo clippy --all-features --all-targets -- -Dwarnings`
---

Fixing some derps in my previous PR :)

- Removed `Send + Sync` from Scalar type; as far as I can tell these constraints are only needed if the trait uses threads. Removing Send+Sync should not prevent a user from using the trait with a Send+Sync type.
- Switched to `cmath::hypot` for `magnitude()` for instant gain in robustness, and fixed tests to reflect

### DIFF
--- a/geo/src/algorithm/vector_ops.rs
+++ b/geo/src/algorithm/vector_ops.rs
@@ -323,23 +323,22 @@ mod test {
     }
 
     #[test]
-    fn test_try_normalize_edge_cases() {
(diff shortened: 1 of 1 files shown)


michaelkirk:
Thanks!

bors r+

Issue #1049, opened 2023-08-02
simplify_vw_preserve causing self-intersecting ring


audunska:
We have some geometry code which would benefit from avoiding self-intersections after simplification. I tried to switch from the rdp algorithm to the `simplify_vw_preserve` function, and wrote some property tests. After fuzzing a bit, quickcheck came up with this counterexample:

```rust
use geo::polygon;
use geo::SimplifyVwPreserve;
let pol = polygon![
    (x: 1., y: 4.),
    (x: 3., y: 4.),
    (x: 1., y: 1.),
    (x: 7., y: 0.),
    (x: 1., y: 0.),
    (x: 0., y: 1.),
    (x: 1., y: 4.),
];
let simplified = pol.simplify_vw_preserve(&2.25);
assert_eq!(simplified, polygon![
    (x: 1., y: 4.),
    (x: 3., y: 4.),
    (x: 1., y: 1.),
    (x: 7., y: 0.),
    (x: 1., y: 0.),
    (x: 1., y: 4.),
]);
```
This simplified object has a self-intersection at `(x: 1., y: 1.)`. 

Tested on geo version 0.25 and 0.26.

audunska:
If I cyclically permute the coordinates, there is no self-intersection. So this is probably related to handling of the exterior being a linear ring.

audunska:
I guess the `cartesian_intersect` function does not detect intersections with endpoints of lines?

urschrei:
Thanks for catching the possible source of the problem @audunska. I'll take a look as soon as I can.

urschrei:
Do you have a specific example of an intersection detection miss? I'd like to test it against some variations in the algorithm.

audunska:
On a side note, I ended up implementing the algorithm described in this paper instead, but optimized using an r* tree in the same way as SimplifyVwPreserve:
https://doi.org/10.1016/j.cageo.2013.08.011

Would there be interest in contributing that implementation, maybe?

urschrei:
I've tuned the algorithm a bit to:

- be more aggressive in deciding what is an intersection, by using the standard `geo` intersection check for new lines and candidate existing lines.
- Reduced the minimum points from 6 to 5, as using 6 was causing your example to exit early while the polygon was in an invalid state. I suspect we might still run into this problem, but I'm not sure it can be completely fixed without implementing some sort of full "rewind" if we hit the minimum point limit while the polygon is invalid.

[...]

PR #1052, opened 2023-08-09
Vw refactor

urschrei:
---

This is a refactor of both VW and VwPreserve, partially in response to #1049:

- The intersection check logic is now more robust, and uses the existing line-line intersection check from `geo`
- The new triangle calculation logic has been refactored into a function which is now used by both VW and VwPreserve
- Rather than `continuing` with a fresh triangle if a triangle area above epsilon is popped from the min-heap we simply `break` now: there's no need to check any more triangles, since the min-heap ensures that all remaining elements are above epsilon 
- We allow processing to continue until we hit n + 1 now
- Many helpful comments have been added

### DIFF
--- a/geo/src/algorithm/simplify_vw.rs
+++ b/geo/src/algorithm/simplify_vw.rs
@@ -8,18 +8,18 @@ use std::collections::BinaryHeap;
 
 use rstar::{RTree, RTreeNum};
 
-/// Store triangle information
-// current is the candidate point for removal
+/// Store triangle information. Area is used for ranking in the priority queue and determining removal
 #[derive(Debug)]
 struct VScore<T, I>
 where
     T: CoordFloat,
 {
     left: usize,
+    /// The current [Point] index in the original [LineString]: The candidate for removal
     current: usize,
     right: usize,
     area: T,
-    // `visvalingam_preserve` uses `intersector`, `visvalingam` does not
+    // `visvalingam_preserve` uses `intersector`, `visvalingam` does not, so it's always false
     intersector: I,
 }
 
@@ -180,6 +180,7 @@ fn recompute_triangles<T>(
         )
         .unsigned_area();
 
+        // This logic only applies to VW-Preserve
         // The current point causes a self-intersection, and this point precedes it
         // we ensure it gets removed next by demoting its area to negative epsilon
         // we check that current_point is less than smallest.current because
--- a/geo/src/algorithm/simplify_vw.rs
+++ b/geo/src/algorithm/simplify_vw.rs
@@ -116,13 +116,12 @@ where
     // While there are still points for which the associated triangle
     // has an area below the epsilon
     while let Some(smallest) = pq.pop() {
-        // This triangle's area is above epsilon, so skip it
         if smallest.area > *epsilon {
             // no need to keep trying: the min-heap ensures that we process triangles in order
             // so if we see one that exceeds the tolerance we're done: everything else is too big
             break;
         }
-        //  This triangle's area is below epsilon: eliminate the associated point
+        //  This triangle's area is below epsilon: the associated point is a candidate for removal
         let (left, right) = adjacent[smallest.current];
         // A point in this triangle has been removed since this VScore
         // was created, so skip it
@@ -181,11 +180,11 @@ fn recompute_triangles<T>(
         .unsigned_area();
 
         // This logic only applies to VW-Preserve
-        // The current point causes a self-intersection, and this point precedes it
+        // smallest.current's removal causes a self-intersection, and this point precedes it
         // we ensure it gets removed next by demoting its area to negative epsilon
         // we check that current_point is less than smallest.current because
         // if it's larger the point in question comes AFTER smallest.current: we only want to remove
-        // the point that comes BEFORE
+        // the point that comes BEFORE smallest.current
         let area = if smallest.intersector && (current_point as usize) < smallest.current {
             -*epsilon
         } else {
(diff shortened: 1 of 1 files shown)


frewsxcv:
@urschrei Is this ready for review?

urschrei:
Nope: it turns out that JTS and GEOS have different interpretations of the input `epsilon`: JTS uses `epsilon ^ 2` for the VW algorithm, and GEOS uses `epsilon`. We _currently_ use `epsilon`, but as @michaelkirk pointed out, GEOS is downstream of JTS and there was discussion of whether to bring our implementation in line with JTS (and also why GEOS made that decision, which also affects Shapely and PostGIS) or stick with what we have. Discussion has stalled, so I guess we can revisit next week and make a decision, at which point this'll be reviewable.

urschrei:
This is ready for review if anyone's feeling inclined!

geo/src/algorithm/simplify_vw.rs
```diff
@@ -6,20 +6,21 @@ use crate::{
...
     T: CoordFloat,
 {
     left: usize,
+    /// The current [Point] index in the original [LineString]: The candidate for removal
     current: usize,
     right: usize,
     area: T,
-    // `visvalingam_preserve` uses `intersector`, `visvalingam` does not
+    // `visvalingam_preserve` uses `intersector`, `visvalingam` does not, so it's always false
     intersector: I,
```
michaelkirk, line 24: IIRC `intersector` is now always a `bool` and the `I` generic param has become inexplicable. Can we remove it, or is there a benefit to keeping this struct generic?

geo/src/algorithm/simplify_vw.rs
```diff
@@ -110,17 +111,18 @@ where
...
+        .collect::<BinaryHeap<VScore<T, bool>>>();
     // While there are still points for which the associated triangle
     // has an area below the epsilon
     while let Some(smallest) = pq.pop() {
-        // This triangle's area is above epsilon, so skip it
         if smallest.area > *epsilon {
-            continue;
+            // no need to keep trying: the min-heap ensures that we process triangles in order
+            // so if we see one that exceeds the tolerance we're done: everything else is too big
+            break;
```
michaelkirk, line 123: I'm not very familiar with this algorithm, but is it true that we can guarantee no smaller triangles will be added in subsequent calls to `recompute_triangles`?

geo/src/algorithm/simplify_vw.rs
```diff
@@ -250,7 +288,7 @@ where
 fn visvalingam_preserve<T, const INITIAL_MIN: usize, const MIN_POINTS: usize>(
     orig: &LineString<T>,
     epsilon: &T,
-    tree: &mut RTree<Line<T>>,
+    tree: &mut RTree<CachedEnvelope<Line<T>>>,
```
michaelkirk, line 291: I'd be a little surprised if there was much benefit to caching the bounding rect of a line, since there's only ever two points in it, but I'm also not that worried about keeping it in.

geo/src/algorithm/simplify_vw.rs
```diff
@@ -310,14 +350,18 @@ where
         }
         // if removal of this point causes a self-intersection, we also remove the previous point
         // that removal alters the geometry, removing the self-intersection
-        // HOWEVER if we're within 2 points of the absolute minimum, we can't remove this point or the next
+        // HOWEVER if we're within 1 point of the absolute minimum, we can't remove this point or the next
```
michaelkirk, line 353: Does this doc change reflect a bug in the existing documentation, or does this reflect a change in behavior from this PR?

michaelkirk (approved):
I'm not very familiar with this algorithm, so my comments are probably not very helpful. Feel free to take or leave them.

geo/src/algorithm/simplify_vw.rs
urschrei: It's a change in the behaviour: https://github.com/georust/geo/pull/1052/commits/8e9aa6babe00863c8f62134d49e5dbc0c9751c91#diff-ba8c428d032bd24b196bd2649c0c7985dff438f05dfebd31814400adf9b39d71R583-R584

geo/src/algorithm/simplify_vw.rs
urschrei: There was a perf improvement, which I naturally didn't include in the commit 🤦‍♂️

geo/src/algorithm/simplify_vw.rs
michaelkirk: Right, `pq` is a min heap, so we will always get the smallest triangle *currently* in the min heap.

But then we add more triangles to the head on line 142.
`recompute_triangles(&smallest, orig, &mut pq, ll, left, right, rr, max, epsilon);`

Is it possible that by `continuing` instead of `breaking`, we'd add triangles on line 142 that would be worth considering, or is there some property of this algorithm that ensures subsequent calls to `recompute_triangles` will not add such triangles?

geo/src/algorithm/simplify_vw.rs
michaelkirk: Oh, great!

geo/src/algorithm/simplify_vw.rs
urschrei: > Is it possible that by continuing instead of breaking, we'd add triangles on line 142 that would be worth considering, or is there some property of this algorithm that ensures subsequent calls to recompute_triangles will not add such triangles?

I don't think so: once we hit the first triangle above epsilon we never even make it as far as line 125 again, let alone line 142, because `continue` restarts the loop with a new triangle, which is immediately discarded.

michaelkirk: Oh right, thanks for walking me through that.

urschrei: I keep going back to check my assumptions, but I'm pretty happy that this logic is correct, and also borne out elsewhere:
- [This is the original implementation, which uses `break`](https://github.com/huonw/isrustfastyet/blob/25e7a68ff26673a8556b170d3c9af52e1c818288/mem/line_simplify.rs#L54-L55)
- [A Go implementation](https://github.com/paulmach/orb/blob/master/simplify/visvalingam.go#L93)

PR #1055, opened 2023-08-19
Add line segmentize trait

JosiahParry:
---

This PR adds a new trait `LineStringSegmentize` which provides a method `line_segmentize(n: usize)` to split `LineString`s into`n` equal length LineStrings which are returned as an `Option<MultiLineString>` . This is inspired by the [lwgeom R package](https://r-spatial.github.io/lwgeom/) which wraps liblwgeom from PostGIS.

### DIFF
--- a/geo/src/algorithm/linestring_segment.rs
+++ b/geo/src/algorithm/linestring_segment.rs
@@ -19,7 +19,6 @@ impl LineStringSegmentize for LineString {
             let total_len = self.euclidean_length();
             let densified = self.densify(total_len / (n as f64));
             return densified.line_segmentize(n);
-            // return Some(MultiLineString::new(vec![densified]))
         } else if n_lines == n {
             // if the number of line segments equals n then return the
             // lines as LineStrings
@@ -57,7 +56,10 @@ impl LineStringSegmentize for LineString {
 
         // iterate through each line segment in the LineString
         for (i, segment) in lns.enumerate() {
-            // first line string push the first coord immediately
+            // All iterations only keep track of the second coordinate
+            // in the Line. We need to push the first coordinate in the 
+            // first line string to ensure the linestring starts at the 
+            // correct place
             if i == 0 {
                 ln_vec.push(segment.start)
             }
--- a/geo/src/algorithm/linestring_segment.rs
+++ b/geo/src/algorithm/linestring_segment.rs
@@ -81,9 +81,7 @@ impl LineStringSegmentize for LineString {
             if (cum_length >= segment_length) && (i != (n_lines - 1)) {
                 let remainder = cum_length - segment_length;
                 // if we get None, we exit the function and return None
-                let Some(endpoint) =  segment.line_interpolate_point((length - remainder) / length) else {
-                    return None
-                };
+                let endpoint = segment.line_interpolate_point((length - remainder) / length)?;
 
                 // add final coord to ln_vec
                 ln_vec.push(endpoint.into());
--- a/geo/src/algorithm/linestring_segment.rs
+++ b/geo/src/algorithm/linestring_segment.rs
@@ -8,6 +8,7 @@ use crate::{Coord, Densify, EuclideanLength, LineString, LinesIter, MultiLineStr
 ///
 /// # Examples
 /// ```
+/// use geo::{LineString, MultiLineString, LineStringSegmentize, Coord};
 /// // Create a simple line string
 /// let lns: LineString<f64> = vec![[0.0, 0.0], [1.0, 2.0], [3.0, 6.0]].into();
 /// // Segment it into 6 LineStrings inside of a MultiLineString
--- a/geo/src/algorithm/linestring_segment.rs
+++ b/geo/src/algorithm/linestring_segment.rs
@@ -4,6 +4,32 @@ use crate::{Coord, Densify, EuclideanLength, LineString, LinesIter, MultiLineStr
 /// Segments a LineString into `n` equal length LineStrings as a MultiLineString.
 /// `None` will be returned when `n` is equal to 0 or when a point
 /// cannot be interpolated on a `Line` segment.
+///
+///
+/// # Examples
+/// ```
+/// // Create a simple line string
+/// let lns: LineString<f64> = vec![[0.0, 0.0], [1.0, 2.0], [3.0, 6.0]].into();
+/// // Segment it into 6 LineStrings inside of a MultiLineString
+/// let segmentized = lns.line_segmentize(6).unwrap();
+///
+/// // Recreate the MultiLineString from scratch
+/// // this is the inner vector used to create the MultiLineString
+/// let all_lines = vec![
+///     LineString::new(vec![Coord { x: 0.0, y: 0.0 }, Coord { x: 0.5, y: 1.0 }]),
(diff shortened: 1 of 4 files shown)


JosiahParry:
The almighty clippy should be satisfied now.

urschrei:
Looks like Clippy still isn't happy. Do you see those failures when you run it locally?

JosiahParry:
@urschrei I do not get them locally, I get different ones! These ones seem simple enough that I should be able to handle them. I just don't want to consume all of the GHA hours 
image

JosiahParry:
I pulled the most recent changes from main and CI is failing unrelated to my changes. 

```
  failed to query replaced source registry `crates-io`
```

michaelkirk:
A couple lines further up in the failed action output shows:
```
   Updating crates.io index
warning: spurious network error (3 tries remaining): [28] Timeout was reached (Operation too slow. Less than 10 bytes/sec transferred the last 30 seconds)
warning: spurious network error (2 tries remaining): [28] Timeout was reached (Operation too slow. Less than 10 bytes/sec transferred the last 30 seconds)
warning: spurious network error (1 tries remaining): [28] Timeout was reached (Operation too slow. Less than 10 bytes/sec transferred the last 30 seconds)
```

So, presumably this is some kind of transient network failure. I've restarted the jobs. 🤞

geo/src/algorithm/linestring_segment.rs
```diff
@@ -0,0 +1,186 @@
+use crate::line_interpolate_point::LineInterpolatePoint;
+use crate::{Coord, EuclideanLength, LineString, LinesIter, MultiLineString};
+
+/// Segments a LineString into `n` LineStrings as a MultiLineString.
```
michaelkirk, line 4:
```suggestion
/// Segments a LineString into `n` equal length LineStrings as a MultiLineString.
```

geo/src/algorithm/linestring_segment.rs
```diff
@@ -0,0 +1,186 @@
...
+/// `None` will be returned when `n` is equal to 0, `n` is larger than the
+///  number of `Line`s that make up the `LineString`, or when a point
```
michaelkirk, line 6: I want to make sure I understand the intent of the code before I try to evaluate it.

Am I understanding correctly that each individual segment will have the same length as all the other segments?

> `n` is larger than the number of `Line`s that make up the `LineString`

This "n is larger than the number of Lines" condition kind of surprises me. I'd expect that we'd split edges as necessary to maintain the "equal length" invariant. Is there a reason we wouldn't?

JosiahParry: That is correct based on how it is written right now. 
Say `x.line_iter()` has 5 lines in it, we cannot partition it beyond 5 in that case.

It could be altered to check if n is greater than the number of lines and then using `x.densify(x.euclidean_length() / n))`. I suspect that could be done by taking total line length, dividing by the desired number of segments to calculate the appropriate value for densify.

michaelkirk:
BTW have you seen https://github.com/georust/geo/pull/1050?

It's not quite the same thing, but it seems like maybe you could leverage the functionality within. Though, I'm not sure if you'd want to, I haven't thought about it very hard.

JosiahParry:
#1050 is interesting and likely a bit more challenging than what i endeavored for here! I could give it a further look after this.

JosiahParry:
Recent commits allow for `n` to be greater than the number of `Line`s that compose a `LineString` which is done using the `Densify` trait to ensure that the distance between `Coord`s is `x.euclidean_length() / n`.

JosiahParry:
I've updated the branch with main. Please let me know what more is needed for this PR. Thanks!

JosiahParry:
Just following up here. Please let me know what more is needed on this PR. Thanks!

JosiahParry:
@michaelkirk / @urschrei, please let me know what tasks are remaining for this PR. Thanks! R bindings to geo are based on my fork with this functionality so a downstream package can improve their peformance by 4 orders of magnitude. I'd love to be able to point to main and keep up with new PRs :)

geo/src/algorithm/linestring_segment.rs
```diff
@@ -0,0 +1,186 @@
...
+                ln_vec.push(endpoint.into());
+
+                // now we drain all elements from the vector into an iterator
+                // this will be collected into a vector to be pushed into the
+                // results coord vec of vec
+                let to_push = ln_vec.drain(..);
+
+                // now collect & push this vector into the results vector
+                res_coords.push(to_push.collect::<Vec<Coord>>());
+
```
urschrei, line 85: Is the `drain` followed by `collect` intended to reuse `ln_vec`s allocation?

geo/src/lib.rs
```diff
@@ -150,13 +150,14 @@
 //! ## Miscellaneous
 //!
 //! - **[`Centroid`](Centroid)**: Calculate the centroid of a geometry
+//! - **[`ChaikinSmoothing`](ChaikinSmoothing)**: Smoothen `LineString`, `Polygon`, `MultiLineString` and `MultiPolygon` using Chaikins algorithm.
```
urschrei, line 153: Should be `Chaikin's`

geo/src/algorithm/linestring_segment.rs
```diff
@@ -0,0 +1,186 @@
+use crate::line_interpolate_point::LineInterpolatePoint;
+use crate::{Coord, Densify, EuclideanLength, LineString, LinesIter, MultiLineString};
+
+/// Segments a LineString into `n` equal length LineStrings as a MultiLineString.
+/// `None` will be returned when `n` is equal to 0 or when a point
+/// cannot be interpolated on a `Line` segment.
+pub trait LineStringSegmentize {
```
urschrei, line 7: We like to have a concise example in the form of a doctest for traits, under the `Examples` header. See e.g. `Densify`.

urschrei (changes requested):
Some minor comments and questions from me.

geo/src/lib.rs
JosiahParry: Okay, but this wasn't my typo! ;)

geo/src/algorithm/linestring_segment.rs
JosiahParry: Correct. I had to (or, didn't know another way to) allocate the vector outside of the forloop so elements in the vector could persist across iterations. The checks inside the loop determined when to push the Coords.

geo/src/lib.rs
urschrei: Modify a line --> fix all conceivable problems with the line. I don't make the rules (I do).

JosiahParry:
All comments are resolved. Praying to the CI overlords 🙏🏼

urschrei:
Guessing that was an inadvertent close.

JosiahParry:
oops! Correct. Thank you @urschrei. CI caught an example issue.

JosiahParry:
Changes should be resolved and CI is satisfied. @urschrei when you get a chance, let me know if I've missed anything! Thanks again for the feedback

urschrei (approved):
LGTM. I'll leave it open for another couple of days in case anyone else wants to take a look.

JosiahParry:
Sounds good! Enjoy you're weekend

urschrei:
[tolling bell]
Last call!

PR #1056, opened 2023-08-23
Performance improvements for Graham's scan

b4l:
---

### DIFF
--- a/geo/src/algorithm/convex_hull/graham.rs
+++ b/geo/src/algorithm/convex_hull/graham.rs
@@ -1,5 +1,6 @@
-use super::{swap_remove_to_first, trivial_hull};
+use super::trivial_hull;
 use crate::kernels::*;
+use crate::utils::lex_cmp;
 use crate::{Coord, GeoNum, LineString};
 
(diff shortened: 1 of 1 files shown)


urschrei:
Similar to my comment in #1057, could you say a little about how this improvement works?

b4l:
Theoretically, it should be an improvement as it may use fewer orientation invocations. The benches show mixed results though, with gains in the synthetic one while regressing the real-world ones, hence a draft. It may need a certain amount of points to become beneficial.

frewsxcv:
Thanks for opening this @b4l. My opinion is that we'll only consider a change like this if we can consistently see it being a performance win in the benches. So considering that, and that momentum has slowed on this pull request, I'm going to close it. If you feel like picking this back up, feel free to reopen.

b4l:
@frewsxcv, that is fine. I just left it like this in case someone else follows this path. Should the branch be deleted or kept?

frewsxcv:
I'll delete it but we can easily restore it if someone else picks it up.

PR #1057, opened 2023-08-24
Point in triangle and rect performance improvements

b4l:
---

### DIFF
--- a/geo/src/algorithm/contains/triangle.rs
+++ b/geo/src/algorithm/contains/triangle.rs
@@ -11,9 +11,33 @@ where
     T: GeoNum,
 {
     fn contains(&self, coord: &Coord<T>) -> bool {
-        let ls = LineString::new(vec![self.0, self.1, self.2, self.0]);
-        use crate::utils::{coord_pos_relative_to_ring, CoordPos};
-        coord_pos_relative_to_ring(*coord, &ls) == CoordPos::Inside
+        // leverageing robust predicates
+        // use crate::algorithm::kernels::Kernel;
+        // use crate::Orientation
+        // self.to_lines()
+        //     .map(|l| T::Ker::orient2d(l.start, l.end, *coord))
+        //     .windows(2)
+        //     .all(|win| win[0] == win[1] && win[0] != Orientation::Collinear)
+
+        // neglecting robust prdicates, hence faster
+        let p0x = self.0.x.to_f64().unwrap();
+        let p0y = self.0.y.to_f64().unwrap();
+        let p1x = self.1.x.to_f64().unwrap();
+        let p1y = self.1.y.to_f64().unwrap();
+        let p2x = self.2.x.to_f64().unwrap();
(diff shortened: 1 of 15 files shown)


urschrei:
Thanks for the PR @b4l! Could you say a little about how this perf improvement works? Is it purely down to switching away from using Robust? Are there any drawbacks (e.g. precision issues?)

b4l:
From my rather superficial understanding, the previous versions involved a conversion to a polygon, which allocates a vector that is considered expensive. Then subsequent, in the case of `rect` robust predicates are suboptimal as there is no interpolation in play since it is axis-aligned and a simple comparison is adequate. For the triangle cases, this implementation leverages ideas around barycentric coordinates and half-plane calculations, which are somewhat robust in terms of not involving any division, though presumably not as robust as `robust predicates`, as the name suggests. For the latter implementations based on robust predicates without allocation are commented out offering lesser performance improvements given the robustness trade-off.

Regarding precision issues, I have only a very basic understanding of the robustness and no idea of the implications this has in practice. The test /bench with the point on the triangle line works as expected though actually looking for reviewers' input regarding that.

urschrei:
Given the explanation (thank you!) and that it's passing tests currently, I'm inclined to merge this. Anyone else got input for or against?

urschrei:
Let's make a decision on this as it's a perf improvement. As per the author's comments on Discord:

> I can not vouch for the robustness, though there is a version using robust predicates for the triangle case commented out which still offers major performance improvements compared to the existing implementation. so maybe we can use this one for confidence?

So the conservative solution is to go with the robust predicate if there's some doubt about robustness. Alternatively, someone dreams up some more tests.

frewsxcv (approved):

rmanoka:
Thanks for the PR @b4l  ;  looks good except the non-robust impl for `Triangle`;  let's use the slightly slower robust impl. on that too.  

+1 for removing the `Vec` allocations though.

geo/src/algorithm/contains/triangle.rs
```diff
@@ -11,9 +11,33 @@ where
     T: GeoNum,
 {
     fn contains(&self, coord: &Coord<T>) -> bool {
-        let ls = LineString::new(vec![self.0, self.1, self.2, self.0]);
-        use crate::utils::{coord_pos_relative_to_ring, CoordPos};
-        coord_pos_relative_to_ring(*coord, &ls) == CoordPos::Inside
+        // leverageing robust predicates
```
rmanoka, line 14: Let's go with this impl. if it's correct, rather than the non-robust one.

For quick intro. to robust, you may want to refer [this example](https://github.com/georust/robust/tree/main/examples/predicate-map) that explains why it's important.

b4l:
Thanks for reviewing and sorry for the late response! I changed to the versions using robust predicates.

Should I remove the commented-out versions?

frewsxcv:
The ahash MSRV issue was resolved upstream https://github.com/tkaitchuck/aHash/pull/208

PR #1058, opened 2023-08-25
Minor: clippy - removed calls to .into_iter() where no needed.

martinfrances107:
The clippy ruleset has improved.

We can remove calls to .into_iter() in a few places.


michaelkirk (approved):
Thanks!

PR #1063, opened 2023-09-20
wkt macro to create geo-types

michaelkirk:
---

Inspired by #1061. I don't think it necessarily replaces #1061, but it might obviate it. It's arguable we might want both (or maybe neither 😢)... how do people feel?

I've applied the new macro to a sampling of tests to give people an idea of where it's nice / less nice.

### DIFF
--- a/geo-types/CHANGES.md
+++ b/geo-types/CHANGES.md
@@ -4,6 +4,8 @@
 
 * Add `Polygon::try_exterior_mut` and `Polygon::try_interiors_mut`.
   <https://github.com/georust/geo/pull/1071>
+* Add `wkt!` macro to define geometries at compile time.
+  <https://github.com/georust/geo/pull/1063>
 
 ## 0.7.11
 * Bump rstar dependency
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -26,6 +26,8 @@
   * <https://github.com/georust/geo/pull/1090>
 * Fix coordinate wrapping in `HaversineDestination`
   * <https://github.com/georust/geo/pull/1091>
+* Add `wkt!` macro to define geometries at compile time.
+  <https://github.com/georust/geo/pull/1063>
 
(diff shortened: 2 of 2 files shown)


geo-types/src/geometry/multi_point.rs
```diff
@@ -235,53 +235,50 @@ mod test {
...
     #[test]
     fn test_abs_diff_eq() {
         let delta = 1e-6;
 
-        let multi = MultiPoint::new(vec![point![x: 0., y: 0.], point![x: 10., y: 10.]]);
+        let multi = wkt! { MULTIPOINT(0. 0.,10. 10.) };
 
-        let multi_x = MultiPoint::new(vec![point![x: 0., y: 0.], point![x: 10.+delta, y: 10.]]);
+        let mut multi_x = multi.clone();
+        *multi_x.0[0].x_mut() += delta;
```
michaelkirk, line 267: Here's an example of where it's less nice. Unlike the `serde_json::json!` macro, you can't embed expressions, the `wkt!` macro only accepts literals.

culebron:
I like this, I think this is better that what I wrote.

urschrei:
This is great!

geo-types/src/macros.rs
```diff
@@ -262,15 +260,11 @@ macro_rules! polygon {
             $crate::line_string![
                 $($exterior_coord), *
             ],
-            <[_]>::into_vec(
```
michaelkirk, line 265: With the nostd build, we need to give a fully qualified path to the vec macro. That's the primary change in this commit.

But also I noticed these into_vec calls. I don't think we were gaining anything by building an array and then boxing it into a Vec vs. building the `vec!` directly, so I've changed it.

urschrei:
I'd like to merge this. Anyone else want to chime in? Speak now (or any time before tomorrow morning)

urschrei:
now what

PR #1065, opened 2023-09-20
Make `Polygon` fields public.

gibbz00:
`Polygon::{exterior,interiors}_mut()` are for example insufficient when using closures that return a `Result`.

I do realize that this might cause issues with polygons that aren't closed if instantiation is created without the proper `new` constructor, so there's probably a better way to go about this. I simply chose the method that resulted in the least amount of changes possible.

Example:

```rust
/// Old
impl Transform for Polygon {
    fn transform_coordinates<F: TransformClosure>(&mut self, f: &mut F) -> Result<()> {
        let (mut exterior, mut interiors) = self.clone().into_inner();
        exterior.transform_coordinates(f)?;
        interiors
            .iter_mut()
            .try_for_each(|line_string| line_string.transform_coordinates(f))?;
        *self = Polygon::new(exterior, interiors);
        Ok(())
    }
}

/// New
impl Transform for Polygon {
    fn transform_coordinates<F: TransformClosure>(&mut self, f: &mut F) -> Result<()> {
        self.exterior.transform_coordinates(f)?;
        self.interiors
            .iter_mut()
            .try_for_each(|interior| interior.transform_coordinates(f))
    }
}
``` 
---

### DIFF
--- a/geo-types/CHANGES.md
+++ b/geo-types/CHANGES.md
@@ -1,5 +1,9 @@
 # Changes
 
+## 0.7.12 (Unrealeased)
+
+* Make `Polygon` fields public.
+
 ## 0.7.11
 * Bump rstar dependency
   <https://github.com/georust/geo/pull/1030>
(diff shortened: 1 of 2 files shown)


frewsxcv:
Right now `Polygon` is _guaranteed_ to be close, and by publicizing the fields we lose that guarantee. I'd rather explore other options to increase ergonomics and keep our guarantee before resorting to this.

gibbz00:
> Right now `Polygon` is _guaranteed_ to be close, and by publicizing the fields we lose that guarantee. I'd rather explore other options to increase ergonomics and keep our guarantee before resorting to this.

Gotcha. Got an alternative solution to work. Sending in a new PR soon.

PR #1066, opened 2023-09-21
Add `ToRadians` and `ToDegrees` traits. 

gibbz00:
Previous struct methods `Point::to_degrees(self)` and `Point::to_radians(self)` are now behind the respective traits.

---

### DIFF
--- a/geo-types/CHANGES.md
+++ b/geo-types/CHANGES.md
@@ -1,5 +1,10 @@
 # Changes
 
+## Unreleased
+
+* Add `ToRadians` and `ToDegrees` traits. Previous struct methods `Point::to_degrees(self)` and `Point::to_radians(self)` are now behind the respective traits.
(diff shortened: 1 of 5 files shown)


michaelkirk:
Since geo-types is an interop crate, we try to keep it pretty svelte, and add functionality in the geo crate when possible. My instinct is that this new trait should live in geo, and we should keep the old (redundant) methods where they were for backwards compatibility, but what do other people think?

gibbz00:
My project only uses geo_types, but I'm fine with moving it to geo. Might just become a bit odd that the to_radians/to_degrees methods exists for Point in geo_types but that one would have to pull in `geo` for the rest.

frewsxcv:
It's less than ideal since `geo` is heavier than `geo-types`, but I'd rather see this get added to `geo`. Then we could implement `ToRadians` (and maybe `FromRadians`?) for any `G: CoordsIter`.

PR #1067, opened 2023-09-21
Add 1.70, 1.71 to CI, bump MSRV to 1.65

michaelkirk:
---

1.65 was released more than 10 months ago: https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html

Notably (for me anyway because I keep trying to use it) this means we can use [let/else syntax](https://rust-lang.github.io/rfcs/3137-let-else.html). 

As a reminder, our policy has been to support at least the last 6 months (STABLE + 3 prior), but I usually bump it conservatively to whatever feature is causing annoyance.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,7 +53,7 @@ jobs:
       matrix:
         container_image:
           # Use the latest stable version. No need for older versions.
-          - "georust/geo-ci:proj-9.2.1-rust-1.70"
+          - "georust/geo-ci:proj-9.2.1-rust-1.72"
     container:
       image: ${{ matrix.container_image }}
     steps:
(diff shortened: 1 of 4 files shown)


lnicola:
We should also bump `rust-version` in the manifest files.

michaelkirk:
> We should also bump rust-version in the manifest files.

Good catch! Done.

lnicola:
There's also `geo-postgis` and maybe even others. And the error here:

```
error: private item shadows public glob re-export
   --> geo/src/lib.rs:206:1
    |
206 | extern crate proj;
    | ^^^^^^^^^^^^^^^^^^
    |
note: the name `proj` in the type namespace is supposed to be publicly re-exported here
   --> geo/src/lib.rs:209:9
    |
209 | pub use crate::algorithm::*;
    |         ^^^^^^^^^^^^^^^^^^^
note: but the private item here shadows it
   --> geo/src/lib.rs:206:1
    |
206 | extern crate proj;
    | ^^^^^^^^^^^^^^^^^^
    = note: `-D hidden-glob-reexports` implied by `-D warnings`
```

lnicola:
Heads-up: I'm gonna try the rebase button :smile:.

lnicola:
Not sure what's up with the failing tests. `cs2cs EPSG:2230 EPSG:26946` works fine in `georust/geo-ci:proj-9.2.1-rust-1.72`.

michaelkirk:
I'm not sure what's going on with the proj build.

These lines are suspicious:
```
proj_create: Cannot find proj.db
test algorithm::map_coords::test::test_fallible_proj ... FAILED
proj_create: no database context specified
```

I think that rust-1.72 is the first release on the bookworm distro. Not sure if that's relevant. It did force me to change the libtiff dependency to something that would (I hoped) work across releases.

frewsxcv (approved):
once tests pass ✅

michaelkirk:
Fixed with https://github.com/georust/docker-images/pull/26/commits/112764c97c46e8d7ffc9e94e4921ffa3877820a1

I'm not sure what's wrong with our setup vs. bookworm, but I'm happy to kick that can down the road.

PR #1068, opened 2023-09-21
Remove unnecessary `extern crate` declarations.

michaelkirk:
---

`extern crate` used to be needed for any dependency and to use macros. As of rust 2018, this is no longer true.

The remaining use cases we have for `extern crate` are rare:

1. importing macros into the global namespace for frequently used macros
2. for sysroot crates like alloc


lnicola (approved):
I think we could even drop the rest (except `alloc`), but it's fine for now.

PR #1070, opened 2023-09-25
Add `ToDegrees`, `ToRadians`.

gibbz00:
New version from feedback given in https://github.com/georust/geo/pull/1066.

---

### DIFF
--- a/geo/src/algorithm/convert_angle_unit.rs
+++ b/geo/src/algorithm/convert_angle_unit.rs
@@ -1,0 +1,84 @@
+use geo_types::Coord;
+use geo_types::CoordFloat;
+
+use crate::{MapCoords, MapCoordsInPlace};
+
+pub trait ToRadians<T: CoordFloat>:
+    Sized + MapCoords<T, T, Output = Self> + MapCoordsInPlace<T>
+{
+    fn to_radians(&self) -> Self {
+        self.map_coords(|Coord { x, y }| Coord {
+            x: x.to_radians(),
+            y: y.to_radians(),
+        })
+    }
+
+    fn to_radians_in_place(&mut self) {
+        self.map_coords_in_place(|Coord { x, y }| Coord {
+            x: x.to_radians(),
+            y: y.to_radians(),
+        })
+    }
+}
+impl<T: CoordFloat, G: MapCoords<T, T, Output = Self> + MapCoordsInPlace<T>> ToRadians<T> for G {}
+
+pub trait ToDegrees<T: CoordFloat>:
+    Sized + MapCoords<T, T, Output = Self> + MapCoordsInPlace<T>
+{
+    fn to_degrees(&self) -> Self {
+        self.map_coords(|Coord { x, y }| Coord {
+            x: x.to_degrees(),
+            y: y.to_degrees(),
+        })
+    }
+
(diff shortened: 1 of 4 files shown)


geo/src/algorithm/convert_angle_unit.rs
```diff
@@ -0,0 +1,83 @@
...
+        self.map_coords(|Coord { x, y }| Coord {
+            x: x.to_radians(),
+            y: y.to_radians(),
+        })
+    }
+}
+impl<T: CoordFloat, G: MapCoords<T, T, Output = Self>> ToRadians<T> for G {}
+
+pub trait ToRadiansInPlace<T: CoordFloat>: MapCoordsInPlace<T> {
+    fn to_radians_in_place(&mut self) {
```
michaelkirk, line 17: Can you put this method in `ToRadians` and get rid of the `ToRadiansInPlace` trait?

e.g. here's how we do it with the [`AffineOps`](https://github.com/georust/geo/blob/main/geo/src/algorithm/affine_ops.rs#L47) trait:

```
impl<T: CoordNum, M: MapCoordsInPlace<T> + MapCoords<T, T, Output = Self>> AffineOps<T> for M
```

My motivation is that people looking for one method might actually want the other and IMO it's easier to find when they're in the same trait. 

For the same reason, I wish we could merge `MapCoords` and `MapCoordsInPlace` but it gets weird because `MapCoordsInPlace` methods don't utilize the NT in `MapCoords<_, NT>`.

michaelkirk (changes requested):

geo/src/algorithm/convert_angle_unit.rs
```diff
@@ -0,0 +1,83 @@
...
+#[cfg(test)]
+mod tests {
+    use std::f64::consts::PI;
+
+    use approx::assert_relative_eq;
+    use geo_types::Line;
+
+    use super::*;
+
+    lazy_static::lazy_static! {
```
frewsxcv, line 55: Rather than adding this dependency, can you just turn these into functions that return the `Line`s? Just having `lazy_static` used in this one spot doesn't seem worth the addition of a dependency

geo/src/algorithm/convert_angle_unit.rs
gibbz00: Fixed :)

geo/src/algorithm/convert_angle_unit.rs
gibbz00: Fair, resolved 🤗 Using lazy statics for mocking is a "design pattern" I've gotten used to liking. Incorporating that pattern should be probably be done in all tests where it would be appropriate, and in separate (low-priority) issue+pr.

michaelkirk (approved):
Thanks! I'll give it another day before merging case anyone wants to weigh in further.

frewsxcv (approved):

frewsxcv:
Actually one more thing before this merges– can you add this to the index of algorithms here? https://docs.rs/geo/0.26.0/geo/#algorithms

gibbz00:
> Actually one more thing before this merges– can you add this to the index of algorithms here? https://docs.rs/geo/0.26.0/geo/#algorithms

Fixed!

lnicola (approved):

PR #1076, opened 2023-10-01
Segmentize tests

Robinlovelace:


Robinlovelace:
Not intended to be merged, more as a sanity check related to https://github.com/JosiahParry/rsgeo/issues/28 so heads-up on that @JosiahParry: seems that the Rust output is correct.

JosiahParry:
The x and y values are places sequentially in your test. Use this LineString and the result returns 3 instead of 4. I'm looking into it! 
```rust
    let lns = vec![
        coord!{ x: 324957.7, y: 673670.1 },
        coord!{ x: 324957.9, y: 673680.1 },
        coord!{ x: 324959.9, y: 673686.8 },
        coord!{ x: 324961.9, y: 673693.4 },
        coord!{ x: 324963.8, y: 673699.0 },
        coord!{ x: 324969.6, y: 673710.0 },
        coord!{ x: 324976.7, y: 673722.1 },
        coord!{ x: 324996.4, y: 673742.9 }
      ]; 
```

Robinlovelace:
Aha, seems like there IS a bug on the Rust side then.

Robinlovelace:
Coord order updated. Curious to know the cause, it's a bit of a niche one, 100+ similar linestrings worked fine, shows value of testing on real data!

JosiahParry:
In the case of this bug it is important to go into the if statement when `i < (nlines - 1)` but it breaks other tests in some cases. So there is some condition in this if statement that I haven't figured out. To fix the bug remove `- 1` but introduces bugs elsewhere. 

https://github.com/georust/geo/blob/9edfae99f1f51f3b582132da5bf9105c9635f24a/geo/src/algorithm/linestring_segment.rs#L99

JosiahParry:
Got it sorted. Will make a PR shortly. Thank you!

Robinlovelace:
Bingo! Quick work.

Issue #1075, opened 2023-10-01
Add more tests for `line_segmentize`


Robinlovelace:
As shown in https://github.com/JosiahParry/rsgeo/issues/28 there's an example where n is set to 4 but only 3 geometries are returned.

Not sure what the underlying cause is but thought it worth adding more tests to ensure it's not on the Rust side.

urschrei:
Just checking: are these tests in addition to what's in #1078?

JosiahParry:
I think the tests I've added in #1078 should cover the cases

PR #1077, opened 2023-10-01
Fix LineSegmentize where number of linestrings is less than n

JosiahParry:
---

The bug: `LineStringSegmentize` trait can return `n - 1` LineStrings. This happens when the last `Line` being iterated over needs to be split into 2 separate Lines. 

The fix: previously, the definition did not permit splitting the last `Line` iterator. This PR allows the last Line to be split and checks after the for loop has completed iteration if the length is equal to `n`. 

Closes: https://github.com/georust/geo/issues/1075

### DIFF
--- a/geo/src/algorithm/linestring_segment.rs
+++ b/geo/src/algorithm/linestring_segment.rs
@@ -180,7 +180,6 @@ mod test {
         assert_eq!(segments.0.len(), 4);
 
         assert_eq!(segments.euclidean_length(), linestring.euclidean_length());
-        
     }
     #[test]
     // that 0 returns None and that usize::MAX returns None
--- a/geo/src/algorithm/linestring_segment.rs
+++ b/geo/src/algorithm/linestring_segment.rs
@@ -76,6 +76,7 @@ impl LineStringSegmentize for LineString {
         // calculate the target fraction for the first iteration
         // fraction will change based on each iteration
         let segment_prop = (1_f64) / (n as f64);
+
         let segment_length = total_length * segment_prop;
 
         // instantiate the first Vec<Coord>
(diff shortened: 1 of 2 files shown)


geo/src/algorithm/linestring_segment.rs
```diff
@@ -96,7 +97,9 @@ impl LineStringSegmentize for LineString {
             // update cumulative length
             cum_length += length;
 
-            if (cum_length >= segment_length) && (i != (n_lines - 1)) {
+            // if the cumulative lenght is greater than or equal to the segment length we
+            // must cut the line in the middle and push the coords
+            if (cum_length >= segment_length) || (cum_length == segment_length) {
```
JosiahParry, line 102: Here we remove the check for the i = n_lines to allow the last iteration to enter the if statement.

geo/src/algorithm/linestring_segment.rs
```diff
@@ -147,6 +155,32 @@ mod test {
...
+        .into();
+        let segments = linestring.line_segmentize(2).unwrap();
+        assert_eq!(segments.0.len(), 2);
+        let segments = linestring.line_segmentize(3).unwrap();
+        assert_eq!(segments.0.len(), 3);
+        let segments = linestring.line_segmentize(4).unwrap();
+        assert_eq!(segments.0.len(), 4);
+
+        assert_eq!(segments.euclidean_length(), linestring.euclidean_length());
+    }
```
JosiahParry, line 183: ensuring that the length of the multilinestring and the original linestring are identical. This ensures that we're not omitting any segments of the linestring that we need.

JosiahParry:
Closing in favor of #1078

PR #1078, opened 2023-10-01
Address LineStringSegmentize bugs

JosiahParry:
---

Thanks to @Robinlovelace for doing tests on _real_ data. Brought to light 2 categories of edge cases where the number of LineStrings in the resultant MultiLineString was not the same quantity as requested by the `n` parameter.

Bug occurred when: 

- There were only 2 coordinates in the LineString and
- When the `Line` length was larger than the maximum size of a segment

This has been addressed through the use of the `Densify` trait. 

LineStrings are densified to ensure that a Line segment never has to be partitioned more than once. Additionally, after densification the lines are counted. If the number of lines is exactly `n`, then the function is exited early. 

An additional change was made which incorrectly would return the `Line` of a LineString if there were `n` lines. This was wrong because the length may have not been equivalent. I have removed this check and early return. I've removed a test that tested for this equivalence since it was a bad assumption.

I've also removed the unneeded `peekable()` call.

------

Aids downstream: 

- https://github.com/JosiahParry/rsgeo/issues/28
- https://github.com/ropensci/stplanr/issues/546

### DIFF
--- a/geo/src/algorithm/linestring_segment.rs
+++ b/geo/src/algorithm/linestring_segment.rs
@@ -65,7 +65,7 @@ impl LineStringSegmentize for LineString {
         // Line more than once.
         let densified = self.densify(segment_length);
 
-        // if the densified lines is exactly equal to the number of requested
+        // if the densified line is exactly equal to the number of requested
         // segments, return early. This will happen when a LineString has
         // exactly 2 coordinates
         if densified.lines().count() == n {
--- a/geo/src/algorithm/linestring_segment.rs
+++ b/geo/src/algorithm/linestring_segment.rs
@@ -71,7 +71,7 @@ impl LineStringSegmentize for LineString {
         if densified.lines().count() == n {
             let linestrings = densified
                 .lines()
-                .map(|xi| LineString::from(xi))
+                .map(LineString::from)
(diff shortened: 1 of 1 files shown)


geo/src/algorithm/linestring_segment.rs
```diff
@@ -78,6 +60,27 @@ impl LineStringSegmentize for LineString {
         let segment_prop = (1_f64) / (n as f64);
         let segment_length = total_length * segment_prop;
 
+        // densify the LineString so that each `Line` segment is not longer
+        // than the segment length ensuring that we will never partition one
+        // Line more than once.
+        let densified = self.densify(segment_length);
+
+        // if the densified lines is exactly equal to the number of requested
```
urschrei, line 68: Should be "line"

urschrei (approved):
Just a tiny typo and then I'm happy to merge.

geo/src/algorithm/linestring_segment.rs
JosiahParry: et voila! Thank you @urschrei

JosiahParry:
Thanks for your speedy help, @urschrei !!

PR #1080, opened 2023-10-04
Add example for `scale_around_point`.

frewsxcv:
---


michaelkirk (approved):
Good catch!

PR #1081, opened 2023-10-06
add DensifyHaversine trait

JosiahParry:
---

This PR adds a `DensifyHaversine` trait which enables users to densify line based spherical geometries. 

It provides the ground work for enabling linestring segmentation on geographical coordinates which can be used by downstream libraries such as [stplanr](https://github.com/ropensci/stplanr/issues/522).

### DIFF
--- a/geo/src/algorithm/densify_haversine.rs
+++ b/geo/src/algorithm/densify_haversine.rs
@@ -19,13 +19,13 @@ use crate::{HaversineIntermediate, HaversineLength};
 /// ```
 /// use geo::{coord, Line, LineString};
 /// use geo::DensifyHaversine;
-/// 
+///
 /// let line = Line::new(coord! {x: 0.0, y: 0.0}, coord! { x: 0.0, y: 1.0 });
 /// // known output
 /// let output: LineString = vec![[0.0, 0.0], [0.0, 0.5], [0.0, 1.0]].into();
 /// // densify
 /// let dense = line.densify_haversine(100000.0);
-/// assert_relative_eq!(dense, output, epsilon = f64::EPSILON);
+/// assert_eq!(dense, output);
 ///```
 pub trait DensifyHaversine<F: CoordFloat> {
     type Output;
--- a/geo/src/algorithm/densify_haversine.rs
+++ b/geo/src/algorithm/densify_haversine.rs
@@ -1,0 +1,277 @@
+use num_traits::FromPrimitive;
+
+use crate::{
+    CoordFloat, Line, LineString, MultiLineString, MultiPolygon, Point, Polygon, Rect, Triangle,
+};
+
+use crate::{HaversineIntermediate, HaversineLength};
+
+/// Returns a new spherical geometry containing both existing and new interpolated coordinates with
+/// a maximum distance of `max_distance` between them.
+///
+/// Note: `max_distance` must be greater than 0.
+///
+/// ## Units
+///
+/// `max_distance`: meters
+///
+/// # Examples
+/// ```
+/// use geo::{coord, Line, LineString};
+/// use geo::DensifyHaversine;
+///
+/// let line = Line::new(coord! {x: 0.0, y: 0.0}, coord! { x: 10.0, y: 10.0 });
+///
+/// // known output
+/// let output: LineString = vec![
+///    [ 0.0, 0.0 ],
+///    [ 1.0998995245640133, 1.1166572478004086 ],
+///    [ 2.2006349325565924, 2.232902789521755 ],
+///    [ 3.3030434132933286, 3.348323819287168 ],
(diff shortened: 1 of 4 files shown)


michaelkirk:
re: the CI failure - You might want to check out our usage of `approx` elsewhere for the sloppy floating point matching.

JosiahParry:
Oh thank goodness. That was an adventure!

urschrei (approved):
LGTM. I'll leave it open in case anyone else wants to kick the tires.

geo/src/algorithm/densify_haversine.rs
```diff
@@ -0,0 +1,251 @@
...
+    LineString<T>: HaversineLength<T>,
+{
+    type Output = LineString<T>;
+
+    fn densify_haversine(&self, max_distance: T) -> Self::Output {
+        let mut new_line = vec![];
+        self.lines()
+            .for_each(|line| densify_line(line, &mut new_line, max_distance));
+        // we're done, push the last coordinate on to finish
+        new_line.push(self.points().last().unwrap());
```
michaelkirk, line 126: We'll panic here if we have an empty linestring, right? I don't think we want that. Was this copied from the existing densify trait? Seems like we need to fix that too, but we could do it in a separate PR.

I think it's reasonable to return an empty line string when trying to densify an empty line string.

It looks to me like empty geometries of other types are already being handled reasonably.

geo/src/algorithm/densify_haversine.rs
```diff
@@ -0,0 +1,251 @@
...
+        let dense = linestring.densify_haversine(110.0);
+        assert_relative_eq!(dense, output, epsilon = f64::EPSILON);
+    }
+
+    #[test]
+    fn test_line_densify() {
+        let output: LineString = vec![[0.0, 0.0], [0.0, 0.5], [0.0, 1.0]].into();
+        let line = Line::new(coord! {x: 0.0, y: 0.0}, coord! { x: 0.0, y: 1.0 });
+        let dense = line.densify_haversine(100000.0);
+        assert_relative_eq!(dense, output, epsilon = f64::EPSILON);
```
michaelkirk, line 249: Take it or leave it feedback: f64::EPSILON is already the default epsilon. You only need to specify epsilon if you want a default epsilon.

https://docs.rs/approx/latest/src/approx/abs_diff_eq.rs.html#87

michaelkirk (changes requested):

geo/src/algorithm/densify_haversine.rs
JosiahParry: This was taken directly from the existing Densify trait. What would you like to do to proceed?

geo/src/algorithm/densify_haversine.rs
JosiahParry: I'll take it :)

geo/src/algorithm/densify_haversine.rs
urschrei: Something like?

```
fn densify_haversine(&self, max_distance: T) -> Self::Output {
    if self.coords_count() == 0 {
        return LineString::new(vec![]);
    }
    // keep going
```

I can add the same to `densify` as a separate PR

JosiahParry:
```rust

use geo::{LineString, Densify};

fn main() {
  let linestring: LineString<f64> = LineString::new(vec![]);
  linestring.densify(30.0);
}
```
panics as you expected

geo/src/algorithm/densify_haversine.rs
```diff
@@ -0,0 +1,263 @@
...
+    T: CoordFloat + FromPrimitive,
+    Line<T>: HaversineLength<T>,
+    LineString<T>: HaversineLength<T>,
+{
+    type Output = LineString<T>;
+
+    fn densify_haversine(&self, max_distance: T) -> Self::Output {
+        if self.coords_count() == 0 {
+            return LineString::new(vec![]);
+        }
```
JosiahParry, line 125: @michaelkirk I've only added this check here as it seems to be called everywherer else. Should this suffice?

geo/src/algorithm/densify_haversine.rs
```diff
@@ -0,0 +1,263 @@
...
+        let dense = line.densify_haversine(100000.0);
+        assert_relative_eq!(dense, output);
+    }
+
+    #[test]
+    fn test_empty_linestring() {
+        let linestring: LineString<f64> = LineString::new(vec![]);
+        let dense = linestring.densify_haversine(10.0);
+        assert_eq!(0, dense.coords_count());
+    }
```
JosiahParry, line 262: Added an additional test for linestrings. I don't think its possible to have an empty `Line` so i didnt add a test for it.

michaelkirk (approved):
Thanks!

PR #1083, opened 2023-10-09
Integrate `spade`

RobWalt:
This PR aims at bringing [`spade`](https://docs.rs/spade/latest/spade/) triangulations to `geo`. More specifically I'm looking into implementing [constrained delauny triangulations](https://docs.rs/spade/latest/spade/struct.ConstrainedDelaunayTriangulation.html) for `geo` since it seems handy in a lot of situations and opens up possibilities for new algorithms.

## Todos 

- [x] more tests?

## Commit summary

---

### chore: add `spade` as a dependency 

although spade is a optional dependency, it looks like a good fit for the default features. Including it has the effect of bringing in 3 new dependencies:

- `robust` 0.2 (we already use this on version 1.1)
- `optional` 0.5
- `spade` 2.2

I'll try to speak to the spade maintainers. Maybe we can bump the robust version in their crate to reduce dependencies further. Also it might make sense to ask them if the optional crate provides any real value.

Update: I created PRs for the changes

- the `robust` version bump should make it since it comes with performance improvements
- the `optional` removal might make it depending on the maintainers taste

Update: Both PRs got merged, so this change doesn't add any additional dependencies

---

### feat: initial implementation of `TriangulateSpade` trait

includes all the code for the integration of `spade` into `geo` and also features some tests

---

---

### DIFF
--- a/geo/src/algorithm/mod.rs
+++ b/geo/src/algorithm/mod.rs
@@ -252,7 +252,7 @@ pub mod triangulate_earcut;
 #[cfg(feature = "earcutr")]
 pub use triangulate_earcut::TriangulateEarcut;
 
-/// Triangulate polygons using an (un)constrained [delaunay triangulation](https://en.wikipedia.org/wiki/Delaunay_triangulation) algorithm.
+/// Triangulate polygons using an (un)constrained [Delaunay Triangulation](https://en.wikipedia.org/wiki/Delaunay_triangulation) algorithm.
 #[cfg(feature = "spade")]
 pub mod triangulate_spade;
 #[cfg(feature = "spade")]
--- a/geo/src/algorithm/mod.rs
+++ b/geo/src/algorithm/mod.rs
@@ -252,6 +252,12 @@ pub mod triangulate_earcut;
 #[cfg(feature = "earcutr")]
 pub use triangulate_earcut::TriangulateEarcut;
 
+/// Triangulate polygons using an [ear-cutting algorithm](https://www.geometrictools.com/Documentation/TriangulationByEarClipping.pdf).
+#[cfg(feature = "spade")]
+pub mod triangulate_spade;
+#[cfg(feature = "spade")]
+pub use triangulate_spade::TriangulateSpade;
+
 /// Vector Operations for 2D coordinates
 mod vector_ops;
 pub use vector_ops::Vector2DOps;
--- a/geo/src/algorithm/mod.rs
+++ b/geo/src/algorithm/mod.rs
@@ -252,7 +252,7 @@ pub mod triangulate_earcut;
 #[cfg(feature = "earcutr")]
 pub use triangulate_earcut::TriangulateEarcut;
 
-/// Triangulate polygons using an [ear-cutting algorithm](https://www.geometrictools.com/Documentation/TriangulationByEarClipping.pdf).
+/// Triangulate polygons using an (un)constrained [delaunay triangulation](https://en.wikipedia.org/wiki/Delaunay_triangulation) algorithm.
 #[cfg(feature = "spade")]
 pub mod triangulate_spade;
 #[cfg(feature = "spade")]
--- a/geo/src/algorithm/triangulate_spade.rs
+++ b/geo/src/algorithm/triangulate_spade.rs
@@ -93,7 +93,7 @@ where
     /// This outer triangulation also contains triangles that are not included in the input
     /// geometry if it wasn't convex. Here's an example:
     ///
-    /// ```ignore
+    /// ```text
     /// ┌──────────────────┐
     /// │\              __/│
     /// │ \          __/ / │
@@ -141,7 +141,7 @@ where
     /// This triangulation only contains triangles that are included in the input geometry.
     /// Here's an example:
     ///
-    /// ```ignore
+    /// ```text
     /// ┌──────────────────┐
     /// │\              __/│
     /// │ \          __/ / │
--- a/geo/src/algorithm/triangulate_spade.rs
+++ b/geo/src/algorithm/triangulate_spade.rs
@@ -55,6 +55,19 @@ mod private {
         fn get_points(&'a self) -> Vec<Coord<T>>;
         /// define a predicate that decides if a point is inside of the object (used for constrained triangulation)
         fn contains_point(&'a self, p: Point<T>) -> bool;
+
+        // processing of the lines that prepare the lines for triangulation.
+        //
+        // `spade` has the general
+        // limitation that constraint lines cannot intersect or else it will panic. This is why we need
+        // to manually split up the lines into smaller parts at the intersection point
+        //
+        // there's also a preprocessing step which tries to minimize the risk of failure of the algo
+        // through edge cases (thin/flat triangles are prevented as much as possible & lines are deduped, ...)
+        fn cleanup_lines(lines: Vec<Line<T>>) -> TriangulationResult<Vec<Line<T>>> {
+            let (known_points, lines) = preprocess_lines(lines);
+            prepare_intersection_contraint(lines, known_points)
+        }
     }
 }
 
@@ -63,26 +76,35 @@ pub trait TriangulateSpade<'a, T>: private::TriangulationRequirementTrait<'a, T>
 where
     T: SpadeTriangulationFloat,
 {
-    // processing of the lines that prepare the lines for triangulation.
-    //
-    // `spade` has the general
-    // limitation that constraint lines cannot intersect or else it will panic. This is why we need
-    // to manually split up the lines into smaller parts at the intersection point
-    //
-    // there's also a preprocessing step which tries to minimize the risk of failure of the algo
-    // through edge cases (thin/flat triangles are prevented as much as possible & lines are deduped, ...)
-    fn cleanup_lines(lines: Vec<Line<T>>) -> TriangulationResult<Vec<Line<T>>> {
-        let (known_points, lines) = preprocess_lines(lines);
-        prepare_intersection_contraint(lines, known_points)
-    }
-
     /// returns a triangulation that's solely based on the points of the geometric object
     ///
     /// The triangulation is guaranteed to be Delaunay
     ///
     /// Note that the lines of the triangulation don't necessarily follow the lines of the input
     /// geometry. If you wish to achieve that take a look at the `constrained_triangulation` and the
-    /// `constrained_outer_triangulation` functions
+    /// `constrained_outer_triangulation` functions.
+    ///
+    /// ```rust
+    /// use geo::TriangulateSpade;
+    /// use geo::{Polygon, LineString, Coord};
+    /// let u_shape = Polygon::new(
+    ///     LineString::new(vec![
+    ///         Coord { x: 0.0, y: 0.0 },
+    ///         Coord { x: 1.0, y: 0.0 },
+    ///         Coord { x: 1.0, y: 1.0 },
+    ///         Coord { x: 2.0, y: 1.0 },
+    ///         Coord { x: 2.0, y: 0.0 },
+    ///         Coord { x: 3.0, y: 0.0 },
+    ///         Coord { x: 3.0, y: 3.0 },
+    ///         Coord { x: 0.0, y: 3.0 },
+    ///     ]),
+    ///     vec![],
+    /// );
+    /// let unconstrained_triangulation = u_shape.unconstrained_triangulation().unwrap();
+    /// let num_triangles = unconstrained_triangulation.len();
+    /// assert_eq!(num_triangles, 8);
+    /// ```
+    ///
     fn unconstrained_triangulation(&'a self) -> TriangulationResult<Triangles<T>> {
         let points = self.get_points();
         points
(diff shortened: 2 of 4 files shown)


urschrei:
This is great, @RobWalt, thank you! We'll wait to see how the proposed changes to spade go.

RobWalt:
@urschrei Do you think it would be feasible to review & merge this as is and keep a tracking issue open for updating the dependencies which track the two spade PRs? I assume these two PRs are not going to be included anytime soon and this issue here is required to merge the spade boolops at some point

geo/src/algorithm/triangulate_spade.rs
```diff
@@ -0,0 +1,568 @@
...
+
+    /// returns triangulation that's based on the points of the geometric object and also
+    /// incorporates the lines of the input geometry
+    ///
+    /// The triangulation is not guaranteed to be Delaunay because of the constraint lines
+    ///
+    /// This outer triangulation also contains triangles that are not included in the input
+    /// geometry if it wasn't convex. Here's an example:
+    ///
+    /// ```ignore
```
urschrei, line 96: I think these diagram blocks should be marked as `text`, not `ignore`.

geo/src/algorithm/triangulate_spade.rs
```diff
@@ -0,0 +1,568 @@
...
+// ======= Float trait ========
+
+pub trait SpadeTriangulationFloat: GeoFloat + SpadeNum {}
+impl<T: GeoFloat + SpadeNum> SpadeTriangulationFloat for T {}
+
+// ======= Triangulation trait =========
+
+pub type Triangles<T> = Vec<Triangle<T>>;
+
+/// Triangulate polygons using a [denlauny triangulation](https://en.wikipedia.org/wiki/Delaunay_triangulation)
```
urschrei, line 40: denlauny -> Delaunay

geo/src/algorithm/triangulate_spade.rs
```diff
@@ -0,0 +1,568 @@
...
+    }
+}
+
+// ========== Triangulation trait impl helpers ============
+
+fn prepare_intersection_contraint<T: SpadeTriangulationFloat>(
+    mut lines: Vec<Line<T>>,
+    mut known_points: Vec<Coord<T>>,
+) -> Result<Vec<Line<T>>, TriangulationError> {
+    // safety net. We can't prove that the `while let` loop isn't going to run infinitely, so
```
urschrei, line 255: Could you document the conditions under which we might see an infinite loop? How did you arrive at the 1000 loops number?

geo/src/algorithm/triangulate_spade.rs
```diff
@@ -0,0 +1,568 @@
...
+            {
+                lines.push(line)
+            }
+
+            lines
+        });
+    (known_points, lines)
+}
+
+/// convertes Line to something somewhat similar in the spade world
```
urschrei, line 447: convertes -> converts

geo/src/algorithm/triangulate_spade.rs
```diff
@@ -0,0 +1,568 @@
...
+fn to_spade_line<T: SpadeTriangulationFloat>(line: Line<T>) -> [Point2<T>; 2] {
+    [to_spade_point(line.start), to_spade_point(line.end)]
+}
+
+/// convertes Coord to something somewhat similar in the spade world
```
urschrei, line 452: convertes -> converts

urschrei (changes requested):

geo/src/algorithm/triangulate_spade.rs
urschrei: (You'll see the difference if you run `cargo doc --open` locally.)

urschrei:
I'm in favour of merging this, both because it's a pre-requisite for the new boolean ops functionality, and because `geo` should support Delaunay triangulation as a feature more generally. For some reason it's not currently showing up as a top-level item under the algorithms module – again, you can check for yourself by running `cargo doc --open`.

geo/src/algorithm/triangulate_spade.rs
```diff
@@ -0,0 +1,568 @@
...
+        let (known_points, lines) = preprocess_lines(lines);
+        prepare_intersection_contraint(lines, known_points)
+    }
+
+    /// returns a triangulation that's solely based on the points of the geometric object
+    ///
+    /// The triangulation is guaranteed to be Delaunay
+    ///
+    /// Note that the lines of the triangulation don't necessarily follow the lines of the input
+    /// geometry. If you wish to achive that take a look at the `constrained_triangulation` and the
```
urschrei, line 74: achive -> achieve

geo/src/algorithm/mod.rs
```diff
@@ -240,6 +240,12 @@ pub mod triangulate_earcut;
 #[cfg(feature = "earcutr")]
 pub use triangulate_earcut::TriangulateEarcut;
 
+/// Triangulate polygons using an (un)constrained [delaunay triangulation](https://en.wikipedia.org/wiki/Delaunay_triangulation) algorithm.
```
urschrei, line 243: Delaunay should always be capitalised in docs as it refers to someone's name.

geo/src/algorithm/triangulate_spade.rs
RobWalt: Thanks for that suggestion. I'll apply it immediately 👍🏼

RobWalt:
- applied the review comments suggestions
- rebased on main
- sealed the helper methods on a separate private trait since they probably shouldn't be public in the first place

geo/src/algorithm/triangulate_spade.rs
```diff
@@ -0,0 +1,710 @@
...
+        // there's also a preprocessing step which tries to minimize the risk of failure of the algo
+        // through edge cases (thin/flat triangles are prevented as much as possible & lines are deduped, ...)
+        fn cleanup_lines(lines: Vec<Line<T>>) -> TriangulationResult<Vec<Line<T>>> {
+            let (known_points, lines) = preprocess_lines(lines);
+            prepare_intersection_contraint(lines, known_points)
+        }
+    }
+}
+
+/// Triangulate polygons using a [Denlauny Triangulation](https://en.wikipedia.org/wiki/Delaunay_triangulation)
```
urschrei, line 74:
```suggestion
/// Triangulate polygons using a [Delaunay Triangulation](https://en.wikipedia.org/wiki/Delaunay_triangulation)
```

urschrei (approved):

geo/src/algorithm/triangulate_spade.rs
```diff
@@ -0,0 +1,710 @@
...
+    use super::*;
+    pub trait TriangulationRequirementTrait<'a, T>
+    where
+        T: SpadeTriangulationFloat,
+    {
+        /// collect all the lines that are relevant for triangulations from the geometric object that
+        /// should be triangulated.
+        ///
+        /// intersecting lines are allowed
+        fn get_lines(&'a self) -> Vec<Line<T>>;
```
michaelkirk, line 52:
```suggestion
        fn lines(&'a self) -> Vec<Line<T>>;
```

I try to follow these guidelines for cohesion with the rest of the rust world
https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

I realize it's private, so it's not a big deal, but might as well get used to it.

geo/src/algorithm/triangulate_spade.rs
```diff
@@ -0,0 +1,710 @@
...
+        /// collect all the points that are relevant for triangulations from the geometric object that
+        /// should be triangulated
+        fn get_points(&'a self) -> Vec<Coord<T>>;
```
michaelkirk, line 55:
```suggestion
        fn points(&'a self) -> Vec<Coord<T>>;
```

geo/src/algorithm/triangulate_spade.rs
```diff
@@ -0,0 +1,710 @@
...
+        .filter(|l| !existing_lines.contains(&Line::new(l.end, l.start)))
+        .collect::<Vec<_>>()
+}
+
+/// snap point to the nearest existing point if it's close enough
+fn snap_or_register_point<T: SpadeTriangulationFloat>(
+    point: Coord<T>,
+    known_points: &mut Vec<Coord<T>>,
+) -> Coord<T> {
+    let epsilon_range = <T as std::convert::From<f32>>::from(0.0001);
```
michaelkirk, line 481: Hard coded epsilons like this are tough when we don't know what projection our data is in. 

e.g. in wgs84 this could be ~10 meters. Is that too much? Is there a way we can infer a more appropriate magnitude based on the input?

geo/src/algorithm/triangulate_spade.rs
```diff
@@ -0,0 +1,710 @@
...
+    known_points
```
michaelkirk, line 482: Does it make sense to use something from our existing `Closest` trait here?

geo/src/algorithm/triangulate_spade.rs
```diff
@@ -0,0 +1,710 @@
...
+// ========== Triangulation trait impl helpers ============
+
+fn prepare_intersection_contraint<T: SpadeTriangulationFloat>(
+    mut lines: Vec<Line<T>>,
+    mut known_points: Vec<Coord<T>>,
+) -> Result<Vec<Line<T>>, TriangulationError> {
+    // safety net. We can't prove that the `while let` loop isn't going to run infinitely, so
+    // we abort after a fixed amount of iterations
+    let mut loop_count = 1000;
+    // in case of an error we have something to return (the scenario that triggered the infinite
```
michaelkirk, line 340: I'm not sure what's meant by this comment. Where do we return the scenario that triggered the infinite loop?

geo/src/algorithm/triangulate_spade.rs
```diff
@@ -0,0 +1,710 @@
...
+        fn get_lines(&'a self) -> Vec<Line<T>>;
+        /// collect all the points that are relevant for triangulations from the geometric object that
+        /// should be triangulated
+        fn get_points(&'a self) -> Vec<Coord<T>>;
+        /// define a predicate that decides if a point is inside of the object (used for constrained triangulation)
+        fn contains_point(&'a self, p: Point<T>) -> bool;
+
+        // processing of the lines that prepare the lines for triangulation.
+        //
+        // `spade` has the general
```
michaelkirk, line 61: nit: weird line wrap here. 

Could you also do a once over to use more conventional punctuation and capitalization (like you're writing a letter or other document) in your method documentation? 

It helps me with comprehension - especially when faced with weird line wraps.

geo/src/algorithm/triangulate_spade.rs
michaelkirk, line 55: Minor perf thing: Can we pass through the underlying iter to avoid allocating a Vec? It seems like it's only usage is here:

```
       let points = self.get_points();
        points
            .into_iter()
```

So it seems like there's no need to collect the iter into a vec since we immediately turn it back into an iter.


It'd be nice if the same applied to `get_lines`, but it seems harder to accomplish because `prepare_intersection_contraint` inserts and removes lines. It still might be possible with filter_map/flatten but it might not be worth the hassle in that case.

michaelkirk (approved):
Nice work! None of my comments should be considered blocking.

geo/src/algorithm/triangulate_spade.rs
RobWalt: Good point! I really don't like these epsilons anyways, but it was necessary at that time to introduce it. I can take a look at how this influences the triangulation and re-evaluate again. After that I could:

- tune the value again to it's limits
- write more docs to make the constant less magic

I would also be open to make it a top level argument, but then it would probably make the API a bit weird. If tuning the default value fails then I think I'll take this route though.

geo/src/algorithm/triangulate_spade.rs
RobWalt: I'll try it out and rebase the spade-boolops after that change locally to test if it introduces any significant differences. Thanks for that suggestion!

geo/src/algorithm/triangulate_spade.rs
RobWalt: Yeah, that is just poor wording. I was just trying to implement [Rule 2 of "The Power of 10" rules of NASA](https://en.wikipedia.org/wiki/The_Power_of_10%3A_Rules_for_Developing_Safety-Critical_Code). It just returns the `LoopTrap` Error to prevent indefinite loops although that didn't occur yet. But you never know 🤷🏼

geo/src/algorithm/triangulate_spade.rs
RobWalt: But I like the idea in general! Maybe I'll have the nerves for this another day 😅

geo/src/algorithm/triangulate_spade.rs
RobWalt: I like the idea to reuse the existent trait! I think it most likely isn't going to work out though. If we want to utilize the trait we probably need to clone the `known_points` to construct a `MultiPoint` to apply the trait on.

Looking a bit into the trait: For points it isn't doing anything drastically different then what is already there. It calculates the `euclidean_distance` and calculates the minimum by comparison. So I think I won't even need to performance test this any further.

RobWalt:
Thanks for the thorough review @michaelkirk! 🏅 I like your taste 👍🏼

RobWalt:
Ok, the proposed changes to spade were accepted and are merged as of now. There's nothing major in the way now. The only things left are:

- [ ] explore review comment about epsilon
- [ ] add more tests (?)

geo/src/algorithm/triangulate_spade.rs
michaelkirk: I made a PR for this here: https://github.com/RobWalt/geo/pull/1

Because different iterators are involved, it'd need to be a boxed trait (unless/until "return `impl Trait` from trait" ever gets stabilized).

geo/src/algorithm/triangulate_spade.rs
```diff
@@ -0,0 +1,710 @@
...
+        T: SpadeTriangulationFloat,
+    {
+        /// collect all the lines that are relevant for triangulations from the geometric object that
+        /// should be triangulated.
+        ///
+        /// intersecting lines are allowed
+        fn lines(&'a self) -> Vec<Line<T>>;
+        /// collect all the points that are relevant for triangulations from the geometric object that
+        /// should be triangulated
+        fn points(&'a self) -> Vec<Coord<T>>;
```
michaelkirk, line 55:
```suggestion
        fn coords(&'a self) -> Vec<Coord<T>>;
```

Sorry for not noticing this on my first review pass. We have two similar-but-different types `Coord` and `Point`, as such it's good to be careful and consistent with our naming.

geo/src/algorithm/triangulate_spade.rs
rmanoka: Agree with @michaelkirk .  Is it possible to make this a parameter of the algorithm, so that it's user input?

RobWalt: I just made it configurable by adding a `SpadeTriangulationConfig` struct which holds a documented `snap_radius` field and which implements `Default` with the previous epsilon value if you don't care about setting that value yourself.

rmanoka (approved):
LGTM;  thanks @RobWalt .

RobWalt:
😎

PR #1090, opened 2023-10-16
Add rhumb-line operations

apendleton:
---
This PR adds a bunch of operations that are roughly equivalent to most of the current Haversine operations in `geo`, but using [rhumb line](https://en.wikipedia.org/wiki/Rhumb_line) (a.k.a. "loxodrome") geometry instead of great-circle geometry. In particular, this PR adds `RhumbBearing`, `RhumbDestination`, `RhumbDistance`, `RhumbIntermediate`, and `RhumbLength` traits.

The code layout is a little different from the haversine ones (everything is in a new `rhumb` module), mainly because I wanted to share code between several of these since there's a lot of overlap. In particular, all of the operations that work on two points (so, bearing, distance, and intermediate) use a common calculations struct that does a bunch of the up-front shared setup. Obviously this is all rearrange-able though if there's a different preferred layout.

The `RhumbBearing`, `RhumbDestination`, and `RhumbDistance` implementations are heavily inspired by their Turf.js equivalents ([bearing](https://github.com/Turfjs/turf/blob/master/packages/turf-rhumb-bearing/index.ts), [destination](https://github.com/Turfjs/turf/blob/master/packages/turf-rhumb-destination/index.ts), and [distance](https://github.com/Turfjs/turf/blob/master/packages/turf-rhumb-distance/index.ts), respectively), which seem in turn to largely be inspired by the [Movable Type](https://www.movable-type.co.uk/scripts/latlong.html) rhumb line formula descriptions.

Turf doesn't have an `intermediate` operation for rhumb lines, so what's I've done here is "find the rhumb distance and rhumb bearing, calculate a fraction along that distance, and then calculate a rhumb destination," with (hopefully) most of the redundant trig factored out, as there's a fair bit of overlap between those operations. There might well be a more-direct trigonometric solution to this problem, though? Movable Type has a "rhumb midpoint" formula, but it doesn't seem to generalize to fractions other than 0.5; [pygeodesy](https://github.com/mrJean1/PyGeodesy/blob/82c42efa4c36335ec48c31d379c13663cd96d971/pygeodesy/sphericalBase.py#L494) _does_ seem to implement this concept, and special-cases 0.5 with the formula from Movable Type but also has a more general one, but honestly I found the code to be very difficult to understand. Hopefully what I've got is good enough for a first pass.

Tests are mostly appropriated from the equivalent Haversine traits and then adjusted so the values look right. I also separately confirmed that at least for the functions where Turf has equivalents, these implementations produce (approximately, allowing for float differences, etc.) the same values.

A couple of other odds and ends:
- **Naming and methodology:** all of the methods here use spherical geometry. Iterative methods for calculating these values on an oblate spheroid also exist (see, e.g., the ones in [`GeographicLib`](https://geographiclib.sourceforge.io/cgi-bin/RhumbSolve)), so there may be some future in which it might make sense to include those as well, as with `HaversineX` vs `GeodesicX`; if so, it may make sense to do something more qualified than `RhumbX` to make it clear what kind of rhumb/loxodrome operations these are? I'm not sure what that would be though.
- **Attribution:** as these are ports and refactors of existing code, I'm not sure if I should put references in comments in these files to where they came from and what their licenses originally were. It doesn't look like existing code that's of similar provenance does so now for the most part, but if that would be a good thing to do, let me know where it's preferred that it go, and I can add it.

### DIFF
--- a/geo/src/algorithm/rhumb/mod.rs
+++ b/geo/src/algorithm/rhumb/mod.rs
@@ -1,3 +1,11 @@
+//! This module provides rhumb-line (a.k.a. loxodrome) geometry operations.
+//! The distance, destination, and bearing implementations are adapted in part
+//! from their equivalents in [Turf.js](https://turfjs.org/), which in turn are
+//! adapted from the Movable Type
+//! [spherical geodesy tools](https://www.movable-type.co.uk/scripts/latlong.html).
+//! Turf.js is copyright its authors and the geodesy tools are copyright Chris
+//! Veness; both are available under an MIT license.
+
 use crate::{point, utils::normalize_longitude, CoordFloat, Point};
 use num_traits::FromPrimitive;
 
--- a/geo/src/algorithm/rhumb/mod.rs
+++ b/geo/src/algorithm/rhumb/mod.rs
@@ -1,0 +1,157 @@
+use crate::{point, utils::normalize_longitude, CoordFloat, Point};
+use num_traits::FromPrimitive;
+
+mod distance;
+pub use distance::RhumbDistance;
+
+mod bearing;
+pub use bearing::RhumbBearing;
+
+mod destination;
+pub use destination::RhumbDestination;
+
+mod intermediate;
+pub use intermediate::RhumbIntermediate;
+
+mod length;
+pub use length::RhumbLength;
+
+struct RhumbCalculations<T: CoordFloat + FromPrimitive> {
+    from: Point<T>,
+    to: Point<T>,
+    phi1: T,
+    delta_lambda: T,
+    delta_phi: T,
+    delta_psi: T,
+}
+
+impl<T: CoordFloat + FromPrimitive> RhumbCalculations<T> {
+    fn new(from: &Point<T>, to: &Point<T>) -> Self {
+        let pi = T::from(std::f64::consts::PI).unwrap();
+        let two = T::one() + T::one();
+        let four = two + two;
+
+        let phi1 = from.y().to_radians();
+        let phi2 = to.y().to_radians();
+        let mut delta_lambda = (to.x() - from.x()).to_radians();
+        // if delta_lambda is over 180° take shorter rhumb line across the anti-meridian:
+        if delta_lambda > pi {
+            delta_lambda = delta_lambda - (two * pi);
+        }
+        if delta_lambda < -pi {
+            delta_lambda = delta_lambda + (two * pi);
+        }
+
+        let delta_psi = ((phi2 / two + pi / four).tan() / (phi1 / two + pi / four).tan()).ln();
+        let delta_phi = phi2 - phi1;
+
+        RhumbCalculations {
+            from: *from,
+            to: *to,
+            phi1,
+            delta_lambda,
+            delta_phi,
+            delta_psi,
+        }
+    }
+
+    fn theta(&self) -> T {
+        self.delta_lambda.atan2(self.delta_psi)
+    }
+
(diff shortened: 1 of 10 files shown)


urschrei:
> Attribution: as these are ports and refactors of existing code, I'm not sure if I should put references in comments in these files to where they came from and what their licenses originally were. It doesn't look like existing code that's of similar provenance does so now for the most part, but if that would be a good thing to do, let me know where it's preferred that it go, and I can add it.

Any time this question arises, I think about https://macwright.com/2013/02/18/literate-jenks (Even though [nobody should be using Jenks in this day and age](https://observablehq.com/@visionscarto/natural-breaks), but I digress). So in answer to your question: you never know who's going to be spelunking in this code in 30 years, so any breadcrumbs you can leave will be appreciated – if you have time, please add them.

Otherwise, this seems like a clean well-organised module to me, and a great addition to the crate. Thank you!

geo/src/algorithm/rhumb/mod.rs
```diff
@@ -0,0 +1,157 @@
...
+) -> Point<T> {
+    let pi = T::from(std::f64::consts::PI).unwrap();
+    let two = T::one() + T::one();
+    let four = two + two;
+    let threshold = T::from(10.0e-12).unwrap();
+
+    let delta_phi = delta * theta.cos();
+    let mut phi2 = phi1 + delta_phi;
+
+    // check for some daft bugger going past the pole, normalise latitude if so
```
urschrei, line 133: I assume this comment is copied verbatim from the movable-type implementation?

apendleton: Indeed, it was in the movable-type one and Turf kept it, so I did too. Happy to clean it up though if you prefer something more... professional :sweat_smile:.

urschrei: Oh no, was just checking!

apendleton:
@urschrei added an attribution block to `rhumb/mod.rs` .

geo/src/lib.rs
```diff
@@ -157,8 +160,10 @@
...
 //! - **[`GeodesicDestination`](GeodesicDestination)**: Given a start point, bearing, and distance, calculate the destination point on a [geodesic](https://en.wikipedia.org/wiki/Geodesics_on_an_ellipsoid)
 //! - **[`GeodesicIntermediate`](GeodesicIntermediate)**: Calculate intermediate points on a [geodesic](https://en.wikipedia.org/wiki/Geodesics_on_an_ellipsoid)
-//! - **[`HaversineDestination`]**: Given a start point, bearing, and distance, calculate the destination point on a sphere
-//! - **[`HaversineIntermediate`](HaversineIntermediate)**: Calculate intermediate points on a sphere
+//! - **[`HaversineDestination`]**: Given a start point, bearing, and distance, calculate the destination point on a sphere assuming travel on a great circle
+//! - **[`HaversineIntermediate`](HaversineIntermediate)**: Calculate intermediate points on a sphere along a great-circle line
+//! - **[`RhumbDestination`]**: Given a start point, bearing, and distance, calculate the destination point on a sphere assuming travel along a rhumb line
+//! - **[`RhumbIntermediate`](HaversineIntermediate)**: Calculate intermediate points on a sphere along a rhumb line
```
frewsxcv, line 166:
```suggestion
//! - **[`RhumbIntermediate`](RhumbIntermediate)**: Calculate intermediate points on a sphere along a rhumb line
```

frewsxcv (approved):

PR #1091, opened 2023-10-16
Fix wrapping of longitudes in HaversineDestination

### LINKED ISSUE #1074: HaversineDestination results don't wrap at the antimeridian as expected
During a recent refactor, we switched some logic that had previously used `HaversineIntermediate` to use `HaversineDestination` instead, in a way that should have been equivalent (calculating great-circle intermediate values between two points). The results do seem in general to be equivalent, but HaversineDestination sometimes produces result values that are outside `[-180,180]` longitude even when the outputs are in-bounds, which I found somewhat surprising.

Looking at the docs, it doesn't seem like `geo` actually commits to any particular longitude representation (`[-180,180]` vs `[0,360]` in particular), so maybe this is left undefined on purpose and applications should just normalize? But notably, the equivalent `HaversineIntermediate` operation _does_ seem to wrap longitudes correctly as I'd expect. Seems like there's some value in consistency, or at least in documenting expected behavior.

Minimal `rust-script` reproduction that compares in-theory-equivalent `HaversineDestination` and `HaversineIntermediate` operations calculating points along an AM-crossing line:

```rust
#!/usr/bin/env rust-script
//! ```cargo
//! [dependencies]
//! geo = "0.26"
//! ```

use geo::{ point, HaversineDistance, HaversineDestination, HaversineBearing, HaversineIntermediate };

fn main() {
    let pt1 = point! { x: 170.0, y: -30.0 };
    let pt2 = point! { x: -170.0, y: -30.0 };

    for (start, end) in [(pt1, pt2), (pt2, pt1)] {
        println!("Going from {:?} to {:?}...", start.x_y(), end.x_y());

        let bearing = start.haversine_bearing(end);
        let total_distance = start.haversine_distance(&end);

        let mut dist = 0.;
        while dist < total_distance {
            let dest_point = start.haversine_destination(bearing, dist);
            let inter_point = start.haversine_intermediate(&end, dist / total_distance);

            println!("destination: {:?}, intermediate: {:?}", dest_point.x_y(), inter_point.x_y());

            dist += 250_000.;
        }
    }
}
```

prints:
```
Going from (170.0, -30.0) to (-170.0, -30.0)...
destination: (170.0, -29.999999999999996), intermediate: (170.0, -29.999999999999996)
destination: (172.5908050903293, -30.172086178256627), intermediate: (172.5908050903293, -30.172086178256635)
destination: (175.18932441403777, -30.293113311281495), intermediate: (175.18932441403777, -30.29311331128149)
destination: (177.79289734601838, -30.362706690084703), intermediate: (177.79289734601838, -30.362706690084703)
destination: (180.39880075944293, -30.380649596934468), intermediate: (-179.6011992405571, -30.380649596934468)
destination: (183.00428252156786, -30.346886007528365), intermediate: (-176.99571747843217, -30.346886007528365)
destination: (185.6065957562343, -30.261521293335424), intermediate: (-174.39340424376573, -30.261521293335424)
destination: (188.2030329670417, -30.124820901911836), intermediate: (-171.79696703295832, -30.12482090191183)
Going from (-170.0, -30.0) to (170.0, -30.0)...
destination: (-170.0, -29.999999999999996), intermediate: (-170.0, -29.999999999999996)
destination: (-172.5908050903293, -30.172086178256627), intermediate: (-172.5908050903293, -30.172086178256635)
destination: (-175.18932441403777, -30.293113311281495), intermediate: (-175.18932441403777, -30.29311331128149)
destination: (-177.79289734601838, -30.362706690084703), intermediate: (-177.79289734601838, -30.362706690084703)
destination: (-180.39880075944293, -30.380649596934468), intermediate: (179.6011992405571, -30.380649596934468)
destination: (-183.00428252156786, -30.346886007528365), intermediate: (176.99571747843217, -30.346886007528365)
destination: (-185.6065957562343, -30.261521293335424), intermediate: (174.39340424376573, -30.261521293335424)
destination: (-188.2030329670417, -30.124820901911836), intermediate: (171.79696703295832, -30.12482090191183)
```

Happy to help out with a PR or whatever if there's some consensus about what, if anything, ought to be done here, but just wanted to flag since it caught me off-guard.

Also FWIW, I compared the behavior to `turf.destination` in turf.js, and again somewhat to my surprise, it exhibits the same past-180 wrapping behavior, so maybe this is normal?

```js
> turf.destination([-170.0, -30.0], 1750000.0, -95.03836877329749, {units: 'meters'})
{
  type: 'Feature',
  properties: {},
  geometry: {
    type: 'Point',
    coordinates: [ -188.2030329670417, -30.124820901911832 ]
  }
}
```

apendleton:
---
Currently `HaversineDestination` sometimes produces output points with longitudes that aren't within [-180,180]. This PR adds logic to ensure that they're correctly wrapped.

**Note:** the wrapping logic is branchless so it's nice and speedy, but a drawback is that it slightly twiddles some floating-point results out like 10 places after the decimal, even for longitudes that don't need wrapping, which required adding some `epsilon = whatever` to some existing tests. If a branch-y version that only touches out of bounds longitudes is preferred, let me know, and I can make the swap (probably in this branch and also in #1090 , which uses the same new function).

fixes #1074

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -18,6 +18,8 @@
   * <https://github.com/georust/geo/pull/908>
 * Add `ToDegrees` and `ToRadians` traits.
   * <https://github.com/georust/geo/pull/1070>
+* Fix coordinate wrapping in `HaversineDestination`
+  * <https://github.com/georust/geo/pull/1091>
 
 ## 0.26.0
 
--- a/geo/src/algorithm/haversine_closest_point.rs
+++ b/geo/src/algorithm/haversine_closest_point.rs
@@ -25,10 +25,14 @@ use num_traits::FromPrimitive;
 /// ```
 /// # use geo::HaversineClosestPoint;
 /// # use geo::{Point, Line, Closest};
+/// use approx::assert_relative_eq;
 /// let line = Line::new(Point::new(-85.93942, 32.11055), Point::new(-84.74905, 32.61454));
 /// let p_from = Point::new(-84.75625, 31.81056);
-/// assert_eq!(line.haversine_closest_point(&p_from),
-///         Closest::SinglePoint(Point::new(-85.13337428852164, 32.45365659858937)));
+/// if let Closest::SinglePoint(pt) = line.haversine_closest_point(&p_from) {
+///     assert_relative_eq!(pt, Point::new(-85.13337428852164, 32.45365659858937), epsilon = 1e-6);
+/// } else {
+///     panic!("Closest::SinglePoint expected");
+/// }
 /// ```
 pub trait HaversineClosestPoint<T>
 where
@@ -458,7 +462,11 @@ mod test {
         let p_from = Point::new(8.15172, 77.40041);
 
         if let Closest::SinglePoint(pt) = line.haversine_closest_point(&p_from) {
-            assert_relative_eq!(pt, Point::new(5.481_094_923_165_54, 82.998_280_987_615_33));
+            assert_relative_eq!(
+                pt,
+                Point::new(5.481_094_923_165_54, 82.998_280_987_615_33),
+                epsilon = 1.0e-6
+            );
         } else {
             panic!("Did not get Closest::SinglePoint!");
         }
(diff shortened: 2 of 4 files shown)


michaelkirk (approved):
I'll leave this another day or so in case anyone else wants to review.

PR #1092, opened 2023-10-16
Add inverse method to AffineTransform

urschrei:
---
<s>This is technically breaking due to the addition of an additional trait bound (`Neg`). In practice, it's unlikely to affect library consumers.</s>

### DIFF
--- a/geo/src/algorithm/affine_ops.rs
+++ b/geo/src/algorithm/affine_ops.rs
@@ -279,7 +279,7 @@ impl<T: CoordNum> AffineTransform<T> {
     }
 }
 
-impl<T: CoordNum + Neg<Output = T>> AffineTransform<T> {
+impl<T: CoordNum + Neg> AffineTransform<T> {
     /// Return the inverse of a given transform. Composing a transform with its inverse yields
     /// the [identity matrix](Self::identity)
     #[must_use]
@@ -300,16 +300,20 @@ impl<T: CoordNum + Neg<Output = T>> AffineTransform<T> {
         if determinant == T::zero() {
             return None; // The matrix is not invertible
         }
-
         let inv_det = T::one() / determinant;
-        Some(Self::new(
-            e * inv_det,
-            -b * inv_det,
-            (b * yoff - e * xoff) * inv_det,
-            -d * inv_det,
-            a * inv_det,
-            (d * xoff - a * yoff) * inv_det,
-        ))
+
+        // If conversion of either the b or d matrix value fails, bail out
+        match (T::from(-b * inv_det), T::from(-d * inv_det)) {
+            (Some(inv_b), Some(inv_d)) => Some(Self::new(
+                e * inv_det,
+                inv_b,
+                (b * yoff - e * xoff) * inv_det,
+                inv_d,
+                a * inv_det,
+                (d * xoff - a * yoff) * inv_det,
+            )),
+            _ => None,
+        }
     }
 }
 
--- a/geo/src/algorithm/affine_ops.rs
+++ b/geo/src/algorithm/affine_ops.rs
@@ -37,7 +37,7 @@ use std::{fmt, ops::Mul, ops::Neg};
 ///     (x: -0.5688687, y: 5.5688687)
 /// ], max_relative = 1.0);
 /// ```
-pub trait AffineOps<T: CoordNum + Neg> {
+pub trait AffineOps<T: CoordNum> {
     /// Apply `transform` immutably, outputting a new geometry.
     #[must_use]
     fn affine_transform(&self, transform: &AffineTransform<T>) -> Self;
@@ -46,9 +46,7 @@ pub trait AffineOps<T: CoordNum + Neg> {
     fn affine_transform_mut(&mut self, transform: &AffineTransform<T>);
 }
 
-impl<T: CoordNum + Neg, M: MapCoordsInPlace<T> + MapCoords<T, T, Output = Self>> AffineOps<T>
-    for M
-{
+impl<T: CoordNum, M: MapCoordsInPlace<T> + MapCoords<T, T, Output = Self>> AffineOps<T> for M {
     fn affine_transform(&self, transform: &AffineTransform<T>) -> Self {
         self.map_coords(|c| transform.apply(c))
     }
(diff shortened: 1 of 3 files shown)


geo/src/algorithm/affine_ops.rs
```diff
@@ -183,6 +187,37 @@ impl<T: CoordNum> AffineTransform<T> {
...
+        let determinant = a * e - b * d;
+
+        if determinant == T::zero() {
+            return None; // The matrix is not invertible
+        }
+
+        let inv_det = T::one() / determinant;
+        Some(Self::new(
+            e * inv_det,
+            T::from(-b * inv_det).unwrap(),
```
michaelkirk, line 213: These `T::from` are kind of unfortunate.

To get rid of them, we could constrain these newly added `Neg`'s to instead be `Neg<Output=T>`? 

Or do we actually want to support Neg w/ other Outputs? I don't think it really even occurred to me before that you could specify the output of Neg. I'm guessing it's for when you want to do something like:
```
// unsigned
impl Neg for MyUnsignedNumber {
    // output of negation can be negative, so output is signed
   type Output= MySignedNumber;
    ...
}
```

maybe that's why? I have no idea.

Anyway, if you'd rather keep it, it's NBD. I wouldn't be super surprised if it's all optimized away anyway. 😅

michaelkirk (approved):
This is clearly useful. I had one take-it-or-leave it thing, but LGTM.

geo/src/algorithm/affine_ops.rs
```diff
@@ -111,16 +115,16 @@ impl<T: CoordNum, M: MapCoordsInPlace<T> + MapCoords<T, T, Output = Self>> Affin
 /// ], max_relative = 1.0);
 /// ```
 #[derive(Copy, Clone, PartialEq, Eq)]
-pub struct AffineTransform<T: CoordNum = f64>([[T; 3]; 3]);
+pub struct AffineTransform<T: CoordNum + Neg = f64>([[T; 3]; 3]);
```
lnicola, line 118: Do we actually need to add these `Neg` bounds?

urschrei: In some form, because they're needed by the arithmetic that produces the determinant (according to the compiler).
But I haven't tried @michaelkirk's suggestion (see his previous comment) yet.

lnicola: But it's only needed for `inverse`, right? It still seems to build if I split out `inverse` into its own `impl` block and drop the new `Neg` bounds from everywhere else:

```rust
impl<T: CoordNum + Neg> AffineTransform<T> {
    /// Return the inverse of a given transform. Composing a transform with its inverse yields
    /// the [identity matrix](Self::identity)
    pub fn inverse(&self) -> Option<Self>
    where
        <T as Neg>::Output: Mul<T>,
        <<T as Neg>::Output as Mul<T>>::Output: ToPrimitive,
    {
        let a = self.0[0][0];
        let b = self.0[0][1];
        let xoff = self.0[0][2];
        let d = self.0[1][0];
        let e = self.0[1][1];
        let yoff = self.0[1][2];

        let determinant = a * e - b * d;

        if determinant == T::zero() {
            return None; // The matrix is not invertible
        }

        let inv_det = T::one() / determinant;
        Some(Self::new(
            e * inv_det,
            T::from(-b * inv_det).unwrap(),
            (b * yoff - e * xoff) * inv_det,
            T::from(-d * inv_det).unwrap(),
            a * inv_det,
            (d * xoff - a * yoff) * inv_det,
        ))
    }
}
```

PS: should probably be `#[must_use]` too.

michaelkirk: Both of these are good points!

michaelkirk: (Though I'd recommend doing 

```
impl<T: CoordNum + Neg<Output=T>> AffineTransform<T> {
```

urschrei: > But it's only needed for `inverse`, right? It still seems to build if I split out `inverse` into its own `impl` block and drop the new `Neg` bounds from everywhere else:
> 
> 
> 
> ```rust
> 
> impl<T: CoordNum + Neg> AffineTransform<T> {
> 
>     /// Return the inverse of a given transform. Composing a transform with its inverse yields
> 
>     /// the [identity matrix](Self::identity)
> 
>     pub fn inverse(&self) -> Option<Self>
> 
>     where
> 
>         <T as Neg>::Output: Mul<T>,
> 
>         <<T as Neg>::Output as Mul<T>>::Output: ToPrimitive,
> 
>     {
> 
>         let a = self.0[0][0];
> 
>         let b = self.0[0][1];
> 
>         let xoff = self.0[0][2];
> 
>         let d = self.0[1][0];
> 
>         let e = self.0[1][1];
> 
>         let yoff = self.0[1][2];
> 
> 
> 
>         let determinant = a * e - b * d;
> 
> 
> 
>         if determinant == T::zero() {
> 
>             return None; // The matrix is not invertible
> 
>         }
> 
> 
> 
>         let inv_det = T::one() / determinant;
> 
>         Some(Self::new(
> 
>             e * inv_det,
> 
>             T::from(-b * inv_det).unwrap(),
> 
>             (b * yoff - e * xoff) * inv_det,
> 
>             T::from(-d * inv_det).unwrap(),
> 
>             a * inv_det,
> 
>             (d * xoff - a * yoff) * inv_det,
> 
>         ))
> 
>     }
> 
> }
> 
> ```
> 
> 
> 
> PS: should probably be `#[must_use]` too.

Ahh, it didn't occur to me to split it into its own impl. This is much better I think.

lnicola: We could also return `None` if the cast fails..

michaelkirk: Oh and breaking out this impl means it's no longer a breaking change, nice!

geo/src/algorithm/affine_ops.rs
```diff
@@ -312,7 +279,41 @@ impl<T: CoordNum + Neg> AffineTransform<T> {
     }
 }
 
-impl<T: CoordNum + Neg> fmt::Debug for AffineTransform<T> {
+impl<T: CoordNum + Neg<Output = T>> AffineTransform<T> {
```
lnicola, line 282: Should we leave this as `Neg` and return `None` if the conversion fails? It makes sense to me.

urschrei: Probably safer, although I'm not sure _how_ it could fail in the first place (`NaN` appearing somewhere?). Have a look at the latest…

lnicola: One more nit :⁠-⁠).

lnicola (approved):

PR #1100, opened 2023-10-27
Fix broken links in docs and fail CI unless docs build cleanly

michaelkirk:
- [-] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

Two changes: 

The first fails CI unless docs build cleanly. This should help us avoid broken links etc. in our docs.

The second commit (forthcoming) fixes the CI failure. I pushed up the first commit separately so you can see what it looks like when CI fails for this reason.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -173,3 +173,15 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
       - run: cargo bench --no-run
+
+  docs:
+    name: build documentation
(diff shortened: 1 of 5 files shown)


michaelkirk:
Ok! You can check the first commit's status to see what a failed docs build looks like, and then you can see it resolved in the second build.

urschrei:
This is _wonderful_

urschrei (approved):

frewsxcv (approved):
🎉

PR #1101, opened 2023-10-28
Utilize trait return impl (stable on nightly) to simplify CoordsIter

frewsxcv:
---

TODO:

* Run benchmarks and compare before/after

### DIFF
--- a/geo/src/algorithm/coords_iter.rs
+++ b/geo/src/algorithm/coords_iter.rs
@@ -1,20 +1,10 @@
-use std::fmt::Debug;
-
 use crate::geometry::*;
 use crate::{coord, CoordNum};
 
-use std::{fmt, iter, marker, slice};
-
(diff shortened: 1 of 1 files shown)


kylebarron:
Wow that's exciting!

geo/src/algorithm/coords_iter.rs
```diff
@@ -36,7 +26,7 @@ pub trait CoordsIter {
     /// assert_eq!(Some(geo::coord! { x: 30., y: 40. }), iter.next());
     /// assert_eq!(None, iter.next());
     /// ```
-    fn coords_iter(&self) -> Self::Iter<'_>;
+    fn coords_iter(&self) -> impl Iterator<Item = Coord<Self::Scalar>>;
```
kylebarron, line 29: I'm looking forward to this for geometry traits too

PR #1102, opened 2023-10-29
Add comments to BoolOps implementation.

andriyDev:
---
This part of the code relies on a lot of assumptions that are not obviously until you really dig into the code, and have a lot of components that are not really clear (what's the difference between a `Ring` and a `Snake`?).

I've taken a stab at documenting these things and adding a few more inline comments to hopefully make it more clear to future readers.

There are a couple parts I still don't really understand, which I've marked as TODOs. Hopefully with a review, someone else might have more insight and we can document this even better. I will also likely expand the scope of comments as I make more progress (so far this is just `assembly.rs`, but I may try to capture more of the BoolOps).

### DIFF
--- a/geo/src/algorithm/bool_ops/assembly.rs
+++ b/geo/src/algorithm/bool_ops/assembly.rs
@@ -58,7 +58,8 @@ impl<T: GeoFloat> RegionAssembly<T> {
                 // are part of a hole or a shell.
                 iter.prev_active(first)
                     .map(|(_, seg)| (seg.region.get(), Some(seg.snake_idx.get())))
-                    // TODO: Figure out why this should be defaulted to 0.
+                    // If the expression is not Some, the snake will complete to an exterior ring
+                    // anyway. Therefore we can set an arbitrary value for the parent snake index.
                     .unwrap_or_else(|| (false, Some(0)))
             } else {
                 // The first crossing is on the right. Since crossings are ordered from
@@ -87,7 +88,9 @@ impl<T: GeoFloat> RegionAssembly<T> {
                         parent_snake_idx = Some(
                             iter.prev_active(c)
                                 .map(|(_, seg)| seg.snake_idx.get())
-                                // TODO
+                                // If the expression is not Some, the snake will complete to an
+                                // exterior ring anyway. Therefore can set an arbitrary value for
+                                // the parent snake index.
                                 .unwrap_or(0),
                         );
                     }
@@ -98,7 +101,9 @@ impl<T: GeoFloat> RegionAssembly<T> {
                         parent_snake_idx = Some(
                             iter.prev_active(d)
                                 .map(|(_, seg)| seg.snake_idx.get())
-                                // TODO
+                                // If the expression is not Some, the snake will complete to an
+                                // exterior ring anyway. Therefore can set an arbitrary value for
+                                // the parent snake index.
                                 .unwrap_or(0),
                         );
                     }
@@ -431,7 +436,7 @@ impl<T: GeoFloat> Snake<T> {
 struct Segment<T: GeoFloat> {
     /// The geometry of the segment.
     geom: LineOrPoint<T>,
-    /// The "region" this segment belongs to. Unused by `LineAssembly`. TODO
+    /// The "region" this segment belongs to. Unused by `LineAssembly`.
     region: Cell<bool>,
     /// The index of the `Snake` that this segment belongs to. Initially 0 as an
     /// "invalid" value. Unused by `LineAssembly`.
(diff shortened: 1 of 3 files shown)


rmanoka:
Thanks @andriyDev ;  the poor docs in this impl. is primarily my fault.  We built this impl. incrementally, trying to fix up the floating point precision issues.  I somewhat gave-up midway, as I couldn't figure how to provably fix the Martinez-Rueda family of algorithms to work with f64/f32.

That said, the Assembly module, iirc, does not have floating point precision issues as it doesn't actually need to break any segments.  In fact, I think, we can even get this module to work without using `GeoFloat` at all (only `GeoNum`).

I'll do my best to review this in the next week or two.

andriyDev:
Hey no blame! The only way to make the code better is to work together :)

As far floating point issues, I had been working on my own Martinez-Rueda implementation and had also been struggling heavily with floating point issues myself. I've got an idea for why those are happening so I'll create a separate issue.

andriyDev:
I rebased onto main, and I added more comments for the `bool_ops` folder.

There are many changes we could make to simplify the code, but I think just documenting it is the first step.

In addition, this isn't the whole story - the main thing that's missing now is documenting `CrossingsIter` which mostly means going through `Sweep` with a fine-toothed comb.

geo/src/algorithm/bool_ops/op.rs
```diff
@@ -173,15 +209,22 @@ impl<T: Float, S: Spec<T>> Proc<T, S> {
...
 }
 
+/// An edge of a shape.
 #[derive(Clone)]
 struct Edge<T: Float, S: Spec<T>> {
+    /// The geometry of the edge.
     geom: LineOrPoint<T>,
+    /// The index of the shape this edge belongs to.
     idx: usize,
+    /// The region of this edge when the piece is going in the same direction as
```
andriyDev, line 219: I'm pretty sure this is wrong. I can sorta understand the intent, but it doesn't make sense to me exactly what the check below means.

frewsxcv:
@andriyDev This is still marked as draft. Is that intentional or is this ready for review?

andriyDev:
I'm happy to call this ready for review. More comments can always be another PR :)

geo/src/algorithm/bool_ops/assembly.rs
```diff
@@ -85,14 +98,15 @@ impl<T: GeoFloat> RegionAssembly<T> {
                         parent_snake_idx = Some(
                             iter.prev_active(d)
                                 .map(|(_, seg)| seg.snake_idx.get())
+                                // TODO
```
frewsxcv, line 101: Can you remove these TODOs?

andriyDev: I can, but these TODOs really need an explanation. They make no sense to me. I tried to remove these defaults as well and tests started to fail, so clearly they are necessary but I don't have a logical reason *how*.

frewsxcv: @rmanoka Do you recall why this `.unwrap_or(0)` is here?

rmanoka: I don't think the number here matters.  If the expression is not `Some()`, then the "snake" will complete to an exterior ring, which doesn't need a parent idx.  I tried verifying by setting random numbers inside `.unwrap_or()`, and tests seem to pass.

geo/src/algorithm/bool_ops/op.rs
urschrei: @rmanoka Can you shed any light on this?

geo/src/algorithm/bool_ops/assembly.rs
urschrei: @andriyDev Are you OK with this explanation?

urschrei:
I'd really like to see this merged in order to facilitate further investigation into `RegionAssembly`'s tendency to panic (see https://github.com/georust/geo/issues/1104) for a possible avenue to investigate. I think there's only one minor question left to be resolved!

andriyDev:
> I'd really like to see this merged in order to facilitate further investigation into `RegionAssembly`'s tendency to panic (see #1104 for a possible avenue to investigate). I think there's only one minor question left to be resolved!

I would be shocked if this code is "responsible" for the panic. The RegionAssembly is fairly straightforward. What's more likely in my mind is that bad data is coming in from Proc or Spec in boolean ops since those are much more complex IMO. But that's just my speculation haha.

urschrei:
> Regardless we (mostly I) should push this through to make progress on that

Either that or outright replacement with your Martinez-Rueda implementation, but that's all a separate discussion.

andriyDev:
> Either that or outright replacement with your Martinez-Rueda implementation, but that's all a separate discussion.

To be clear, mine is also riddled with panics LMAO. Maybe one day I'll find the "smoking gun".

geo/src/algorithm/bool_ops/assembly.rs
andriyDev: I've added the explanation to the comments. I believe this is ready to go then (unless there are other concerns). :)

frewsxcv (approved):
@andriyDev thank you so much for adding these comments!

PR #1105, opened 2023-11-03
Allow an AffineTransform to be constructed from a borrowed array

urschrei:
---

"borrowed array" was formerly "slice", which was wrong!

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -28,6 +28,8 @@
   * <https://github.com/georust/geo/pull/1091>
 * Add `wkt!` macro to define geometries at compile time.
   <https://github.com/georust/geo/pull/1063>
+* Allow affine transforms to be constructed from borrowed arrays.
+  <https://github.com/georust/geo/pull/1105>
 
 ## 0.26.0
 
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -42,7 +42,7 @@
 * Add `compose_many` method to `AffineOps`
   * <https://github.com/georust/geo/pull/1148>
 * Allow affine transforms to be constructed from borrowed arrays.
-  <https://github.com/georust/geo/pull/1105>
+  * <https://github.com/georust/geo/pull/1105>
 
 ## 0.27.0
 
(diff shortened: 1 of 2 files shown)


lnicola:
That's not a slice, is it?

urschrei:
Well this is embarrassing

urschrei:
how about now?

lnicola:
Better. This could probably be `A: AsRef<[T; 6]>`, but whatever.

Still, as prior art, `Ipv4Addr` only implements [`From<[u8; 4]>`](https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#impl-From%3C%5Bu8;+4%5D%3E-for-Ipv4Addr). And `[T; 6]` will be `Copy`, so I'm not sure it's worth adding it.

urschrei:
> Better. This could probably be `A: AsRef<[T; 6]>`, but whatever.

I actually tried this but got the annoying `trait objects must include the `dyn` keyword` error, which also doesn't help since the compiler says  `dyn AsRef<[T; 6]> + 'static` doesn't impl `Sized`. I gave up at that point, rightly or wrongly.

lnicola:
I meant

```rust
impl<T: CoordNum, A: AsRef<[T; 6]>> From<A> for AffineTransform<T> {
    fn from(arr: A) -> Self {
        let arr = arr.as_ref();
        Self::new(arr[0], arr[1], arr[2], arr[3], arr[4], arr[5])
    }
}
```

This would subsume both existing implementations (`[T; 6]` and `(T, T, T, T, T, T)`), as well as the one for `&[T; 6]`.

But as I mentioned, I don't think it's necessary. Do you have some example where it helps? Because the following already works today:

```rust
fn foo(arr: &[f64; 6]) {
    let _tr = AffineTransform::from(*arr);
}
```

urschrei:
> Do you have some example where it helps? Because the following already works today

Only in avoiding the *, which was annoying me…

lnicola:
Dunno, if `From<[u8; 4]>` is good enough for the standard library, maybe it's good enough for us? :slightly_smiling_face:.

urschrei:
Fair.

frewsxcv (approved):

frewsxcv:
https://github.com/georust/geo/issues/1149

PR #1107, opened 2023-11-05
Add LineStringSegmentizeHaversine

JosiahParry:
---

Supercedes https://github.com/georust/geo/pull/1106. 

This PR: 

- adds a new trait `LineStringSegmentizeHaversine` to segment a linestring using haversine distance instead of euclidean distance
  - this gives downstream users an option to segment linestrings on geographic as well as projected coordinates
- fixes a bug in `LineStringSegmentize` where `n - 1` linestrings were created for _very_ small distances
  - this was caused by precision in the 6-ish decimal place. It was fixed by subtracting `f64::EPSILON` from the calculated segment length that is passed to `densify()`
- turns the segmentize implementation into a macro.
  - the benefit of this is that segmentize for haversine does not need to duplicate code and other future implementations (hopefully geodesic) will not need more than a Densify and Length trait of the same type.

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -3,6 +3,8 @@
 ## Unreleased
 * Fix a panic when calculating the haversine closest point to a point intersecting the geometry
   * <https://github.com/georust/geo/pull/1119>
+* Add `LineStringSegmentizeHaversine` trait as a an alternative to `LineStringSegmentize` for geographic coordinates.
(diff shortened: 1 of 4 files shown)


JosiahParry:
Please let me know if this needs any additional work. :)

frewsxcv (approved):
lgtm! will see if others have thoughts before merging

JosiahParry:
thanks for your help getting this across the line! Thank you!

PR #1109, opened 2023-11-13
Add `len()` and `is_empty()` to MultiPoint.

### LINKED ISSUE #1099: MultiPoint::len
Nice crate! I'm enjoying testing it out.

My small feature request would be to add a `.len()` method to `MultiPoint` and similar types. Seems like something that should be really easy to do, considering it can be created from a `Vec`.

andriyDev:
---
Adds `len()` and `is_empty()` to `MultiPoint`.

Fixes #1099.


lnicola (approved):
Might be worth adding a `CHANGES.md` entry.

andriyDev:
> Might be worth adding a `CHANGES.md` entry.

Done! I wasn't sure if this was a big enough change to add to the changelist haha

RobWalt (approved):

PR #1114, opened 2023-11-17
Remove unneeded trait lifetime in geo-traits

kylebarron:
- [ ] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/main/CODE_OF_CONDUCT.md).
- [ ] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

Ref https://github.com/georust/geo/pull/1021#discussion_r1395166770.

I've been incubating traits in geoarrow-rs, so this matches the same PR I made there: https://github.com/geoarrow/geoarrow-rs/pull/253, which also has updated trait implementations on WKB objects.

Issue #1113, opened 2023-11-15
Thoughts on algorithms for trait-based geometries


kylebarron:
A continuation of previous trait-based geometry discussions https://github.com/georust/geo/discussions/838, https://github.com/georust/geo/pull/1021, https://github.com/georust/geo/pull/1019. 

Just putting some words down on the page because @urschrei nerdsniped me! On discord we were talking about coordinate representations for geo and geo-types and @urschrei asked what the blockers were on trait-based geometries. I responded that I thought figuring out the best way to implement algorithms on top of geometries was hard.

I started to play around with a bit of scratch traits, got something to compile, and figured I should jot down some notes in case it resonates with anyone else.

The main idea is having algorithm traits _per geometry type_ that rely on that type's geometry trait definition. Then a blanket implementation over every geometry type that implements that trait.

```rs
use crate::geo_traits::{self, GeometryTrait, PointTrait};
use arrow_array::OffsetSizeTrait;
use geo::CoordNum;

pub trait AreaPoint<T>: PointTrait
where
    T: CoordNum,
{
    fn signed_area(&self) -> T {
        // signed area implemented solely in terms of PointTrait's methods
        todo!()
    }

    fn unsigned_area(&self) -> T {
        todo!()
    }
}

impl<T: CoordNum, P: PointTrait<T = T>> AreaPoint<T> for P {}
```

to keep a top-level `Area` trait as well, we could have

```rs
pub trait Area<'a, T>: GeometryTrait<'a, T = T>
where
    T: CoordNum,
{
    fn signed_area(&'a self) -> T {
        use geo_traits::GeometryType::*;
        match self.as_type() {
            Point(x) => AreaPoint::signed_area(x),
            _ => todo!(),
        }
    }

    fn unsigned_area(&'a self) -> T {
        use geo_traits::GeometryType::*;
        match self.as_type() {
            Point(x) => AreaPoint::unsigned_area(x),
            _ => todo!(),
        }
    }
}

impl<'a, T: CoordNum, P: GeometryTrait<'a, T = T>> Area<'a, T> for P {}
```

The main downside of this might be the proliferation of public traits? Is there a way to get around `AreaPoint`, `AreaLineString`, etc?

Who knows, this might fall apart on more complex traits like `Intersection` that require two geometries? Idk, thoughts?

PR #1115, opened 2023-11-17
Implement Area based on traits

kylebarron:
---

An exploration of https://github.com/georust/geo/issues/1113, based on top of #1114. Having a bunch of traits is verbose 😕 . Is there any way to group lots of traits as one? Wishing for something like
```rs
pub mod AreaTraits {
    pub use super::AreaPoint;
    pub use super::AreaLineString;
}
```
where importing `AreaTraits` would add all of the internal traits to scope. Guessing that doesn't exist, and probably for a good reason.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,18 +6,17 @@ on:
       - trying
       - release/**
   pull_request:
+  merge_group:
   schedule: [cron: "45 6 * * *"]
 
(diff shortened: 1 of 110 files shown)


frewsxcv:
I hope it's okay I converted this to a draft PR, since it's not really ready for a review until we figure out a plan for geo-traits

PR #1120, opened 2023-11-21
feat: implement `Validify` repair trait

RobWalt:
# Description

While discouraging the handling of invalid geometries and prioritizing the maintenance of geometric validity as a fundamental principle, it is recognized that geometries may occasionally deviate from perfection. In such instances, we have to deal with checking the geometry and repairing it all on our own. That's why it's time for a trait in `geo` which does the job for us.

This pull request introduces a new feature by implementing the `Validify` for the first two candidates: Self intersecting `Polygon`s and `MultiPolygon`s. The Validity trait serves to repair these invalid geometries by splitting them up into smaller polygons which fulfill the implicit invariants.

# Changes Made

- Implemented the `Validify` trait
- Add `split_into_valid` method on the trait and implement it for `Polygon` and `MultiPolygon`
- Also handles splitting up banana polygons 🍌 

# Usage Example

```rust
use geo::Validify;

// Assuming `your_geometry` is an instance of a self intersecting `Polygon`
let valid_multi_polygon = your_geometry.split_into_valid();
```

# Testing

- Added a few unit tests to ensure the correct behavior of the `Validify` trait and methods used for the first impls

# Documentation

- Document everything about the `Validify` trait quiet heavily for the afterworld

- [ ] Usage examples in trait docs

### DIFF
--- a/geo/src/algorithm/mod.rs
+++ b/geo/src/algorithm/mod.rs
@@ -258,6 +258,10 @@ pub mod triangulate_spade;
 #[cfg(feature = "spade")]
 pub use triangulate_spade::TriangulateSpade;
 
+/// Operations on invalid polygons to make them valid
+mod validity;
+pub use validity::Validify;
+
 /// Vector Operations for 2D coordinates
 mod vector_ops;
 pub use vector_ops::Vector2DOps;
--- a/geo/src/algorithm/validity/impls.rs
+++ b/geo/src/algorithm/validity/impls.rs
@@ -1,0 +1,32 @@
+use super::*;
+
+use geo_types::{MultiPolygon, Polygon};
+
+use crate::GeoFloat;
+
+/// Splits self intersecting polygons into individual valid polygons
+///
+/// Self intersecting here includes banana polygons
+///
+/// There are no real guarantees in terms of minimality we can give since the algorithms is a
+/// heuristic. In practice it turned out to work quiet well though.
+///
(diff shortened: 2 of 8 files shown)


urschrei:
Very nice. Also worth having a look at https://github.com/tudelft3d/prepair (cc @hugoledoux) which uses a constrained triangulation approach for further inspiration if you're interested…

RobWalt:
Lol that sounds a lot like what I did with SpadeBoolops. Now I wonder if I could just do a single polygon union to achieve the same 😅

Thanks a lot for the link! That's super useful!!

hugoledoux:
Watch out: the machinery used to achieve what we did is CGAL, which allows us to use a *robust* constrained Delaunay triangulation where we can intersect constraints and attach attributes to edges and triangles, and etc etc.

A standard triangulation doesn't allow you to do this, although if https://github.com/Stoeoef/spade is used it would be a good basis (I think, I haven't thoroughly tested it).

Since you have GEOS in rust, perhaps it's simpler to translate the algo of [`ST_MakeValid`](https://postgis.net/docs/ST_MakeValid.html) (no docs AFAIK, the code is the ultimate docs here...).

RobWalt:
I think I overengineered this a bit. I'm not seeing myself finishing it anytime soon, so I'll just gonna close this now. Feel free to pick it up or talk to me about the theoretical side of things

PR #1121, opened 2023-11-28
Remove `clone` in geometry trait implementation

kylebarron:
---

I struggled with this initially because I thought I _had_ to implement `Iter` in terms of `Self::ItemType`, which resulted in a double `&&Self::ItemType`.

In geoarrow-rs this resulted in a massive speedup (see https://github.com/geoarrow/geoarrow-rs/pull/278#issuecomment-1829024334) for converting `geo` objects to `geoarrow` (which is entirely implemented on top of these traits):

```
convert Vec<geo::Polygon> to PolygonArray
                        time:   [31.564 µs 31.608 µs 31.661 µs]
                        change: [-52.414% -52.271% -52.131%] (p = 0.00 < 0.05)
                        Performance has improved.
```

In the short term, in geoarrow-rs this also means all algorithms that result in geometries will get faster (though this probably isn't a huge part of the overall time) because currently all algorithms are implemented by converting each geometry to `geo`, applying the operation, and converting the output vec back to geoarrow.

### DIFF
--- a/geo-traits/src/geometry.rs
+++ b/geo-traits/src/geometry.rs
@@ -110,3 +110,41 @@ impl<'a, T: CoordNum + 'a> GeometryTrait for Geometry<T> {
         }
     }
 }
+
+impl<'a, T: CoordNum + 'a> GeometryTrait for &'a Geometry<T> {
+    type T = T;
+    type Point<'b> = Point<Self::T> where Self: 'b;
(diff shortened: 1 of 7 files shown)


RobWalt (approved):
Good catch! 👍🏼

frewsxcv (approved):

PR #1133, opened 2024-01-01
Implement topological equality check for IntersectionMatrix

### LINKED ISSUE #1132: Add is_equal_topo method to IntersectionMatrix to determine topological equality
This should test whether the matrix matches `[T*F**FFF*]`, returning `true` if the two geometries are _topologically_ equal.
This is different from `Eq`, which is an exact equality comparison. For example, these two `LineString`s satisfy `[T*F**FFF*]` even though they're obviously not `Eq`:

```rust
let ls1 = line_string![(x: 1.0, y: 1.0), (x: 2.0, y: 2.0)];
let ls2 = line_string![(x: 2.0, y: 2.0), (x: 1.0, y: 1.0)];
let lsrelation = ls1.relate(&ls2);
let de9im_eq = "T*F**FFF*";
assert_eq!(lsrelation.matches(&de9im_eq).unwrap(), true);
```

(please feel free to bikeshed the method name)

urschrei:
---

Fixes #1132

### DIFF
--- a/geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
+++ b/geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
@@ -231,6 +231,7 @@ impl IntersectionMatrix {
             && self.0[CoordPos::Inside][CoordPos::Outside] == Dimensions::Empty
             && self.0[CoordPos::Outside][CoordPos::Inside] == Dimensions::Empty
             && self.0[CoordPos::Outside][CoordPos::OnBoundary] == Dimensions::Empty
+            && self.0[CoordPos::OnBoundary][CoordPos::Outside] == Dimensions::Empty
     }
 
     /// Directly accesses this matrix
--- a/geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
+++ b/geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
@@ -223,6 +223,16 @@ impl IntersectionMatrix {
             && self.0[CoordPos::Outside][CoordPos::OnBoundary] == Dimensions::Empty
     }
 
+    /// Tests whether this matrix matches `[T*F**FFF*]`.
+    ///
+    /// returns `true` if the first geometry is *topologically* equal to the second.
+    pub fn is_equal_topo(&self) -> bool {
+        self.0[CoordPos::Inside][CoordPos::Inside] != Dimensions::Empty
+            && self.0[CoordPos::Inside][CoordPos::Outside] == Dimensions::Empty
+            && self.0[CoordPos::Outside][CoordPos::Inside] == Dimensions::Empty
+            && self.0[CoordPos::Outside][CoordPos::OnBoundary] == Dimensions::Empty
+    }
+
     /// Directly accesses this matrix
     ///
     /// ```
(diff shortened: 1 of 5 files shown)


urschrei:
@michaelkirk There are some cases that will be useful for this in the `jts-test-runner` `Relate` suite (`TestRelateLC` and `TestRelateLL`), but even after looking at the README I can't figure out how to activate / write them.

michaelkirk:
All the files in textxml/general are parsed and processed. We only skip the operations we don't know how to support which may live across files or be a subset of a particular file. This seemed like the best way to get as much test coverage as possible.

I hope you don't mind, but I pushed a commit to your branch that prints a little bit more about the stats of how many test cases are being skipped and why in jts_test_runner. You can see a lot more detail (too much detail probably) if you run with `RUST_LOG=jts_test_runner=debug cargo test -p jts-test-runner`

Feel free to remove it if you don't want to include it in your branch.

michaelkirk:
Just to be clear - the tests in TestRelateLC and TestRelateLL are already being run. You can convince yourself by watching the tests fail after you tweak the expected output.

urschrei:
Ahhh OK. Thank you! I'll have a further poke around.

michaelkirk:
re: the CI failure, I'll open a PR soon that bumps our MSRV to 1.70 (>6 months old) and updates the "latest" containers.

geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
```diff
@@ -223,6 +223,16 @@ impl IntersectionMatrix {
...
     }
 
+    /// Tests whether this matrix matches `[T*F**FFF*]`.
+    ///
+    /// returns `true` if the first geometry is *topologically* equal to the second.
+    pub fn is_equal_topo(&self) -> bool {
+        self.0[CoordPos::Inside][CoordPos::Inside] != Dimensions::Empty
+            && self.0[CoordPos::Inside][CoordPos::Outside] == Dimensions::Empty
+            && self.0[CoordPos::Outside][CoordPos::Inside] == Dimensions::Empty
+            && self.0[CoordPos::Outside][CoordPos::OnBoundary] == Dimensions::Empty
```
michaelkirk, line 233: I think we need one more condition: 

`&& self.0[CoordPos::OnBoundary][CoordPos::Outside] == Dimensions::Empty`

urschrei: Damn, thought I had them all. I looked at the other is_ methods for inspiration, but evidently got a bit confused. Which matrix entry does it correspond to?

michaelkirk: TBH, I firstly just intuited that any checks for an "is equal" relationship should be symmetrical - and noticed they weren't.

Then I noticed there are 5 constraints in the de-9im spec string, but you only had 4 comparisons and deduced which non-symmetrical check was missing.

But to try to answer your question directly, I think it's like this:

## Spec for "is topo equal": `[T*F**FFF*]`

|| Inside_a | OnBoundary_a | Outside_a |
| ------------- | ------------- |------ |------ |
| Inside_b |T | * | F|
| OnBounary_b | *| *| F (missing)|
| Outside_b | F | F| *|

## Corresponding Indexes into de-9im string

|| Inside_a | OnBoundary_a | Outside_a |
| ------------- | ------------- |------ |------ |
| Inside_b |0 | 1 | 2|
| OnBounary_b | 3| 4| 5 (missing)|
| Outside_b | 6 | 7| 8|

So we were missing a check for the "idx 5th" (the 6th element)

```
[T*F**FFF*]
[     ^   ]
```

urschrei: Amazing, thank you!

michaelkirk (approved):

PR #1134, opened 2024-01-02
Introduce `total_cmp` on GeoNum to avoid unwraps

michaelkirk:
---

In particular the latest `home` crate used by bindgen, required by proj, needs rust-1.70+

Also updates CI containers to latest


urschrei:
Is the TotalOrd trait requirement a result of the MSRV bump?

urschrei:
Ah, just read the commit message. It's due to a lint introduced in Rust 1.75

urschrei (approved):

geo/src/algorithm/simplify_vw.rs
```diff
@@ -241,7 +241,7 @@ fn vwp_wrapper<T, const INITIAL_MIN: usize, const MIN_POINTS: usize>(
     epsilon: &T,
 ) -> Vec<Vec<Coord<T>>>
 where
-    T: CoordFloat + RTreeNum + HasKernel,
+    T: GeoFloat + RTreeNum,
```
michaelkirk, line 244: Without changing this change I get:
```
error[E0599]: no method named `intersects` found for struct `CachedEnvelope<geo_types::Line<T>>` in the current scope
```

This error was happening because the previous constraint, `CoordFloat + HasKernel` was enough to hit the `impl GeoNum for T` blanket impl. Since GeoFloat is basically `CoordFloat + HasKernel` it makes sense to just use GeoFloat to fix this and similar future problems if we add other traits.

geo/src/algorithm/sweep/point.rs
```diff
@@ -29,17 +29,17 @@ impl<T: GeoNum> std::fmt::Debug for SweepPoint<T> {
...
-        }
+        Some(self.cmp(other))
     }
 }
 
 /// Derive `Ord` from `PartialOrd` and expect to not fail.
 impl<T: GeoNum> Ord for SweepPoint<T> {
     fn cmp(&self, other: &Self) -> Ordering {
-        self.partial_cmp(other).unwrap()
+        match self.0.x.total_cmp(&other.0.x) {
```
lnicola, line 39: Should we implement `TotalOrd` on some tuples, at least, to make this look nicer? The users might expect it to work similar to `Ord`

lnicola (approved):

geo/CHANGES.md
```diff
@@ -1,6 +1,17 @@
 # Changes
 
 ## Unreleased
+* POSSIBLY BREAKING: New `TotalOrd` trait is required for our base numeric
+  trait GeoNum, this avoids some potential crashes when working with geometries
+  that contain NaN points. This shouldn't break for any common numeric types,
+  but if you are using something exotic you'll need to manually implement
+  TotalOrd for your numeric type.
+  * <https://github.com/georust/geo/pull/1134>
+* POSSIBLY BREAKING: Visvalingam trait implementation moved from
```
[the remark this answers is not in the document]
michaelkirk: I've updated the changelog to clarify it's only `SimplifyVwPreserve` that changed.

geo/src/algorithm/sweep/point.rs
michaelkirk: This reasonable piece of feedback lead me down a rabbit hole. It might have been an over reaction on my part...

Ultimately, the reason for these changes are that I want to remove an `unwrap` from the code for the geometries we support. I introduced TotalOrd to do that. I don't actually care to support a general purpose TotalOrd trait. I realized:

1. GeoNum requires HasKernel, so arguably we only need to implement this functionality for the same types that we've defined HasKernel for (f32, f64, i16, i32, i64, isize).

2. Rather than try to keep in sync the implementations for HasKernel and GeoNum, since there's so little you can do with HasKernel without GeoNum, I though it made sense to simplify and and merge HasKernel's functionality into GeoNum.

I think this approach makes sense, but I'm prepared to back it out if it's controversial. @rmanoka in particular might have thoughts about that since the `HasKernel` trait was his devising (IIRC).

geo/CHANGES.md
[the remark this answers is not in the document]
lnicola: Of course, not sure what I was thinking.

geo/src/algorithm/sweep/point.rs
lnicola: Yeah, I guess implementing it for tuples isn't that easy. I'm fine with not having that, at least until someone complains.

https://doc.rust-lang.org/std/cmp/enum.Ordering.html#method.then_with can make this look nicer, but it doesn't really matter.

frewsxcv (approved):

michaelkirk:
@rmanoka - did you want to weigh in on this? I've gotten a few positive responses so far so I'm going to go ahead and merge it in a couple days if I don't hear from you.

rmanoka (approved):

rmanoka:
Looks good to me too!  +1 for merge.

PR #1136, opened 2024-01-13
Add all remaining predicates: covered by, covers, crosses, overlaps, and touches

urschrei:
---

<s>I want to do a little more checking with regard to the logic we use to determine whether the input dimensions differ (which is a requirement of the `Crosses` predicate), and conversely checking whether they're the same (a requirement of the `Overlaps` predicate). I'm currently doing this by interrogating the IM for the following two relations: `a-interior / b-exterior` and `a-exterior / b-interior`. I assert that if they're **not** equal the input geometry dimensions differ, and vice versa, but I'm not yet 100 % certain that this is true for all valid input geometries.</s>

<s>Addressed in https://github.com/georust/geo/pull/1136/commits/af8a50cf1da7ba014b3778751e16c95b1995bfae for all non multi-geometries.</s>

### DIFF
--- a/geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
+++ b/geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
@@ -222,9 +222,12 @@ impl IntersectionMatrix {
     //     && self.0[CoordPos::Outside][CoordPos::OnBoundary] == Dimensions::Empty`
     // ==================================================================
 
-    /// Tests if this matrix matches `[FF*FF****]`.
+    /// Returns `true` if geometries `a` and `b` are disjoint: they have no point in common,
+    /// forming a set of disconnected geometries.
     ///
-    /// returns `true` if the two geometries related by this matrix are disjoint
+    /// # Notes
+    /// - Matches `[FF*FF****]`
+    /// - This predicate is **anti-reflexive**
     pub fn is_disjoint(&self) -> bool {
         self.0[CoordPos::Inside][CoordPos::Inside] == Dimensions::Empty
             && self.0[CoordPos::Inside][CoordPos::OnBoundary] == Dimensions::Empty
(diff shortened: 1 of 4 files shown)


geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
```diff
@@ -268,6 +268,161 @@ impl IntersectionMatrix {
...
+    ///     (x: 124., y: 182.),
+    ///     (x: 106., y: 146.),
+    ///     (x: 162., y: 159.),
+    ///     (x: 124., y: 182.),
+    /// ];
+    ///
+    /// let intersection = poly1.relate(&poly2);
+    /// assert_eq!(intersection.is_coveredby(), true);
+    /// ```
+    #[allow(clippy::nonminimal_bool)]
```
urschrei, line 299: Clippy has suggested simplified boolean logic but I want to maintain what's here as it's easy to check against the documented relations.

geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
```diff
@@ -268,6 +268,161 @@ impl IntersectionMatrix {
...
+    ///     (x: 126., y: 179.),
+    ///     (x: 110., y: 150.),
+    ///     (x: 161., y: 160.),
+    ///     (x: 126., y: 179.),
+    /// ];
+    ///
+    /// let intersection = poly1.relate(&poly2);
+    /// assert_eq!(intersection.is_overlaps(), true);
+    /// ```
+    #[allow(clippy::nonminimal_bool)]
```
urschrei, line 399: Clippy has suggested simplified boolean logic but I want to maintain what's here as it's easy to check against the documented relations.

urschrei:
We now additionally check that neither of the input geometry dimensions are empty for `Crosses` and `Overlaps`

geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
```diff
@@ -268,6 +278,212 @@ impl IntersectionMatrix {
...
+
+    /// Returns `true` if `a` touches `b`: they have at least one point in common, but their
+    /// interiors do not intersect.
+    /// # Notes
+    /// - This predicate is **symmetrice**
+    #[allow(clippy::nonminimal_bool)]
+    pub fn is_touches(&self) -> bool {
+        // [FT*******]
+        self.0[CoordPos::Inside][CoordPos::Inside] == Dimensions::Empty
+        && self.0[CoordPos::Outside][CoordPos::Inside] != Dimensions::Empty ||
```
michaelkirk, line 364: This line seems wrong, right?

geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
```diff
@@ -268,6 +278,212 @@ impl IntersectionMatrix {
...
+    /// ```
+    /// # Notes
+    /// - If any of the following do not hold, the function will return `false`:
+    ///     - The intersection of the interiors of the geometries must be non-empty
+    ///     - The intersection must have dimension less than the maximum dimension of the two input geometries
+    ///     - The intersection of the two geometries must not equal either geometry
+    /// - This predicate is **symmetric**
+    pub fn is_crosses(&self) -> bool {
+        if self.0[CoordPos::Inside][CoordPos::Outside]
+            != self.0[CoordPos::Outside][CoordPos::Inside]
```
michaelkirk, line 399: There are a few checks in this method that I'm having a hard time reconciling with the logic from the wikipedia page. Did you have another source?

After thinking about it for a while, I think I'm starting to grasp the argument, but I haven't gotten completely through the exercise.

Something like:

```
A's interior cannot be wholly within B's interior unless dims(A.I) <= dims(B.I)
B's interior cannot be wholly within A's interior unless dims(B.I) <= dims(A.I)
```

e.g. a point can never contain a line. a line can never contain a polygon.

So unless `dims(A) == dims(B)` A.I and/or B.I __must__ "leak" into the other's exterior. A

And then we can prove that, when A doesn't contain B, `dims(A.I v B.E) == dims(A.I)` 

Is that about right?

geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
```diff
@@ -268,6 +278,212 @@ impl IntersectionMatrix {
...
+                return self.0[CoordPos::Inside][CoordPos::Inside] != Dimensions::Empty
+                    && self.0[CoordPos::Inside][CoordPos::Outside] != Dimensions::Empty;
+            }
+            if a_greaterthan_b {
+                // [T*****T**]
+                return self.0[CoordPos::Inside][CoordPos::Inside] != Dimensions::Empty
+                    && self.0[CoordPos::Outside][CoordPos::Inside] != Dimensions::Empty;
+            }
+        } else {
+            // dimensions are equal: this is only permitted for line / line
```
michaelkirk, line 422: Can you clarify what you mean by `permitted` here? I understand there is a special case in the definition of `is_crosses` for line/line geometries, but do you mean that the previous `if` check precludes this block from executing on `line` /`line` geometries or that if we are in this block, anything that is not line/line will return false?

michaelkirk (changes requested):

geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
urschrei: My understanding is that the previous check is supposed to preclude this block from executing. That is, it's (obviously) possible to check this predicate on any geometry combination, but the result will be wrong (or ill-defined?) for any combination that isn't line / line as other geometries with equal dimensions cannot "cross".

geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
```diff
@@ -268,6 +278,212 @@ impl IntersectionMatrix {
...
+            && self.0[CoordPos::Outside][CoordPos::Inside] != Dimensions::Empty
+        {
+            // input geom dimensions differ and neither are empty: proceed to calculating relationships between dimensions
+            let ii = self.0[CoordPos::Inside][CoordPos::Inside];
+            let ie = self.0[CoordPos::Inside][CoordPos::Outside];
+            let ei = self.0[CoordPos::Outside][CoordPos::Inside];
+            let a_lessthan_b = ii == ie;
+            let a_greaterthan_b = ii == ei;
+            // something has gone horribly wrong
+            assert!(a_lessthan_b != a_greaterthan_b);
```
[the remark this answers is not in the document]
urschrei: Yep, makes sense!

geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
urschrei: the first check: `if self.0[CoordPos::Inside][CoordPos::Outside] != self.0[CoordPos::Outside][CoordPos::Inside]`, establishes that the input geometry dimensions differ, and neither are empty (this works, as demonstrated by the `combination_dimension` test).

If that check fails, they _are_ equal, and we proceed to check whether they're both dimension 1, as that's the only equal dimension geometry that can cross (see the note on https://postgis.net/docs/ST_Crosses.html).

If that first check passes, we proceed to figure out whether dimension A is greater than dimension B, or vice versa, and apply the correct predicate.

michaelkirk: Are you asserting that `self.0[CoordPos::Inside][CoordPos::Outside] == dims(A)`? That seems like it'd often be true, but I'm not sure what the rules are for when it's true or not. Certainly not true when `B.contains(A)`. What's less clear to me are the boundary conditions.

Is there a way we can make this implementation look more like the definitions on postgis or wikipedia so I wouldn't have to think so hard? 😆

michaelkirk: My earlier suggestion was something like:
```
let dims_a = max(self.0[CoordPos::Inside][CoordPos::Inside], self.0[CoordPos::Inside][CoordPos::Boundary, self.0[CoordPos::Inside][CoordPos::Exterior]);

let dims_b = max(self.0[CoordPos::Inside][CoordPos::Inside], self.0[CoordPos::Boundary][CoordPos::Inside], self.0[CoordPos::Exterior][CoordPos::Inside]);
```

Then, now that we have `dims_a`/`dims_b`, it seems like we can have a case analysis that aligns with the definitions given in wikipedia and postgis.

If you feel more confident with your approach, I'm ok with going your way, but I'll try to propose some additional comments that cover my gaps in understanding.

geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
```diff
@@ -268,6 +278,225 @@ impl IntersectionMatrix {
...
+            let ii = self.0[CoordPos::Inside][CoordPos::Inside];
+            let ie = self.0[CoordPos::Inside][CoordPos::Outside];
+            let ei = self.0[CoordPos::Outside][CoordPos::Inside];
+            let a_lessthan_b = ii == ie;
+            let a_greaterthan_b = ii == ei;
+            // something has gone horribly wrong
+            debug_assert!(a_lessthan_b != a_greaterthan_b,
+                "Input geometry dimensions cannot simultaneously be greater than and less than each other. Are your geometries valid?"
+            );
+            if a_lessthan_b {
```
michaelkirk, line 417:
```suggestion
            if dims_a < dims_b {
```

This can be simplified now that we have `dims_` - I think it'll be easier to track than computing the ii/ie/ei.

Additionally, it might be really slick to have all the cases in a match, but I leave that to your judgement.

(Introducing a `match` might be overly baroque, but we should at least get rid of `a_less_than_b`)

geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
```diff
@@ -268,6 +278,225 @@ impl IntersectionMatrix {
...
+            .max(self.0[CoordPos::Outside][CoordPos::Inside]);
+        if dims_a == dims_b {
+            if dims_a == Dimensions::OneDimensional && dims_b == Dimensions::OneDimensional {
+                // line / line
+                self.0[CoordPos::Inside][CoordPos::Outside] == Dimensions::OneDimensional
+                    && self.0[CoordPos::Inside][CoordPos::Outside] != Dimensions::Empty
+                    && self.0[CoordPos::Outside][CoordPos::Inside] != Dimensions::Empty
+            } else {
+                // points or polygons
+                let ii = self.0[CoordPos::Inside][CoordPos::Inside];
```
michaelkirk, line 483: Similarly, is there still value in these checks now that we have dims_a/dims_b? It's ok to leave them if you think the debug_assert is checking something useful, but I wanted to double check since we no longer use ii/ie/ei to compute results.

michaelkirk (changes requested):

geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
urschrei: I do like a match…

geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
```diff
@@ -268,6 +283,226 @@ impl IntersectionMatrix {
...
+        let dims_a = self.0[CoordPos::Inside][CoordPos::Inside]
+            .max(self.0[CoordPos::Inside][CoordPos::OnBoundary])
+            .max(self.0[CoordPos::Inside][CoordPos::Outside]);
+
+        let dims_b = self.0[CoordPos::Inside][CoordPos::Inside]
+            .max(self.0[CoordPos::OnBoundary][CoordPos::Inside])
+            .max(self.0[CoordPos::Outside][CoordPos::Inside]);
+        match (dims_a, dims_b) {
+            // a < b
+            (Dimensions::ZeroDimensional, Dimensions::OneDimensional)
```
michaelkirk, line 423: Alternatively, you can use an if rather than spelling out all the cases:
```
_ if dims_a < dims_b => {
                // [T*T******]
                self.0[CoordPos::Inside][CoordPos::Inside] != Dimensions::Empty
                    && self.0[CoordPos::Inside][CoordPos::Outside] != Dimensions::Empty
}
```

But what you have also looks correct.

geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
```diff
@@ -268,6 +283,226 @@ impl IntersectionMatrix {
...
+            .max(self.0[CoordPos::Inside][CoordPos::OnBoundary])
+            .max(self.0[CoordPos::Inside][CoordPos::Outside]);
+
+        let dims_b = self.0[CoordPos::Inside][CoordPos::Inside]
+            .max(self.0[CoordPos::OnBoundary][CoordPos::Inside])
+            .max(self.0[CoordPos::Outside][CoordPos::Inside]);
+        match (dims_a, dims_b) {
+            // line / line: [1*T***T**]
+            (Dimensions::OneDimensional, Dimensions::OneDimensional) => {
+                self.0[CoordPos::Inside][CoordPos::Outside] == Dimensions::OneDimensional
```
michaelkirk, line 491:
```suggestion
                self.0[CoordPos::Inside][CoordPos::Inside] == Dimensions::OneDimensional
```

I think?

michaelkirk (approved):
There is what looks to be a straight forward typo in `is_overlaps`, but other than that LGTM.

One take it or leave it (truly) style thing.

Thanks for seeing this through!

geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
urschrei: Ahh, I suspected there there was  a way to use `_` and an `if` statement to bundle these together.

geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
urschrei: 🤦‍♂️

urschrei:
> There is what looks to be a straight forward typo in `is_overlaps`, but other than that LGTM.
> 
> One take it or leave it (truly) style thing.
> 
> Thanks for seeing this through!

Many thanks for the reviews. Our long national nightmare is over.

PR #1139, opened 2024-01-15
Fix typos

jayvdb:
  (IMO this isnt of sufficient value to be mentioned)
---


urschrei (approved):
Thanks!

PR #1144, opened 2024-01-30
Add Voronoi diagrams with zero dependency delaunay triangulation

doc-E-brown:
This PR adds the `VoronoiDiagram` trait to create Voronoi diagrams from `Polygon` and `MultiPoint` types.  The Voronoi diagram is created using the property that [voronoi diagrams are the dual of delaunay triangulation](https://en.wikipedia.org/wiki/Voronoi_diagram#Properties).

The `VoronoiDiagram` trait uses an added `TriangulateDelaunay` trait that computes Delaunay triangles without any additional dependencies.  As per the conversation [here](https://github.com/georust/geo/issues/1002#issuecomment-1807505844) this work was started before the introduction of the spade methodology.  As requested in the conversation I've added a bench performance test to compare the different methods and these are the results when running on my laptop.

```
Delaunay Triangulation  time:   [2.2839 µs 2.3687 µs 2.4780 µs]
                        change: [-21.682% -14.728% -7.0971%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) high mild

Spade Unconstrained Triangulation
                        time:   [1.4216 µs 1.4831 µs 1.5582 µs]
                        change: [+39.411% +46.250% +53.808%] (p = 0.00 < 0.05)
                        Performance has regressed.
  5 (5.00%) high mild
  2 (2.00%) high severe

Constrained Outer Triangulation
                        time:   [3.5185 µs 3.7830 µs 4.1260 µs]

Constrain Triangulation time:   [3.5962 µs 3.7355 µs 3.8970 µs]
Found 15 outliers among 100 measurements (15.00%)
  2 (2.00%) low mild
  4 (4.00%) high mild
  9 (9.00%) high severe
```

the `TriangulateDelaunay` trait is faster then the `Constrained Outer Triangulation` method but not the `Unconstrained Method`.

The original intent behind implementing Delaunay triangulation in `geo` arose when we needed a method that was compliant with the Apache license using `geo-types` and was not otherwise available.

Thank you for considering this PR.

---

### DIFF
--- a/geo/src/algorithm/triangulate_delaunay.rs
+++ b/geo/src/algorithm/triangulate_delaunay.rs
@@ -1,0 +1,645 @@
+use std::{fmt, fmt::Debug};
+
+use crate::coord;
+use crate::{BoundingRect, Coord, GeoFloat, Line, MultiPoint, Polygon, Triangle};
+
+/// The default expansion value for creating the super triangle.
+/// A bounding rectangle is computed for the points and the super triangle
+/// is expanded by this value by default.
+pub const DEFAULT_SUPER_TRIANGLE_EXPANSION: f64 = 20.;
+
+type Result<T> = std::result::Result<T, DelaunayTriangulationError>;
+
+/// Delaunay Triangulation for a given set of points using the
+/// [Bowyer](https://doi.org/10.1093%2Fcomjnl%2F24.2.162)-[Watson](https://doi.org/10.1093%2Fcomjnl%2F24.2.167)
+/// algorithm
+pub trait TriangulateDelaunay<T: GeoFloat> {
+    /// # Example
+    ///
+    /// ```rust
+    /// use geo::{coord, polygon, Triangle, TriangulateDelaunay};
+    ///
+    /// let points = polygon![
+    ///     (x: 0., y: 0.),
+    ///     (x: 1., y: 2.),
+    ///     (x: 2., y: 4.),
+    ///     (x: 2., y: 0.),
+    ///     (x: 3., y: 2.),
+    ///     (x: 4., y: 0.),
+    /// ];
+    ///
+    /// let tri_force = points.delaunay_triangulation().unwrap();
+    ///
+    /// assert_eq!(vec![
+    ///     Triangle(
+    ///         coord! {x: 1., y: 2.},
+    ///         coord! {x: 0., y: 0.},
+    ///         coord! {x: 2., y: 0.},
+    ///     ),
+    ///     Triangle(
+    ///         coord! {x: 2., y: 4.},
+    ///         coord! {x: 1., y: 2.},
+    ///         coord! {x: 3., y: 2.},
+    ///     ),
+    ///     Triangle(
+    ///         coord! {x: 1., y: 2.},
+    ///         coord! {x: 2., y: 0.},
+    ///         coord! {x: 3., y: 2.},
+    ///     ),
+    ///     Triangle(
+    ///         coord! {x: 3., y: 2.},
+    ///         coord! {x: 2., y: 0.},
+    ///         coord! {x: 4., y: 0.},
+    ///     )],
+    ///     tri_force
+    /// );
+    ///
+    /// ```
+    ///
+    fn delaunay_triangulation(&self) -> Result<Vec<Triangle<T>>>;
+}
+
+impl<T: GeoFloat> TriangulateDelaunay<T> for Polygon<T>
+where
+    f64: From<T>,
+{
+    fn delaunay_triangulation(&self) -> Result<Vec<Triangle<T>>> {
+        let super_triangle = DelaunayTriangle(create_super_triangle(self)?);
+
+        let mut triangles: Vec<DelaunayTriangle<T>> = vec![super_triangle.clone()];
+        let mut processed_points: Vec<Coord<T>> = Vec::new();
+        for pt in self.exterior().coords() {
+            // If we have already added the point, do not process again
+            // to avoid producing intersecting triangles
+            if !processed_points.contains(pt) {
+                add_coordinate(&mut triangles, pt);
+                processed_points.push(*pt);
+            }
+        }
+        let triangles = remove_super_triangle(&triangles, &super_triangle);
+        Ok(triangles.iter().map(|x| x.0).collect())
+    }
+}
+
+impl<T: GeoFloat> TriangulateDelaunay<T> for MultiPoint<T>
+where
+    f64: From<T>,
+{
+    fn delaunay_triangulation(&self) -> Result<Vec<Triangle<T>>> {
+        let poly = Polygon::new(self.into_iter().cloned().collect(), vec![]);
+        poly.delaunay_triangulation()
+    }
+}
+
+fn create_super_triangle<T: GeoFloat>(geometry: &Polygon<T>) -> Result<Triangle<T>>
+where
+    f64: From<T>,
+{
+    let expand_factor = T::from(DEFAULT_SUPER_TRIANGLE_EXPANSION)
+        .ok_or(DelaunayTriangulationError::FailedToConstructSuperTriangle)?;
+    let bounds = geometry
+        .bounding_rect()
+        .ok_or(DelaunayTriangulationError::FailedToConstructSuperTriangle)?;
+    let width = bounds.width() * expand_factor;
+    let height = bounds.height() * expand_factor;
+    let bounds_min = bounds.min();
+    let bounds_max = bounds.max();
+
+    Ok(Triangle(
+        coord! {x: bounds_min.x - width, y: bounds_min.y},
+        coord! {x: bounds_max.x + width, y: bounds_min.y - height},
+        coord! {x: bounds_max.x + width, y: bounds_max.y + height},
+    ))
+}
+
+fn find_line<T: GeoFloat>(line: &Line<T>, lines: &[Line<T>]) -> Option<usize> {
+    for (idx, sample_line) in lines.iter().enumerate() {
+        if line == sample_line {
+            return Some(idx);
+        }
+    }
+    None
+}
+
+fn add_coordinate<T: GeoFloat>(triangles: &mut Vec<DelaunayTriangle<T>>, c: &Coord<T>)
+where
+    f64: From<T>,
+{
+    let mut bad_triangles: Vec<&DelaunayTriangle<T>> = Vec::new();
+
+    // Check for the triangles where the point is present within the
+    // corresponding circumcircle.
+    for tri in triangles.iter() {
+        if tri.is_in_circumcircle(c) {
+            bad_triangles.push(tri);
+        }
+    }
+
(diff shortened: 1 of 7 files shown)


geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -0,0 +1,645 @@
...
+        }
+    }
+
+    let mut bad_triangle_edges: Vec<Line<T>> = Vec::new();
+    let mut bad_triangle_edge_count: Vec<u128> = Vec::new();
+    for x in &bad_triangles {
+        for y in x.0.to_lines() {
+            let flipped_y = Line::new(y.end, y.start);
+            let idx =
+                find_line(&y, &bad_triangle_edges).or(find_line(&flipped_y, &bad_triangle_edges));
```
RobWalt, line 144: Just a premature perf tip so feel free to ignore this maybe:

In another PR of mine, I did more or less the same (maybe that's an idication we should add this somewhere generally, but that's for another day). While reviewing, someone asked some perf questions and I tried to compare the coodinates directly, which gave me a good performance imporvement. You can read a little more about the numbers here in the commit message https://github.com/georust/geo/pull/1087/commits/49c2f040f51adbe3b37ef2cc9b367866ecc05961

I would suggest to use this here instead

```rust
bad_triangle_edges.iter().position(|other| is_line_shared(y, other))
```

and test out the performance just for fun. Btw this also avoids iterating through the vec twice.

geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -0,0 +1,645 @@
...
+    for tri in triangles.iter() {
+        if tri.is_in_circumcircle(c) {
+            bad_triangles.push(tri);
+        }
+    }
+
+    let mut bad_triangle_edges: Vec<Line<T>> = Vec::new();
+    let mut bad_triangle_edge_count: Vec<u128> = Vec::new();
+    for x in &bad_triangles {
+        for y in x.0.to_lines() {
```
RobWalt, line 141: Let's please use more human readable variable names. I'd suggest

- `x => tri`
- `y => line`

geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -0,0 +1,645 @@
...
+    // Remove the bad triangles
+    let mut new_triangles: Vec<DelaunayTriangle<T>> = triangles
+        .iter()
+        .filter(|x| !bad_triangles.contains(x))
+        .cloned()
+        .collect();
+
+    polygon
+        .iter()
+        .for_each(|x| new_triangles.push(DelaunayTriangle(Triangle(x.start, x.end, *c))));
```
RobWalt, line 174: Not sure, but: Can't we inline `polygon` here and not collect above into a new `Vec`?

Edit: Ah just saw that we push to the `new_triangles`. Then please try this instead to avoid needless allocation (you'll probably have to add a `.cloned()` to the iterator somewhere.

```rust
   // Shared edges are those with a count of > 1
    let polygon = bad_triangle_edge_count
        .iter()
        .enumerate()
        .filter(|(_, count)| **count < 2)
        .map(|(idx, _)| bad_triangle_edges[idx]); // <- no collect here anymore

    // Remove the bad triangles
    let mut new_triangles: Vec<DelaunayTriangle<T>> = triangles
        .iter()
        .filter(|x| !bad_triangles.contains(x))
        .cloned()
        .chain(polygon) // <- maybe polygon.cloned()
        .collect();
```

geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -0,0 +1,645 @@
...
+
+    // Replace the triangles
+    triangles.clear();
+    new_triangles.iter().for_each(|x| triangles.push(x.clone()));
```
RobWalt, line 178: We can probably save a clone here as well if we just do

```rust
*triangles = new_triangles;
```

geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -0,0 +1,645 @@
...
+        let mut processed_points: Vec<Coord<T>> = Vec::new();
+        for pt in self.exterior().coords() {
+            // If we have already added the point, do not process again
+            // to avoid producing intersecting triangles
+            if !processed_points.contains(pt) {
+                add_coordinate(&mut triangles, pt);
+                processed_points.push(*pt);
+            }
+        }
+        let triangles = remove_super_triangle(&triangles, &super_triangle);
```
RobWalt, line 79: Questions: 

- Can it happen that the `super_triangle` gets removed during the algorithm?
- If not, can it happen that the `super_triangle` gets moved away from the first position to somewhere else in the Vec

Depending on the answers, the removal might be easier than a full search through the whole vec as done in `remove_super_triangle`

geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -0,0 +1,645 @@
...
+    pub(crate) fn shares_edge(&self, other: &DelaunayTriangle<T>) -> Option<Line<T>> {
+        let other_lines = other.0.to_lines();
+        for line in self.0.to_lines().iter() {
+            for other_line in other_lines.iter() {
+                if is_line_shared(line, other_line) {
+                    return Some(*line);
+                }
+            }
+        }
+        None
```
RobWalt, line 240: This can be simplified to

```rust
let other_lines = other.0.to_lines();
self.0.to_lines()
  .iter()
  .find(|line| {
    other_lines.iter()
      .any(|other| is_line_shared(line, other))
})
```

geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -0,0 +1,645 @@
...
+        //     [a_d_x, a_d_y, a_d_x_d_y],
+        //     [b_d_x, b_d_y, b_d_x_d_y],
+        //     [c_d_x, c_d_y, c_d_x_d_y],
+        // ]
+        //
+        let determinant = a_d_x * ((b_d_y * c_d_x_d_y) - (b_d_x_d_y * c_d_y))
+            - a_d_y * ((b_d_x * c_d_x_d_y) - (b_d_x_d_y * c_d_x))
+            + a_d_x_d_y * (b_d_x * c_d_y - b_d_y * c_d_x);
+
+        determinant > 0.0
```
RobWalt, line 270: Why is the `f64` necessary? If it's just because of the comparison with `0.0` at the end, you can use `T::zero()` instead of that literal value

RobWalt (changes requested):
Sorry it took me a while to review this. I started with just the Delauny triangulation part. This is just a first iteration cleaning the code up here and there. I didn't really go through how the algorithm works yet. Will do that in a follow up review.

Overall, even though I left some comments, I think the code is really clean and the comments were really helpful so far! Also thanks for taking the time to create nice Errors which implement Display!

geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -0,0 +1,645 @@
...
+                find_line(&y, &bad_triangle_edges).or(find_line(&flipped_y, &bad_triangle_edges));
+            if let Some(idx) = idx {
+                // The unwrap is acceptable due to the push lines below.
+                let count = bad_triangle_edge_count.get_mut(idx).unwrap();
+                *count += 1;
+            } else {
+                // These two vectors must be updated at the same time
+                bad_triangle_edges.push(y);
+                bad_triangle_edge_count.push(1);
+            }
```
RobWalt, line 153: It looks like it might make sense to use a Vec of tuples here: `Vec<(Line<T>, usize)>`.

We can then

```rust
if let Some(idx) = idx {
    let (_, count) = bad_edges_with_count.get_mut(idx).expect("valid since idx comes from .position()");
    *count +=1;
} else {
    bad_edges_with_count.push((y,1));
}
```

and further below

```rust
    let polygon: Vec<Line<T>> = bad_triangle_edge_count
        .into_iter()
        .enumerate()
        .filter(|(_, count)| *count < 2)
        .map(|(line, _)| line)
        .map(|edge| DelaunayTriangle(Triangle(edge.start, edge.end, *c)));
```

doc-E-brown:
Thanks so much for your thorough review @RobWalt  I'll work through them and have an update soon

geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -0,0 +1,645 @@
...
+}
+
+fn find_line<T: GeoFloat>(line: &Line<T>, lines: &[Line<T>]) -> Option<usize> {
+    for (idx, sample_line) in lines.iter().enumerate() {
+        if line == sample_line {
+            return Some(idx);
+        }
+    }
+    None
+}
```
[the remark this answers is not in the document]
doc-E-brown: yes it is. TIL! thanks!

geo/src/algorithm/triangulate_delaunay.rs
doc-E-brown: I used the suggestions provided as per `bad_triangle_edges.iter().find(|other| is_line_shared(y, other))` and below.  It is definitely much cleaner now!

geo/src/algorithm/triangulate_delaunay.rs
doc-E-brown: yes with a couple of modifications to account for the temporary value that is freed while in use

geo/src/algorithm/triangulate_delaunay.rs
doc-E-brown: it is not required.  I've removed.

geo/src/algorithm/triangulate_delaunay.rs
doc-E-brown: No the `super_triangle` isn't removed during the process.  The scaffolding provided by the `super_triangle` will remain as well as other triangles formed from it, an example.  In this diagram the `super_triangle` and corresponding artefacts are shown in red while the final delaunay triangulation is shown in below.

The process of removing the `super_triangle` also requires removing triangles that share vertices with the `super_triangle` and these triangles can be located anywhere in the vec.

doc-E-brown:
Hi @RobWalt sorry for the delay, I've updated the PR as per your comments.  It definitely looks much cleaner! Thanks!

RobWalt:
Hey, sorry for leaving this unattended for so long. I kind of forgot about it. Here's my new suggestion: The PR is kind of big. Could we split it up into two parts?

1. the triangulation part, which I'm happy to approve instantly
2. the vonoroi part, which still needs some review

urschrei:
@doc-E-brown @RobWalt Apologies, I thought I had already asked this question: given that we already depend on `spade` – which provides Voronoi tessellation (though we haven't exposed it via `geo`) and Delaunay triangulation, and have an ear-cut triangulation module. As it stands, I can't imagine merging this PR:

1. It duplicates functionality
2. It's not tested in use to the same extent as the functionality it's duplicating (Spade is mature and in fairly wide use)

Issue #706, opened 2021-12-30
Add `num_rings` or `rings_count` to `Polygon`


frewsxcv:

callpraths:
Thanks for adding starter bugs for the likes of me :)
Is there interest in adding this?

A few clarifications would be useful:

1. Naming: I'd vote for `ring_count` to go with existing `coord_count` (PostGIS uses `ST_NRings`, geos crate has `get_num_interior_rings` and nothing that includes the exterior ring).
2. Should this be added directly to struct impl(s)?
3. It should probably be added to: `Polygon`, `MultiPolygon`, `Rect` (always 1) and `Triangle` (always 1).

callpraths:
I want to answer 2 with: Add `geo::algorithm::RingCount` trait, even though it will only be implemented for a few of the core types. There are plenty of examples where algorithms only apply to subset of types (e.g. `FrechetDistance` only applies to `LineString`.

frewsxcv:
Resolved in https://github.com/georust/geo/pull/1145

PR #1145, opened 2024-01-31
Ring counts

carlospadron:
---
This implements 2 methods for polygons: num_rings as total of rings in a polygon, and num_interior_rings as the count of internal rings. This hopefully fulfils issue [706](https://github.com/georust/geo/issues/706)

### DIFF
--- a/geo-types/src/geometry/polygon.rs
+++ b/geo-types/src/geometry/polygon.rs
@@ -415,6 +415,56 @@ impl<T: CoordNum> Polygon<T> {
     {
         (current_vertex + (self.exterior.0.len() - 1) - 1) % (self.exterior.0.len() - 1)
     }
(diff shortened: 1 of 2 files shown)


geo-types/src/geometry/polygon.rs
```diff
@@ -415,6 +415,56 @@ impl<T: CoordNum> Polygon<T> {
     {
         (current_vertex + (self.exterior.0.len() - 1) - 1) % (self.exterior.0.len() - 1)
     }
+
+    /// Count the number of rings in the polygon
```
lnicola, line 419:
```suggestion
    /// Count the total number of rings (interior and exterior) in the polygon
```

lnicola (approved):

PR #1146, opened 2024-02-05
Implement `CoordsIter` for arrays and slices

grovesNL:
---

I found this useful so I could use `extremes` etc. on some kind of arbitrary coordinate list I'm already storing somewhere.

Many of the coordinate lists I'm using could technically form triangles but the points could be colinear or identical, so they don't follow the guarantees for `Triangle`. I also don't want to use `MultiPoint` for these because it would require creating another `Vec` (potentially thousands in my case).

I'm not sure if it makes sense to implement this for arrays and slices of `Coord`, `Point`, or both. For now it's only implemented at the `Coord` level.

### DIFF
--- a/geo/src/algorithm/coords_iter.rs
+++ b/geo/src/algorithm/coords_iter.rs
@@ -447,6 +447,50 @@ impl<T: CoordNum> CoordsIter for Geometry<T> {
     }
 }
 
+// ┌──────────────────────────┐
+// │ Implementation for Array │
+// └──────────────────────────┘
+
+impl<const N: usize, T: CoordNum> CoordsIter for [Coord<T>; N] {
+    type Iter<'a> = iter::Copied<slice::Iter<'a, Coord<T>>> where T: 'a;
+    type ExteriorIter<'a> = Self::Iter<'a> where T: 'a;
+    type Scalar = T;
+
+    fn coords_iter(&self) -> Self::Iter<'_> {
+        self.iter().copied()
+    }
+
+    fn coords_count(&self) -> usize {
+        N
+    }
+
+    fn exterior_coords_iter(&self) -> Self::ExteriorIter<'_> {
+        self.coords_iter()
+    }
+}
+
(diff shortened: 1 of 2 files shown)


grovesNL:
Actually I'll rework this to simplify it a bit more

grovesNL:
I thought I might be able to use `AsRef` and `ExactSizeIterator` here instead (or even just as an iterator adapter with `IntoIterator`), but I wasn't sure how to make the types work out with `CoordsIter`, so I've kept it as two separate implementations (array and slices) for now

frewsxcv (approved):

geo/src/algorithm/coords_iter.rs
```diff
@@ -447,6 +447,50 @@ impl<T: CoordNum> CoordsIter for Geometry<T> {
     }
 }
 
+// ┌──────────────────────────┐
+// │ Implementation for Array │
+// └──────────────────────────┘
+
+impl<const N: usize, T: CoordNum> CoordsIter for [Coord<T>; N] {
+    type Iter<'a> = iter::Copied<slice::Iter<'a, Coord<T>>> where T: 'a;
```
kylebarron, line 455: Is there a way to not copy each item? It seems like you don't get much of a benefit by implementing this on slices and arrays if you're still copying every coordinate?

grovesNL: `CoordsIter` works with `Coord` items so I don't think there's a way around copying each coordinate for now. The copying of `Coord` should be really cheap compared to allocating a `Vec` though, which is the cost I'm trying to avoid.

kylebarron: Oh sorry I had a rust brain-fart and forgot copied != cloned 🙈 . Ignore me!

grovesNL: 🙂 all good!

michaelkirk (approved):

PR #1151, opened 2024-02-20
attempt to better document Simplify traits `epsilon` parameter

michaelkirk:
---

I'm actually not sure if this is what the parameter really means, but it's my best guess at this point. Is anyone more familiar willing to clarify?

### DIFF
--- a/geo/src/algorithm/simplify.rs
+++ b/geo/src/algorithm/simplify.rs
@@ -153,7 +153,12 @@ where
 ///
 /// Multi* objects are simplified by simplifying all their constituent geometries individually.
 ///
-/// An epsilon less than or equal to zero will return an unaltered version of the geometry.
+/// A larger `epsilon` means being more aggressive about removing points with less concern for
+/// maintaining the existing shape. Specifically, when you consider whether to remove a point, you
+/// can draw a triangle consisting of the candidate point and the points before and after it.
+/// If the area of this triangle is less than `epsilon`, we will remove the point.
+///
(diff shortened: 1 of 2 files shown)


urschrei:
The triangle explanation (which is correct!) only applies to the Visvalingam-Whyatt algorithm. The RDP algorithm (which underlies Simplify) is distance-based…

michaelkirk:
Thank you! Please take another look @urschrei

geo/src/algorithm/simplify.rs
```diff
@@ -192,7 +198,13 @@ pub trait Simplify<T, Epsilon = T> {
 /// This operation uses the [Ramer–Douglas–Peucker algorithm](https://en.wikipedia.org/wiki/Ramer–Douglas–Peucker_algorithm)
 /// and does not guarantee that the returned geometry is valid.
 ///
-/// An epsilon less than or equal to zero will return an unaltered version of the geometry.
```
michaelkirk, line 195: A related question is can we alter our algorithm such that `epsilon == 0` will remove collinear points, but nothing else?

michaelkirk: Maybe with distance and area calculations it's not feasible to have epsilon == 0 from a robustness perspective...

urschrei: Hmm I don't know that either algorithm can be made to work precisely like that. I can have a look at "the literature".

urschrei (approved):

PR #1155, opened 2024-02-23
Prepare for 0.28.0 release

frewsxcv:
---

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -1,6 +1,9 @@
 # Changes
 
 ## Unreleased
+
+## 0.28.0
+
 * BREAKING: The `HasKernel` trait was removed and it's functionality was merged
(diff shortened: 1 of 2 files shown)


michaelkirk:
Thanks for moving this along. I just queued #1151 for merge. Can we include that? 

Are we dependent on anything in geo-types being released?

michaelkirk:
> Are we dependent on anything in geo-types being released?

Answering my own question... it looks like we're relying on the new unreleased rstar feature, so I think we have to do a geo-types release first.

frewsxcv:
> > Are we dependent on anything in geo-types being released?
> 
> Answering my own question... it looks like we're relying on the new unreleased rstar feature, so I think we have to do a geo-types release first.

https://github.com/georust/geo/pull/1156

urschrei (approved):

PR #1159, opened 2024-03-12
Implement getter methods on AffineTransform to access internal elements

### LINKED ISSUE #1158: Expose fields in AffineTransform struct (or have a way to access the internal values)?
Currently as of `geo=0.28.0`, the AffineTransform struct's fields (implemented in #866) are currently not publicly exposed:

https://github.com/georust/geo/blob/08d3f394a9552e90474b4dacd508927cba84a6d2/geo/src/algorithm/affine_ops.rs#L118-L119

The nine values in the tuple struct are:

```
[[a, b, xoff],
 [d, e, yoff],
 [0, 0, 1]]
```

where `a`/`e` are the x/y resolution, `b`/`d` are x/y rotation (typically 0), and `xoff`/`yoff` are the x/y origin coordinates. See https://www.perrygeo.com/python-affine-transforms.html which is a good reference on this.

Personally, I'd like to have a way to access these individual fields to build up an array of x and y coordinates for pixels in a raster image, e.g. when reading from GeoTIFFs (currently implementing this in https://github.com/weiji14/cog3pio/pull/8, but will do further work that will require access to `a`/`xoff`/`e`/`yoff`). I'm aware of https://docs.rs/gdal/0.16.0/gdal/type.GeoTransform.html which implements something similar, but would prefer to not depend on GDAL if possible :slightly_smiling_face:

Specifically, it would be nice to be able to do something like:

```rust
use geo::AffineTransform;

let transform = AffineTransform::new(200.0, 0.0, 499980.0, 0.0, -200.0, 5300040.0);

let x_scale: f64 = transform.a;
let x_rotation: f64 = transform.b;
let x_origin: f64 = transform.c;
let y_rotation: f64 = transform.d;
let y_scale: f64 = transform.e;
let y_origin: f64 = transform.f;
```

Yes, the affine transformation field names could possibly be renamed to something more easily interpretable.

Would be happy to open a PR for this if this sounds ok. Alternatively, if making the fields public would cause breaking changes, I'd also be ok with implementing a trait that can convert `AffineTransform` to a [`HashMap`](https://doc.rust-lang.org/std/collections/struct.HashMap.html) or some other type that allows for variable-based indexing.

weiji14:
Enable internal elements of the affine transform matrix to be retrieved using getter methods. Included some short documentation and a unit test.

Example usage:

```rust
use geo::AffineTransform;

let transform = AffineTransform::new(10.0, 0.0, 400_000.0, 0.0, -10.0, 500_000.0);

let a: f64 = transform.a();
let b: f64 = transform.b();
let xoff: f64 = transform.xoff();
let d: f64 = transform.d();
let e: f64 = transform.e();
let yoff: f64 = transform.yoff();

assert_eq!(transform, AffineTransform::new(a, b, xoff, d, e, yoff))

```

---

Fixes #1158

### DIFF
--- a/geo/src/algorithm/affine_ops.rs
+++ b/geo/src/algorithm/affine_ops.rs
@@ -323,27 +323,27 @@ impl<T: CoordNum> AffineTransform<T> {
         Self([[a, b, xoff], [d, e, yoff], [T::zero(), T::zero(), T::one()]])
     }
 
-    /// Get the x-resolution value.
+    /// See [AffineTransform::new] for this value's role in the affine transformation.
     pub fn a(&self) -> T {
         self.0[0][0]
     }
-    /// Get the x-rotation value.
+    /// See [AffineTransform::new] for this value's role in the affine transformation.
     pub fn b(&self) -> T {
         self.0[0][1]
     }
-    /// Get the x-offset value.
+    /// See [AffineTransform::new] for this value's role in the affine transformation.
     pub fn xoff(&self) -> T {
         self.0[0][2]
     }
-    /// Get the y-rotation value.
+    /// See [AffineTransform::new] for this value's role in the affine transformation.
     pub fn d(&self) -> T {
         self.0[1][0]
     }
-    /// Get the y-resolution value.
+    /// See [AffineTransform::new] for this value's role in the affine transformation.
     pub fn e(&self) -> T {
         self.0[1][1]
     }
-    /// Get the y-offset value.
+    /// See [AffineTransform::new] for this value's role in the affine transformation.
     pub fn yoff(&self) -> T {
         self.0[1][2]
     }
--- a/geo/src/algorithm/affine_ops.rs
+++ b/geo/src/algorithm/affine_ops.rs
@@ -624,13 +624,13 @@ mod tests {
         assert_eq!(expected, poly);
     }
     #[test]
-    fn test_affine_transform_indexing() {
+    fn test_affine_transform_getters() {
         let transform = AffineTransform::new(10.0, 0.0, 400_000.0, 0.0, -10.0, 500_000.0);
         assert_eq!(transform.a(), &10.0);
         assert_eq!(transform.b(), &0.0);
         assert_eq!(transform.xoff(), &400_000.0);
         assert_eq!(transform.d(), &0.0);
-        assert_eq!(transform.e(), -&10.0);
+        assert_eq!(transform.e(), &-10.0);
         assert_eq!(transform.yoff(), &500_000.0);
     }
 }
--- a/geo/src/algorithm/affine_ops.rs
+++ b/geo/src/algorithm/affine_ops.rs
@@ -321,28 +321,28 @@ impl<T: CoordNum> AffineTransform<T> {
     }
 
     /// Get the x-resolution value.
-    pub fn a(&self) -> &T {
-        &self.0[0][0]
+    pub fn a(&self) -> T {
+        self.0[0][0]
     }
     /// Get the x-rotation value.
-    pub fn b(&self) -> &T {
-        &self.0[0][1]
+    pub fn b(&self) -> T {
+        self.0[0][1]
     }
     /// Get the x-offset value.
-    pub fn xoff(&self) -> &T {
-        &self.0[0][2]
+    pub fn xoff(&self) -> T {
+        self.0[0][2]
     }
     /// Get the y-rotation value.
-    pub fn d(&self) -> &T {
-        &self.0[1][0]
+    pub fn d(&self) -> T {
+        self.0[1][0]
     }
     /// Get the y-resolution value.
-    pub fn e(&self) -> &T {
-        &self.0[1][1]
+    pub fn e(&self) -> T {
+        self.0[1][1]
     }
     /// Get the y-offset value.
-    pub fn yoff(&self) -> &T {
-        &self.0[1][2]
+    pub fn yoff(&self) -> T {
+        self.0[1][2]
     }
 }
 
@@ -626,11 +626,11 @@ mod tests {
     #[test]
     fn test_affine_transform_getters() {
         let transform = AffineTransform::new(10.0, 0.0, 400_000.0, 0.0, -10.0, 500_000.0);
-        assert_eq!(transform.a(), &10.0);
-        assert_eq!(transform.b(), &0.0);
-        assert_eq!(transform.xoff(), &400_000.0);
-        assert_eq!(transform.d(), &0.0);
-        assert_eq!(transform.e(), &-10.0);
-        assert_eq!(transform.yoff(), &500_000.0);
+        assert_eq!(transform.a(), 10.0);
+        assert_eq!(transform.b(), 0.0);
+        assert_eq!(transform.xoff(), 400_000.0);
+        assert_eq!(transform.d(), 0.0);
+        assert_eq!(transform.e(), -10.0);
+        assert_eq!(transform.yoff(), 500_000.0);
     }
 }
--- a/geo/src/algorithm/affine_ops.rs
+++ b/geo/src/algorithm/affine_ops.rs
@@ -4,7 +4,7 @@ use num_traits::ToPrimitive;
 use approx::{AbsDiffEq, RelativeEq};
 
 use crate::{Coord, CoordFloat, CoordNum, MapCoords, MapCoordsInPlace};
-use std::{fmt, ops::Mul, ops::Neg};
+use std::{fmt, ops::Index, ops::Mul, ops::Neg};
 
 /// Apply an [`AffineTransform`] like [`scale`](AffineTransform::scale),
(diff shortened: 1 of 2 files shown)


geo/src/algorithm/affine_ops.rs
```diff
@@ -453,6 +453,34 @@ impl<U: CoordFloat> AffineTransform<U> {
...
+    /// let f: f64 = transform["yoff"];
+    /// ```
+    fn index(&self, idx: &str) -> &Self::Output {
+        match idx {
+            "a" => &self.0[0][0],
+            "b" => &self.0[0][1],
+            "xoff" => &self.0[0][2],
+            "d" => &self.0[1][0],
+            "e" => &self.0[1][1],
+            "yoff" => &self.0[1][2],
```
weiji14, line 478: Need to decide if using these names are ok, or maybe just do `a`, `b`, `c`, `d`, `e`, `f`. See discussion at #1158.

geo/src/algorithm/affine_ops.rs
```diff
@@ -453,6 +453,34 @@ impl<U: CoordFloat> AffineTransform<U> {
...
+            _ => &f64::NAN,
```
weiji14, line 479: Not ideal to return NaN here, would prefer if the possible indexes are limited to just 6 possible values, maybe with an enum?

lnicola:
I personally prefer methods instead of `Index`.

geo/src/algorithm/affine_ops.rs
lnicola: With getters this wouldn't be an issue.

urschrei: Out-of-bounds indexing in Rust is usually a panic, right? I'm not sure if this is strictly out of bounds, though.

weiji14: Yeah on second thought, getter methods might actually be better. I've refactored things in commit a0fdea520dbd501cc177960fe824e1450676f613

geo/src/algorithm/affine_ops.rs
```diff
@@ -300,12 +313,37 @@ impl<T: CoordNum> AffineTransform<T> {
...
-    /// [0, 0, 1]] <-- not part of the input arguments
+    ///  [d, e, yoff],
+    ///  [0, 0, 1]] <-- not part of the input arguments
     /// ```
     pub fn new(a: T, b: T, xoff: T, d: T, e: T, yoff: T) -> Self {
         Self([[a, b, xoff], [d, e, yoff], [T::zero(), T::zero(), T::one()]])
     }
+
+    /// Get the x-resolution value.
+    pub fn a(&self) -> &T {
```
kylebarron, line 324: CoordNum implements `Copy`, so is it better to return `T` here instead?

geo/src/algorithm/affine_ops.rs
```diff
@@ -115,6 +115,19 @@ impl<T: CoordNum, M: MapCoordsInPlace<T> + MapCoords<T, T, Output = Self>> Affin
...
+/// ## Create affine transform manually, and access elements using getter methods
+/// ```
+/// let transform = AffineTransform::new(10.0, 0.0, 400_000.0, 0.0, -10.0, 500_000.0);
+///
+/// let a: f64 = transform.a();
+/// let b: f64 = transform.b();
+/// let c: f64 = transform.xoff();
+/// let d: f64 = transform.d();
+/// let e: f64 = transform.e();
+/// let f: f64 = transform.yoff();
```
kylebarron, line 128: To make this an actual doctest, you could create a second transform with these values and assert it's equal?

```rs
assert_eq!(transform, AffineTransform::new(a, b, c, d, e, f))
```

geo/src/algorithm/affine_ops.rs
```diff
@@ -115,6 +115,19 @@ impl<T: CoordNum, M: MapCoordsInPlace<T> + MapCoords<T, T, Output = Self>> Affin
...
 /// ], max_relative = 1.0);
 /// ```
+///
+/// ## Create affine transform manually, and access elements using getter methods
+/// ```
+/// let transform = AffineTransform::new(10.0, 0.0, 400_000.0, 0.0, -10.0, 500_000.0);
+///
+/// let a: f64 = transform.a();
+/// let b: f64 = transform.b();
+/// let c: f64 = transform.xoff();
```
kylebarron, line 125: Maybe name this variable binding `xoff`? Not sure

geo/src/algorithm/affine_ops.rs
weiji14: Yeah, finding out that I need to deref or call `.to_owned()` on the value in order to use it elsewhere. Maybe it is better to return a cloned copy?

kylebarron: since T is copy, you should be able to just remove the `&` I think

geo/src/algorithm/affine_ops.rs
weiji14: Ok, done at 771061c96a5a0ac6314e129db67d5e3907867f09!

geo/src/algorithm/affine_ops.rs
urschrei: Yes, `xoff` please.

geo/src/algorithm/affine_ops.rs
```diff
@@ -115,6 +115,22 @@ impl<T: CoordNum, M: MapCoordsInPlace<T> + MapCoords<T, T, Output = Self>> Affin
...
+/// use geo::AffineTransform;
+///
+/// let transform = AffineTransform::new(10.0, 0.0, 400_000.0, 0.0, -10.0, 500_000.0);
+///
+/// let a: f64 = transform.a();
+/// let b: f64 = transform.b();
+/// let c: f64 = transform.xoff();
+/// let d: f64 = transform.d();
+/// let e: f64 = transform.e();
+/// let f: f64 = transform.yoff();
```
urschrei, line 130: and `yoff` here.

geo/src/algorithm/affine_ops.rs
```diff
@@ -300,12 +316,37 @@ impl<T: CoordNum> AffineTransform<T> {
...
+        self.0[1][0]
+    }
+    /// Get the y-resolution value.
+    pub fn e(&self) -> T {
+        self.0[1][1]
+    }
+    /// Get the y-offset value.
+    pub fn yoff(&self) -> T {
+        self.0[1][2]
+    }
```
urschrei, line 349: These doc comments – if they're needed at all – should not refer to the terminology ("resolution", "rotation" etc) used in Matt's blog post, as these methods aren't solely intended for vector <--> raster coordinate transformations.

urschrei (changes requested):

geo/src/algorithm/affine_ops.rs
weiji14: Done in a6138f18791d079029fb59b453506f3f754c57bd

geo/src/algorithm/affine_ops.rs
weiji14: Yes, the methods here are only used to retrieve the element values, not for coordinate transformation. So what would be good descriptions for each of these getter methods? Do we simply don't assign any meaning to them, and assume that the user can figure out what the `a`, `b`, `xoff`, `c`, `d`, `yoff` values represent?

urschrei: That depends on whether it's reasonable to assume that someone who's trying to build a custom affine transform is familiar with its mechanics, and has (or is likely to) look at https://docs.rs/geo/latest/geo/algorithm/affine_ops/struct.AffineTransform.html, where the position and thus meaning of a, b, xoff etc are obvious.

weiji14: Ideally, there should be some description when a user hovers over the method in their IDE, so that they don't have to refer to the online documentation. Having a method with a single letter like `.a()` or `.b()` unfortunately isn't very descriptive...

michaelkirk: How about just a link back to the constructor docs, which has an ascii diagram of the matrix?

// See [new] for this value's role in the affine transformation.

weiji14: > How about just a link back to the constructor docs, which has an ascii diagram of the matrix?

Tried this locally, and clicking on the `AffineTransform::new` link in my editor still takes me to the online documentation page.

Screenshot at 2024-04-26 16-42-12

But I'll take this as a good-enough compromise, unless anyone has any other suggestions?

kylebarron: In the rendered docs online the link should point to the correct anchor.

kylebarron (approved):

weiji14:
Anything else I can do to move this PR along?

michaelkirk (approved):
LGTM - and it looks like @urschrei's feedback has been addressed.

urschrei (approved):

PR #1161, opened 2024-03-13
GeodesicBearing and HaversineBearing return result

cojmeister:
---

I'm stuck in the comparison, I'm trying to implement the trait but I'm having issues. Any help would be appreciated.

### DIFF
--- a/geo-types/src/geometry/point.rs
+++ b/geo-types/src/geometry/point.rs
@@ -255,6 +255,21 @@ impl<T: CoordNum> Point<T> {
         self.set_y(lat)
     }
 
+    fn check_x_in_limits(&self) -> Result<(), String> {
+        let x_in_64 =  self.x().to_f64().unwrap();
+        if x_in_64 < -180.0 || x_in_64 > 180.0 {
+            return Err("x is out of bounds: [ -180, 180 ]".into());
+        }
+        Ok(())
+    }
+
+    fn check_y_in_limits(&self) -> Result<(), String> {
+        let y_in_64 =  self.y().to_f64().unwrap();
+        if y_in_64 < -90.0 || y_in_64 > 90.0 {
+            return Err("y is out of bounds: [ -90, 90 ]".into());
+        }
+        Ok(())
+    }
 
     /// Check whether latitude and longitude of a point are within the allowed values of
     ///  - lat [ -90,90 ]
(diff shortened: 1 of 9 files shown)


geo-types/src/geometry/point.rs
```diff
@@ -254,8 +254,46 @@ impl<T: CoordNum> Point<T> {
...
+    /// p0.check_coordinate_limits().unwrap();
+    /// ```
+    pub fn check_coordinate_limits(&self) -> Result<(), String> {
+        self.check_y_in_limits(-90f64, 90f64)?;
+        self.check_x_in_limits(-180f64, 180f64)?;
+        Ok(())
+    }
 }
 
+impl<T: CoordNum, U: PartialOrd<T> + PartialEq<T> + Copy> Point<T> {
```
cojmeister, line 281: This function is preventing the code to compile, bringing the error:
```
error[E0207]: the type parameter `U` is not constrained by the impl trait, self type, or predicates
   --> geo-types/src/geometry/point.rs:281:19
    |
281 | impl<T: CoordNum, U: PartialOrd<T> + PartialEq<T> + Copy> Point<T> {
    |                   ^ unconstrained type parameter
```

geo-types/src/geometry/point.rs
```diff
@@ -254,8 +254,47 @@ impl<T: CoordNum> Point<T> {
...
+
+    fn check_x_in_limits(&self) -> Result<(), String> {
+        let x_in_64 =  self.x().to_f64().unwrap();
+        if x_in_64 < -180.0 || x_in_64 > 180.0 {
+            return Err("x is out of bounds: [ -180, 180 ]".into());
+        }
+        Ok(())
+    }
+
+    fn check_y_in_limits(&self) -> Result<(), String> {
```
michaelkirk, line 266: Points can be projected - they might not be lat/lon.

michaelkirk:
I'm not sure it makes sense to fail when longitude is > 180. Maybe it does, but maybe it's fine to just use modulo since that's probably(?) what the user would want to happen.

Further, as I mentioned previously, we can't assume that coords are lon/lat. 

I'm going to close this for now since it's received no activity, but if you'd like to continue working on it, I'd suggesting first starting a discussion to see if there's any kind of consensus for making this change, and include how other popular libraries handle this case.

PR #1166, opened 2024-03-30
Add MeanCenter trait

JosiahParry:
---

This PR adds a new trait `MeanCenter` which calculates the euclidean mean center of a set of coordinates. It is implemented for all geometry types.

### DIFF
--- a/geo/src/algorithm/mean_center.rs
+++ b/geo/src/algorithm/mean_center.rs
@@ -33,13 +33,12 @@ where
     fn mean_center(&self) -> Point {
         let mut x_sum = T::Scalar::zero();
         let mut y_sum = T::Scalar::zero();
-        let mut n = 0_usize;
+        let n = self.coords_count();
 
         for coord in self.coords_iter() {
             let (xi, yi) = coord.x_y();
             x_sum += xi;
             y_sum += yi;
-            n += 1_usize;
         }
 
         let denominator = n as T::Scalar;
@@ -55,6 +54,7 @@ where
 mod test {
 
     use super::*;
+    use crate::Centroid;
     use geo_types::{
         Coord, Geometry, GeometryCollection, LineString, MultiLineString, MultiPoint, MultiPolygon,
         Point, Polygon, Rect, Triangle,
@@ -75,8 +75,10 @@ mod test {
             Coord { x: 1.0, y: 1.0 }.into(),
         ];
 
-        let mean_center = MultiPoint::new(coords).mean_center();
+        let mpnt = MultiPoint::new(coords);
+        let mean_center = mpnt.mean_center();
         assert_eq!(mean_center, Point::new(0.5, 0.5));
+        assert_eq!(mean_center, mpnt.centroid().unwrap());
     }
 
     #[test]
@@ -91,6 +93,7 @@ mod test {
         let lns = LineString::new(coords);
         let mean_center = lns.mean_center();
         assert_eq!(mean_center, Point::new(0.5, 0.5));
+        assert_eq!(mean_center, lns.centroid().unwrap());
     }
 
     #[test]
@@ -148,6 +151,7 @@ mod test {
 
         let mean_center = mlns.mean_center();
         assert_eq!(mean_center, Point::new(1.5, 1.5));
+        assert_eq!(mean_center, mlns.centroid().unwrap());
     }
 
     #[test]
@@ -162,6 +166,7 @@ mod test {
         let lns = LineString::new(coords);
         let mean_center = lns.mean_center();
         assert_eq!(mean_center, Point::new(0.25, 0.25));
+        assert_eq!(mean_center, lns.centroid().unwrap());
     }
 
     #[test]
@@ -187,6 +192,7 @@ mod test {
 
         let mean_center = mpoly.mean_center();
         assert_eq!(mean_center, Point::new(1.25, 1.25));
+        assert_eq!(mean_center, mpoly.centroid().unwrap());
     }
 
     #[test]
@@ -198,6 +204,7 @@ mod test {
         let triangle = Triangle::new(v1, v2, v3);
         let mean_center = triangle.mean_center();
         assert_eq!(mean_center, Point::new(1.0 / 3.0, 1.0 / 3.0));
+        assert_eq!(mean_center, triangle.centroid());
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


geo/src/algorithm/mean_center.rs
```diff
@@ -0,0 +1,253 @@
...
+}
+
+impl<T> MeanCenter for T
+where
+    T: CoordsIter<Scalar = f64>,
+{
+    fn mean_center(&self) -> Point {
+        let mut x_sum = T::Scalar::zero();
+        let mut y_sum = T::Scalar::zero();
+        let mut n = 0_usize;
```
lnicola, line 36:
```suggestion
        let mut n = self.coords_count();
```

lnicola:
I think this is normally called a [centroid](https://en.wikipedia.org/wiki/Centroid).

JosiahParry:
There are subtle differences between centroids and the mean center of point sets. I've added some tests that compare the centroid for each geometry with the mean center. There are differences, notably for a polygons. 

Here's [a blog post](https://glenbambrick.com/tag/weighted-mean-center/) that (briefly) describes its utility. I was planning on adding a ` fn weighted_mean_center()` as well as a follow up. I can add that to this PR if you think it would make the trait more motivating .

urschrei:
> There are subtle differences between centroids and the mean center of point sets. I've added some tests that compare the centroid for each geometry with the mean center. There are differences, notably for a polygons. 
> 
> 
> 
> Here's [a blog post](https://glenbambrick.com/tag/weighted-mean-center/) that (briefly) describes its utility. I was planning on adding a ` fn weighted_mean_center()` as well as a follow up. I can add that to this PR if you think it would make the trait more motivating .
> 
> 
> 
> 

It's obviously an automatic approval from me because the blog post uses Irish CSO data, but I'd love to see the weighted mean centre method too.

JosiahParry:
🤣 heard. I've already pushed the weighted_mean_center() method and doc changes. I just need to add tests which I'll find time for tonight or tomorrow. 🫡

frewsxcv (commented):
Can you add an entry for this in the algorithm index? https://docs.rs/geo/latest/geo/#algorithms

JosiahParry:
> I think this is normally called a [centroid](https://en.wikipedia.org/wiki/Centroid).

Yup, you're right! OOPS! 🙈 

The mean center for `MultiLineString`, `MultiPolygon`, and `GeomtryCollection` is where this differs. Since this is a point pattern technique, there must be one point per feature. Each geometry's centroid is used to represent it. Then the mean center (centroid) of those is used. 

This technique is most useful for feature collections. For example, finding the weighted mean center of 311 reports grouped by category (MultiPoint per category) and the weight would be days since report or estimate cost to fix etc. 

[Related doc from arcgis implementation](https://pro.arcgis.com/en/pro-app/latest/tool-reference/spatial-statistics/h-how-mean-center-spatial-statistics-works.htm)

I've updated the trait doc and implementation to reflect this.

geo/src/algorithm/mean_center.rs
```diff
@@ -0,0 +1,694 @@
...
+    type Output = Option<Point>;
+    fn mean_center(&self) -> Self::Output {
+        // take the centroid of each geometry
+        // the mean center is calculated from the centroids
+        // any geometry that does not have a centroid is filtered out
+        // this should be rare
+        let centroids = self
+            .iter()
+            .map(|g| g.centroid())
+            .filter(|c| c.is_some())
```
michaelkirk, line 246: You can replace the `map/filter/map(unwrap)` with `flat_map(|g| g.centroid())`

(this occurs in several places in this PR)

geo/src/algorithm/mean_center.rs
```diff
@@ -0,0 +1,694 @@
...
+/// mean center.
+///
+/// For [`Line`], [`Triangle`], [`Rect`], [`LineString`], [`Polygon`], and [`MultiPoint`], weights
+/// are applied  to each coordinate in the geometry. For [`MultiLineString`], [`MultiPolygon`],
+/// and [`GeometryCollection`], the weights are applied to the centroid of each component
+/// geometry. If a geometry does not have a centroid, it is filtered out.
+///
+/// The `weights` argument takes a slice which is used as an iterator which is cycled.
+/// If there are fewer weights than coordinates, the weights will be recycled until the
+/// calculation has completed.
```
michaelkirk, line 28: > If there are fewer weights than coordinates, the weights will be recycled until the calculation has completed.

Is there a useful reason for this cycle behavior or some precedent? 

Or is this more like "do something, anything, so long as we don't explode"?

geo/src/algorithm/mean_center.rs
```diff
@@ -0,0 +1,694 @@
...
+        let mut x_sum = 0.0;
+        let mut y_sum = 0.0;
+        let mut weight_sum = 0.0;
+        let n = self.coords_count();
+
+        if n == 0 {
+            return None;
+        }
+
+        for ((i, coord), weight) in self.coords_iter().enumerate().zip(weights.iter().cycle()) {
```
michaelkirk, line 146: Hm. The blog post uses centroids here as well right?:

            xy_arr[i] = (ft_geom.Centroid().GetX() * weight, ft_geom.Centroid().GetY() * weight)

geo/src/algorithm/mean_center.rs
```diff
@@ -0,0 +1,694 @@
...
+                let x = x_sum / weight_sum;
+                let y = y_sum / weight_sum;
+
+                Some(Point::new(x, y))
+            }
+        }
+    };
+}
+
+impl_mean_center_option!(LineString);
```
michaelkirk, line 127: It looks like you're computing centroid here. 

Two things:

1. The blog uses one weight per geometry, not per coord in the geometry.

2. For linear features like a LineString, the blog post uses midpoint instead of centroid. Though I can't think of an application of this algorithm where using a midpoint would be more useful than a centroid. I know there are applications where you want the "middle" of the geometry to be within a geometry, which might not happen with centroid  (which is why people love polylabel), but in this case, we're eventually averaging a bunch of points together, so there's no guarantee that the output point will be inside of any geometry anyway.

michaelkirk (changes requested):
Do you have an application for this algorithm? I haven't really grokked its utility and am a little skeptical of its specification.

I'm not allergic to merging it if you have a use for it and can fix it up.

PR #1157, opened 2024-02-26
Add geo-traits crate

kylebarron:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/main/CODE_OF_CONDUCT.md).
- [ ] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

Updated version of #1019 that's clean against the latest `main`. Aside from having no extra diff from `main`, it also has:

- Automatic iterators. See https://github.com/geoarrow/geoarrow-rs/pull/427. This adds a `Sized` required bound, but means that all sources get iterators out of the box with minimal required impls.

PR #1169, opened 2024-04-13
Poof of concept: `geo_traits::Coord`

frewsxcv:
---

### DIFF
--- a/geo-types/src/error.rs
+++ b/geo-types/src/error.rs
@@ -23,7 +23,7 @@ impl fmt::Display for Error {
 
 #[cfg(test)]
 mod test {
-    use crate::{Geometry, Point, Rect};
+    use crate::{Geometry, Point, Rect, Coord};
     use alloc::string::ToString;
     use core::convert::TryFrom;
 
@@ -32,7 +32,7 @@ mod test {
         let point = Point::new(1.0, 2.0);
         let point_geometry = Geometry::from(point);
 
-        let rect = Rect::new(Point::new(1.0, 2.0), Point::new(3.0, 4.0));
+        let rect = Rect::<Coord>::new(Point::new(1.0, 2.0), Point::new(3.0, 4.0));
         let rect_geometry = Geometry::from(rect);
 
         Point::try_from(point_geometry).expect("failed to unwrap inner enum Point");
--- a/geo-types/src/geometry/line_string.rs
+++ b/geo-types/src/geometry/line_string.rs
@@ -1,8 +1,8 @@
 #[cfg(any(feature = "approx", test))]
 use approx::{AbsDiffEq, RelativeEq};
 
-use crate::geo_traits;
-use crate::{Coord, CoordNum, Line, Point, Triangle};
+use crate::geo_traits::{self, Coord};
+use crate::{Line, Point, Triangle};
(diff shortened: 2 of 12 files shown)


geo-types/src/private_utils.rs
```diff
@@ -58,10 +60,13 @@ fn get_min_max<T: PartialOrd>(p: T, min: T, max: T) -> (T, T) {
     }
 }
 
-pub fn line_segment_distance<T, C>(point: C, start: C, end: C) -> T
+pub fn line_segment_distance<T, C>(
+    point: impl Into<C>,
```
michaelkirk, line 64: Do we need the `Into`? Can't we call the methods directly on our `C`?

geo-types/src/error.rs
```diff
@@ -32,7 +32,7 @@ mod test {
         let point = Point::new(1.0, 2.0);
         let point_geometry = Geometry::from(point);
 
-        let rect = Rect::new(Point::new(1.0, 2.0), Point::new(3.0, 4.0));
+        let rect = Rect::<Coord>::new(Point::new(1.0, 2.0), Point::new(3.0, 4.0));
```
michaelkirk, line 35: Hm... was this not compiling without this change?

geo-types/src/geometry/line_string.rs
```diff
@@ -134,42 +136,48 @@ use core::ops::{Index, IndexMut};
...
-pub struct LineString<T: CoordNum = f64>(pub Vec<Coord<T>>);
+pub struct LineString<C: geo_traits::Coord = Point<f64>>(pub Vec<C>);
 
 /// A [`Point`] iterator returned by the `points` method
-#[derive(Debug)]
-pub struct PointsIter<'a, T: CoordNum + 'a>(::core::slice::Iter<'a, Coord<T>>);
+pub struct PointsIter<'a, C: geo_traits::Coord + 'a>(::core::slice::Iter<'a, C>);
+
+impl<'a, C: fmt::Debug + geo_traits::Coord + 'a> fmt::Debug for PointsIter<'a, C> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
```
michaelkirk, line 145: Do you know why we have to roll our own Debug here?

michaelkirk (commented):
I'd be potentially interested in merging a Coord trait (though I have a bikeshed preference for calling it `Point`, see: https://github.com/georust/geo/pull/1157#issuecomment-2118489464

One thought is we could incubate it in `geo`, which has breaking releases relatively often, and promote it to `geo_types` once we feel it's stable.

I did have a couple of concerns with this PR about the debug and explicit type annotations. In the actual code in this PR it's obviously NBD, but I'm wondering if they are indicative of changes that will break for downstream users and if there's anything we could do to prevent them.

geo-types/src/geometry/line_string.rs
frewsxcv: Fixed in https://github.com/georust/geo/pull/1169/commits/559dc725c104ae87b1b994b124c7aebe7dc98de4

frewsxcv:
Closing in favor of https://github.com/georust/geo/pull/1157

PR #1170, opened 2024-04-15
fix: error in earth radius

louisgreiner:
Just a typo in the documentation, const is ok (according to IUGG R1 radius https://en.wikipedia.org/wiki/Earth_radius)

See const here : https://github.com/georust/geo/blob/main/geo/src/lib.rs#L245

---


urschrei (approved):
Well spotted. Thanks!

PR #1191, opened 2024-06-12
Fix issue in Debug impl for AffineTransform where yoff is shown instead of xoff

yasamoka:
---

Thank you for your great work on this crate!


michaelkirk (approved):
Good catch, thank you!

Issue #1195, opened 2024-06-25
Semantics for AffineTransform composition is reversed


adampowerdeciphex:
When composing multiple AffineTransforms, one would expect that the composed transform is the equivalent of applying the individual transforms from left to right. Instead it is the same as applying right to left.  

The solution would be to reverse the matrix multiplication of the [compose()](https://docs.rs/geo/0.28.0/src/geo/algorithm/affine_ops.rs.html#133) function. Alternatively this could be made clear in the docs as there may be people that depend on this right to left behaviour.

# Sample code
````rust
use geo::{AffineOps, AffineTransform, Point};

fn main() {
    let point = Point::new(1.,0.);

    let translate = AffineTransform::translate(1., 0.);
    let scale = AffineTransform::scale(4., 1., [0., 0.]);
    let composed = translate.compose(&scale);

    assert_eq!(point.affine_transform(&translate), Point::new(2., 0.));
    assert_eq!(point.affine_transform(&scale), Point::new(4., 0.));
    assert_eq!(point.affine_transform(&translate).affine_transform(&scale), Point::new(8., 0.));

    // This assert fails, as the composed transform instead produces Point::new(5., 0.)
    assert_eq!(point.affine_transform(&composed), Point::new(8., 0.));
}
````

lnicola:
For prior art, see https://gdal.org/api/raster_c_api.html#_CPPv424GDALComposeGeoTransformsPKdPKdPd.

lnicola:
But I don't understand, why do you say it's right-to-left? It looks left-to-right to me (`composed` is equivalent to `translate`, then `scale`).

adampowerdeciphex:
Sorry I should have mentioned that the final assert here fails. 
The expected result would be `Point::new(8., 0.)`, but instead produces `Point::new(5., 0.)`

michaelkirk:
Fixed in #1196 - please take a look. Thank you for catching this and providing a test!

adampowerdeciphex:
Fantastic, thanks for the quick fix! 😃

PR #1196, opened 2024-06-25
Affine flip `compose` args to be conventional

michaelkirk:
---

FIX: https://github.com/georust/geo/issues/1195

e.g. for "conventional" https://gdal.org/api/raster_c_api.html#_CPPv424GDALComposeGeoTransformsPKdPKdPd

```
/// The resulting geotransform is the equivalent to padfGT1 and then padfGT2 being applied to a point.
/// Parameters:
///     padfGT1 -- the first geotransform, six values.
///     padfGT2 -- the second geotransform, six values.
///     padfGTOut -- the output geotransform, six values, may safely be the same array as padfGT1 or padfGT2.
void GDALComposeGeoTransforms(const double *padfGeoTransform1, const double *padfGeoTransform2, double *padfGeoTransformOut)
```

Note: 

> `padfGT1` **and then** `padfGT2`

... previously we were effectively doing `padfGT2` **and then** `padfGT1`.

I also simplified some of the examples to have more understandable
input/output. I'm not good enough at mental matrix math to know what to
expect from a rotated skew.

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -6,6 +6,8 @@
   * <https://github.com/georust/geo/pull/1159>
 * Fix issue in Debug impl for AffineTransform where yoff is shown instead of xoff
   * <https://github.com/georust/geo/pull/1191>
+* Fix `AffineTransform::compose` ordering to be conventional - such that the argument is applied *after* self.
+  * <https://github.com/georust/geo/pull/1196>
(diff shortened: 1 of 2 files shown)


lnicola (approved):
LGTM, I guess, but there's a conflict on the changelog.

Issue #1184, opened 2024-05-18
performance for multipolygon contains much worse than for geos


gauteh:
Hi,

I'm trying to convert some code for roaring-landmask (https://github.com/gauteh/roaring-landmask) from geos to geo (https://github.com/gauteh/roaring-landmask/pull/27). However, checking if a point is contained by a large multipolygon (with all countries from GSHHG) is far slower with geo than with geos. With geos I used `PreparedGeometry`, does something similar exist?

geos:
```
test shapes::tests::benches::test_contains_in_ocean ... bench:       1,586 ns/iter (+/- 45)
test shapes::tests::benches::test_contains_on_land  ... bench:       1,499 ns/iter (+/- 169)
```

geo:
```
test shapes::tests::benches::test_contains_in_ocean ... bench:  10,713,374 ns/iter (+/- 356,454)
test shapes::tests::benches::test_contains_on_land  ... bench:   6,709,172 ns/iter (+/- 255,050)
```

urschrei:
No, but you can decompose the multipolygon into polygons and add them to an rstar spatial index: https://docs.rs/rstar/latest/rstar/struct.RTree.html. The rtree is serializable, re-exported by geo (so it should "just work" for bulk loading the multipolygon Vec), and you can associate extra data with each polygon by using https://docs.rs/rstar/latest/rstar/primitives/struct.GeomWithData.html

gauteh:
Thanks, for the tip. Do you know if that is how geos is doing it?

urschrei:
A geos prepared geometry is an rtree behind the scenes, yes. So things like point in multipolygon query the tree transparently.

gauteh:
I'm trying to add all my Polygons to an RTree, and then check if a point is contained by any of the polygons. But I'm getting this error:

```
   Compiling roaring-landmask v0.7.3 (/home/gauteh/met/roaring-landmask)
error[E0599]: the method `nearest_neighbor` exists for struct `RTree<Polygon>`, but its trait bounds were not satisfied
  --> src/shapes.rs:74:28
   |
74 |         let po = self.geom.nearest_neighbor(&p);
   |                            ^^^^^^^^^^^^^^^^ method cannot be called on `RTree<Polygon>` due to unsatisfied trait bounds
   |
  ::: /home/gauteh/.cargo/registry/src/index.crates.io-6f17d22bba15001f/geo-types-0.7.13/src/geometry/polygon.rs:71:1
   |
[...]

PR #1198, opened 2024-07-08
graph send

gauteh:
- **existing bench**
- **Add `PreparedGeometry` to speed up repeated `Relate` operations.**
- **geometry_graph: make Send (rm Rc + RefCell)**
- **more &mut's rather than Rc<RefCells>.**


geo/src/algorithm/relate/geomgraph/index/rstar_edge_set_intersector.rs
```diff
@@ -1,105 +1,64 @@
...
-            .iter()
-            .flat_map(|edge| {
-                let start_of_final_segment: usize = RefCell::borrow(edge).coords().len() - 1;
-                (0..start_of_final_segment).map(|segment_i| Segment::new(segment_i, edge))
-            })
-            .collect();
-        let tree = RTree::bulk_load(segments);
-
-        for (edge0, edge1) in tree.intersection_candidates_with_other_tree(&tree) {
-            if check_for_self_intersecting_edges || edge0.edge.as_ptr() != edge1.edge.as_ptr() {
```
gauteh, line 71: Sending two mutable references to `add_intersection` doesn't work without the Rc<RefCell<Edge>>. But this results in intersections being added to the same edge twice. Is that intentional?

geo/src/algorithm/relate/geomgraph/geometry_graph.rs
```diff
@@ -44,44 +50,106 @@ impl<F> GeometryGraph<'_, F>
 where
     F: GeoFloat,
 {
-    pub fn edges(&self) -> &[Rc<RefCell<Edge<F>>>] {
+    pub(crate) fn set_tree(&mut self, tree: Arc<RTree<Segment<F>>>) {
+        self.tree = Some(tree);
+    }
+
+    pub(crate) fn get_or_build_tree(&self) -> Arc<RTree<Segment<F>>> {
```
gauteh, line 57: The tree should now always be built.

geo/src/algorithm/relate/geomgraph/index/rstar_edge_set_intersector.rs
michaelkirk: > But this results in intersections being added to the same edge twice. Is that intentional?

I'm not sure I understand what you mean, but it sounds like a bug. Can you provide a little example?

gauteh: Here: https://github.com/gauteh/geo/blob/graph-send/geo/src/algorithm/relate/geomgraph/index/segment_intersector.rs#L139 both edge0 and edge1 are pointing to the same object for the case when `check_for_self_intersecting_edges` is `true`.

michaelkirk: Ah thanks for the link. Based on the comment it seems like it wasn't an accident, but it seems weird, and to be honest, I don't remember why we'd want that.

It seems to be present in the JTS source that this was ported from. That means it's probably, but not necessarily correct 😄:

https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/geomgraph/index/SegmentIntersector.java#L179

It might be a deeper dive than I have time for today to answer this question, but if you'd like to dig deeper and document any findings that could be helpful. i.e. do all the tests still pass if we "fix" it?

gauteh: I will probably just have to reproduce the behaviour for now, can put in a comment for the future.

michaelkirk:
#1197 has been merged - could you do a rebase or merge to minimize the diff?

gauteh:
I have rebased, but I haven't had time to work any further on this yet.

gauteh:
Some of the test fails seem like numeric precision errors, they are also present on `main`. There is one failing test left (debug assert), which I don't know why is failing.

```
 3  failed: algorithm::interior_point::test::polygon_cell_test                          ▐
 thread 'algorithm::interior_point::test::polygon_cell_test' panicked at geo/src/algorit▐
 hm/relate/geomgraph/edge_end_bundle_star.rs:112:21:                                    ▐
 side_location conflict with coordinate: Coord { x: 2.0, y: 1.0 }, right_location: Insid▐
 e, current_location: Outside                                                           ▐

```

geo/src/algorithm/relate/geomgraph/index/rstar_edge_set_intersector.rs
gauteh: Made separate methods for testing against self.

gauteh:
The lifetime in prepared geometry is tedious to work with. Could that be replaced by a Cow or maybe a owned version of the geometrygraph?

gauteh:
> The lifetime in prepared geometry is tedious to work with. Could that be replaced by a Cow or maybe a owned version of the geometrygraph?

Just realized that is exactly how it is now...

gauteh:
This seems to be working now.

It appears that a new geometrygraph + intersection matrix is created for every point I want to check against the preparedgeometry.

urschrei:
> It appears that a new geometrygraph + intersection matrix is created for every point I want to check against the preparedgeometry.

Can you expand on this and / or illustrate it with a minimal test case?

gauteh:
My use case is roaring-landmask, here's a PR where I am running against these changes:  https://github.com/gauteh/roaring-landmask/pull/28, in https://github.com/gauteh/roaring-landmask/blob/georust-prepared/src/shapes.rs#L129 I check if a point is in any of the polygons using `is_covers`:

```
self.geom.relate(&p).is_covers()
```

this creates a geometrygraph of the point `p` and tests it against the PreparedGeometry. I believe geos has a PreparedPolygon/PreparedMultiPolygon with specializations for checking various relates towards a point.

In roaring-landmask this may be tested with e.g.: `cargo test --features nightly --release  -j 1 many -- --nocapture`

michaelkirk:
> It appears that a new geometrygraph + intersection matrix is created for every point I want to check against the preparedgeometry.

That's right. The difference with `PreparedGeometry` is that the Polygon's geometry graph isn't created from scratch every time.

The point's graph will be created from scratch, since it's a different point every time, but it should also be a pretty trivial graph.

> I check if a point is in any of the polygons using is_covers:

Note that in the case of points and polygons, `polygon.relate(points).is_covers()` is the same as `polygon.instersects(point)`. 

The GeometryGraph is a very general, built to handle all the de-9im relations for all geometry types. If you are only interested in "point intersects polygon" checks, there is likely a faster approach involving less machinery - maybe ray casting, but I don't think we have an implementation of that yet.

gauteh:
> > It appears that a new geometrygraph + intersection matrix is created for every point I want to check against the preparedgeometry.
> 
> That's right. The difference with `PreparedGeometry` is that the Polygon's geometry graph isn't created from scratch every time.
> 
> The point's graph will be created from scratch, since it's a different point every time, but it should also be a pretty trivial graph.

Indeed, the tree has no members.

> 
> > I check if a point is in any of the polygons using is_covers:
> 
> Note that in the case of points and polygons, `polygon.relate(points).is_covers()` is the same as `polygon.instersects(point)`.

Is there any performance difference? Or exactly the same?

> 
> The GeometryGraph is a very general tool, built to handle all the de-9im relations for all geometry types. If you are only interested in "point intersects polygon" checks, there is likely a faster approach involving less machinery - maybe ray casting, but I don't think we have an implementation of that yet.

So it appears, and that makes sense (to start with). I think that this PR is mostly functional now. I had to make a few slightly messy changes to satisfy borrowing rules, without doing a much greater re-structuring. There is some cleanup to be done, but it should be a useful feature (and actually a pretty cool plus over geos, where it is not safe to share the PreparedGeometry) to other applications. Sadly, I don't think it will make much difference to my issue (https://github.com/georust/geo/issues/1184).

geo/src/algorithm/relate/relate_operation.rs
```diff
@@ -85,27 +82,48 @@ where
             .compute_self_nodes(Box::new(self.line_intersector.clone()));
 
         // compute intersections between edges of the two input geometries
-        let segment_intersector = self
-            .graph_a
-            .compute_edge_intersections(&self.graph_b, Box::new(self.line_intersector.clone()));
+        // let segment_intersector = self
```
urschrei, line 85: I'm not clear on what's happening here.

urschrei:
@michaelkirk I've rebased and tidied this up and appeased young master Clipperton – I only have one query regarding the "satisfy rust's borrowing rules section" but otherwise this hasn't added any obvious complexity to our `GeomGraph` machinery (though it's difficult to imagine how that would be possible…), and the `Send` bound will probably come in useful at some point. WDYT?

geo/src/algorithm/relate/geomgraph/geometry_graph.rs
```diff
@@ -35,7 +34,7 @@ where
 {
     arg_index: usize,
     parent_geometry: GeometryCow<'a, F>,
-    tree: Option<Rc<RTree<Segment<F>>>>,
+    tree: Option<RTree<Segment<F>>>,
```
michaelkirk, line 37: There are some positive changes in this PR, but there is a serious perf regression with:
```
 relate prepared polygons
                        time:   [142.78 ms 143.14 ms 143.63 ms]
                        change: [+192.56% +193.59% +194.76%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

I haven't zeroed in on the cause yet - but this line is suspect. The point of the `Rc<Tree>` is that it was cheap to clone. It seems like we've removed that and now have a "cache the tree and invalidate it as necessary" approach.

Note there are also some nice performance wins to the *non*-prepared geometry operations in this PR, but I personally don't think it's worth it as phrased.

geo/src/algorithm/relate/geomgraph/index/prepared_geometry.rs
```diff
@@ -5,7 +5,7 @@ use crate::GeometryCow;
 use crate::{GeoFloat, Relate};
 
 use std::cell::RefCell;
-use std::rc::Rc;
+use std::sync::Arc;
```
michaelkirk, line 8: All these Arc imports seem unused. Mr. Clipperton is so idiosyncratic with his hangups.

michaelkirk (approved):

michaelkirk (changes requested):
whoops - didn't mean to approve yet.

geo/src/algorithm/relate/geomgraph/geometry_graph.rs
urschrei: Ooof yeah that regression is a show-stopper at the moment.

geo/src/algorithm/relate/geomgraph/geometry_graph.rs
```diff
@@ -49,23 +47,33 @@ impl<F> GeometryGraph<'_, F>
...
     }
 
-    pub(crate) fn build_tree(&self) -> RTree<Segment<F>> {
+    pub(crate) fn update_tree(&mut self) {
+        if self.tree.is_none() {
+            self.tree = Some(self.build_tree());
+        }
+    }
+
+    fn invalidate_tree(&mut self) {
```
michaelkirk, line 67: This got a little more complex.

geo/src/algorithm/relate/geomgraph/index/rstar_edge_set_intersector.rs
```diff
@@ -15,42 +15,64 @@ where
...
             if check_for_self_intersecting_edges || segment_0.edge_idx != segment_1.edge_idx {
-                let edge_0 = &edges[segment_0.edge_idx];
-                let edge_1 = &edges[segment_1.edge_idx];
-                segment_intersector.add_intersections(
-                    edge_0,
-                    segment_0.segment_idx,
-                    edge_1,
-                    segment_1.segment_idx,
-                );
+                if segment_1.edge_idx == segment_0.edge_idx {
```
michaelkirk, line 25: This got quite a lot more complex.

geo/src/algorithm/relate/geomgraph/index/segment_intersector.rs
```diff
@@ -95,25 +94,85 @@ where
         false
     }
 
+    // Copy of `add_intersections` specialized for a single 'edge' against itself.
+    pub fn add_intersections_against_self(
```
michaelkirk, line 98: This got quite a lot more complex.

geo/src/algorithm/relate/relate_operation.rs
```diff
@@ -85,27 +82,48 @@ where
...
+        //     .graph_a
+        //     .compute_edge_intersections(&mut self.graph_b, Box::new(self.line_intersector.clone()));
+
+        // this is a copy of the above functionality to satisfy rust borrowing rules.
+        // from here [...
+        let mut segment_intersector =
+            SegmentIntersector::new(Box::new(self.line_intersector.clone()), false);
```
michaelkirk, line 92: This got  a bit more complex.

michaelkirk (changes requested):
It's nice to get rid of the `Rc<RefCell<T>>`, but there is a lot of new complexity for the sake of Send. In the meanwhile, people can Send their geometry and make a PreparedGeometry on each thread.  

Is there a way it can be simplified?

frewsxcv:
@gauteh I'm going to convert this to a draft until the comments and merge conflicts are addressed

frewsxcv:
Development seems to have stalled on this, so I'm going to close this pull request. @gauteh If you get around to resuming work on this, please reopen!

PR #1199, opened 2024-07-10
Linear implementation of Frechet distance

gliderkite:
---

This PR is due to the fact that when using the `frechet_distance` methods on our tools with long routes (200km+) we were not able to run it due to `fatal runtime error: stack overflow`.

This patch includes a new linear implementation (inspired from [here](https://github.com/joaofig/discrete-frechet/tree/master)) that removes recursion and allows to avoid the stack overflow fatal errors, while improving overall performances (~70% faster when testing 1000 10km-long routes).

### DIFF
--- a/geo/src/algorithm/frechet_distance.rs
+++ b/geo/src/algorithm/frechet_distance.rs
@@ -47,12 +47,12 @@ where
 {
     fn frechet_distance(&self, ls: &LineString<T>) -> T {
         if self.coords_count() != 0 && ls.coords_count() != 0 {
-            let mut data = Data {
-                cache: vec![vec![T::nan(); ls.coords_count()]; self.coords_count()],
+            Data {
+                cache: vec![T::zero(); self.coords_count() * ls.coords_count()],
                 ls_a: self,
                 ls_b: ls,
-            };
-            data.compute(self.coords_count() - 1, ls.coords_count() - 1)
+            }
+            .compute_linear()
         } else {
             T::zero()
         }
@@ -63,7 +63,7 @@ struct Data<'a, T>
 where
     T: GeoFloat + FromPrimitive,
 {
-    cache: Vec<Vec<T>>,
+    cache: Vec<T>,
     ls_a: &'a LineString<T>,
     ls_b: &'a LineString<T>,
 }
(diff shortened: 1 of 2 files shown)


geo/src/algorithm/frechet_distance.rs
```diff
@@ -72,19 +72,26 @@ impl<'a, T> Data<'a, T>
...
-        if self.cache[i][j].is_nan() {
-            let eucl = Point::from(self.ls_a[i]).euclidean_distance(&Point::from(self.ls_b[j]));
-            self.cache[i][j] = match (i, j) {
-                (0, 0) => eucl,
-                (_, 0) => self.compute(i - 1, 0).max(eucl),
-                (0, _) => self.compute(0, j - 1).max(eucl),
-                (_, _) => ((self.compute(i - 1, j).min(self.compute(i - 1, j - 1)))
-                    .min(self.compute(i, j - 1)))
-                .max(eucl),
-            };
```
urschrei, line 85: Please leave a comment pointing to the source of this implementation (https://github.com/joaofig/discrete-frechet/tree/master? Or a more direct link), and a brief explanation of how it works.

urschrei:
Oh and if you could contrive a test that fails using the old algorithm but passes now, that would be useful…

geo/src/algorithm/frechet_distance.rs
gliderkite: The implementation is not exactly the same, as I use a single vector instead of a matrix to reduce number of allocations and improve data locality, but I will add a link to it if you prefer. I did leave all the other links that point to the resources that explain the theory behind it.

gliderkite:
> Oh and if you could contrive a test that fails using the old algorithm but passes now, that would be useful…

That would be tricky to do, because we would need to generate a huge route, store it somewhere so that it can be accessed locally and on CI to run tests. If you have suggestions or prior examples on how to do this as part of this repo (not sure if you already use git lfs, or have other tests that read from test resources, eg `include_str`, etc), I will be happy to add this.

urschrei:
> > Oh and if you could contrive a test that fails using the old algorithm but passes now, that would be useful…
> 
> That would be tricky to do, because we would need to generate a huge route, store it somewhere so that it can be accessed locally and on CI to run tests. If you have suggestions or prior examples on how to do this as part of this repo (not sure if you already use git lfs, or have other tests that read from test resources, eg `include_str`, etc), I will be happy to add this.

Ah, in that case probably not worth the hassle.

geo/src/algorithm/frechet_distance.rs
gliderkite: I have added a link to the reference implementation, let me know if that looks good to you.

gliderkite:
> Ah, in that case probably not worth the hassle.

I can create a deterministic test where we generate two very big `LineString` stored in memory to then compute the Frechet distance. We are not interested in the result, just that the test completes without panicking. This test will fail with the current implementation (but pass with the new one). If this sounds good to you I can add it as part of this PR.

urschrei:
> If this sounds good to you I can add it as part of this PR.

If you have time / capacity, that would be great.

urschrei (approved):
Thanks! I'll leave this open for another day or so in case anyone else wants to chime in.

PR #1201, opened 2024-07-26
Use latest Docker images (max 1.79)

urschrei:
---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,7 +53,7 @@ jobs:
       matrix:
         container_image:
           # Use the latest stable version. No need for older versions.
-          - "georust/geo-ci:proj-9.4.0-rust-1.79"
+          - "ghcr.io/georust/geo-ci:proj-9.4.0-rust-1.79"
     container:
       image: ${{ matrix.container_image }}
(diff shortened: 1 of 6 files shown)


urschrei:
NB: I had to add an "allow dead code" exception to our Sweep-line impl to allow the lint to pass as it became more strict at some point post Rust 1.75, so happy to add a Changelog entry for this PR if need be.

michaelkirk:
I think we should keep the MSRV in Cargo.toml synced with whatever we're running on CI.

And then I'd say bumping MSRV deserves a change log entry.

urschrei:
> I think we should keep the MSRV in Cargo.toml synced with whatever we're running on CI.
> 
> And then I'd say bumping MSRV deserves a change log entry.

Done (and thanks for fortuitously updating the Docker images yesterday!)

michaelkirk:
It was no coincidence! Thanks for finishing the job.

michaelkirk (approved):
I pushed up another commit to bump `rust-version` in Cargo.toml.

urschrei:
Ah, I just…didn't commit my Cargo changes. Very good.

PR #1202, opened 2024-07-26
Generic implementation of geodesic_destination for Point

arpadav:
---

Previously was only implemented for `f64`

Question: Not too happy about the `unwrap`'s, any way to avoid? Or will the trait bounds of `CoordFloat` prevent `to_f64` and `from` from ever being `None`?

### DIFF
--- a/geo/src/algorithm/geodesic_destination.rs
+++ b/geo/src/algorithm/geodesic_destination.rs
@@ -1,5 +1,5 @@
 use crate::Point;
-use geo_types::CoordNum;
+use geo_types::{CoordFloat, CoordNum};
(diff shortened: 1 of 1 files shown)


arpadav:
Ultimately agree with your sentiment. The last thing I want is to be misleading for the sake of convenience, and `MapCoords` looks promising despite the overhead

Can close, but can always revisit if [geographiclib](https://docs.rs/geographiclib/latest/geographiclib/) updates the WGS84 impl to use generics rather than `f64`. Or if someone re-writes the calculation without said dependency

michaelkirk:
Closing this based on previous discussion.

PR #1204, opened 2024-07-29
Relax constraints on `Coord` and associated structures

dsgallups:
---
Hello! I am currently working on a rendering library for leptos, which uses geo. I would like to make interactive `svg`s in the dom, but that requires that some of the inner types of several `geo-types` structs should have their constraints relaxed.

This change does not affect current implementations in other crates, as these changes only allow types that are not `CoordNum` (but potentially implement `Add` or `PartialEq`, etc.) to access crate implementations.

The reasoning for keeping generics unconstrained on structs is identified in the [Rust API Guide](https://rust-lang.github.io/api-guidelines/future-proofing.html#c-struct-bounds). This change is part of a series of processes to relax bounds as the `geo` ecosystem is extended.

- `Coord` no longer has an impl constraint of `CoordNum` for its inner values
- `Rect<T>`, `LineString<T>`, `Line<T>`,`PointsIter<T>` generic constraints have been relaxed.
- Certain implementations have been relaxed for `T` across types to allow values to access identical functionality written by `geo`

Relates to  #1169 and #1157.
Relates to https://github.com/georust/geo-svg/issues/13
Thank you for your consideration!

### DIFF
--- a/geo-types/src/geometry/coord.rs
+++ b/geo-types/src/geometry/coord.rs
@@ -25,7 +25,7 @@ use approx::{AbsDiffEq, RelativeEq, UlpsEq};
 /// [vector space]: //en.wikipedia.org/wiki/Vector_space
 #[derive(Eq, PartialEq, Clone, Copy, Debug, Hash, Default)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub struct Coord<T: CoordNum = f64> {
+pub struct Coord<T = f64> {
     pub x: T,
     pub y: T,
 }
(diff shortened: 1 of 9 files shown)


michaelkirk:
I think I understand how this change makes the code more mathematically "pure", but are you actually running into a problem with the current implementation? Can you succinctly describe that problem or point to some code which fails to compile?

dsgallups:
> I think I understand how this change makes the code more mathematically "pure", but are you actually running into a problem with the current implementation? Can you succinctly describe that problem or point to some code which fails to compile?

Absolutely, will create an issue describing several examples of this. Thank you for considering my PR!

dsgallups:
Ran out of time and forgot about this

PR #1205, opened 2024-07-30
Relax restrictions on bool ops between polygon and multipolygon

### LINKED ISSUE #1203: Boolean ops should be available between differing geometry types
In keeping with JTS and GEOS, boolean ops should be available between all the "official" geometries. Currently, they're only available between two geometries of the same types, and this, coupled with the `union` op's output of `MultiPolygon`, is causing difficulties in extending the API, as `fold` and `reduce` operations are greatly complicated by this asymmetry.

When #1089 has merged, we should revisit the public API with a view to a) extending it and b) seeing whether it's possible to change the default output of `union`.

urschrei:
Whereas previously, boolean ops could only be performed between two geometries of the same type, it should now be possible to perform them between any combination of two Polygons and MultiPolygons

Closes #1203

---

### DIFF
--- a/geo/src/algorithm/bool_ops/tests.rs
+++ b/geo/src/algorithm/bool_ops/tests.rs
@@ -124,9 +124,9 @@ fn single_and_multi() {
     let wkt2 = "MULTIPOLYGON (((80 260, 90 260, 90 250, 80 250, 80 260)))";
     // From JTS union op
     let res = "MULTIPOLYGON (((110 310, 220 310, 220 210, 110 210, 110 310)), ((80 260, 90 260, 90 250, 80 250, 80 260)))";
-    let poly = Polygon::<f64>::try_from_wkt_str(&wkt1).unwrap();
-    let mpoly = MultiPolygon::<f64>::try_from_wkt_str(&wkt2).unwrap();
-    let respoly = MultiPolygon::<f64>::try_from_wkt_str(&res).unwrap();
+    let poly = Polygon::<f64>::try_from_wkt_str(wkt1).unwrap();
+    let mpoly = MultiPolygon::<f64>::try_from_wkt_str(wkt2).unwrap();
+    let respoly = MultiPolygon::<f64>::try_from_wkt_str(res).unwrap();
     let union = mpoly.union(&poly);
     let intersection_matrix = respoly.relate(&union);
     // coords will be arranged differently, but we only care about topology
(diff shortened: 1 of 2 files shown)


urschrei:
@rmanoka I'd love some feedback here to make sure I haven't accidentally tied myself into knots w/r/t/ types – the existing tests weren't modified and all pass, however.

The main reason we need this is in order to implement a unary union: a `t::f(g) -> multipoly` bool op isn't an obstacle, but the current restriction which says `t` and `g` must be the same type (`Polygon` or `MultiPolygon`) does cause an issue, because it requires an intermediate conversion of `Polygon` to `MultiPolygon` during the fold and reduction ops, and those conversions needlessly allocate because `MultiPolygon` is `Vec` under the hood.

WDYT?

urschrei:
Yeah I'm open to it – where would that leave the public boolops API?

geo/src/algorithm/bool_ops/tests.rs
```diff
@@ -116,6 +116,39 @@ fn test_complex_rects() -> Result<()> {
...
 }
+
+#[test]
+fn single_and_multi() {
+    let wkt1 = "POLYGON ((110 310, 220 310, 220 210, 110 210, 110 310))";
+    // multipolygon containing a single polygon
+    let wkt2 = "MULTIPOLYGON (((80 260, 90 260, 90 250, 80 250, 80 260)))";
+    // From JTS union op
+    let res = "MULTIPOLYGON (((110 310, 220 310, 220 210, 110 210, 110 310)), ((80 260, 90 260, 90 250, 80 250, 80 260)))";
+    let poly = Polygon::<f64>::try_from_wkt_str(wkt1).unwrap();
```
michaelkirk, line 127:
```suggestion
    let poly = geo::wkt!(POLYGON ((110 310, 220 310, 220 210, 110 210, 110 310)));
```

reminder about the geo::wkt! macro: https://docs.rs/geo/latest/geo/macro.wkt.html

michaelkirk: lol I'm sorry that I sabotaged you. 

The macro will try to build an integer geometry if you give it integers. Forget it!

urschrei: Oh noooooo

urschrei:
I don't think we need this anymore now that #1234 has merged. I _think_ it also enables the major cascaded union use case: https://gist.github.com/urschrei/cd80b4d2ec3c75f12fa541a5bdbf6489

michaelkirk:
I totally forgot about this one. Sorry — we should have merged it months ago.

Issue #1181, opened 2024-05-17
Unify line measurement traits


michaelkirk:
We have a lot of algorithms (bearing, destination, distance, etc.)  implemented across different metric spaces (haversine, euclidean, geodesic), and the current arrangement, arrived at organically, can be pretty confusing.

e.g. It took me a while to realize that we already have haversine corollaries to some euclidean methods. Same thing in #1208 

I think we can arrange things differently to make this more obvious and discoverable to users

**EDIT:** I've deleted my original proposal, and have a new one.

## New Proposal

Introduce a `struct` per metric space:

```
struct Haversine;
struct Rhumb;
struct Geodesic;
struct Euclidean;
```

Introduce a unified trait per algorithm, and implement that trait on each struct:
```
trait Bearing<T: CoordFloat> {
  fn bearing(p1: Point<T>, p2: Point<T>) -> T
}

impl Bearing for Haversine {
  fn bearing(p1: Point<T>, p2: Point<T>) -> T {
    todo!()
  }
}
impl Bearing for Rhumb {
  fn bearing(p1: Point<T>, p2: Point<T>) -> T {
    todo!()
  }
}
impl Bearing for Geodesic {
  fn bearing(p1: Point<T>, p2: Point<T>) -> T {
    todo!()
  }
}
impl Bearing for Euclidean {
  fn bearing(p1: Point<T>, p2: Point<T>) -> T {
    todo!()
  }
}
```

## Current Measurement Traits

Here our existing traits that seem at least somewhat related to the notion at hand. I don't intend to unify all of these, but there are some definite patterns that I think we can capture to make things more intuitive to people.

As a csv:  [measurements.csv](https://github.com/user-attachments/files/17081924/measurements.csv)

Rendered:
| Trait                  | area_signed               | area_unsigned             | bearing                | closest_point            | densify                  | destination              | distance                           | intermediate              | intermediate_fill          | length               | (➡Oddballs ➡)         | interpolate_point          | cross_track_distance                   | bearing_distance          | perimeter            | perimeter_and_area_signed       | perimeter_and_area_unsigned       |
|------------------------|---------------------------|---------------------------|------------------------|--------------------------|--------------------------|--------------------------|------------------------------------|---------------------------|----------------------------|----------------------|-----------------------|-----------------------------|-------------------------------------|---------------------------|----------------------|---------------------------------|-----------------------------------|
[...]

PR #1208, opened 2024-08-10
Add `DensifyGeodesic` trait

grevgeny:
---

This PR adds `DensifyGeodesic` trait that follows logic similar to `DensifyHaversine`. However, the trait is only implemented for `T = f64` as underlying Geodesic traits are implemented only for this type.

### DIFF
--- a/geo/src/algorithm/densify_geodesic.rs
+++ b/geo/src/algorithm/densify_geodesic.rs
@@ -255,4 +255,14 @@ mod tests {
         let dense = linestring.densify_geodesic(10.0);
         assert_eq!(0, dense.coords_count());
     }
+
+    #[test]
+    fn test_no_op_densify() {
+        let linestring: LineString = vec![[0.0, 0.0], [1.0, 0.0], [2.0, 0.0], [3.0, 0.0]].into();
+        let output: LineString = vec![[0.0, 0.0], [1.0, 0.0], [2.0, 0.0], [3.0, 0.0]].into();
+        // Use a very large max_meters to ensure no points are added.
+        let dense = linestring.densify_geodesic(1000000.0);
+        // Check that the densified linestring is identical to the original.
+        assert_relative_eq!(dense, output, epsilon = 1.0e-6);
+    }
 }
--- a/geo/src/algorithm/densify_geodesic.rs
+++ b/geo/src/algorithm/densify_geodesic.rs
@@ -1,6 +1,6 @@
 use num_traits::ToPrimitive;
 
-use crate::{CoordFloat, Line, Point};
+use crate::{CoordFloat, CoordsIter, Line, LineString, MultiLineString, MultiPolygon, Point, Polygon, Rect, Triangle};
 
 use crate::{GeodesicIntermediate, GeodesicLength};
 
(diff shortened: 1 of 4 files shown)


geo/src/algorithm/densify_geodesic.rs
```diff
@@ -0,0 +1,243 @@
...
+/// let output: LineString = vec![[10.0, 20.0], [65.879360, 37.722253], [125.0, 25.00]].into();
+/// // densify
+/// let dense = line.densify_geodesic(5703861.471800622);
+///
+/// assert_relative_eq!(dense, output, epsilon = 1.0e-6);
+///```
+pub trait DensifyGeodesic<F: CoordFloat> {
+    type Output;
+
+    fn densify_geodesic(&self, max_distance: F) -> Self::Output;
```
michaelkirk, line 34:
```suggestion
    fn densify_geodesic(&self, max_meters: f64) -> Self::Output {
```

I know it's in the docs, but it's nice to be redundant sometimes.

geo/src/algorithm/densify_geodesic.rs
```diff
@@ -0,0 +1,243 @@
...
+    fn densify_geodesic(&self, max_distance: f64) -> Self::Output {
+        self.to_polygon().densify_geodesic(max_distance)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::coord;
+
```
michaelkirk, line 167: Can you add a test where `max_meters` is larger than each of the segment lengths?   i.e. where the algorithm is a no-op.

geo/src/algorithm/mod.rs
```diff
@@ -84,6 +82,10 @@ pub use densify::Densify;
 pub mod densify_haversine;
 pub use densify_haversine::DensifyHaversine;
 
+/// Densify geometry components on a spheroid;
```
michaelkirk, line 85:
```suggestion
/// Densify geometry components on a geodesic.
```

nit: period, not semicolon.

And I don't know that it's better, but "geodesic"  is consistent with our other doc strings, so let's use it everywhere.

michaelkirk (changes requested):
I have some minor nits, but overall it looks good. Thank you!

grevgeny:
@michaelkirk thanks for your review! I have added proposed changes

michaelkirk:
I think you'll need to run cargo fmt, otherwise LGTM

I'll leave it open for a couple days before merging in case anyone else wants to weigh in.

michaelkirk:
Darn it. I'm sorry for not realizing this earlier. This seems related to https://github.com/georust/geo/issues/1181

michaelkirk:
Assuming #1216 gets merged, I've got a different approach that should supersede this, generically implementing Densify for all line measures (Haversine, Geodesic, Euclidean, etc.)

You can see a WIP here: https://github.com/georust/geo/compare/mkirk/unify-densify?expand=1

michaelkirk:
My alternative to this was merged in #1231.

With that you can call:
```rust
use geo::{Geodesic, Densify, Euclidean};
line_string.densify::<Geodesic>();
line_string.densify::< Euclidean >();
```

Let me know if that works for your purposes @grevgeny.

michaelkirk:
With #1298 (not yet released) this changed to:

```rust
use geo::{Geodesic, Densify, Euclidean};
Geodesic.densify(&line_string);
Euclidean.densify(&line_string);
```

PR #1214, opened 2024-09-15
docs: fixed typo iff -> if

btdevine:
---

### DIFF
--- a/geo/src/algorithm/winding_order.rs
+++ b/geo/src/algorithm/winding_order.rs
@@ -68,12 +68,12 @@ pub trait Winding {
     /// `None` otherwise.
     fn winding_order(&self) -> Option<WindingOrder>;
 
-    /// True iff this is wound clockwise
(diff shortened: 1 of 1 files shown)


geo/src/algorithm/winding_order.rs
```diff
@@ -68,12 +68,12 @@ pub trait Winding {
     /// `None` otherwise.
     fn winding_order(&self) -> Option<WindingOrder>;
 
-    /// True iff this is wound clockwise
+    /// True if this is wound clockwise
```
kylebarron, line 71: I assume `iff` means "if an only if", but it might be easier to follow for newcomers if that's spelled out.

kylebarron: I figure whoever wrote that had a math-y background and it was second nature to them, but my own vote would be to spell it out in public-facing documentation

PR #1217, opened 2024-09-26
Simplify line_segment_distance a little

lnicola:
---

Just happened to come across this, hopefully it's not going to be too contentious.

### DIFF
--- a/geo-types/src/private_utils.rs
+++ b/geo-types/src/private_utils.rs
@@ -72,14 +72,15 @@ where
     }
     let dx = end.x - start.x;
     let dy = end.y - start.y;
-    let r = ((point.x - start.x) * dx + (point.y - start.y) * dy) / (dx.powi(2) + dy.powi(2));
+    let d_squared = dx * dx + dy * dy;
+    let r = ((point.x - start.x) * dx + (point.y - start.y) * dy) / d_squared;
     if r <= T::zero() {
         return line_euclidean_length(Line::new(point, start));
     }
     if r >= T::one() {
         return line_euclidean_length(Line::new(point, end));
     }
-    let s = ((start.y - point.y) * dx - (start.x - point.x) * dy) / (dx * dx + dy * dy);
+    let s = ((start.y - point.y) * dx - (start.x - point.x) * dy) / d_squared;
     s.abs() * dx.hypot(dy)
 }
 


geo-types/src/private_utils.rs
```diff
@@ -72,14 +72,15 @@ where
...
+    let r = ((point.x - start.x) * dx + (point.y - start.y) * dy) / d;
     if r <= T::zero() {
         return line_euclidean_length(Line::new(point, start));
     }
     if r >= T::one() {
         return line_euclidean_length(Line::new(point, end));
     }
-    let s = ((start.y - point.y) * dx - (start.x - point.x) * dy) / (dx * dx + dy * dy);
+    let s = ((start.y - point.y) * dx - (start.x - point.x) * dy) / d;
     s.abs() * dx.hypot(dy)
```
michaelkirk, line 84:
```suggestion
    s.abs() * d.sqrt()
```

right? I have no idea if it's faster though.

geo-types/src/private_utils.rs
```diff
@@ -72,14 +72,15 @@ where
     }
     let dx = end.x - start.x;
     let dy = end.y - start.y;
-    let r = ((point.x - start.x) * dx + (point.y - start.y) * dy) / (dx.powi(2) + dy.powi(2));
+    let d = dx * dx + dy * dy;
```
michaelkirk, line 75: I guess this is `d_squared`?

michaelkirk (changes requested):
> Just happened to come across this, hopefully it's not going to be too contentious.

Apparently I can make anything contentious. 😆 

I think it seems like a good idea though.

geo-types/src/private_utils.rs
urschrei: I _think_ explicit var * var is faster (or thought to be faster, in a folkloric way) than var.powf(T::one() + T::one()), or certainly was for a long time, but who really knows the compiler's inner thoughts (godbolt does)

michaelkirk: To be clear, I'm just picking nits about the variable naming.

```
let d_squared = dx * dx + dy * dy;
```

I think d implies:
```
let d = (dx * dx + dy + dy).sqrt()
```

But maybe I'm reading in too deep.

urschrei: Ohhhh sorry

geo-types/src/private_utils.rs
lnicola: It might be faster, but it's probably less precise. Compare that to the [previous and current](https://github.com/bminor/glibc/commit/6c848d70383e1dbe932ef41723ac0abfdeec7ca8) glibc implementations.

geo-types/src/private_utils.rs
lnicola: Done. Regarding performance, at least in isolation, `x * x` and `x.powi(2)` produce the same code, this is for consistency reasons, since it can make people think it's intentional. I actually tracked the code down to https://github.com/georust/geo/pull/139 to see if there was any discussion of it.

geo-types/src/private_utils.rs
michaelkirk: Now that you say it, I'm not surprised, but wow, there are so many gotchas with this stuff!

michaelkirk (approved):

Issue #256, opened 2018-06-04
One Length trait to rule them all 💍


frewsxcv:
The following code is blocked on https://github.com/rust-lang/rust/issues/51344. Relevant discussion happened in https://github.com/georust/rust-geo/issues/47.

```rust
pub struct Line<T> {
    start: (T, T),
    end: (T, T),
}

////////////

pub trait LengthAlgo<T> {
    fn length(line: Line<T>) -> T;
}

////////////

pub enum Haversine {}

impl<T> LengthAlgo<T> for Haversine {
    fn length(line: Line<T>) -> T {
        unimplemented!()
    }
}

pub enum Vincenty {}

impl<T> LengthAlgo<T> for Vincenty {
    fn length(line: Line<T>) -> T {
        unimplemented!()
    }
}

////////////

pub trait Length<T, Algo: LengthAlgo<T> = Haversine> {
    fn length(self) -> T;
}

impl<T, Algo: LengthAlgo<T>> Length<T, Algo> for Line<T> {
    fn length(self) -> T {
        Algo::length(self)
    }
}

////////////

fn main() {
    let line: Line<i32> = Line {
        start: (4, 6),
        end: (2, 23),
    };
    
    // This doesn't compile – error message says `Algo` type needs to be specified
    // let haversine_length = line.length();
    
    // This compiles, despite us not specifying the type for `Algo`
    let haversine_length = Length::<i32>::length(line);
}
```

frewsxcv:
There are two topics that need to be discussed here:

- What should the trait look like?
- What should the default algorithm be?

urschrei:
I would say the default should be Euclidean, since `Geo` deals primarily with planar geometry.

rory:
I agree about euclidean distance by default.

The trait here takes ownership of a Line. Would it be better to take a reference to 2 points? To prevent creating and discarding a Line struct each time? (or am I misunderstanding something about rust? 🙂) 

On 4 June 2018 14:38:44 CEST, "Stephan Hügel" <notifications@github.com> wrote:
>I would say the default should be Euclidean, since `Geo` deals
>primarily with planar geometry.

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

frewsxcv:
I have opinions about euclidean as a default, but short on time right now to expand.

Regarding taking ownership, you're right, it should take `&self`. Unfortunately can't make much progress here until my Rust issue is addressed :-/

rory:
[...]

PR #1220, opened 2024-09-30
Proof of concept: Allow custom earth radii

frewsxcv:
---

### DIFF
--- a/geo/src/algorithm/rhumb/distance.rs
+++ b/geo/src/algorithm/rhumb/distance.rs
@@ -1,4 +1,4 @@
-use crate::{CoordFloat, Point, MEAN_EARTH_RADIUS};
+use crate::{earth_radius::{MeanEarthRadius, Iugg}, CoordFloat, Point};
 use num_traits::FromPrimitive;
 
 use super::RhumbCalculations;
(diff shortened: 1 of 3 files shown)


geo/src/algorithm/rhumb/distance.rs
```diff
@@ -37,16 +37,20 @@ pub trait RhumbDistance<T, Rhs = Self> {
     /// ```
     ///
     /// [rhumb line]: https://en.wikipedia.org/wiki/Rhumb_line
-    fn rhumb_distance(&self, rhs: &Rhs) -> T;
+    fn rhumb_distance(&self, rhs: &Rhs) -> T {
```
frewsxcv, line 40: Sadly rust doesn't allow us to do something like this:

```
fn rhumb_distance<Radius: MeanEarthRadius = Iugg>(&self, rhs: &Rhs) -> T {
```

geo/src/earth_radius.rs
```diff
@@ -0,0 +1,57 @@
...
+    const VALUE: f64 = 6378137.0;
+}
+
+impl MeanEarthRadius for Iugg {
+    const VALUE: f64 = 6371008.8;
+}
+
+impl MeanEarthRadius for Sphere {
+    const VALUE: f64 = 6371000.0;
+}
```
frewsxcv, line 57: I am not at all attached to these particular struct names or this particular set of options

michaelkirk:
With my work on #1216, I've been thinking about this too! 

I was imagining something kind of like what you've laid out here. Here's a POC with configurable haversine: b89c07e0

It's built on top of #1216 + the unified length trait I mentioned in #256  (I haven't opened a PR for #256 yet, I'm waiting on #1216 first).

One thing that my POC does not account for are methods on geometries - e.g. I haven't come up with a nice way to use a parameterized instance of a metric space (Haversine/Geodesic) to implement the unified `Length` trait. It seems likely doing so would require moving the implementation off of the geometry and onto an instance of `Haversine`.

PR #1222, opened 2024-10-08
deprecate legacy line measure traits 

michaelkirk:
---

Deprecating the old call sites of {Haversine,Rhumb,Geodesic}x{Distance,Bearing,Destination,Intermediate} in favor of the new methods introduced in #1216

Moved some implementation details from HaversineIntermediateFill to Haversine.

### DIFF
--- a/geo/src/algorithm/line_measures/metric_spaces/geodesic.rs
+++ b/geo/src/algorithm/line_measures/metric_spaces/geodesic.rs
@@ -1,5 +1,6 @@
 use super::super::{Bearing, Destination, Distance, InterpolatePoint};
 use crate::Point;
+use geographiclib_rs::{DirectGeodesic, InverseGeodesic};
 
 /// An ellipsoidal model of the earth, using methods given by [Karney (2013)].
 ///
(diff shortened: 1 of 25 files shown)


geo/src/algorithm/line_measures/metric_spaces/haversine.rs
```diff
@@ -174,13 +199,114 @@ impl<F: CoordFloat + FromPrimitive> InterpolatePoint<F> for Haversine {
...
+        if include_ends {
+            points.push(end);
+        }
+
+        points.into_iter()
+    }
+}
+
+#[allow(clippy::many_single_char_names)]
+struct HaversineIntermediateFillCalculation<T> {
```
michaelkirk, line 236: The logic from HaversineIntermediateFillCalculation  was cut/paste from

https://github.com/georust/geo/pull/1222/files#diff-634e7da640fe8307cf8b1ef08ca56b7bfd954216ed7a41bee8e5d06b2e660d1fL94

Though I did re-organize the methods, which were previously free methods, now instance methods. The actual implementation logic is identical.

geo/src/algorithm/line_measures/metric_spaces/geodesic.rs
```diff
@@ -36,7 +37,13 @@ impl Bearing<f64> for Geodesic {
     /// [geodesic line]: https://en.wikipedia.org/wiki/Geodesics_on_an_ellipsoid
     /// [Karney (2013)]:  https://arxiv.org/pdf/1109.4448.pdf
     fn bearing(origin: Point<f64>, destination: Point<f64>) -> f64 {
```
michaelkirk, line 39: All these implementations were copy/pasted (with some variable renaming) from the deprecated traits.

So, instead of delegating to the deprecated traits, the deprecated traits now delegate to the modern ones.

michaelkirk: > So, instead of delegating to the deprecated traits, the deprecated traits now delegate to the modern ones.

Oh! The 2 exceptions to this are that the deprecated `GeodesicBearing` and `HaversineBearing` do *not* delegate to the modern traits. They have kept their legacy implementation in tact to preserve the fact that their output range is [-180...180], whereas all the modern Bearing traits have unified around [0...360].

See #1210 for more details.

frewsxcv (approved):

PR #1224, opened 2024-10-10
Include is_equal_topo tests from jts test suite

michaelkirk:
---

No change in behavior here - I just realized there were some additional tests we can add to our test suite.


michaelkirk:
merging after making some trivial clippy fixes

PR #1231, opened 2024-10-18
unify Densify trait (adds support for Geodesic, Rhumb)

michaelkirk:
---

Supersedes #1208 
Part of #1181 

Densify is now generic, and implemented for anything that implements `Distance<Point, Point>`  and `Interpolate<Point>`. 

Specifically that means we continue to support Euclidean and Haversine densification, and now additionally support Geodesic and Rhumb.


michaelkirk:
Pushed up some missing docs - sorry 🙇

kylebarron (approved):

PR #1232, opened 2024-10-22
deprecate EuclideanDistance

michaelkirk:
---

Part of #1181

In #1222, I moved only the `Point x Point` distance calculations. EuclideanDistance is implemented for all other geometry types too. This PR moves the rest of those calculations to the new `Euclidean` line measure struct for consistency.

It's a lot of code movement, but no change in behavior (unless I messed up 😬).

### DIFF
--- a/geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
+++ b/geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
@@ -1,122 +1,39 @@
-use crate::{
-    Coord, GeoFloat, Geometry, GeometryCollection, Line, LineString, MultiLineString, MultiPoint,
-    MultiPolygon, Point, Polygon, Rect, Triangle,
-};
-use crate::{Distance, Euclidean};
-use num_traits::{float::FloatConst, Bounded, Signed};
-
+use super::{Distance, Euclidean};
+use crate::algorithm::Intersects;
+use crate::coordinate_position::{coord_pos_relative_to_ring, CoordPos};
+use crate::geometry::*;
+use crate::{CoordFloat, GeoFloat, GeoNum};
+use num_traits::{Bounded, Float};
 use rstar::primitives::CachedEnvelope;
 use rstar::RTree;
-use rstar::RTreeNum;
 
-#[deprecated(
-    since = "0.29.0",
-    note = "Please use the `Euclidean::distance` method from the `Distance` trait instead"
-)]
-/// Returns the distance between two geometries.
-pub trait EuclideanDistance<T, Rhs = Self> {
-    /// Returns the distance between two geometries
-    ///
-    /// If a `Point` is contained by a `Polygon`, the distance is `0.0`
-    ///
-    /// If a `Point` lies on a `Polygon`'s exterior or interior rings, the distance is `0.0`
-    ///
-    /// If a `Point` lies on a `LineString`, the distance is `0.0`
-    ///
-    /// The distance between a `Point` and an empty `LineString` is `0.0`
-    ///
-    /// # Examples
-    ///
-    /// `Point` to `Point`:
-    ///
-    /// ```
-    /// use approx::assert_relative_eq;
-    /// use geo::EuclideanDistance;
-    /// use geo::point;
-    ///
-    /// let p1 = point!(x: -72.1235, y: 42.3521);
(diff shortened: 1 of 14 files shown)


geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -0,0 +1,1087 @@
...
+        }
+    }
+}
+
+impl<F: GeoFloat> Distance<F, &LineString<F>, &Polygon<F>> for Euclidean {
+    fn distance(line_string: &LineString<F>, polygon: &Polygon<F>) -> F {
+        if line_string.intersects(polygon) {
+            F::zero()
+        } else if !polygon.interiors().is_empty()
+            // FIXME: Explodes on empty line_string
```
michaelkirk, line 190: re: `FIXME` - this panic on the next line isn't new behavior in this PR, but I noticed it while moving the code over.

Note that currently `nearest_neighbor_distance` returns `Float::max_value` if the input geometries are empty.

So, since we're already returning `Float::max_value` for _some_ degenerate input, I'd argue it'd be preferable to **always** return `Float::max_value` for **any** degenerate input in this algorithm, rather than sometimes `panic`'ing and sometimes returning `Float::max_value`.

A larger breaking change could be to make distance return an `Option`.

❤️ empty geometries ❤️

geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -0,0 +1,1087 @@
...
+    }
+}
+
+impl<F: GeoFloat> Distance<F, &Line<F>, &Polygon<F>> for Euclidean {
+    fn distance(line: &Line<F>, polygon: &Polygon<F>) -> F {
+        if line.intersects(polygon) {
+            return F::zero();
+        }
+
+        // REVIEW: This impl changed slightly.
```
michaelkirk, line 159: Heads up that this impl changed slightly, though the behavior should be the same. I couldn't resist the simplification. 😇 

All the other implementations were simply cut+paste+renamed.

The original impl is here: https://github.com/georust/geo/pull/1232/files#diff-23bef671023e471a3228d4ac73548395b9073026c10fb07f418ac3e81df8cdddL242
```rust
fn euclidean_distance(&self, other: &Polygon<T>) -> T {
  if self.intersects(other) {
    return T::zero();
  }
  // line-line distance between each exterior polygon segment and the line
  let exterior_min = other
    .exterior()
    .lines()
    .fold(<T as Bounded>::max_value(), |acc, point| {
      acc.min(self.euclidean_distance(&point))
    });
  // line-line distance between each interior ring segment and the line
  // if there are no rings this just evaluates to max_float
  let interior_min = other
    .interiors()
    .iter()
    .map(|ring| {
      ring.lines().fold(<T as Bounded>::max_value(), |acc, line| {
        acc.min(self.euclidean_distance(&line))
      })
    })
    .fold(<T as Bounded>::max_value(), |acc, ring_min| {
      acc.min(ring_min)
    });
  // return smaller of the two values
  exterior_min.min(interior_min)
}
```

geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -0,0 +1,1087 @@
...
+
+fn ring_contains_coord<T: GeoNum>(ring: &LineString<T>, c: Coord<T>) -> bool {
+    match coord_pos_relative_to_ring(c, ring) {
+        CoordPos::Inside => true,
+        CoordPos::OnBoundary | CoordPos::Outside => false,
+    }
+}
+
+#[cfg(test)]
+mod test {
```
michaelkirk, line 388: These are the tests copied from the legacy euclidean_distance trait, updated to use the new trait. I didn't review them very carefully, except to make sure that they still passed.

geo/src/utils.rs
```diff
@@ -93,9 +93,6 @@ pub fn partial_min<T: PartialOrd>(a: T, b: T) -> T {
     }
 }
 
-// Moved to their own module, but we re-export to avoid breaking the API.
-pub use crate::coordinate_position::{coord_pos_relative_to_ring, CoordPos};
```
michaelkirk, line 97: The `util` module is not publicly accessible, so there's no need to worry about breaking an API's **within** it.

frewsxcv (approved):

PR #1233, opened 2024-10-24
change the boolops trait to return result

TotalKrill:
This would enable us to start working on removing the crashes due to ".unwrap" and instead let a user handle them gracefully in the code

I intentionally did not solve any actual crashes as of now, more checking if this is a direction that is likely to be implemented

---

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -1,7 +1,8 @@
 # Changes
 
 ## Unreleased
-
+* Change BooleanOps trait to return an ErrorType instead of crashing when given bad input
+  * <https://github.com/georust/geo/pull/1233>
 * Implement getter methods on `AffineTransform` to access internal elements.
   * <https://github.com/georust/geo/pull/1159>
 * Fix issue in Debug impl for AffineTransform where yoff is shown instead of xoff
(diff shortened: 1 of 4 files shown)


michaelkirk:
We've had some long standing crashing issues with our boolean ops, but I'd prefer to sit on this for just a bit longer (I'm talking days, not weeks or months). 

I'm exploring integrating `i_overlay` as a replacement engine for our boolean operations and it's going quite well. 

My understanding is that i_overlay is intended to be infallible, so it'd be a shame to introduce a fallible API if we can avoid it.

For those not following on discord, the crash that elicited this PR has been resolved in i_overlay just now.
https://github.com/iShape-Rust/iOverlay/issues/10#issuecomment-2435872344

michaelkirk:
> I intentionally did not solve any actual crashes as of now, more checking if this is a direction that is likely to be implemented

In any case, there's no sense in returning a result until an actual crash is addressed, so I'm going to mark this as a draft for now.

lnicola:
I don't think it's worth changing the signature of that method just because of an implementation bug. Users can always `catch_unwind` as a workaround.

TotalKrill:
Looking through the code, there are far less unwraps in i_overlay code, compared to the code currently in geo. And I have not managed to break that code either.

I dont think that suggesting catch_unwind is a very user-friendly strategy, especially not in Rust, that actually has solutions for handling code that can fail, such as code that is liberally sprinkled with unwraps...

With that being said, im not sure if this way is actually needed if i_overlay solves all crashes people are experiencing. from what I can tell from other users on the bevy discord. It does seem to be the case.

michaelkirk:
See #1234

Issue #1060, opened 2023-08-27
Improve Documentation for HaversineIntermediate


JosiahParry:
Documentation for `HaversineIntermediate` trait can be improved. 

Notably the parts that can be improved are the function arguments `f` for `haversine_intermediate()` and `max_dist` and `include_ends` for `haversine_intermediate_fill()`. 

- I _think_ the `f` is short for fraction of the distance between self and other. 
- I think `haversine_intermediate_fill()` returns a vector of the points that are interpolated between self and other

If some clarity can be provided in this issue I'd be happy to make a PR 

https://github.com/georust/geo/blob/d13e70c6438d4b8a845a1403afb8db4ec146f251/geo/src/algorithm/haversine_intermediate.rs#L6

JayKickliter:
I was just about to open this same exact issue for `GeodesicIntermediate`.

JayKickliter:
I'd be happy to open a PR if someone more familiar with domain can confirm @JosiahParry's deductions.

michaelkirk:
I don't use either of those traits, but taking a quick look at the code, your interpretation seems correct to me @JosiahParry. 

Looking at the git archeology, it seems probable that the docs/names for the GeodesicIntermediate were copy/pasted from HaversineIntermediate.

If either of you want to open PRs with better docs, that'd be excellent.

frewsxcv:
Update here– `HaversineIntermediate` is now deprecated https://github.com/georust/geo/commit/5f6bc012c364ef35ad2c55a62ef6ca04cd3e8fa8

michaelkirk:
Thanks for making the connection @frewsxcv 

The new `InterpolatePoint` trait that replaces the deprecated `HaversineIntermediate` trait has some finishing touches in
https://github.com/georust/geo/pull/1235

I think the new methods are much clearer, but there's always room for improvement.

PR #1235, opened 2024-10-25
Add InterpolatePoint::point_at_distance_between for line_measures.

michaelkirk:
---

Part of #1181 

As well as the new method, also
FIXES https://github.com/georust/geo/pull/1060 - by having better docs.


michaelkirk:
Merged in #1239

PR #1242, opened 2024-10-30
[traits] rename CoordTrait::nth_unchecked -> nth_or_panic to match conventional semantics

michaelkirk:
---

`unchecked` implies it's an unsafe method, like `Vec::get_unsafe`. It's how we use the term elsewhere in this api, e.g. see this discussion https://github.com/georust/geojson/pull/245#discussion_r1821255299

I also cleaned up some stale docs.

I opted not to add `foo_or_panic` methods to the various collections in this PR, as it seemed like a slightly different issue that could be followed up separately.

### DIFF
--- a/geo-traits/src/coord.rs
+++ b/geo-traits/src/coord.rs
@@ -16,10 +16,10 @@ pub trait CoordTrait {
 
     /// Access the n'th (0-based) element of the CoordinateTuple.
     /// Returns `None` if `n >= DIMENSION`.
-    /// See also [`nth_unchecked()`](Self::nth_unchecked).
+    /// See also [`nth_or_panic()`](Self::nth_or_panic).
     fn nth(&self, n: usize) -> Option<Self::T> {
         if n < self.dim().size() {
-            Some(self.nth_unchecked(n))
+            Some(self.nth_or_panic(n))
         } else {
             None
         }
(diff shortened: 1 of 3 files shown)


kylebarron (approved):
Should there also be an _actually unsafe_ `nth_unchecked` that doesn't even do a bounds check? (I don't know; I don't usually use any unchecked variants)

michaelkirk:
I've restored `nth_unchecked`, marked it `unsafe` and by default it just calls `nth_or_panic`.

Please take another look.

geo-traits/src/coord.rs
```diff
@@ -16,10 +16,16 @@ pub trait CoordTrait {
 
     /// Access the n'th (0-based) element of the CoordinateTuple.
     /// Returns `None` if `n >= DIMENSION`.
-    /// See also [`nth_unchecked()`](Self::nth_unchecked).
+    ///
+    /// See also [`nth_or_panic()`](Self::nth_or_panic) and [`nth_unchecked()`](Self::nth_unchecked).
+    ///
+    /// # Panics
```
michaelkirk, line 22: Added a panic condition.

frewsxcv (approved):

kylebarron (approved):

PR #1243, opened 2024-10-31
Add `Haversine:distance` for Line vs. Point (same as CrossTrackDistance)

michaelkirk:
---

Part of #1181 

At first I also made CrossTrackDistance generic so that you could:
```diff
- let distance = point.cross_track_distance(line_start, line_end);
+ let distance = point.cross_track_distance::<Haversine>(line_start, line_end);
+ let distance = point.cross_track_distance::<Euclidean>(line_start, line_end);
```

But I decided to avoid the breaking change for now and rolled it back (since we just released a breaking change 2 hours ago 😅)

At this point, I consider CrossTrackDistance to be wholly redundant with Line x Point distance, but maybe it's worth keeping around just because it's a well know (?) algorithm name people will be searching for. I'm willing to be convinced otherwise though.

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-* BREAKING: Make `CrossTrackDistance` generic both to clarify when it is using Haversine, but also implement it for Euclidean.
+* Add `Haversine:distance` for Line vs. Point (same as CrossTrackDistance)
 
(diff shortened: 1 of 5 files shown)


michaelkirk:
I didn't realize that CrossTrackDistance applies to an (**infinite**) line passing through the given line points, not a line segment. This change is incorrect.

PR #1246, opened 2024-10-31
Cascaded / Unary union

urschrei:
---

<s>This is a draft since it currently has unknown perf characteristics, and there may be better ways of defining the trait implementers</s>


geo/src/algorithm/bool_ops/mod.rs
```diff
@@ -125,6 +126,54 @@ pub enum OpType {
...
+/// Returns the [BooleanOps::union] of all contained geometries
+///
+/// This is more efficient than a manual union, but may result in increased memory use due to the use of an R*-tree
+pub trait UnaryUnion {
+    type Scalar: BoolOpsNum;
+
+    /// Construct a tree of all the input geometries and progressively union them from the "bottom up"
+    ///
+    /// This is considerably more efficient than successively adding new [Polygon]s to an ever more complex [Polygon].
+    /// The output [MultiPolygon] will contain a single member.
```
michaelkirk, line 138: How can you guarantee there will be only one Polygon? What if the two polygons are disjoint?

lnicola:
Did you ask for some polygons? They're not WKT, but here's a bunch of them I just happened to have on hand: https://data.europa.eu/data/datasets/d18f6f95-1d96-4b86-a4ba-291a416e45bb?locale=en.

geo/src/algorithm/bool_ops/mod.rs
urschrei: …then there will be more than one Polygon, yes.

urschrei:
Some preliminary perf data:

a [73.8 mb WKT file](https://www.dropbox.com/scl/fi/yiohzokedom522hk7pfvw/parcels.wkt?rlkey=mmxzj6clmxqk803e5kpgdou4c&dl=0), containing 39490 polygons (extract from large-scale agricultural data, thanks @lnicola!) can be unioned into a `MultiPolygon` in around 11 seconds in `release` mode. The naïve approach:

```rust
let unioned = data
    .iter()
    .fold(MultiPolygon::<f64>::new(vec![]), |acc, poly| {
        acc.union(poly)
    });
```

Hadn't terminated after 18 minutes, so this is at least a 100x improvement (?).

Next I need to verify that algorithm produces valid output (no reason to think it doesn't)

cc @dabreegster please take this for a spin with more real-world data when you get a chance!

urschrei:
Since we can `union` `MultiPolygon`s, I'd like to make this available for them too, but I don't _think_ it's possible to generically express that, because you need concrete types for the `init`, `fold`, and `reduce` function definitions and they vary slightly between `Polygon` and `MultiPolygon`.

I'd also be happy to write a second impl for `MultiPolygon`, but rustc also complains about that for some reason, even though the bounds are different:


```rust
impl<T, Container> UnaryUnion for Container
where
    T: BoolOpsNum,
    Container: IntoIterator<Item = MultiPolygon<T>> + Clone,
    MultiPolygon<T>: RTreeObject,
{ ... }
```

dabreegster:
I'm seeing great results from INSPIRE parcels. https://www.dropbox.com/scl/fi/8w9ufilq2c7wq1vp4fei0/London_Borough_of_Southwark.zip?rlkey=ejcug1i5fnkf35d2xjwcogz9m&dl=0

Using `mapshaper v1.geojson -dissolve -explode -o v2.geojson format=geojson geojson-type=FeatureCollection`, it takes 1.3s to dissolve 40k features into 7.5k. Using the new API in this PR (via https://github.com/acteng/will-it-fit/tree/use_unary_union), it takes 5.1s to dissolve to 1.1k. This implementation handles inputs that already overlap **better** than mapshaper.

Mapshaper output, darker area is features overlapping originally that weren't dissolved:
image
Rust output handles these:
image

There's more, larger INSPIRE inputs. I'm on battery power right now, so not going to push it :)  But a larger run in Birmingham on 60k features ran in about 30s (I didn't time it exactly) and produced 17k outputs. I have even larger OS data I could try, but can't share the inputs.
https://github.com/acteng/will-it-fit/blob/use_unary_union/data_prep/inspire_one_area.sh

It's particularly impressive/useful that the base boolean ops work on both projected and un-projected inputs!

No feedback on this API; it seems fine to me. I have a more niche use case of limiting the maximum area of unioned polygons, so I'll probably adapt what's here to do that. I don't think baking in support for that level of customization makes sense.

michaelkirk:
[speaking of a multithreading flag](https://github.com/georust/geo/pull/1250#discussion_r1826064335) - this feature might be a good candidate for rayon. (probably not in this PR, but it'd be interesting to see)

kylebarron:
> this feature might be a good candidate for rayon

It might be good to have a separate discussion thread about rayon API design. In particular, I'd love it if any rayon integration could use a rayon thread pool provided by the caller, so that higher level apps that already use rayon don't have work split across multiple rayon pools.

michaelkirk:
Could you implement it in terms of  `impl BooleanOps<Scalar = Self::Scalar>` (both MultiPolygon and Polygon implement that)

Something like:

```diff
- Container: IntoIterator<Item = Polygon<T>> + Clone,
+ Container: IntoIterator<Item = impl BooleanOps<Scalar = T>> + Clone,
```

This should work for both `Iterator<Item=Polygon>` and `Iterator<Item=MultiPolygon>`, but if you wanted a single iterator, you'd need to do something like `Iterator<Item=Box<dyn BooleanOps>>`

urschrei:
> > Since we can union MultiPolygons, I'd like to make this available for them too,
> 
> Could you implement it in terms of `impl BooleanOps<Scalar = Self::Scalar>` (both MultiPolygon and Polygon implement that)
> 
> Something like:
> 
> ```diff
> - Container: IntoIterator<Item = Polygon<T>> + Clone,
> + Container: IntoIterator<Item = impl BooleanOps<Scalar = T>> + Clone,
> ```
> 
> This should work for both `Iterator<Item=Polygon>` and `Iterator<Item=MultiPolygon>`, but if you wanted a single iterator, you'd need to do something like `Iterator<Item=Box<dyn BooleanOps>>`

You can't use `impl` in trait bounds, so that won't work as written (and it has to be an `RTreeObject` but that can probably be tagged on.

I think we will actually need separate impls even if I can figure out the above because I'm struggling to imagine how we'd specialise the `fold` function definition inside `unary_union`:

it's currently

```rust
let fold = |mut accum: MultiPolygon<T>, poly: &Polygon<T>| -> MultiPolygon<T> {
            accum = accum.union(poly);
            accum
        };
```

but for MultiPolygon it would have to be 

```rust
let fold = |mut accum: MultiPolygon<T>, mpoly: &MultiPolygon<T>| -> MultiPolygon<T> {
            accum = accum.union(mpoly);
            accum
        };
```

urschrei:
I wonder if this would work if Polygon and MultiPolygon were enum members and BooleanOps was impl'd on the enum (followed by UnaryUnion).

urschrei:
No point in tying myself in knots playing type Tetris: `MultiPolygon` doesn't impl `RTreeObject` anyway (we should probably do that) so it'll have to be a manual op for now (it's not hugely complex: you `flat_map` the MultiPolygon containers).

> I have a more niche use case of limiting the maximum area of unioned polygons, so I'll probably adapt what's here to do that. I don't think baking in support for that level of customization makes sense.

I think this might be tricky to implement either on your side or in `geo` as the fold operation definitionally has a scalar output: there's no way to skip geometries or end early – those geometries then wouldn't appear in the output at all, which isn't what you want.

urschrei:
For follow-on work, a preliminary parallel version was proposed in https://github.com/georust/rstar/issues/80#issuecomment-988681923:


```rust
fn parallel_bottom_up_fold_reduce<T, S, I, F, R>(tree: &RTree<T>, init: I, fold: F, reduce: R) -> S
where
    T: RTreeObject,
    RTreeNode<T>: Send + Sync,
    S: Send,
    I: Fn() -> S + Send + Sync,
    F: Fn(S, &T) -> S + Send + Sync,
    R: Fn(S, S) -> S + Send + Sync,
{
    fn inner<T, S, I, F, R>(parent: &ParentNode<T>, init: &I, fold: &F, reduce: &R) -> S
    where
        T: RTreeObject,
        RTreeNode<T>: Send + Sync,
        S: Send,
        I: Fn() -> S + Send + Sync,
        F: Fn(S, &T) -> S + Send + Sync,
        R: Fn(S, S) -> S + Send + Sync,
    {
        parent
            .children()
            .into_par_iter()
            .fold(init, |accum, child| match child {
                RTreeNode::Leaf(value) => fold(accum, value),
                RTreeNode::Parent(parent) => {
                    let value = inner(parent, init, fold, reduce);

                    reduce(accum, value)
                }
            })
            .reduce(init, reduce)
    }

    inner(tree.root(), &init, &fold, &reduce)
}
```

geo/src/algorithm/bool_ops/mod.rs
```diff
@@ -125,6 +132,83 @@ pub enum OpType {
...
+    mut fold: F,
+    mut reduce: R,
+) -> S
+where
+    T: RTreeObject,
+    I: FnMut() -> S,
+    F: FnMut(S, &T) -> S,
+    R: FnMut(S, S) -> S,
+{
+    fn inner<T, S, I, F, R>(parent: &ParentNode<T>, init: &mut I, fold: &mut F, reduce: &mut R) -> S
```
adamreichold, line 190: In my experience, recursive algorithms like this one can benefit from grouping those parameters which are constant over the whole algorithm to reduce the overhead of the recursive calls. So something like

```rust
struct Ops<I, F, R> {
  init: I,
  fold: F,
  reduce: R,
}

fn inner<T, S, I, F, R>(ops: &mut Ops<I, F, R>, parent: &ParentNode<T>) -> S /* .. */

inner(&mut Ops { init, fold, reduce }, tree.root())
```

geo/src/algorithm/bool_ops/mod.rs
```diff
@@ -140,3 +224,27 @@ impl<T: BoolOpsNum> BooleanOps for MultiPolygon<T> {
...
+where
+    T: BoolOpsNum,
+    Container: IntoIterator<Item = Polygon<T>> + Clone,
+    Polygon<T>: RTreeObject,
+{
+    type Scalar = T;
+
+    fn unary_union(&self) -> MultiPolygon<Self::Scalar> {
+        // these three functions drive the union operation
+        let init = || MultiPolygon::<T>::new(vec![]);
```
adamreichold, line 239: This looks like `MultiPolygon` might benefit from a `Default` impl so this could be written as

```suggestion
        let init = MultiPolygon::<T>::default;
```

geo/src/algorithm/bool_ops/mod.rs
```diff
@@ -140,3 +224,27 @@ impl<T: BoolOpsNum> BooleanOps for MultiPolygon<T> {
...
+        let fold = |mut accum: MultiPolygon<T>, poly: &Polygon<T>| -> MultiPolygon<T> {
+            accum = accum.union(poly);
+            accum
+        };
+        let reduce = |accum1: MultiPolygon<T>, accum2: MultiPolygon<T>| -> MultiPolygon<T> {
+            accum1.union(&accum2)
+        };
+        let rtree = RTree::bulk_load(self.clone().into_iter().collect());
```
adamreichold, line 247: I wonder whether it might be more general to require cloneability only of the items instead of the whole container (which might unnecessarily clone auxiliary data structures), e.g.

 ```suggestion
        let rtree = RTree::bulk_load(self.iter().cloned().collect());
```

with the trait bounds adjusted accordingly.

**Comment by michaelkirk** [comment]
>  type Tetris

The type tetris isn't too bad. But concerningly the test doesn't pass for multipolygon. 

I'm not sure if this represents a bug in BooleanOps or in UnaryUnion (or something else): https://github.com/georust/geo/commit/39a2f62e88f0976109b4442cef1f77eec3442107

I'm OK without merging support for MultiPolygons at present, but wanted to highlight this behavior in case it's meaningful to you.

File: geo/src/algorithm/bool_ops/mod.rs (hunk 3 above)
Comment (michaelkirk) [reply geo/src/algorithm/bool_ops/mod.rs:239]: I think the same goes for all our multi-types, but probably better to address that outside of this PR.

**Comment by urschrei** [comment]
> > type Tetris
> 
> The type tetris isn't too bad. But concerningly the test doesn't pass for multipolygon.
> 
> I'm not sure if this represents a bug in BooleanOps or in UnaryUnion (or something else): [39a2f62](https://github.com/georust/geo/commit/39a2f62e88f0976109b4442cef1f77eec3442107)
> 
> I'm OK without merging support for MultiPolygons at present, but wanted to highlight this behavior in case it's meaningful to you.

Oh nice job, I spent a couple of hours getting nowhere with the types today. Oh that's interesting.

> > type Tetris
> 
> The type tetris isn't too bad. But concerningly the test doesn't pass for multipolygon.
> 
> I'm not sure if this represents a bug in BooleanOps or in UnaryUnion (or something else): [39a2f62](https://github.com/georust/geo/commit/39a2f62e88f0976109b4442cef1f77eec3442107)
> 
> I'm OK without merging support for MultiPolygons at present, but wanted to highlight this behavior in case it's meaningful to you.

Oh nice job, I burned a couple of hours on these today. The test result _should_ be the same AFAIK. What's the actual output geometry??

**Comment by adamreichold** [comment]
> Some preliminary perf data:

One thing I wonder about the performance is whether the [`CachedEnvelope`] combinator can help with the R tree construction due the relatively expensive envelope computation?

**Comment by urschrei** [comment]
> > > type Tetris
> > 
> > 
> > The type tetris isn't too bad. But concerningly the test doesn't pass for multipolygon.
> > I'm not sure if this represents a bug in BooleanOps or in UnaryUnion (or something else): [39a2f62](https://github.com/georust/geo/commit/39a2f62e88f0976109b4442cef1f77eec3442107)
> > I'm OK without merging support for MultiPolygons at present, but wanted to highlight this behavior in case it's meaningful to you.
> 
> Oh nice job, I burned a couple of hours on these today. The test result _should_ be the same AFAIK. What's the actual output geometry??

The correct output _should_ be:


```rust
Polygon { exterior:
    LineString([
        Coord { x: 200.38308698623, y: 288.3387931645567 },
        Coord { x: 204.07584924189865, y: 288.2701221168692 },
        Coord { x: 210.99999999939024, y: 291.99999999857494 },
        Coord { x: 209.99999999939024, y: 289.99999999857494 },
        Coord { x: 212.24082540659725, y: 285.47846008694717 },
        Coord { x: 205.630617245422, y: 287.73853614038774 },
        Coord { x: 204.00000000684082, y: 287.0000000060255 },
        Coord { x: 200.38308698623, y: 288.3387931645567 }]), interiors: [] }]
```

But what we get is:


```rust
[Polygon { exterior: 
    LineString([
        Coord { x: 200.38308698977124, y: 288.3387931571061 },
        Coord { x: 204.07584924543988, y: 288.2701221094186 },
        Coord { x: 205.63061724896323, y: 287.73853613293716 },
        Coord { x: 204.00000001038205, y: 287.0000000060255 },
        Coord { x: 200.38308698977124, y: 288.3387931571061 }]), interiors: [] },
Polygon { exterior: LineString([
    Coord { x: 204.0758492379893, y: 288.2701221168692 },
    Coord { x: 210.9999999954809, y: 291.99999999857494 },
    Coord { x: 209.9999999954809, y: 289.99999999857494 },
    Coord { x: 212.24082541013848, y: 285.47846008694717 },
    Coord { x: 212.24082539523732, y: 285.47846009439775 },
    Coord { x: 212.2408254026879, y: 285.47846008694717 },
    Coord { x: 205.63061724896323, y: 287.73853613293716 },
    Coord { x: 205.6306172564138, y: 287.73853614038774 },
    Coord { x: 204.07584926034104, y: 288.2701221094186 },
    Coord { x: 204.07584924543988, y: 288.2701221094186 },
    Coord { x: 204.0758492379893, y: 288.2701221168692 }]), interiors: [] }
];
```

That's not any kind of correct. Hm.

urschrei:
> > > > type Tetris
> > > 
> > > 
> > > The type tetris isn't too bad. But concerningly the test doesn't pass for multipolygon.
> > > I'm not sure if this represents a bug in BooleanOps or in UnaryUnion (or something else): [39a2f62](https://github.com/georust/geo/commit/39a2f62e88f0976109b4442cef1f77eec3442107)
> > > I'm OK without merging support for MultiPolygons at present, but wanted to highlight this behavior in case it's meaningful to you.
> > 
> > 
> > Oh nice job, I burned a couple of hours on these today. The test result _should_ be the same AFAIK. What's the actual output geometry??
> 
> The correct output _should_ be:
> 
> ```rust
> Polygon { exterior:
>     LineString([
>         Coord { x: 200.38308698623, y: 288.3387931645567 },
>         Coord { x: 204.07584924189865, y: 288.2701221168692 },
>         Coord { x: 210.99999999939024, y: 291.99999999857494 },
>         Coord { x: 209.99999999939024, y: 289.99999999857494 },
>         Coord { x: 212.24082540659725, y: 285.47846008694717 },
>         Coord { x: 205.630617245422, y: 287.73853614038774 },
>         Coord { x: 204.00000000684082, y: 287.0000000060255 },
>         Coord { x: 200.38308698623, y: 288.3387931645567 }]), interiors: [] }]
> ```
> 
> But what we get is:
> 
> ```rust
> [Polygon { exterior: 
>     LineString([
>         Coord { x: 200.38308698977124, y: 288.3387931571061 },
>         Coord { x: 204.07584924543988, y: 288.2701221094186 },
>         Coord { x: 205.63061724896323, y: 287.73853613293716 },
>         Coord { x: 204.00000001038205, y: 287.0000000060255 },
>         Coord { x: 200.38308698977124, y: 288.3387931571061 }]), interiors: [] },
> Polygon { exterior: LineString([
>     Coord { x: 204.0758492379893, y: 288.2701221168692 },
>     Coord { x: 210.9999999954809, y: 291.99999999857494 },
>     Coord { x: 209.9999999954809, y: 289.99999999857494 },
>     Coord { x: 212.24082541013848, y: 285.47846008694717 },
>     Coord { x: 212.24082539523732, y: 285.47846009439775 },
>     Coord { x: 212.2408254026879, y: 285.47846008694717 },
>     Coord { x: 205.63061724896323, y: 287.73853613293716 },
>     Coord { x: 205.6306172564138, y: 287.73853614038774 },
>     Coord { x: 204.07584926034104, y: 288.2701221094186 },
>     Coord { x: 204.07584924543988, y: 288.2701221094186 },
>     Coord { x: 204.0758492379893, y: 288.2701221168692 }]), interiors: [] }
> ];
> ```
> 
> That's not any kind of correct. Hm.

I think this is a bug in the `MultiPolygon` `union` algorithm (not clear if it stems from `i_overlay`):

Using the test polygons, you should get the same result if you do:

```rust
let empty = MultiPolygon::<f64>::new(vec![]);
// multi_poly_1 is [poly1, poly2]
let empty_union = empty.union(&multi_poly_1);

// and
let mp1 = MultiPolygon::new(vec![poly1.clone()]);
let mp2 = MultiPolygon::new(vec![poly2.clone()]);
let mpu = &mp1.union(&mp2);
```

…but you don't. That's a bug unless I'm missing a detail about how `MultiPolygon` `union` is supposed to work.

michaelkirk:
I'm not sure if it's relevant, but it might be interesting that `MultiPolygon(vec![poly1, poly2])` is not a valid multipolygon

Screenshot 2024-11-02 at 21 20 58

It'd be nice to be able to handle this case - but it might be a hint towards resolving it.

michaelkirk:
I was able to get the tests passing by make the multipolygon: 590056341930fb747cb847a5362e310742dd1f8e

Note the usage of `assert_relative_eq`, some of the output points are *slightly* shifted (1e-10), but I think that's expected with the grid snapping in i_overlay.

urschrei:
This will pass when #1254 merges. 

> I was able to get the tests passing by making the multipolygon valid: [5900563](https://github.com/georust/geo/commit/590056341930fb747cb847a5362e310742dd1f8e)
> 
> Note the usage of `assert_relative_eq`, some of the output points are _slightly_ shifted (1e-10), but I think that's expected with the grid snapping in i_overlay.

Urgh I should have checked validity before anything else. Sorry.

urschrei:
Tests will pass when #1254 merges.

urschrei:
> > Some preliminary perf data:
> 
> One thing I wonder about the performance is whether the [`CachedEnvelope`] combinator can help with the R tree construction due the relatively expensive envelope computation?

I implemented this in d82d2412710b48591777f5d8469903027d5b2355. Any perf change seems to be dwarfed by the test time, but I think it makes sense to keep it given the inherent characteristics of real-world geographic Polygons.

geo/src/algorithm/bool_ops/mod.rs
urschrei: Done in ba95b2c86fa43d88dad32c48167c12975d64cd36

geo/src/algorithm/bool_ops/mod.rs
urschrei: Unless I'm misunderstanding you, we would have to restrict the container types to `Vec` and slices if we wanted `iter. Even if the `Container` items impl `Clone`, if we stick with `IntoIter` we have to clone the entire value because it's behind a shared reference:

```
impl<T, Boppable, BoppableCollection> UnaryUnion for BoppableCollection
where
    T: BoolOpsNum,
    Boppable: BooleanOps<Scalar = T> + RTreeObject + Clone,
    BoppableCollection: IntoIterator<Item = Boppable>,
{
    ...
    let rtree = RTree::bulk_load(
        self.into_iter()
            .map(|p| CachedEnvelope::new(p.clone()))
            .collect(),
    );    
}
```

Won't work.

adamreichold: I think the trait should avoid imposing the reference and follow `IntoIterator` in being implemented on references instead, e.g. the following diff appears to work:

```diff
diff --git a/geo/src/algorithm/bool_ops/mod.rs b/geo/src/algorithm/bool_ops/mod.rs
index 814b00bb..bf3b898b 100644
--- a/geo/src/algorithm/bool_ops/mod.rs
+++ b/geo/src/algorithm/bool_ops/mod.rs
@@ -168,7 +168,7 @@ pub trait UnaryUnion {
     /// let combined = mp.unary_union();
     /// assert_eq!(&combined, merged);
     /// ```
-    fn unary_union(&self) -> MultiPolygon<Self::Scalar>;
+    fn unary_union(self) -> MultiPolygon<Self::Scalar>;
 }
 
 // This function carries out a full post-order traversal of the tree, building up MultiPolygons from inside to outside.
@@ -228,29 +228,42 @@ impl<T: BoolOpsNum> BooleanOps for MultiPolygon<T> {
     }
 }
 
-impl<T, Boppable, BoppableCollection> UnaryUnion for BoppableCollection
+struct RTreeObjectRef<'a, T>(&'a T);
+
+impl<'a, T> RTreeObject for RTreeObjectRef<'a, T>
+where
+    T: RTreeObject,
+{
+    type Envelope = T::Envelope;
+
+    fn envelope(&self) -> Self::Envelope {
+        self.0.envelope()
+    }
+}
+
+impl<'a, T, Boppable, BoppableCollection> UnaryUnion for &'a BoppableCollection
 where
     T: BoolOpsNum,
-    Boppable: BooleanOps<Scalar = T> + RTreeObject,
-    BoppableCollection: IntoIterator<Item = Boppable> + Clone,
+    Boppable: BooleanOps<Scalar = T> + RTreeObject + 'a,
+    Self: IntoIterator<Item = &'a Boppable>,
 {
     type Scalar = T;
 
-    fn unary_union(&self) -> MultiPolygon<Self::Scalar> {
+    fn unary_union(self) -> MultiPolygon<Self::Scalar> {
         // these three functions drive the union operation
         let init = || MultiPolygon::<T>::new(vec![]);
-        let fold =
-            |mut accum: MultiPolygon<T>, poly: &CachedEnvelope<Boppable>| -> MultiPolygon<T> {
-                accum = accum.union(std::ops::Deref::deref(poly));
-                accum
-            };
+        let fold = |mut accum: MultiPolygon<T>,
+                    poly: &CachedEnvelope<RTreeObjectRef<'a, Boppable>>|
+         -> MultiPolygon<T> {
+            accum = accum.union((&*poly).0);
+            accum
+        };
         let reduce = |accum1: MultiPolygon<T>, accum2: MultiPolygon<T>| -> MultiPolygon<T> {
             accum1.union(&accum2)
         };
         let rtree = RTree::bulk_load(
-            self.clone()
-                .into_iter()
-                .map(|p| CachedEnvelope::new(p))
+            self.into_iter()
+                .map(|p| CachedEnvelope::new(RTreeObjectRef(p)))
                 .collect(),
         );
         bottom_up_fold_reduce(&rtree, init, fold, reduce)
```

and it avoids any cloning at all by building an R tree of short-lived references to the original objects. (I think that `RTreeObjectRef` combinator probably belongs into `rstar` as a blanket impl.)

urschrei: Wow. Yep, that works. We should definitely add this to rstar. If I understand this correctly, it's restricted to situations in which the original collection and its items have the same (`'a`) lifetime as the tree? That's extremely useful.

adamreichold: > We should definitely add this to rstar. 

https://github.com/georust/rstar/pull/178

> If I understand this correctly, it's restricted to situations in which the original collection and its items have the same ('a) lifetime as the tree?

The references need to outlive the tree itself. I'd say that this is not really different from creating a `Vec<&'a T>`, we just have to wire through the trait bounds via the wrapper.

adamreichold:
> Any perf change seems to be dwarfed by the test time, 

Yeah, the R tree construction time is probably not the main contributor here as building it for single operation would hardly amortize otherwise. Considering the multi-second runtimes, the parallel version might really be interesting though, even keeping in mind that tree algorithms are not Rayon's forte due to the significant scheduling overhead.

urschrei:
> > Any perf change seems to be dwarfed by the test time,
> 
> Yeah, the R tree construction time is probably not the main contributor here as building it for single operation would hardly amortize otherwise. Considering the multi-second runtimes, the parallel version might really be interesting though, even keeping in mind that tree algorithms are not Rayon's forte due to the significant scheduling overhead.

I'd be happy to hold off merging until we have a working Rayon version. I don't know if there's a way to use trait bounds to allow passing in either `iter` or `into_par_iter` to `unary_union`: if there is, we can try using the stricter trait bounds from the sketched function above for both variants? That's not a show-stopper though.

urschrei:
```diff
@@ -178,9 +178,10 @@ pub trait UnaryUnion {
 fn bottom_up_fold_reduce<T, S, I, F, R>(tree: &RTree<T>, init: I, fold: F, reduce: R) -> S
 where
     T: RTreeObject,
-    I: FnMut() -> S,
-    F: FnMut(S, &T) -> S,
-    R: FnMut(S, S) -> S,
+    RTreeNode<T>: Send + Sync,
+    I: FnMut() -> S + Send + Sync,
+    F: FnMut(S, &T) -> S + Send + Sync,
+    R: FnMut(S, S) -> S + Send + Sync,
 {
     // recursive algorithms can benefit from grouping those parameters which are constant over
     // the whole algorithm to reduce the overhead of the recursive calls
@@ -246,7 +247,8 @@ fn envelope(&self) -> Self::Envelope {
 impl<'a, T, Boppable, BoppableCollection> UnaryUnion for &'a BoppableCollection
 where
     T: BoolOpsNum,
-    Boppable: BooleanOps<Scalar = T> + RTreeObject + 'a,
+    Boppable: BooleanOps<Scalar = T> + RTreeObject + 'a + Send + Sync,
+    <Boppable as RTreeObject>::Envelope: Send + Sync,
     Self: IntoIterator<Item = &'a Boppable>,
 {
     type Scalar = T;
```

These trait bounds work with the current function definition.

adamreichold:
>  I don't know if there's a way to use trait bounds to allow passing in either iter or into_par_iter to unary_union

I would suggest making it an explicit decision on the caller's part, for example using a wrapper type which signals that parallelism might be useful, e.g.

```diff
diff --git a/geo/Cargo.toml b/geo/Cargo.toml
index d48be60f..f78654d1 100644
--- a/geo/Cargo.toml
+++ b/geo/Cargo.toml
@@ -17,7 +17,7 @@ default = ["earcutr", "spade", "multithreading"]
 use-proj = ["proj"]
 proj-network = ["use-proj", "proj/network"]
 use-serde = ["serde", "geo-types/serde"]
-multithreading = ["i_overlay/allow_multithreading"]
+multithreading = ["i_overlay/allow_multithreading", "dep:rayon"]
 
 [dependencies]
 earcutr = { version = "0.4.2", optional = true }
@@ -32,6 +32,7 @@ robust = "1.1.0"
 rstar = "0.12.0"
 serde = { version = "1.0", optional = true, features = ["derive"] }
 i_overlay = { version = "1.7.2", default-features = false }
+rayon = { version = "1.10.0", optional = true }
 
 [dev-dependencies]
 approx = ">= 0.4.0, < 0.6.0"
diff --git a/geo/src/algorithm/bool_ops/mod.rs b/geo/src/algorithm/bool_ops/mod.rs
index 23f6a0b6..70cd7a29 100644
--- a/geo/src/algorithm/bool_ops/mod.rs
+++ b/geo/src/algorithm/bool_ops/mod.rs
@@ -4,6 +4,9 @@ mod tests;
 
 pub use i_overlay_integration::BoolOpsNum;
 
+#[cfg(feature = "multithreading")]
+use rayon::prelude::*;
+
 use crate::geometry::{LineString, MultiLineString, MultiPolygon, Polygon};
 use rstar::{
     primitives::CachedEnvelope, primitives::ObjectRef, ParentNode, RTree, RTreeNode, RTreeObject,
@@ -214,6 +217,49 @@ where
     inner(&mut ops, tree.root())
 }
 
+#[cfg(feature = "multithreading")]
+fn par_bottom_up_fold_reduce<T, S, I, F, R>(tree: &RTree<T>, init: I, fold: F, reduce: R) -> S
+where
+    T: RTreeObject,
+    RTreeNode<T>: Send + Sync,
+    S: Send,
+    I: Fn() -> S + Send + Sync,
+    F: Fn(S, &T) -> S + Send + Sync,
+    R: Fn(S, S) -> S + Send + Sync,
+{
+    // recursive algorithms can benefit from grouping those parameters which are constant over
+    // the whole algorithm to reduce the overhead of the recursive calls
+    struct Ops<I, F, R> {
+        init: I,
+        fold: F,
+        reduce: R,
+    }
+
+    fn inner<T, S, I, F, R>(ops: &Ops<I, F, R>, parent: &ParentNode<T>) -> S
+    where
+        T: RTreeObject,
+        RTreeNode<T>: Send + Sync,
+        S: Send,
+        I: Fn() -> S + Send + Sync,
+        F: Fn(S, &T) -> S + Send + Sync,
+        R: Fn(S, S) -> S + Send + Sync,
+    {
+        parent
+            .children()
+            .into_par_iter()
+            .fold(&ops.init, |accum, child| match child {
+                RTreeNode::Leaf(value) => (ops.fold)(accum, value),
+                RTreeNode::Parent(parent) => {
+                    let value = inner(ops, parent);
+                    (ops.reduce)(accum, value)
+                }
+            })
+            .reduce(&ops.init, &ops.reduce)
+    }
+    let ops = Ops { init, fold, reduce };
+    inner(&ops, tree.root())
+}
+
 impl<T: BoolOpsNum> BooleanOps for Polygon<T> {
     type Scalar = T;
 
@@ -234,7 +280,7 @@ impl<'a, T, Boppable, BoppableCollection> UnaryUnion for &'a BoppableCollection
 where
     T: BoolOpsNum,
     Boppable: BooleanOps<Scalar = T> + RTreeObject + 'a,
-    Self: IntoIterator<Item = &'a Boppable>,
+    &'a BoppableCollection: IntoIterator<Item = &'a Boppable>,
 {
     type Scalar = T;
 
@@ -255,6 +301,44 @@ where
                 .map(|p| CachedEnvelope::new(ObjectRef::new(p)))
                 .collect(),
         );
+
         bottom_up_fold_reduce(&rtree, init, fold, reduce)
     }
 }
+
+#[cfg(feature = "multithreading")]
+/// Wrapper type which signals to algorithms operating on `T` that utilizing parallelism might be viable.
+pub struct AllowMultithreading<T>(pub T);
+
+#[cfg(feature = "multithreading")]
+impl<'a, T, Boppable, BoppableCollection> UnaryUnion for AllowMultithreading<&'a BoppableCollection>
+where
+    T: BoolOpsNum + Send,
+    Boppable: BooleanOps<Scalar = T> + RTreeObject + 'a + Send + Sync,
+    <Boppable as RTreeObject>::Envelope: Send + Sync,
+    &'a BoppableCollection: IntoParallelIterator<Item = &'a Boppable>,
+{
+    type Scalar = T;
+
+    fn unary_union(self) -> MultiPolygon<Self::Scalar> {
+        // these three functions drive the union operation
+        let init = || MultiPolygon::<T>::new(vec![]);
+        let fold = |mut accum: MultiPolygon<T>,
+                    poly: &CachedEnvelope<ObjectRef<'a, Boppable>>|
+         -> MultiPolygon<T> {
+            accum = accum.union(&***poly);
+            accum
+        };
+        let reduce = |accum1: MultiPolygon<T>, accum2: MultiPolygon<T>| -> MultiPolygon<T> {
+            accum1.union(&accum2)
+        };
+        let rtree = RTree::bulk_load(
+            self.0
+                .into_par_iter()
+                .map(|p| CachedEnvelope::new(ObjectRef::new(p)))
+                .collect(),
+        );
+
+        par_bottom_up_fold_reduce(&rtree, init, fold, reduce)
+    }
+}
diff --git a/geo/src/algorithm/bool_ops/tests.rs b/geo/src/algorithm/bool_ops/tests.rs
index 7cd0fed8..3c51075c 100644
--- a/geo/src/algorithm/bool_ops/tests.rs
+++ b/geo/src/algorithm/bool_ops/tests.rs
@@ -1,4 +1,4 @@
-use super::{BooleanOps, UnaryUnion};
+use super::{AllowMultithreading, BooleanOps, UnaryUnion};
 use crate::{wkt, Convert, MultiPolygon, Polygon, Relate};
 
 #[test]
@@ -8,14 +8,14 @@ fn test_unary_union() {
     let poly3: Polygon = wkt!(POLYGON((211.0 292.0,202.07584923592933 288.2701221108328,212.24082541367974 285.47846008552216,210.0 290.0)));
 
     let polys = vec![poly1.clone(), poly2.clone(), poly3.clone()];
-    let poly_union = polys.unary_union();
+    let poly_union = AllowMultithreading(&polys).unary_union();
     assert_eq!(poly_union.0.len(), 1);
 
-    let multi_poly_12 = MultiPolygon::new(vec![poly1.clone(), poly2.clone()]);
+    /*let multi_poly_12 = MultiPolygon::new(vec![poly1.clone(), poly2.clone()]);
     let multi_poly_3 = MultiPolygon::new(vec![poly3]);
     let multi_polys = vec![multi_poly_12.clone(), multi_poly_3.clone()];
     let multi_poly_union = multi_polys.unary_union();
-    assert_eq!(multi_poly_union.0.len(), 1);
+    assert_eq!(multi_poly_union.0.len(), 1);*/
 }
 
 #[test]
```

This way, the caller can decide whether the scheduling overhead is sufficiently small for her expected inputs.

(I also opted to parallelize the initial envelope computations in this case. And I guess the definitions of init, fold and reduce could be pulled out to slightly reduce the code duplication. The `AllowMultithreading` wrapper might also live outside this module as it could become generally useful.)

urschrei:
The multithreading wrapper reduces the test in https://github.com/georust/geo/pull/1246#issuecomment-2450897799 time from 11.31 s -> ~ 3.42 s on my machine.

Very nice.

geo/src/algorithm/bool_ops/mod.rs
```diff
@@ -2,6 +2,11 @@ mod i_overlay_integration;
 #[cfg(test)]
 mod tests;
 
+#[cfg(feature = "multithreading")]
+use rayon::prelude::IntoParallelIterator;
+#[cfg(feature = "multithreading")]
+use rayon::prelude::*;
```
adamreichold, line 8: These look redundant? If you want to avoid the prelude, importing `IntoParallelIterator` and `ParallelIterator` might suffice.

urschrei: Oops yep

geo/src/algorithm/bool_ops/mod.rs
```diff
@@ -125,6 +137,128 @@ pub enum OpType {
...
+{
+    // recursive algorithms can benefit from grouping those parameters which are constant over
+    // the whole algorithm to reduce the overhead of the recursive calls
+    struct Ops<I, F, R> {
+        init: I,
+        fold: F,
+        reduce: R,
+    }
+
+    fn inner<T, S, I, F, R>(ops: &mut Ops<I, F, R>, parent: &ParentNode<T>) -> S
```
adamreichold, line 198: Sorry for going into code golfing territory here, but I noticed that this separation into outer and inner function here and in the parallel variant is not really necessary when used as implementation detail (in contrast to public API). Meaning the inner function can just become the outer function and called directly with an `Ops` argument and the root node.

urschrei:
(I've snuck in some minor edits to `geo` top-level docs too. Sorry, sorry)

michaelkirk (commented):
I get this when building:

```  --> geo/src/algorithm/bool_ops/mod.rs:12:33
   |
12 |     primitives::CachedEnvelope, primitives::ObjectRef, ParentNode, RTree, RTreeNode, RTreeObject,
   |                                 ^^^^^^^^^^^^^^^^^^^^^ no `ObjectRef` in `primitives`
```

Do you need to bump the minimum version of rstar?

urschrei:
Sorry, done.

urschrei:
The difference in union behaviour (it's not necessarily erroneous) between single-threaded and multi-threaded ops needs to be explained before we can merge this, IMO. I'll dig into it ASAP because I'd like this to land.

urschrei:
I didn't get a conclusive answer about the small mismatch between multi-threaded and single-threaded geometry unions, so let's press on. Please review!

michaelkirk:
Can you clarify how to reproduce the difference between serial vs. multithreaded?

Is it like this:
```rust
let wkt_str = std::fs::read_to_string("parcels.wkt").unwrap();
let multi_polygon: MultiPolygon = MultiPolygon::try_from_wkt_str(&wkt_str).unwrap();
assert_eq!(multi_polygon.0.len(), 39490);

let serial_union= {
  let start = Instant::now();
  let union = multi_polygon.unary_union();
  let duration = start.elapsed();
  println!("Time elapsed: {:.2?}", duration);
  union
};
let parallel_union = {
  let start = Instant::now();
  let union = AllowMultithreading(&multi_polygon).unary_union();
  let duration = start.elapsed();
  println!("Time elapsed: {:.2?}", duration);
  union
};

if serial_union == parallel_union {
  println!(">> output is equal!");
} else {
  println!(">> output not equal!");
}
```

urschrei:
> Can you clarify how to reproduce the difference between serial vs. multithreaded?
> 
> Is it like this:
> 
> ```rust
> let wkt_str = std::fs::read_to_string("parcels.wkt").unwrap();
> let multi_polygon: MultiPolygon = MultiPolygon::try_from_wkt_str(&wkt_str).unwrap();
> assert_eq!(multi_polygon.0.len(), 39490);
> 
> let serial_union= {
>   let start = Instant::now();
>   let union = multi_polygon.unary_union();
>   let duration = start.elapsed();
>   println!("Time elapsed: {:.2?}", duration);
>   union
> };
> let parallel_union = {
>   let start = Instant::now();
>   let union = AllowMultithreading(&multi_polygon).unary_union();
>   let duration = start.elapsed();
>   println!("Time elapsed: {:.2?}", duration);
>   union
> };
> 
> if serial_union == parallel_union {
>   println!(">> output is equal!");
> } else {
>   println!(">> output not equal!");
> }
> ```

Correct!

michaelkirk:
Here's my go at implementing the "simplification" based approach recommended by Nail over in https://github.com/iShape-Rust/iOverlay/issues/16#issuecomment-2542015083

https://github.com/georust/geo/compare/mkirk/unary_union?expand=1


It is quite a bit faster!

Using our existing nl_plots test fixture as input, projected from lng/lat to a local euclidean projection.

    Time elapsed (naive): 28.34ms
    Time elapsed (serial recursive): 4.72ms
    Time elapsed (multithreaded recursive): 2.93ms
    Time elapsed (simplification): 165.13µs

I also ran a test with @urschrei's 72MB parcels.wkt and got similar results (though linearly larger, which makes sense given the much larger dataset):

    Time elapsed (naive): I gave up after waiting 10 minutes.
    Time elapsed (serial recursive): 6.15s
    Time elapsed (multithreaded recursive): 1.62s
    Time elapsed (simplification): 333.45ms

I've tried a few inputs, and haven't found one where the simplification approach is slower.

urschrei:
Kill your darlings 😭; I can't argue with a 5x speedup that has the same quality output. I think most of the doc changes and additions can be copied over to a new PR without much change @michaelkirk, and then we can close this in favour of that and merge.

michaelkirk:
Ok @urschrei - I've pushed up a commit that uses the new simplification based approach. 

I've also reverted the dependency changes, since those are no longer relevant to the new approach.

Mostly as a matter of personal taste, I exposed a free function `fn unary_union` rather than a `UnaryUnion` trait with a `unary_union` method. It seemed more discoverable, less ceremonial, but I could swap back if you prefer.

michaelkirk:
(note CI is failing due to #1283)

geo/src/algorithm/bool_ops/mod.rs
```diff
@@ -109,6 +124,75 @@ pub enum OpType {
     Xor,
 }
 
+/// Efficient [union](BooleanOps::union) of many adjacent / overlapping geometries
+///
+/// This is typically much faster than `union`ing a bunch of geometries together one at a time.
+///
+/// Note: Geometries can be wound in either direction, but the winding order must be consistent,
+/// and the polygon's interiors must be wound opposite to its exterior.
```
urschrei, line 132:
```suggestion
/// and each polygon's interior rings must be wound opposite to its exterior.
```

michaelkirk (approved):

PR #1250, opened 2024-11-01
Allow configuring of the `i_overlay` dep, Rayon transitive dep. Document more features.

frewsxcv:
---

### DIFF
--- a/geo/Cargo.toml
+++ b/geo/Cargo.toml
@@ -13,7 +13,7 @@ rust-version = "1.75"
 categories = ["science::geo"]
 
 [features]
-default = ["earcutr", "spade", "i_overlay", "multithreading"]
+default = ["earcutr", "spade", "multithreading"]
 use-proj = ["proj"]
 proj-network = ["use-proj", "proj/network"]
 use-serde = ["serde", "geo-types/serde"]
@@ -31,7 +31,7 @@ proj = { version = "0.27.0", optional = true }
 robust = "1.1.0"
 rstar = "0.12.0"
 serde = { version = "1.0", optional = true, features = ["derive"] }
-i_overlay = { version = "1.7.2", optional = true }
+i_overlay = "1.7.2"
 
 [dev-dependencies]
 approx = ">= 0.4.0, < 0.6.0"
--- a/geo/Cargo.toml
+++ b/geo/Cargo.toml
@@ -13,10 +13,11 @@ rust-version = "1.75"
 categories = ["science::geo"]
 
 [features]
-default = ["earcutr", "spade"]
+default = ["earcutr", "spade", "i_overlay", "multithreading"]
 use-proj = ["proj"]
 proj-network = ["use-proj", "proj/network"]
 use-serde = ["serde", "geo-types/serde"]
+multithreading = ["i_overlay/allow_multithreading"]
 
 [dependencies]
 earcutr = { version = "0.4.2", optional = true }
@@ -30,7 +31,7 @@ proj = { version = "0.27.0", optional = true }
 robust = "1.1.0"
 rstar = "0.12.0"
 serde = { version = "1.0", optional = true, features = ["derive"] }
-i_overlay = "1.7.2"
+i_overlay = { version = "1.7.2", optional = true }
 
 [dev-dependencies]
 approx = ">= 0.4.0, < 0.6.0"
--- a/geo/src/lib.rs
+++ b/geo/src/lib.rs
@@ -67,7 +67,7 @@
 //!
 //! ## Boolean Operations
 //!
-//! - **[`BooleanOps`]**: combine or split (Multi)Polygons using intersecton, union, xor, or difference operations. Requires the `"i_overlay"` feature, which is enabled by default.
+//! - **[`BooleanOps`]**: combine or split (Multi)Polygons using intersecton, union, xor, or difference operations.
 //!
 //! ## Outlier Detection
 //!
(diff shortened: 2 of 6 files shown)


urschrei (approved):

geo/Cargo.toml
```diff
@@ -30,7 +31,7 @@ proj = { version = "0.27.0", optional = true }
 robust = "1.1.0"
 rstar = "0.12.0"
 serde = { version = "1.0", optional = true, features = ["derive"] }
-i_overlay = "1.7.2"
+i_overlay = { version = "1.7.2", optional = true }
```
kylebarron, line 34: IMO boolean operations are such a core feature that I wouldn't oppose `i_overlay` itself being required. But I think the multithreading should be optional.

Also, don't you want `default-features = false` here, so that the `i_overlay/allow_multithreading` feature here is only applied when `multithreading` is on?

frewsxcv: Ah yes great catch, I meant to add `default-features = false` here

frewsxcv: https://github.com/georust/geo/pull/1250/commits/6e2303904d3b3470e683f194a8e21cbe3786fa67

frewsxcv: >IMO boolean operations are such a core feature that I wouldn't oppose i_overlay itself being required. But I think the multithreading should be optional.

I agree boolean operations are important, but the success of `geo` up until the operations were added to the crate indicates many people don't need that functionality and the dependency tree that comes with it. So they should have the ability to opt-out. Adding the extra `cfg` check here is low maintenance cost for us

geo/src/lib.rs
```diff
@@ -177,9 +177,25 @@
 //!
 //! The following optional [Cargo features] are available:
 //!
-//! - `proj-network`: Enables [network grid] support for the [`proj` crate]. After enabling this feature, [further configuration][proj crate file download] is required to use the network grid
-//! - `use-proj`: Enables coordinate conversion and transformation of `Point` geometries using the [`proj` crate]
-//! - `use-serde`: Allows geometry types to be serialized and deserialized with [Serde]
+//! - `i_overlay`:
+//!     - Enables the `i_overlay` crate, which provides boolean operations on geometries.
```
michaelkirk, line 181: Can we call this feature `bool_ops`? I think it will be more meaningful to our users.

kylebarron: Assuming we won't have a separate boolean operations implementation in the future that doesn't depend on `i_overlay`

geo/Cargo.toml
michaelkirk: Just to be clear, we've had *non optional* boolean operations in geo for years now. So if anyone is building geo with `no-default` features, this will be a breaking change for them. 

I personally have a slight preference for "make rayon optional but bool_ops mandatory" - because of the above breakage and adding configuration and documentation complexity for a feature on average outweighs the benefits of letting people fine tune in this case in my estimation.

But it's only a slight prefrence.

Question: if we break the no-default-features build, do we need to make another bump to `0.X.0`?

geo/src/lib.rs
urschrei: Given how long it's taken to get to this point, that seems unlikely. Not impossible, but a small risk IMO.

geo/Cargo.toml
urschrei: > if anyone is building geo with no-default features, this will be a breaking change for them.

That hadn't occurred to me. I'm now also in the "slight preference for mandatory" camp.

geo/Cargo.toml
```diff
@@ -13,10 +13,11 @@ rust-version = "1.75"
 categories = ["science::geo"]
 
 [features]
-default = ["earcutr", "spade"]
+default = ["earcutr", "spade", "i_overlay", "multithreading"]
 use-proj = ["proj"]
 proj-network = ["use-proj", "proj/network"]
 use-serde = ["serde", "geo-types/serde"]
+multithreading = ["i_overlay/allow_multithreading"]
```
michaelkirk, line 20: This `multithreading` feature also enables `i_overlay` - is that intended? 

If so, then I think we should call it something like `i_overlay_multithreading` (or per my other comment, `bool_ops_multithreading`)

Though I could imagine a use case for a `multithreading` flag in geo more generally. In which case this name is good as is! But we should change the definition to:

`multithreading = ["i_overlay?/allow_multithreading"]`

Which enables `allow_multithreading` on `i_overlay` if we're using `i_overlay`, but otherwise don't enable `i_overlay` just because someone tried to build geo with `multithreading`.

michaelkirk: My own preference is: 
`multithreading = ["i_overlay?/allow_multithreading"]`
(or `multithreading = ["bool_ops?/allow_multithreading"]`)

geo/src/lib.rs
michaelkirk: In case my original concern wasn't clear: I think a user perusing the feature flags will more quickly understand what a `bool_ops` feature does, vs. an `i_overlay` feature. 

The fact that we're using `i_overlay` for bool_ops is an implementation detail that the user probably doesn't care about.

frewsxcv: Cool I'll change it to `bool_ops`

geo/Cargo.toml
frewsxcv: I'll change it! I didn't know about this syntax

geo/src/lib.rs
frewsxcv: Nevermind, I'm just going to remove the feature

geo/Cargo.toml
frewsxcv: I still have a preference for making the bool ops stuff optional but it sounds like there's agreement about making it required so I'll just change it

michaelkirk (approved):

michaelkirk:
Can you push up a changelog entry @frewsxcv?

frewsxcv:
Yep will do that now

PR #1258, opened 2024-11-05
Remove associated type `T` on `geo-traits` traits.

frewsxcv:
---

Do we need to specify `T` for each of the traits?


kylebarron:
I'm on my phone and can't look at this closely, but I find it very useful to have a compile time generic on the `T`. I can't quite picture how this would work without the generics.

PR #1259, opened 2024-11-05
Temporarily pin `arbitrary` to work around upstream bug.

frewsxcv:
---


frewsxcv:
https://github.com/rust-fuzz/arbitrary/issues/203

urschrei:
Even with this change, we're suddenly seeing a test failure on concave hull? https://github.com/georust/geo/actions/runs/11686563527/job/32542774219

PR #1254, opened 2024-11-03
Add rstar MultiPolygon compat

urschrei:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/main/CODE_OF_CONDUCT.md).
- [x] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

This should merge before #1246

PR #1261, opened 2024-11-05
There is a bug in rstar 0.12.1, so pin for now.

### LINKED ISSUE #1260: CI failing after `cargo update`
https://github.com/georust/geo/actions/runs/11689401597/job/32552055508

Failure is in concave hull.

michaelkirk:
---

FIXES #1260 

see: https://github.com/georust/rstar/pull/181

We don't have to do this - we could wait a bit if we think an rstar fix is forthcoming.

### DIFF
--- a/geo-types/CHANGES.md
+++ b/geo-types/CHANGES.md
@@ -2,7 +2,8 @@
 
 ## UNRELEASED
 
-* Pin to rstar 0.12.0, to avoid a bug in 0.12.1
+* Pin to rstar 0.12.2, to avoid a bug in 0.12.1
+  * <https://github.com/georust/geo/pull/1261>
 
 ## 0.7.13
 
--- a/geo-types/CHANGES.md
+++ b/geo-types/CHANGES.md
@@ -1,5 +1,9 @@
 # Changes
 
+## UNRELEASED
+
+* Pin to rstar 0.12.0, to avoid a bug in 0.12.1
+
 ## 0.7.13
 
 * POSSIBLY BREAKING: Minimum supported version of Rust (MSRV) is now 1.70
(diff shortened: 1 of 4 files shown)


urschrei:
Let's hold off until Adam has a look at https://github.com/georust/rstar/pull/181 – #1254 is open to enable #1246, but #1246 depends on a new API in rstar 0.12.1 so would be blocked by this anyway.

michaelkirk:
I'm ok to wait for a little bit, but my primary concern isn't the new feature in #1246, but rather on `geo` users getting incorrect results in the meanwhile.

michaelkirk:
It looks like resolution is imminent with https://github.com/georust/rstar/pull/184

Anyone opposed to my following up with patch releases to geo-types and geo to require `rstar = "0.12.2"` ?

urschrei:
Do we need to do that if we yank rstar 0.12.1?

urschrei:
0.12.1 is yanked

michaelkirk:
I'm not sure!

What happens if you run `cargo update -p geo` in your project? Is it possible to update geo while keeping the yanked version of `rstar` around?

michaelkirk:
Answering my own question: `cargo update -p geo` will *not* proactively update it's rstar dependency, even if the current version has been yanked.

Unless we make a new release, the user would need to either run (unqualified) `cargo update` or know to run `cargo update -p rstar`.

urschrei:
Annoying. Thanks for figuring that out!

urschrei:
Where did we land on this after yesterday's discussion? My understanding is that there was a 24 hour (ish) period during which, if you updated your `geo` dependency, or started a new project and included it, you'd end up with a perf-regressed and buggy `geo`, which WOULD be fixed with a full `cargo update` but WOULDN'T be fixed with a `cargo update -p geo`.

lnicola:
I think Cargo givea a warning if you have a yanked crate in your lockfile.

Issue #1251, opened 2024-11-01
Clarify and decide what the unknown dimension does in `geo-traits`


frewsxcv:
There was some discussion in Discord, but I also want to track this issue in GitHub. Should this block a `geo-traits` 0.2.0 release?

kylebarron:
Perhaps we should make a little progress here before 0.2.0

frewsxcv:
Relevant discussion: https://github.com/georust/geojson/pull/245#issuecomment-2466771255

frewsxcv:
For the time being, that seems alright to me. In which case, that wouldn't require any changes to the current dimensions implementation, right?

kylebarron:
Correct I believe

PR #1255, opened 2024-11-03
Converter functions for geo-traits to geo-types

kylebarron:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/main/CODE_OF_CONDUCT.md).
- [ ] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

Closes #1253

PR #1263, opened 2024-11-05
prepare for geo-traits 0.2.0 release

michaelkirk:
---

I think it makes sense to release these breaking changes before proceeding with integrations like
https://github.com/georust/wkt/pull/124

### DIFF
--- a/geo-traits/CHANGES.md
+++ b/geo-traits/CHANGES.md
@@ -1,6 +1,6 @@
 # Changes
 
-## Unreleased
+## 0.2.0 - 2024.11.06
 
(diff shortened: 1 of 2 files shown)


kylebarron:
It would be nice to include https://github.com/georust/geo/pull/1255.

frewsxcv:
@kylebarron What do you think, should we try to resolve https://github.com/georust/geo/issues/1251 before doing a 0.2.0 release here?

kylebarron:
I don't know. I don't _love_ the idea of making a new variant or of making the dimension return option. But it seems unavoidable to support geojson. I'd love @michaelkirk 's feedback when he's back.

kylebarron (approved):

PR #1265, opened 2024-11-06
Add multithreading support to Multi* geometries

### LINKED ISSUE #1257: Multi-thread support for geo-types using Rayon
#1246 proposes a multi-threaded variant of the unary union algorithm. Out of the box, this works on containers that impl `IntoParallelIterator<Item = &'a Boppable>`. `MultiPolygon` doesn't, because `geo-types` has very few dependencies, let alone Rayon. However, if we had optional Rayon support (and enabled it when it was enabled for `geo`) we could call it like this:

```rust
let poly_union = AllowMultithreading(&parcel_multipolygon).unary_union()
```

Instead of like this:

```rust
let poly_union = AllowMultithreading(&parcel_multipolygon.0).unary_union()
```

Which isn't very nice and forces us to use the exposed-implementation-detail `.0` syntax.

urschrei:
---

I think I've set up the Cargo features correctly – the `multithreading` feature is disabled by default in `geo-types`, but enabled by `geo` when `geo`'s default features are active. I just want to make sure the other `geo-types` default features aren't impacted.

Closes #1257.

This PR should merge before #1246.

### DIFF
--- a/geo-types/src/geometry/multi_line_string.rs
+++ b/geo-types/src/geometry/multi_line_string.rs
@@ -5,6 +5,8 @@ use alloc::vec::Vec;
 #[cfg(any(feature = "approx", test))]
 use approx::{AbsDiffEq, RelativeEq};
 use core::iter::FromIterator;
+#[cfg(feature = "multithreading")]
+use rayon::prelude::*;
 
 /// A collection of
 /// [`LineString`s](line_string/struct.LineString.html). Can
@@ -115,6 +117,36 @@ impl<T: CoordNum> MultiLineString<T> {
 
     pub fn iter_mut(&mut self) -> impl Iterator<Item = &mut LineString<T>> {
         self.0.iter_mut()
+    }
+}
+
+#[cfg(feature = "multithreading")]
+impl<T: CoordNum + Send> IntoParallelIterator for MultiLineString<T> {
+    type Item = LineString<T>;
+    type Iter = rayon::vec::IntoIter<LineString<T>>;
+
+    fn into_par_iter(self) -> Self::Iter {
+        self.0.into_par_iter()
+    }
+}
+
+#[cfg(feature = "multithreading")]
+impl<'a, T: CoordNum + Sync> IntoParallelIterator for &'a MultiLineString<T> {
+    type Item = &'a LineString<T>;
+    type Iter = rayon::slice::Iter<'a, LineString<T>>;
+
+    fn into_par_iter(self) -> Self::Iter {
+        self.0.par_iter()
+    }
+}
+
+#[cfg(feature = "multithreading")]
+impl<'a, T: CoordNum + Send + Sync> IntoParallelIterator for &'a mut MultiLineString<T> {
+    type Item = &'a mut LineString<T>;
+    type Iter = rayon::slice::IterMut<'a, LineString<T>>;
+
+    fn into_par_iter(self) -> Self::Iter {
+        self.0.par_iter_mut()
     }
 }
 
--- a/geo-types/src/geometry/multi_line_string.rs
+++ b/geo-types/src/geometry/multi_line_string.rs
@@ -240,8 +240,8 @@ mod test {
         let mut multimut: MultiLineString<i32> = wkt! {
             MULTILINESTRING((0 0,2 0,1 2,0 0), (10 10,12 10,11 12,10 10))
         };
-        let _ = multi.par_iter().for_each(|_p| ());
-        let _ = multimut.par_iter_mut().for_each(|_p| ());
+        multi.par_iter().for_each(|_p| ());
+        multimut.par_iter_mut().for_each(|_p| ());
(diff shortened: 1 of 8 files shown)


frewsxcv (approved):
Can you also update the doc comment about the `multithreading` feature flag in `geo/lib.rs`? To mention these new ~features~ functionality, and/or indicate it's not just for `rstar`

geo-types/src/lib.rs
```diff
@@ -62,6 +62,8 @@
 //! The following optional [Cargo features] are available:
 //!
 //! - `std`: Enables use of the full `std` library. Enabled by default.
+//! - `multithread`: Enables multi-threaded iteration over `Multi*` geometries. **Disabled**
```
michaelkirk:
```suggestion
//! - `multithreading`: Enables multi-threaded iteration over `Multi*` geometries. **Disabled**
```

geo-types/src/geometry/multi_line_string.rs
```diff
@@ -118,6 +120,36 @@ impl<T: CoordNum> MultiLineString<T> {
...
+impl<T: CoordNum + Send> IntoParallelIterator for MultiLineString<T> {
+    type Item = LineString<T>;
+    type Iter = rayon::vec::IntoIter<LineString<T>>;
+
+    fn into_par_iter(self) -> Self::Iter {
+        self.0.into_par_iter()
+    }
+}
+
+#[cfg(feature = "multithreading")]
```
michaelkirk, line 133: The consuming `into` implementation makes sense to me.

And I can see how the ref and mut flavors below provide necessary **non**-consuming functionality. But what does it look like to use it? Can you include some example code in your PR that uses these new methods to make the new interface easier to evaluate?

My guess is that you'd need to do something like

`(& my_multi_line_string).into_par_iter()`
and
`(&mut my_multi_line_string).into_par_iter()`
?

Which is a bit rough. I'm afraid people will just end up `my_multi_line_string.clone().into_par_iter()` out of confusion.

Would it get better if we got rid of these two ref/mut implementations and instead implemented `rayon::IntoParallelRefIterator` and `rayon::IntoParallelRefMutIterator` which seem purpose built for this?

michaelkirk (changes requested):

michaelkirk (commented):
Feature flags LGTM!

geo-types/src/lib.rs
urschrei: Ha caught it already. Sorry!

geo-types/src/geometry/multi_line_string.rs
urschrei: Hm. These work with the current impls:


```rust
    #[test]
    fn test_multithreading_linestring() {
        let multi: MultiLineString<i32> = wkt! {
            MULTILINESTRING((0 0,2 0,1 2,0 0), (10 10,12 10,11 12,10 10))
        };
        let mut multimut: MultiLineString<i32> = wkt! {
            MULTILINESTRING((0 0,2 0,1 2,0 0), (10 10,12 10,11 12,10 10))
        };
        let _ = multi.par_iter().for_each(|_p| ());
        let _ = multimut.par_iter_mut().for_each(|_p| ());
        let _ = &multi.into_par_iter().for_each(|_p| ());
        let _ = &mut multimut.par_iter_mut().for_each(|_p| ());
    }
```

I admittedly didn't see the `rayon::IntoParallelRefIterator` and just adapted our existing iterators to `rayon::IntoParallelIterator`, but I'm confused by

> This trait is automatically implemented for I where &I: IntoParallelIterator. In most cases, users will want to implement [IntoParallelIterator](https://docs.rs/rayon/latest/rayon/iter/trait.IntoParallelIterator.html) rather than implement this trait directly.

michaelkirk: Oh splendid! They're already implemented by the blanket impl!

I'm glad that users will be able to do:
```
let _ = multi.par_iter().for_each(|_p| ());
let _ = multimut.par_iter_mut().for_each(|_p| ());
```

vs. the slightly more arcane:
```
let _ = &multi.into_par_iter().for_each(|_p| ());
let _ = &mut multimut.into_par_iter().for_each(|_p| ());
```

michaelkirk (approved):

geo-types/src/geometry/multi_line_string.rs
urschrei: Oh yeah the latter two were just to make sure "everything" worked. Thanks for the sanity check!

geo-types/src/geometry/multi_line_string.rs
```diff
@@ -231,6 +231,20 @@ mod test {
     use super::*;
     use crate::{line_string, wkt};
 
+    #[test]
```
michaelkirk, line 234: Don't forget to feature gate theses tests

urschrei: 😅

PR #1274, opened 2024-11-21
feat: make `FrechetDistance` generic over Metric Spaces

grevgeny:
---

Part of https://github.com/georust/geo/issues/1181


grevgeny:
@michaelkirk What do you think about the following approach to `FrechetDistance` with new metric space setup for lines?

lnicola:
This is only a drive-by comment, but wouldn't it make sense to pass the metric as a value, e.g. `g1.frechet_distance(&Euclidean, g2)`? I think some metrics can be parameterized, like Mahalanobis, which takes a point and a covariance matrix.

lnicola:
We could go directly for `Euclidean.frechet_distance(&line_string_1, &line_string`, then.

lnicola:
Yeah, something like that. In my example, `Euclidean` is an empty struct, and there's a single trait method, which takes `&self`. I imagine we could also make constants for the more common parameterizations.

michaelkirk:
Euclidean isn't a very illuminating example, because I don't think you'd ever want to customize it, so let's talk about Haversine.

... maybe this would be better:
```rust
trait FrechetDistance: Distance<Point, Point> {
  fn frechet_distance(&self, lhs: &LineString, rhs: &LineString) -> F {
     // more or less the same impl as it exists today
  }
}

pub struct Haversine;
pub struct CustomHaversine { radius_meters: f64 };

impl Haversine {
  pub fn with_radius(radius_meters: f64) -> CustomHaversine {
    // Is it weird that this returns something other than `Self`?
    CustomHaversine { radius_meters }
  }
}

impl Distance<Point, Point> for CustomHaversine {
  fn to_radians(degrees: f64) -> f64 {
    degrees * PI / 180.0
  }
  fn distance(&self, lhs: &Point, rhs: &Point) -> F {
     let lat1 = Self::to_radians(lat1);
     let lon1 = Self::to_radians(lon1);
     let lat2 = Self::to_radians(lat2);
     let lon2 = Self::to_radians(lon2);

     let dlat = lat2 - lat1;
     let dlon = lon2 - lon1;

     let a = (dlat / 2.0).sin().powi(2) + lat1.cos() * lat2.cos() * (dlon / 2.0).sin().powi(2);
     let c = 2.0 * a.sqrt().atan2((1.0 - a).sqrt());

     self.radius_meters * c  
  }
}

impl Distance<Point, Point> for Haversine {
 fn distance(&self, lhs: &Point, rhs: &Point) -> F {
    CustomHaversine::default().distance(lhs, rhs)
  }
}

// stays concise for the vast majority (I think?) of people who don't want to customize anything
Haversine.frechet_distance(&line_string_1, &line_string)

// Offer a slightly more verbose API for the 0.5% of users who want to customize the metric space
let mars_haversine = Haversine::with_radius(3_389_500);
mars_haversine.frechet_distance(&line_string_1, &line_string)
```

vs. my [previous suggestion](https://github.com/georust/geo/pull/1274#issuecomment-2515597271), this would avoid the weird distinction between `custom_frechet_distance` and `distance`. For something like Euclidean this distinction probably wouldn't even make sense.

Sorry to derail your PR review @grevgeny - it's kind of related though. 😅

grevgeny:
@michaelkirk @lnicola Thanks for your review! 

Should I then adjust my code based on your latest comment @michaelkirk ?

michaelkirk:
I'm not sure. I'm also looking for some design feedback on my proposal.

(/cc @lnicola too)

Do you think the approach I laid out above makes sense - I think it's what you were getting at @lnicola, but I wanted to use a non-euclidean example. 

It would be a breaking change to move all the recently introduced LineMeasure implementations from (e.g.) `Haversine::distance` to `Haversine.distance`, but maybe worth it for the functionality of being able to customize.

michaelkirk:
Sorry to keep you in a holding pattern @grevgeny. 

The reason I'm holding up the train is because I wanted to make some breaking changes to the line measures stuff to support customizable metric spaces.

Rather than release this, and then immediately break it, I'd prefer to make those API-breaking changes first, then adjust this PR.

I've just now opened the PR with those breaking changes here: #1298  - if/when that gets merged, I'd like to see this PR updated to look like:
```rust
Haversine.frechet_distance(line_string_1, line_string_2)
Euclidean.frechet_distance(line_string_1, line_string_2)
CustomHaversine::new(3_389_500.0).frechet_distance(line_string_1, line_string_2)
```

But maybe just hold off for now - I'm not sure how controversial #1298 will be. Feel free to weigh in over there.

lnicola:
> Do you think the approach I laid out above makes sense - I think it's what you were getting at @lnicola, but I wanted to use a non-euclidean example.

```rust
Haversine.frechet_distance(&line_string_1, &line_string_2);
Haversine::with_radius(3_389_500).frechet_distance(&line_string_1, &line_string_2);
```

I.. haven't given this much thought, but the following feel lighter to me compared to having both `Haversine` and `CustomHaversine`:

```rust
EARTH_HAVERSINE.frechet_distance(&line_string_1, &line_string_2);
// or
Haversine::default().frechet_distance(&line_string_1, &line_string_2);
// or
Haversine::earth().frechet_distance(&line_string_1, &line_string_2);

Haversine::with_radius(3_389_500.0).frechet_distance(&line_string_1, &line_string_2);
```

lnicola:
`Euclidean` doesn't need to be customizable, but I'd argue for `EUCLIDEAN`, for consistency. If we ever consider adding Mahalanobis (from my previous example), it could be argued it's more efficient to just transform the data. But the other three could, in principle, be customized.

While `Metric` / `CustomMetric` is a bit annoying to implement, it's not the end of the world either, and I guess we're unlikely to add a lot of other metrics.

michaelkirk:
Ok, #1298 has merged! (sorry for the delay)

Can you base this on what's now in master?

So I'd expect usage to look like:

```
Haversine.frechet_distance(&line_string_1, &line_string_2);
Euclidean.frechet_distance(&line_string_1, &line_string_2);
let mars = HaversineMeasure::new(mars_radius);
mars.frechet_distance(&line_string_1, &line_string_2);
```

grevgeny:
Sure! Thanks for the updates, I'll update the code accordingly

grevgeny:
PR is ready for review. 

@michaelkirk

geo/src/algorithm/frechet_distance.rs
```diff
@@ -1,8 +1,8 @@
-use crate::coords_iter::CoordsIter;
-use crate::line_measures::{Distance, Euclidean};
+use crate::line_measures::Euclidean;
 use crate::{GeoFloat, LineString};
 use num_traits::FromPrimitive;
 
+#[deprecated]
```
michaelkirk, line 5: Please add a note referring them to the new callsite, `Euclidean::frechet_distance` - see other `deprecated` declarations in the repo for an example.

geo/src/algorithm/line_measures/frechet_distance.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+impl<F: CoordFloat> Data<'_, F> {
+    /// [Reference implementation]: https://github.com/joaofig/discrete-frechet/tree/master
+    fn compute_linear(&mut self, metric_space: &impl Distance<F, Point<F>, Point<F>>) -> F {
+        let columns_count = self.ls_b.coords_count();
+
+        for (i, a) in self.ls_a.points().enumerate() {
+            for (j, b) in self.ls_b.points().enumerate() {
+                let dist = metric_space.distance(a, b);
```
michaelkirk, line 81: I haven't scrutinized this implementation, but other than changing this `Euclidean.distance` to a generic `metric_space.distance`, it looks like it was just cut/pasted from the previous implementation.

Which is expected! :+1:

 But let me know if you've changed anything else in the impl that you'd like scrutinized.

michaelkirk (changes requested):
This looks great! Just a couple minor house cleaning nits.

geo/src/algorithm/frechet_distance.rs
grevgeny: Fixed

geo/src/algorithm/line_measures/frechet_distance.rs
grevgeny: You're right. The only thing changed is the distance calculation between points.

geo/src/algorithm/frechet_distance.rs
```diff
@@ -1,8 +1,11 @@
-use crate::coords_iter::CoordsIter;
-use crate::line_measures::{Distance, Euclidean};
+use crate::line_measures::Euclidean;
 use crate::{GeoFloat, LineString};
 use num_traits::FromPrimitive;
 
+#[deprecated(
+    since = "0.30.0",
+    note = "Please use the `Euclidean.frechet_distance` method from the `geo::line_measures::FrechetDistance` trait instead"
```
michaelkirk, line 7: perfect!

michaelkirk (approved):
Looks great! Wow, what a journey. Thanks for your patience!

PR #1276, opened 2024-12-02
Update fmt and clippy for Rust 1.83

kylebarron:
---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,7 +53,7 @@ jobs:
       matrix:
         container_image:
           # Use the latest stable version. No need for older versions.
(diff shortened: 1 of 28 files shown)


kylebarron:
Does the failing test imply that the fmt of rust 1.83 is not a superset of the fmt of rust 1.79? Should we update that CI to use rust 1.83?

michaelkirk:
I'd be fine to update the format version in CI.

kylebarron:
Ah looks like we need to push a new container. That's in https://github.com/georust/docker-images?

michaelkirk:
> Ah looks like we need to push a new container. That's in https://github.com/georust/docker-images?

Ah yes, I'm taking a look at that now.

michaelkirk (approved):

Issue #1268, opened 2024-11-11
Crash in polygon.contains(linestring)


dabreegster:
To reproduce, clone https://github.com/dabreegster/georust_crash and `cargo run`. The input coordinates are Euclidean, relative to a 0,0 origin. I tried a few online WKT viewers, but all of them need some kind of CRS. I'll try the JTS one later and try to understand the problem.

```
geo-0.29.1/src/algorithm/relate/geomgraph/edge_end_bundle_star.rs:112:21:
side_location conflict with coordinate: Coord { x: 306.05652706452844, y: 848.7298084068675 }, right_location: Inside,  current_location: Outside
```

urschrei:
JTS thinks The polygon's invalid:

```
Ring self-intersection at or near point (293.0157983578036, 694.0574518014394, NaN)
```

urschrei:
If you "fix" the input Polygon (which will split it into a MultiPolygon with a little triangular outlier beneath the main geometry):

```
[...]

PR #1279, opened 2024-12-09
Validation::is_valid

### LINKED ISSUE #123: Dealing with Validity

michaelkirk:
---

Fixes #123, #127, #1268 

Much of the original code and test cases are courtesy of https://github.com/mthh/geo-validity-check (thanks @mthh!).

I also enabled the JTS IsValid test cases from our existing JTS test suite.

## My approach:
First, I just copy/pasted the existing implementation from https://github.com/mthh/geo-validity-check in: https://github.com/georust/geo/pull/1279/commits/92d8bfc39119a2c8114e219438c773ac2816bf2f

And then wired it into our crate: https://github.com/georust/geo/pull/1279/commits/5dd1909cab2af518d6b7652f696665215c732226

Most significantly, I did a lot of refactoring in pursuit of deduplicating code in https://github.com/georust/geo/pull/1279/commits/b89e715b84300608feff56f83e65acf760d6f965, so any bugs are my own.

Note to merger: please don't squash these commits, they're somewhat meaningful.


michaelkirk:
@mthh - since this is largely based on your implementation, and you've done some work with validation, would you mind reviewing this?

geo/CHANGES.md
```diff
@@ -4,6 +4,20 @@
 
 - Loosen bounds on `RemoveRepeatedPoints` trait (`num_traits::FromPrimitive` isn't required)
   - <https://github.com/georust/geo/pull/1278>
+- Added: `Validation` trait to check validity of `Geometry`.
+  - https://github.com/georust/geo/pull/1279
+  ```
```
frewsxcv, line 9:
```suggestion
  ```rust
```

geo/src/algorithm/validation/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+/// let text = validation_error.to_string();
+/// assert_eq!(text, "exterior ring must have at least 3 distinct points");
+//
+/// // Get all validation errors
+/// let all_validation_errors = invalid_polygon.validation_errors();
+/// assert_eq!(all_validation_errors.len(), 2);
+/// assert_eq!(all_validation_errors[0].to_string(), "exterior ring must have at least 3 distinct points");
+/// assert_eq!(all_validation_errors[1].to_string(), "interior ring at index 0 is not contained within the polygon's exterior");
+/// ```
+pub trait Validation {
```
frewsxcv, line 62: Can you add an entry for this in the crate root docs? `src/lib.rs`

geo/src/algorithm/validation/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+    type Error: fmt::Debug + fmt::Display;
```
frewsxcv, line 63: Any reason you didn't reach for `std::error::Error` here? (which extends both of these traits)

frewsxcv (commented):
Exciting! Can't wait for this to land

Definitely not done reviewing, just did a first pass and added a few comments

geo/Cargo.toml
```diff
@@ -28,6 +29,7 @@ geographiclib-rs = { version = "0.2.3", default-features = false }
 log = "0.4.11"
 num-traits = "0.2"
 proj = { version = "0.27.0", optional = true }
+geos = { version = "9.1.1", features = ["geo"], optional = true }
```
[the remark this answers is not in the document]
urschrei: Just to check: we only build `geos` if `geos-tests` is enabled (it isn't by default)?

Also: note that libgeos is LGPL. I'm not sure whether "building bindings using the source" counts as use for licensing purposes, but the source is bundled in the geos crate: https://github.com/georust/geos/tree/master/sys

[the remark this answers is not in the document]
michaelkirk: > Just to check: we only build geos if geos-tests is enabled (it isn't by default)?

That's correct.

> Also: note that libgeos is LGPL. I'm not sure whether "building bindings using the source" counts as use for licensing purposes.

Thanks, I somehow never realized geos was LGPL. I'm inclined to remove it since its inclusion is not that critical. Bummer though.

geo/src/algorithm/validation/mod.rs
michaelkirk: Great point - updated!

geo/src/algorithm/validation/geometry.rs
```diff
@@ -0,0 +1,146 @@
...
+    InvalidGeometryCollection, InvalidLine, InvalidLineString, InvalidMultiLineString,
+    InvalidMultiPoint, InvalidMultiPolygon, InvalidPoint, InvalidPolygon, InvalidRect,
+    InvalidTriangle,
+};
+use crate::{GeoFloat, Geometry};
+
+use crate::geometry_cow::GeometryCow;
+use std::fmt;
+
+/// A [`Geometry`] is valid if it's inner variant is valid.
```
frewsxcv, line 12:
```suggestion
/// A [`Geometry`] is valid if its inner variant is valid.
```

geo/src/algorithm/validation/mod.rs
```diff
@@ -0,0 +1,154 @@
...
+    fn check_validation(&self) -> Result<(), Self::Error> {
+        self.visit_validation(Box::new(Err))
+    }
+
+    /// Visit the validation of the geometry.
+    ///
+    /// The closure `handle_validation_error` is called for each validation error.
+    fn visit_validation<T>(
+        &self,
+        handle_validation_error: Box<dyn FnMut(Self::Error) -> Result<(), T> + '_>,
```
frewsxcv, line 99: Presumably something like this wasn't possible?

```
handle_validation_error: FnMut(Self::Error) -> Result<(), T>,
```

geo/src/algorithm/validation/mod.rs
```diff
@@ -0,0 +1,154 @@
...
+        match self {
+            RingRole::Exterior => write!(f, "exterior ring"),
+            RingRole::Interior(idx) => write!(f, "interior ring at index {}", idx),
+        }
+    }
+}
+
+/// The position of the problem in a multi-geometry, starting at 0.
+#[derive(Debug, PartialEq, Clone)]
+pub struct GeometryIndex(usize);
```
frewsxcv, line 121: Is the unnamed field here intentionally not `pub`? As long as this structure is exposed publicly, it might be helpful for users to be able to read the value

frewsxcv (commented):
Another pass

geo/src/algorithm/validation/mod.rs
michaelkirk: Good point - done!

geo/src/algorithm/validation/mod.rs
michaelkirk: Because of the potentially infinite recursion of `GeometryCollection`, we cannot use a true generic `impl FnMut` - it must be boxed. 

Is that what you were asking about?

michaelkirk: Here's the error message when you switch to: `handle_validation_error: impl FnMut(Self::Error) -> Result<(), T>,`

```
error: reached the recursion limit while instantiating `validation::geometry_collection::<impl Validation for geo::GeometryCollection>::visit_validation::<geo::validation::InvalidGeometry, Box<{closure@validation::geometry::<impl Validation for geo::Geometry>::visit_validation<..., ...>::{closure#7}}>>`
  --> /Users/mkirk/src/georust/geo/geo/src/algorithm/validation/geometry.rs:91:48
   |
91 |               Geometry::GeometryCollection(g) => g.visit_validation(Box::new(|err| {
   |  ________________________________________________^
92 | |                 handle_validation_error(InvalidGeometry::InvalidGeometryCollection(err))
93 | |             }))?,
   | |_______________^
   |
note: `validation::geometry_collection::<impl Validation for geo::GeometryCollection<F>>::visit_validation` defined here
  --> /Users/mkirk/src/georust/geo/geo/src/algorithm/validation/geometry_collection.rs:28:5
   |
28 | /     fn visit_validation<T>(
29 | |         &self,
30 | |         mut handle_validation_error: impl FnMut(Self::Error) -> Result<(), T>,
31 | |     ) -> Result<(), T> {
   | |______________________^
   = note: the full type name has been written to '/Users/mkirk/src/georust/geo/target/debug/deps/jts_test_runner-201a42bc420bd6fb.long-type.txt'


```

frewsxcv (approved):
I replied with one more comment but this looks good to me! Thanks for putting the effort into making this happen

geo/src/algorithm/validation/mod.rs
michaelkirk: I don't think that addresses the potentially infinite recursion of the closure - without boxing the closure, the compiler must build a type which can store the state of a closure which validates a `GeometryCollection` inside a `GeometryCollection` inside a `GeometryCollection` inside a...

I confess I'm not that confident in my analysis, but I did have a go at your suggestion [here](https://github.com/georust/geo/tree/mkirk/mthh_is_valid_2-code-review-closure-unboxing), but ultimately hit the same compiler error:

> Reached the recursion limit while instantiating `validation::geometry_collection::<impl Validation for geo::GeometryCollection>::visit_validation::<geo::validation::InvalidGeometry, {closure@validation::geometry::<impl Validation for geo::Geometry>::visit_validation<geo::validation::InvalidGeometry, ...>::{closure#7}}>`

michaelkirk:
Thanks for the review @frewsxcv!

I'll give it another day before merging in case anyone else wants to look.

mthh:
@michaelkirk thank you for inviting me to review it, unfortunately I won't have the time to do so.
Anyway I'm glad to see the Validation trait land in geo, thanks for the effort :)

michaelkirk:
See https://github.com/georust/geo/issues/1283 for details on the unrelated CI failure.

Issue #1283, opened 2024-12-19
CI broken, due to MSRV of `home` crate.


michaelkirk:
CI is currently broken, due to an [MSRV bump](https://github.com/rust-lang/cargo/pull/14871) of the [`home`](https://crates.io/crates/home) crate.

Our direct dependency `proj` has a transitive dependency on the [`which`](https://crates.io/crates/which) crate, which in turn depends on this `home` crate.

Someone recently opened https://github.com/harryfei/which-rs/issues/104, which well summarizes the issue

> [home is not intended for external use (outside of cargo)](https://github.com/notpeter/env-home/blob/main/README.md#as-an-alternative-to-the-home-crate). home has cargo's automatic MSRV-3 policy, which causes build failures for downstream dependants who check against their MSRV and where the package resolver is not MSRV-aware.

They proposed a fix just yesterday: https://github.com/harryfei/which-rs/pull/105 

> [env-home](https://github.com/notpeter/env-home/) is a better solution:
>
> - it is intended for external users
> - it is the proof of concept to replace env::home_dir: https://github.com/rust-lang/libs-team/issues/372
> - it has a much lower MSRV
> - it is MIT or Apache 2.0 licensed (like home)

As a side note, I just learned that rust 2024 intends to have an MSRV aware package resolution, yay!

We can either:
- wait to see if this fix gets merged soon
- pin to an older version of `home`
- abandon our 6+ month MSRV support window  (not my preferred option)

All this to read a couple of env vars 😆

urschrei:
Let's give it a few days, but I can't see a fix landing before Christmas

michaelkirk:
#1285 was merged, which works around the problem, but I'm going to leave this issue open as a reminder that ultimately we should remove the dev-dependency specification for `home`.

Plausible resolutions are:

- https://github.com/harryfei/which-rs/pull/105 gets merged
- or we wait long enough that our MSRV is recent enough for `home`

Xaeroxe:
`which` 7.0.1 is now released which should resolve this problem.

michaelkirk:
Oh geeze, turns bindgen [removed the which crate a while back](https://github.com/rust-lang/rust-bindgen/commit/c03964c0257b03ad4762b57f3ffb96d32d49c0ce).

Sorry for the noise @Xaeroxe, but thanks for the quick response anyway.

For us, the next steps are:
- [ ] update bindgen in proj-sys
- [ ] release proj-sys
[...]

PR #1285, opened 2024-12-19
fix build on rust pre 1.81

michaelkirk:
---

Works around #1283


kylebarron (approved):

michaelkirk:
I'm gonna go ahead and merge this, since we have a backlog of PR's that I don't want to sit on over the holidays.

PR #1288, opened 2024-12-20
[WIP] Example of const generics for N dimension Coords

latot:
Hi, I play a little with the crate, and touched the ```coords.rs``` file, I was able to set everything to use const generics so it could works just using any number of dimensions.

What I did not do, is write the serialize/deserialize part, because are not implement by default for ```[T; N]```, still this is not made to be merged, but as a proof of concept and see how it could be.


michaelkirk:
This seems like a dupe of https://github.com/georust/geo/pull/797 - please review that before opening a PR with similar functionality.

latot:
@michaelkirk I did not checked it! I'll do it on the future.

but this PR is essentially different, here is to do it works with N dimensions, arbitrary one.

PR #1294, opened 2025-01-05
Implement `RTreeObject` for `Triangle`

b4l:
---

Found that this was missing when playing around.


urschrei:
I know it's annoying, but I think splitting it makes sense!

urschrei (approved):

b4l:
> I know it's annoying, but I think splitting it makes sense!

Done in #1295

PR #1295, opened 2025-01-05
Implement `AsRef<Coord>` for `Point` and `Coord`

b4l:
---

Add `AsRef<Coord>` for `Point` and `Coord`. 

There might be other places in the codebase deserving adaptation in case this is accepted as a way forward. 

Emerged and splitted from #1294.

### DIFF
--- a/geo-types/CHANGES.md
+++ b/geo-types/CHANGES.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Implement `RTreeObject` for `Triangle`.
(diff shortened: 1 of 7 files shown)


urschrei (approved):

michaelkirk:
Probably worth calling out (I didn't realize it at the time) but because the method signature in our private_utils has changed, we can't release `geo` until we make a new `geo-types` release and then bump geo's geo-types dependency to it.

urschrei:
Arghh. The Great Uncoupling strikes again.

PR #1298, opened 2025-01-11
Customizable metric space for line measures

michaelkirk:
---

## Motivation

Currently, our [`Haversine`](https://docs.rs/geo/latest/geo/algorithm/line_measures/metric_spaces/struct.Haversine.html) metric space calculation uses the IUGG's "mean radius", but there are several other reasonable values. Similarly, the [`Geodesic`](https://docs.rs/geo/latest/geo/algorithm/line_measures/metric_spaces/struct.Geodesic.html) internally uses Karney's geographic algorithms which support other geoid parameters.

## Proposal

By reorganizing our traits, we can let users take advantage of all the associated algorithms (length, densify, distance) for any given (potentially customizable) metric space. 

The trick is, almost nobody (by my count) actually wants this configurability, so the resultant API must not make it burdensome for those who want the "default" case. 

Some [conversation in #1274](https://github.com/georust/geo/pull/1274#issuecomment-2514454615) and [another in #1140](https://github.com/georust/geo/pull/1140#pullrequestreview-1907267868) lead me to revisit the recently overhauled line_measure traits, and that it's possible to support this kind of customization in a reasonable way.

There are a lot of changes, but they are pretty mechanical. You can get a pretty good idea of the kinds of changes our users will need to make by reviewing this diff. 

As well as being more expressive, I actually prefer where the code is now, without the turbo-fish. I just wish I'd avoided the churn by thinking of it before the big changes in the v0.29.0 (primarily #1216) were released (on 2024-10-30):

```rust
// Before v0.29.0)
p1.haversine_distance(p2);
line_string.haversine_length();
line_string.densify_haversine(100);

// Currently (as of v0.29.0)
Haversine::distance(p1, p2);
line_string.length::<Haversine>();
line_string.densify::<Haversine>(100);

// After (this PR)
Haversine.distance(p1, p2);
Haversine.length(&line_string);
Haversine.densify(&line_string, 100);

// and now you can do this:
let planet_mars = CustomHaversine::new(3_389_500.0);
planet_mars.densify(&line_string, 100);
```

### References 

This solves the same problem as https://github.com/georust/geo/pull/1220, but more generally.
Tracking issue for line_measure overhaul here #1181

### DIFF
--- a/geo/benches/euclidean_distance.rs
+++ b/geo/benches/euclidean_distance.rs
@@ -37,7 +37,7 @@ fn criterion_benchmark(c: &mut criterion::Criterion) {
             (x: -6.064453, y: 68.49604),
         ];
         bencher.iter(|| {
-            criterion::black_box(Euclidean::distance(&poly1, &poly2));
+            criterion::black_box(Euclidean.distance(&poly1, &poly2));
         });
     });
 
@@ -79,7 +79,7 @@ fn criterion_benchmark(c: &mut criterion::Criterion) {
             ]
             .convex_hull();
             bencher.iter(|| {
-                criterion::black_box(Euclidean::distance(&poly1, &poly2));
+                criterion::black_box(Euclidean.distance(&poly1, &poly2));
             });
         },
     );
--- a/geo/benches/geodesic_distance.rs
+++ b/geo/benches/geodesic_distance.rs
@@ -7,7 +7,7 @@ fn criterion_benchmark(c: &mut criterion::Criterion) {
         let b = geo::Point::new(16.372477, 48.208810);
 
         bencher.iter(|| {
-            criterion::black_box(criterion::black_box(Geodesic::distance(a, b)));
+            criterion::black_box(criterion::black_box(Geodesic.distance(a, b)));
         });
     });
 }
--- a/geo/src/algorithm/centroid.rs
+++ b/geo/src/algorithm/centroid.rs
@@ -466,7 +466,7 @@ impl<T: GeoFloat> CentroidOperation<T> {
         match line.dimensions() {
             ZeroDimensional => self.add_coord(line.start),
             OneDimensional => {
-                self.add_centroid(OneDimensional, line.centroid().0, line.length(&Euclidean))
+                self.add_centroid(OneDimensional, line.centroid().0, Euclidean.length(line))
             }
             _ => unreachable!("Line must be zero or one dimensional"),
         }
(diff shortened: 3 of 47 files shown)


geo-types/CHANGES.md
```diff
@@ -2,7 +2,6 @@
 
 ## Unreleased
 
-- Fix page location of citation for mean earth radius used in Haversine calculations
 - Implement `RTreeObject` for `Triangle`.
```
michaelkirk, line 5: I've only moved this to `geo/CHANGES.md` where I think it belongs.

michaelkirk:
Converting this draft while I incorporate some feedback from @lnicola - what do you think, does 
[Make all Haversine customizable, removing CustomHaversine](https://github.com/georust/geo/pull/1298/commits/71038e0503495d333a2fd7bd48a7d9924d850616) make things better or worse?

In summary:

```rust
// "long ago" before 0.29.0
point1.haversine_distance(point2)
point1.rhumb_distance(point2)

// as of 0.29.0
Haversine::distance(point1, point2)
Rhumb::distance(point1, point2)

// v1 of this PR:
CustomHaversine::new(3_389_500.0).distance(point1, point2)
Haversine.distance(point1, point2)
Rhumb.distance(point1, point2)

// current version of this PR (with @lnicola's suggestion - removing "CustomHaversine")
Haversine::new(3_389_500.0).distance(point1, point2)
HAVERSINE.distance(point_1, point_2)
RHUMB.distance(point_1, point_2)
```

michaelkirk:
What do you think @lnicola - based on the diff do you prefer the `HAVERSINE` constant vs having separate `Haversine`/`CustomHaversine { radius }` structs? 

If so, I'd like to push through the rest of the changes.

lnicola:
@michaelkirk sorry for the late reply. It looks.. just as I expected. Seems fine to me, but you might want to ask for a third opinion :grinning:.

michaelkirk:
This is ready for review!

I've incorporated some more feedback (from discord) into this and updated the description with a "Version 3 of this PR".

urschrei:
My concern (perhaps badly articulated on Discord), was that we're shipping public structs that can contain an arbitrary value (e.g. the sphere radius in `HaversineMeasure`), which could be either confusing in downstream libraries / applications or used maliciously. One reasonable counterpoint to this is that downstream users are already trusting us, all kinds of malicious interventions are possible with a newtype wrapper, and that concerned users should be testing and auditing the code they use (`HaversineMeasure`'s `radius` field is public) and / or engaging in other supply-chain security practices that are not our responsibility.

urschrei (approved):

PR #1302, opened 2025-01-16
more legible geo-types debug output (WKT-ish)

### LINKED ISSUE #564: Write custom, more succinct, debug output for all geometries in geo-types
Currently we're just using the derived Debug, which is pretty noisy, especially for our compound structures:

e.g.
        
This one's maybe fine?
```
let coord = Coordinate {x:  1.0, y: 2.0};
println!("coord: {:?}", &coord);

> coord: Coordinate { x: 1.0, y: 2.0 }
```

But elsewhere, "Coordinate" isn't adding anything in my opinion:
```
let point = Point::from(coord);
println!("point: {:?}", &point);

> point: Point(Coordinate { x: 1.0, y: 2.0 })
```

Maybe something like: `Point( x: 1.0, y: 2.0)` would be better?

And that compounds:
```
let line = Line::new(coord, coord);
println!("line: {:?}", &line);

> line: Line { start: Coordinate { x: 1.0, y: 2.0 }, end: Coordinate { x: 3.0, y: 4.0 } }
```
Maybe something like: `Line[(x: 1.0, y: 2.0) -> (x: 3.0, y: 4.0)] ` would be better?

```
let multi_point = MultiPoint(vec![point, point, point]);
println!("multi_point: {:?}", &multi_point);

> multi_point: MultiPoint([Point(Coordinate { x: 1.0, y: 2.0 }), Point(Coordinate { x: 3.0, y: 4.0 }), Point(Coordinate { x: 5.0, y: 6.0 })])
```
Maybe something like: `MultiPoint[(x: 1.0, y: 2.0), (x: 3.0, y: 4.0), (x: 5.0, y: 6.0)] ` would be better?

This isn't intended to be a comprehensive spec - just trying to exemplify what I think is valuable - putting some thought into the specifics for each type will be some work.

michaelkirk:
---

FIXES #564 

I've wanted this for a long time! Inevitably for any project using geo I end up pulling in some other crate to produce debug output (usually wkt). The writing portion of wkt is pretty short (parsing is a little trickier) which is why I opted to copy the logic rather than relying on the wkt crate or something.

Note: Most of the output is valid WKT, but not all of it is.
For example, there is no RECT or TRIANGLE type in standard WKT.

And there are certain invalid representations you can make in geo that
are not valid WKT, but for a debug output format we still need to print
*something*

e.g. a polygon with interiors but no exterior: `POLYGON(EMPTY,(1 2,2 4))`


geo-types/src/debug.rs
```diff
@@ -0,0 +1,466 @@
...
+        }
+        write!(f, ")")?;
+    }
+    Ok(())
+}
+
+#[cfg(feature = "std")]
+#[cfg(test)]
+mod tests {
+    use super::*;
```
michaelkirk, line 185: The tests are pretty comprehensive, and the best place to see what's what.

geo-types/src/debug.rs
```diff
@@ -0,0 +1,466 @@
...
+        );
+    }
+    #[test]
+    fn rect() {
+        let rect = Rect::new((1, 2), (3, 4));
+        assert_eq!("RECT(1 2,3 4)", format!("{rect:?}"));
+
+        let rect = Rect::new((3, 4), (1, 2));
+        // output is always (min, max)
+        assert_eq!("RECT(1 2,3 4)", format!("{rect:?}"));
```
michaelkirk, line 390: Here's an example of output that's not actually WKT, but I think it's still intuitive.

urschrei:
I think the "don't let the perfect be the enemy of the good" is the right approach here w/r/t/ validity. If it's easy to get 95 % of debug output straight into TestBuilder (etc), and the other 5 % requires some light search and replace, that's still a huge improvement.

frewsxcv (approved):

PR #1304, opened 2025-01-17
Rename Spade* to Delaunay in new, identical module

urschrei:
---

The `Spade` suffix is an implementation detail, since we only happen to use the `spade` crate for Delaunay triangulations. It's also not semantically meaningful, whereas `Delaunay` is. I've left some non-public uses of `Spade` in the new code (where it refers to errors etc).

I've marked the existing module as deprecated, which is causing a lot of warnings as module-internal functions are called. I'm open to suggestions to suppress these or otherwise stop them annoying everyone.

### DIFF
--- a/geo/src/algorithm/mod.rs
+++ b/geo/src/algorithm/mod.rs
@@ -263,8 +263,20 @@ pub use triangulate_earcut::TriangulateEarcut;
 
 /// Triangulate polygons using an (un)constrained [Delaunay Triangulation](https://en.wikipedia.org/wiki/Delaunay_triangulation) algorithm.
 #[cfg(feature = "spade")]
+pub mod triangulate_delaunay;
+#[cfg(feature = "spade")]
+pub use triangulate_delaunay::TriangulateDelaunay;
+
+/// Triangulate polygons using an (un)constrained [Delaunay Triangulation](https://en.wikipedia.org/wiki/Delaunay_triangulation) algorithm.
+#[cfg(feature = "spade")]
+#[deprecated(
+    since = "0.29.4",
+    note = "please use the `triangulate_delaunay` module instead"
+)]
 pub mod triangulate_spade;
 #[cfg(feature = "spade")]
+#[deprecated(since = "0.29.4", note = "please use `TriangulateDelaunay` instead")]
+#[allow(deprecated)]
 pub use triangulate_spade::TriangulateSpade;
 
 /// Vector Operations for 2D coordinates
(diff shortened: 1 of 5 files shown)


geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -0,0 +1,786 @@
...
+    }
+}
+
+impl std::error::Error for TriangulationError {}
+
+pub type TriangulationResult<T> = Result<T, TriangulationError>;
+
+// ======= Float trait ========
+
+pub trait SpadeTriangulationFloat: GeoFloat + SpadeNum {}
```
urschrei, line 57: `Spade` has been kept here deliberately.

michaelkirk:
Can you pull in https://github.com/georust/geo/tree/mkirk/allow-deprecated-spade-usage (or something like it) to get rid of the deprecation warnings?

geo/src/algorithm/mod.rs
```diff
@@ -263,8 +263,19 @@ pub use triangulate_earcut::TriangulateEarcut;
 
 /// Triangulate polygons using an (un)constrained [Delaunay Triangulation](https://en.wikipedia.org/wiki/Delaunay_triangulation) algorithm.
 #[cfg(feature = "spade")]
+pub mod triangulate_delaunay;
+#[cfg(feature = "spade")]
```
michaelkirk, line 267: Should we get a more meaningful feature name, like `triangulate_delaunay`?

urschrei:
Clippy still fails because of `--all-targets`. If you run `cargo clippy --all-features --all-targets -- -Dwarnings` locally you'll see the warnings. Also present for Rust-analyzer

michaelkirk:
Huh, you're right. I thought  `#![allow(deprecated)]` would apply to the entire module (and it's submodules)

Further, even when I add another `#![allow(deprecated)]` to the tests module, the warnings persist.

Any idea why?

I don't think we should merge this until those warnings go away. 

One heavy handed approach would be to just delete the test module for this deprecated module, since it's duplicated in the new module.

geo/src/algorithm/mod.rs
michaelkirk: To be clear, I don't feel strongly about this part, and am happy to leave it as is.

urschrei:
I think the tests in the deprecated module are tripping Clippy. Let me try and pull them out (since we're replacing them in the new module)

geo/src/algorithm/mod.rs
urschrei: I'd like to keep `spade` for now, since we'll hopefully also be making use of some other `spade` functionality (Voronoi diagrams) in the near future.

michaelkirk (approved):

PR #1305, opened 2025-01-17
docs: add example to line_measure

### LINKED ISSUE #1303: Better top-level usage examples for line measures
Here's what the top-level docs for the `line_measures` and `Distance` modules look like currently:

Image
Image

To someone who wants to know how to calculate e.g. the Euclidean distance (I chose that example deliberately, since it's probably one of the most common operations) between two geometries, these docs offer _no_ hint of how you might do it. It's an easy fix, but we should urgently fix it.

michaelkirk:
---

FIXES #1303


michaelkirk:
Note, this is going to introduce some conflicts with #1298

urschrei (approved):

PR #1308, opened 2025-01-29
Fix Rect and Triangle to_polygon and iterator winding order

### LINKED ISSUE #1307: Rect::to_polygon creates clockwise-ordered Polygons
[Rect::to_polygon](https://github.com/georust/geo/blob/main/geo-types/src/geometry/rect.rs#L217) creates a polygon that is in clockwise order. I think this is a problem. To quote the documentation of module `orient:`

```
By default, a properly-oriented Polygon has its outer ring oriented counter-clockwise,
```

Now the order needs to be reversed to get the expected behavior

urschrei:
---

Closes #1307


michaelkirk (approved):
Nice - thanks for jumping on these so quickly!

There's also CoordsIter, which appears to be iterating in CW order. AFAIK we don't make any guarantees about winding order for CoordsIter, so maybe it's fine as is.

urschrei:
Oh yeah. I think it's a reasonable expectation.

urschrei:
If we're going to fix `Rect`, we also have to fix `Triangle`, right? A triangle can currently be constructed in either CW or CCW order (we just store the input) _but_ we also say

> The semantics and validity are that of the equivalent Polygon

So I'm proposing to:

- figure out the direction of the input
- store the input ccw by adjusting it if need be
- documenting that the coords are stored in ccw order and yielded in ccw order by iterators (same as we now say about `Rect`)

michaelkirk:
That makes sense to me. Good thinking.On Jan 30, 2025, at 14:31, Stephan Hügel ***@***.***> wrote:﻿
If we're going to fix Rect, we also have to fix Triangle, right? A triangle can currently be constructed in either CW or CCW order (we just store the input) but we also say

The semantics and validity are that of the equivalent Polygon

So I'm proposing to:

figure out the direction of the input
store the input ccw by adjusting it if need be
documenting that the coords are stored in ccw order and yielded in ccw order by iterators (same as we now say about Rect)

geo/src/algorithm/area.rs
```diff
@@ -399,7 +399,8 @@ mod test {
             coord! { x: 0.0, y: 1.0 },
             coord! { x: 1.0, y: 0.0 },
         );
-        assert_relative_eq!(triangle.signed_area(), -0.5);
+        // triangles are always ccw, thus positive
+        assert_relative_eq!(triangle.signed_area(), 0.5);
```
urschrei, line 403: Test is redundant now since you can't check a triangle for negative area. Remove?

geo-types/src/geometry/triangle.rs
```diff
@@ -1,20 +1,31 @@
...
 #[derive(Copy, Clone, Hash, Eq, PartialEq)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct Triangle<T: CoordNum = f64>(pub Coord<T>, pub Coord<T>, pub Coord<T>);
 
 impl<T: CoordNum> Triangle<T> {
     /// Instantiate Self from the raw content value
     pub fn new(v1: Coord<T>, v2: Coord<T>, v3: Coord<T>) -> Self {
-        Self(v1, v2, v3)
+        // determine cross product of input points. NB: non-robust
+        let orientation = Point::from(v1).cross_prod(v2.into(), v3.into());
```
urschrei, line 22: We would be better served by `robust::orient2d` here, but geo-types does not depend on it. It seems like a big addition for one method, but Robust has no dependencies and correctness is important…

geo/src/algorithm/area.rs
michaelkirk: Slight preference to keep it, if only that it documents that re-wind the input. But either is fine with me.

geo-types/src/geometry/triangle.rs
michaelkirk: Darn, that's annoying. But I think this makes more sense than bringing in another dependency.

So there's a world in which some user has given us the appropriate winding, but then we're mis-ordering the points here due to floating point robustness.

It's also worth noting that `Triangle` might be used for tessellation in a graphics rendering contexts - so any work done in the constructor could be a hot path. Cross product seems unlikely to be a bottleneck though (right?).

We *could* add something like `Triangle::unchecked_new(p1, p2, p3)` which works like the old constructor, assumes the user has pre-ordered their coords correctly. Or we could just wait to see if someone actually cares to request it. I'm ok either way.

michaelkirk (approved):

geo-types/src/geometry/triangle.rs
urschrei: If we did that, our iterators would provide different values depending on how the geometry was instantiated (In a third party library you might have triangles coming from different modules, one group constructed with `new` and one constructed with `new_unchecked`: that seems like a recipe for confusion / anger) and if we _didn't_ want that we'd be moving the perf overhead into the iterator logic by ensuring that we always start with e.g. the right-most coordinate.

I think I would rather have consistent behaviour and OGC Simple Features adherence where possible, even if that means a small perf penalty. Graphics applications have different priorities, and the `geo` ecosystem may not be a good fit for those. I realise that we aren't being internally consistent because we don't do this on Polygon, but we might in the future.

michaelkirk: > If we did that, our iterators would provide different values depending on how the geometry was instantiated 

Correct - that's what I was proposing to make available as a non-default escape hatch.

But like I said, I'm ok to omit that escape hatch for now, and revisit it **if** someone actually requests it.

PR #1310, opened 2025-01-30
Fix orientation of boolops geometries

### LINKED ISSUE #1309: BooleanOps creates polygons that are oriented clockwise
Similar to https://github.com/georust/geo/issues/1307, boolean operations in `BooleanOps` such as `intersect` and `union` create polygons whose exterior is oriented **clockwise.** However, "by default, a properly-oriented Polygon has its outer ring oriented counter-clockwise" (doc of module `orient`).

Thus, `BooleanOps` breaks my expectations and I have to `orient` them myself to get the expected results.

Here is a simple unit test to demonstrate the problem

```rust 
   #[test]
    fn test_winding_order() {
        let poly1 = polygon!((x: 0.0, y: 0.0), (x: 1.0, y: 0.0), (x: 1.0, y: 1.0));
        assert!(matches!(poly1.exterior().winding_order(), Some(WindingOrder::CounterClockwise)));

        {
            let union = poly1.union(&polygon!());
            assert_eq!(union.0.len(), 1);
    
            let union = &union.0[0];
            // FAILS
            assert!(matches!(union.exterior().winding_order(), Some(WindingOrder::CounterClockwise)));
        }
        {
            let intersection = poly1.intersection(&poly1);
            assert_eq!(intersection.0.len(), 1);

            let intersection = &intersection.0[0];
            // FAILS
            assert!(matches!(intersection.exterior().winding_order(), Some(WindingOrder::CounterClockwise)));
        }

        let poly2 = polygon!((x: 0.0, y: 0.0), (x: 1.0, y: 1.0), (x: 0.0, y: 1.0));
        assert!(matches!(poly2.exterior().winding_order(), Some(WindingOrder::CounterClockwise)));

        {
            let union = poly1.union(&poly2);
            assert_eq!(union.0.len(), 1);
    
            let union = &union.0[0];
            assert!(union.interiors().is_empty());
            // FAILS
            assert!(matches!(union.exterior().winding_order(), Some(WindingOrder::CounterClockwise)));
        }
    }
```

urschrei:
---

Closes #1309


michaelkirk (approved):
Darn, it was right in the docs:
>     /// Note: Outer boundary paths have a clockwise order, and holes have a counterclockwise order.

Rather than use `Orient` which does some inspection to determine the current winding, we could just *always* reverse the ordering.  It should be more efficient, but I don't know how expensive `Orient` is in the first place - so take it or leave it. We can always come back and improve it.


Good catch though!

urschrei:
> Rather than use Orient which does some inspection to determine the current winding, we could just always reverse the ordering.

Are we depending on upstream never changing the implementation if we do that?

michaelkirk:
>>> Rather than use Orient which does some inspection to determine the current winding, we could just always reverse the ordering.
>>
>> Are we depending on upstream never changing the implementation if we do that?
>
> We are depending on upstream not changing this documented behavior without bumping semver. Which is certainly not impossible but I'd be surprised. Though I've been surprised by i_overlay's semver policy before... 🤔

Actually, because we have pretty strict pinning for i_overlay (because of aforementioned semver suprises), we're only trusting them to not changing winding order in a **patch** version, which I'd feel comfortable with.  If we bump our version constraints and the winding order changes, these tests would fail which would be our cue to revisit.

But like I said, I'm ok to leave this as you have it for now.

urschrei:
Oh good point. No need to leave perf on the table for no reason.

michaelkirk:
If you're looking for inspo, I was testing this out: https://github.com/georust/geo/tree/mkirk/bools-ops-winding

PR #1311, opened 2025-02-03
Add customizable GeodesicMeasure

michaelkirk:
---

Building on #1298, adds the ability to customize your geoid for Karney geodesic calculations.

Note: this bumps MSRV to 1.80 (released 6+ months ago) in order to utilize `LazyLock`.

Perf: No measurable difference in the benchmarks

    geodesic distance f64   time:   [421.15 ns 423.53 ns 426.07 ns]
                            change: [-1.7862% -1.2409% -0.7204%] (p = 0.00 < 0.05)

### DIFF
--- a/geo/src/algorithm/line_measures/metric_spaces/geodesic.rs
+++ b/geo/src/algorithm/line_measures/metric_spaces/geodesic.rs
@@ -14,8 +14,8 @@ use std::sync::LazyLock;
 /// Distances are computed using [geodesic lines] and are measured **in meters**.
 ///
 /// [`GeodesicMeasure`] allows creating a geodesic model with custom values for the equatorial
-/// radius (`A`) and the inverse flattening factor (`F`). Unless you have special needs, just use
-/// [`Geodesic`].
+/// radius (`A`) and the inverse flattening factor (`F`). Unless you have special needs, use
+/// [`Geodesic`], which is simpler than this struct.
 ///
 /// # Example
 ///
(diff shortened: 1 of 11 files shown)


geo/src/algorithm/line_measures/metric_spaces/geodesic.rs
```diff
@@ -1,16 +1,120 @@
...
-/// Distances are computed using [geodesic lines] and are measured in meters.
+/// Measures distance on an ellipsoidal model of the Earth using methods given by [Karney (2013)].
+///
+/// Geodesic measurements are more accurate than [`Haversine`], but a bit slower, because they
+/// account for the fact that the Earth is not a perfect sphere.
+///
+/// Distances are computed using [geodesic lines] and are measured **in meters**.
+///
+/// [`GeodesicMeasure`] allows creating a geodesic model with custom values for the equatorial
+/// radius (`A`) and the inverse flattening factor (`F`). Unless you have special needs, just use
```
urschrei, line 17: The most minor of nits: we should try not to use "just" in our public docs IMO, as it assumes a degree of knowledge / familiarity that the reader may not have (we are actually **very** good about this, barring some instances floating around in `Haversine` examples)

urschrei (approved):

geo/src/algorithm/line_measures/metric_spaces/geodesic.rs
michaelkirk: This is a good nit. "just" is something I'm largely trying to strike from my writing. It usually doesn't add anything, is sometimes ambiguous, and sometimes condescending.

I've removed a few other instances in https://github.com/georust/geo/pull/1311/commits/1d9005e075f33b8c946859bac80423c334fe5424

PR #1312, opened 2025-02-05
impl approx::UlpsEqual for geo-types

michaelkirk:
---

I assumed we implemented all the "approximate equality" traits from approx long ago, but we missed one!

As a reminder, an ULPS comparison is effectively counting "how many floating point numbers are between these two numbers". For the most part it solves the same problems as relative_eq, but it's a little better behaved at the extremes, like when the "max_relative" factor you want to use is itself affected by floating point error.

### DIFF
--- a/geo-types/src/geometry/coord.rs
+++ b/geo-types/src/geometry/coord.rs
@@ -1,8 +1,5 @@
 use crate::{coord, CoordNum, Point};
 
-#[cfg(any(feature = "approx", test))]
-use approx::{AbsDiffEq, RelativeEq, UlpsEq};
-
 /// A lightweight struct used to store coordinates on the 2-dimensional
 /// Cartesian plane.
 ///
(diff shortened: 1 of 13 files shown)


geo-types/src/geometry/coord.rs
```diff
@@ -268,54 +265,57 @@ impl<T: CoordNum> Zero for Coord<T> {
...
-    #[inline]
-    fn default_epsilon() -> T::Epsilon {
-        T::default_epsilon()
-    }
+mod approx_integration {
+    use super::*;
+    use approx::{AbsDiffEq, RelativeEq, UlpsEq};
+
+    impl<T> AbsDiffEq for Coord<T>
+    where
```
michaelkirk, line 273: Other than the new UlpsEq impl below, these changes are formatting changes.

Though I also simplified/standardized some of our type constraints.

e.g. here `T::Epsilon: Copy` is replaced by `AbsDiffEq<Epsilon=T>`

Since `T: CoordNum`, we already know it's `Copy`.

In **theory**, the previous implementation might have allowed something like:

`Coord<f32>.abs_diff(Coord<f64>)`

But because `pub trait AbsDiffEq<Rhs = Self>` we only ever supported comparing to `Self` anyway (and I think that's just fine).

Bottom line, this doesn't change behavior and is more succinct/uniform with the other declarations.

geo-types/src/geometry/geometry_collection.rs
```diff
@@ -248,78 +246,101 @@ impl<'a, T: CoordNum> GeometryCollection<T> {
...
-    ) -> bool {
-        if self.0.len() != other.0.len() {
-            return false;
+mod approx_integration {
+    use super::*;
+    use approx::{AbsDiffEq, RelativeEq, UlpsEq};
+
+    impl<T> RelativeEq for GeometryCollection<T>
+    where
+        T: CoordNum + RelativeEq<Epsilon = T>,
```
michaelkirk, line 255: Another no-op simplification in the type constraints: We don't need to specify `AbsDiffEq` **and** `RelativeEq` because `RelativeEq: AbsDiffEq`, so just `RelativeEq` is sufficient.

geo-types/src/geometry/geometry_collection.rs
```diff
@@ -248,78 +246,101 @@ impl<'a, T: CoordNum> GeometryCollection<T> {
...
-            return false;
+    impl<T> UlpsEq for GeometryCollection<T>
+    where
+        T: CoordNum + UlpsEq<Epsilon = T>,
+    {
+        fn default_max_ulps() -> u32 {
+            T::default_max_ulps()
         }
 
-        let mut mp_zipper = self.into_iter().zip(other);
```
michaelkirk, line 321: I think this `mp_zipper` variable was copy/pasted from `MultiPolygon` or `MultiPoint`. No biggie, but I got rid of it since it's confusing in the context of `GeometryCollection`.

geo-types/src/geometry/geometry_collection.rs
```diff
@@ -248,78 +246,101 @@ impl<'a, T: CoordNum> GeometryCollection<T> {
 }
 
 #[cfg(any(feature = "approx", test))]
```
michaelkirk, line 248: Rather than having several conditional cfgs in the file, I consolidated all the `approx` implementations into one conditional module.

geo-types/src/geometry/coord.rs
michaelkirk: BTW I recommend turning on the option to hide whitespace changes for this diff.

geo-types/src/geometry/geometry_collection.rs
urschrei: Nice!

geo-types/src/geometry/mod.rs
```diff
@@ -270,104 +268,260 @@ where
...
+    {
+        fn default_max_ulps() -> u32 {
+            T::default_max_ulps()
+        }
+
+        /// Approximate equality assertion for floating point geometries based on the number of
+        /// representable floats that fit between the two numbers being compared.
+        ///
+        /// "relative_eq" might be more intuitive, but it does floating point math in its error
+        /// calculation, introducing its **own** error into the error calucation.
```
frewsxcv, line 399:
```suggestion
        /// calculation, introducing its **own** error into the error calculation.
```

PR #1315, opened 2025-02-08
implement `From<&(T, T)>` and `From<&[T; 2]>` for `Point<T>`

kylecarow:
---

This allows users to create points from borrowed tuples and slices, increasing flexibility


kylecarow:
On deeper thought, I think this is just a copy, so anything that uses this should probably just be copying the data for the point explicitly

PR #1316, opened 2025-02-27
Update i_overlay to 2.0.0

AndrewLipscomb:
---

Updates the `i_overlay` internal package to 2.0.0 - 2.1.0 

This is motivated by the fact that `i_overlay` now appears to offer the only implementation of polygon-offset style buffering in the ecosystem right now. 

I would be interested in doing a buffer trait + methods here at some point in the future, but for now a dependency bump is the first step in getting it in there. 

I have run `cargo test` with both current `main` and this change - both had tests fail for 

```
    algorithm::line_measures::metric_spaces::geodesic::tests::destination::interpolate_point::points_along_line_with_endpoints
    algorithm::line_measures::metric_spaces::geodesic::tests::destination::interpolate_point::points_along_line_without_endpoints
    algorithm::line_measures::metric_spaces::geodesic::tests::test_non_standard_geoid
```

which appear to be small floating point precision errors. Updating this library does not appear to have regressed anything else.

### DIFF
--- a/geo/Cargo.toml
+++ b/geo/Cargo.toml
@@ -31,7 +31,7 @@ proj = { version = "0.28.0", optional = true }
 robust = "1.1.0"
 rstar = "0.12.0"
 serde = { version = "1.0", optional = true, features = ["derive"] }
-i_overlay = { version = "1.10.0, < 1.11.0", default-features = false }
+i_overlay = { version = "2.0.0, < 2.1.0", default-features = false }
 
 [dev-dependencies]
 approx = ">= 0.4.0, < 0.6.0"


michaelkirk:
Thanks! 

Yes, I too am excited about the new i_overlay buffering functionality. I imagine we'll have it in geo before too long one way or another. 

In fact, I've recently started my own integration here: https://github.com/georust/geo/commits/mkirk/geo-buffer/

I'm using that branch in a test application now, and want to iterate on the API a bit before I proposed a particular integration with geo. You're free to use that branch (with no stability guarantees in the short term of course) or you can go your own way and propose something completely different.

FYI one of the things I'm undecided but concerned about is re-exporting some/all of the underlying i_overlay API. Typically we avoid this, so we can control our own stability (e.g. a breaking release in i_overaly wouldn't necessitate a breaking release in geo), but the i_overlay API _is_ pretty nice and feature-ful, so re-exporting it is tempting. 

That's one of the things I'm hoping to get an intuition for while I testbed this branch over the next couple weeks.

Btw, those test failures are likely due to some platform inconsistencies. I can't reproduce them (nor can CI it seems). But if you'd like to follow up by adding a small epsilon to the tests that are failing for you, I'd merge it. (search for `epsilon =` for similar examples).

michaelkirk:
In any case, it's fine to make this version increase, though I'll wait to merge until tomorrow in case someone unexpectedly has strong feelings about this for some reason.

AndrewLipscomb:
@michaelkirk Beautiful mate, thanks for the quick attention. 

Yep a days worth of wait is sensible enough. I'll have to have a read on your implementation - as a bigger maintainer here you'd probably have a better idea of whats a good direction for the library implementation. If you want to spin out a job or two my way - I'm keen to lend a hand.

PR #1321, opened 2025-03-04
line interpolate point across MetricSpaces

michaelkirk:
---

`LineInterpolatePoint` is currently Euclidean only.  As part of #1216, I wanted to make this functionality work across metric spaces.

```rust
// geos
line_interpolate_point(line, 0.8, true)
// georust::geo
Euclidean.interpolate_point_at_ratio_from_start(line, 0.8)

// also:
Haversine.interpolate_point_at_ratio_from_start(line, 0.8)
Geodesic.interpolate_point_at_ratio_from_start(line, 0.8)
Rhumb.interpolate_point_at_ratio_from_start(line, 0.8)
```

As a bonus, I also wanted to add some (IMO) long missing niceties to the api.
```rust
// geos communicates "from end" with a negative distance
line_interpolate_point(line, -0.8, true)
// georust::geo
Euclidean.interpolate_point_at_ratio_from_end(line, 0.8)
```
```rust
// geos communicates "distance" rather than "ratio" by passing in `false`
line_interpolate_point(line, 0.8, false)
// georust::geo
Euclidean.interpolate_point_at_distance_from_start(line, 0.8)
```
```rust
// geos
line_interpolate_point(line, -0.8, false)
// georust::geo
Euclidean.interpolate_point_at_distance_from_end(line, 0.8)
```

You can use all those methods across metric spaces:
```rust
Haversine.interpolate_point_at_ratio_from_end(line, 0.8)
Geodesic.interpolate_point_at_distance_from_start(line, 0.8)
Rhumb.interpolate_point_at_distance_from_end(line, 0.8)
// etc.
```

The names are admittedly long. It's my own preference to have verbose method names over geos-like "mode" parameters.

## Bonus

While working on this, and filling out the deprecations, I found and fixed a bug in `LineStringSegmentizeHaversine` which was using the legacy `LineInterpolatePoint`, which was Euclidean only. It's easy to make this kind of mistake, the way we've historically not been very explicit about which metric space an algorithm is for.

### DIFF
--- a/geo-types/src/geometry/line_string.rs
+++ b/geo-types/src/geometry/line_string.rs
@@ -226,30 +226,23 @@ impl<T: CoordNum> LineString<T> {
         self.0
     }
 
-    /// Return an iterator yielding one [Line] for each line segment
+    /// Return an iterator yielding one [`Line`] for each line segment
     /// in the [`LineString`].
     ///
     /// # Examples
     ///
     /// ```
-    /// use geo_types::{coord, Line, LineString};
+    /// use geo_types::{wkt, Line, LineString};
     ///
-    /// let mut coords = vec![(0., 0.), (5., 0.), (7., 9.)];
-    /// let line_string: LineString<f32> = coords.into_iter().collect();
-    ///
+    /// let line_string = wkt!(LINESTRING(0 0,5 0,7 9));
     /// let mut lines = line_string.lines();
+    ///
     /// assert_eq!(
-    ///     Some(Line::new(
-    ///         coord! { x: 0., y: 0. },
-    ///         coord! { x: 5., y: 0. }
-    ///     )),
+    ///     Some(Line::new((0, 0), (5, 0))),
     ///     lines.next()
     /// );
     /// assert_eq!(
-    ///     Some(Line::new(
-    ///         coord! { x: 5., y: 0. },
-    ///         coord! { x: 7., y: 9. }
-    ///     )),
+    ///     Some(Line::new((5, 0), (7, 9))),
     ///     lines.next()
     /// );
     /// assert!(lines.next().is_none());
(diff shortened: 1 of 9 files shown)


michaelkirk:
This is a big PR, but it's overwhelmingly docs and tests, I swear!

geo-types/src/geometry/line_string.rs
```diff
@@ -261,6 +254,37 @@ impl<T: CoordNum> LineString<T> {
...
+    ///     lines.next()
+    /// );
+    /// assert_eq!(
+    ///     Some(Line::new((5, 0), (0, 0))),
+    ///     lines.next()
+    /// );
+    /// assert!(lines.next().is_none());
+    /// ```
+    pub fn rev_lines(&'_ self) -> impl ExactSizeIterator<Item = Line<T>> + '_ {
+        self.0.windows(2).rev().map(|w| {
```
michaelkirk, line 282: At first I assumed I wanted to do `line_string.lines().rev()` but that's not quite what I needed.

Given points: `a, b, c`

```
lines():
1. a->b
2. b->c
```

lines().rev():
```
1. b->c
2. a->b
```

But what I want is:
```
rev_lines()
1. c->b
2. b->a
```

michaelkirk:
I've just pushed up a bike-shedding commit. 

There are two equivalent new APIs:

```rust
Euclidean.point_at_ratio_from_start(&line_string, 0.5)
line_string.point_at_ratio_from_start(&Euclidean, 0.5)
```

Previously, the deprecation pointed you to the first version. I changed that deprecation to now recommend the latter. 

I don't think it matters much, but the second version is closer to the existing API, so I think it'll be slightly easier to migrate to.

geo/src/algorithm/linestring_segment.rs
```diff
@@ -85,9 +84,8 @@ macro_rules! implement_segmentize {
 
                     if (cum_length >= segment_length) && (i != (n_lines - 1)) {
                         let remainder = cum_length - segment_length;
-                        let endpoint =
-                            segment.line_interpolate_point((length - remainder) / length)?;
-
+                        let endpoint = $metric_space
+                            .point_at_ratio_from_start(&segment, (length - remainder) / length);
```
michaelkirk, line 88: This is a bugfix in the case of Haversine

geo/src/lib.rs
```diff
@@ -49,6 +49,7 @@
 //! - **[`Bearing`]**: Calculate the bearing between two points.
 //!
 //! - **[`Destination`]**: Calculate the destination point from an origin point, given a bearing and a distance.
+//! - **[`InterpolateLine`]**: Interpolate a `Point` along a `Line` or `LineString`.
 //! - **[`InterpolatePoint`]**: Interpolate points along a line.
```
dabreegster, line 53: I've looked at https://docs.rs/geo/latest/geo/algorithm/line_measures/trait.InterpolatePoint.html and this PR a few times, but I'm still fuzzy on the difference -- is the old trait deprecated?

geo/src/algorithm/line_measures/interpolate_line.rs
```diff
@@ -0,0 +1,783 @@
...
+
+impl<F: CoordFloat> InterpolatableLine<F> for LineString<F> {
+    type Output = Option<Point<F>>;
+
+    fn point_at_ratio_from_start<MetricSpace: InterpolatePoint<F> + Length<F>>(
+        &self,
+        metric_space: &MetricSpace,
+        ratio: F,
+    ) -> Self::Output {
+        let distance = ratio * metric_space.length(self);
```
dabreegster, line 345: The < start or > end case is handled by the other method, all good

dabreegster (approved):
All of this looks clear to me. The API choices are very explicit and nice, better than geos. There's lots of repetition in docs, but I don't see any copy/paste errors in them

geo/src/lib.rs
michaelkirk: I deprecated it in this PR: https://github.com/georust/geo/pull/1321/files#diff-1e6b7ec38d282e44d7225459b81fa17d1ca9b7529763d44d0550c303077f6826R9

But your comment made me realize I should remove the deprecated one from these front page docs!

PR #1322, opened 2025-03-04
Fix Relate for Point/MultiPoint (broken just yesterday)

michaelkirk:
---

I broke this in https://github.com/georust/geo/commit/44816ebf6e874a6995d6f3f57c887c16d15e074e (merged yesterday, not yet released) by attempting to reuse the bbox from the geometry graph's RTree.

The problem was that the RTree only contains the Geometry's _segments_, but Point and MultiPoint have no segments - they're zero-dimensional geometries.

I don't often use `Relate` with zero dimensional geometries, so it's only luck that an unrelated example I was writing triggered an assert!

To be clear, we're still getting *most* of the perf benefit from #1317 - we're still caching the bbox. Now we have to compute it 1 time. Previously we were trying to compute it zero times by re-using the one from the RTree.

== PERF

There was only a small loss from this change, indicating it was probably a bad
idea from the start.

    build prepared polygons time:   [4.5677 ms 4.5717 ms 4.5764 ms]
                            change: [+1.1094% +1.2751% +1.4241%] (p = 0.00 < 0.05)
                            Performance has regressed.

    relate already prepared polygons
                            time:   [18.889 ms 18.912 ms 18.942 ms]
                            change: [+0.5218% +0.6533% +0.8127%] (p = 0.00 < 0.05)

    build and relate prepared polygons
                            time:   [23.598 ms 23.632 ms 23.671 ms]
                            change: [+0.4791% +0.6917% +0.9029%] (p = 0.00 < 0.05)

    unprepared polygons/relate unprepared polygons
                            time:   [744.51 ms 749.08 ms 754.49 ms]
                            change: [-1.2051% -0.1460% +0.8681%] (p = 0.80 > 0.05)

### DIFF
--- a/geo/src/algorithm/relate/geomgraph/index/prepared_geometry.rs
+++ b/geo/src/algorithm/relate/geomgraph/index/prepared_geometry.rs
@@ -41,19 +41,8 @@ where
 {
     let mut geometry_graph = GeometryGraph::new(0, geometry.clone().into());
     let r_tree = geometry_graph.build_tree();
-    let envelope = r_tree.root().envelope();
-
-    // geo and rstar have different conventions for how to represet an empty bounding boxes
-    let bounding_rect: Option<Rect<F>> = if envelope == rstar::AABB::new_empty() {
-        None
-    } else {
-        Some(Rect::new(envelope.lower(), envelope.upper()))
-    };
-    // They should be equal - but we can save the computation in the `--release` case
-    // by using the bounding_rect from the RTree
-    debug_assert_eq!(bounding_rect, geometry_graph.geometry().bounding_rect());
-
     geometry_graph.set_tree(Rc::new(r_tree));
+    let bounding_rect = geometry_graph.geometry().bounding_rect();
 
     // TODO: don't pass in line intersector here - in theory we'll want pluggable line intersectors
     // and the type (Robust) shouldn't be hard coded here.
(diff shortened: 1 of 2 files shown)


geo/src/algorithm/relate/geomgraph/index/prepared_geometry.rs
```diff
@@ -41,19 +41,8 @@ where
...
-        None
-    } else {
-        Some(Rect::new(envelope.lower(), envelope.upper()))
-    };
-    // They should be equal - but we can save the computation in the `--release` case
-    // by using the bounding_rect from the RTree
-    debug_assert_eq!(bounding_rect, geometry_graph.geometry().bounding_rect());
-
     geometry_graph.set_tree(Rc::new(r_tree));
+    let bounding_rect = geometry_graph.geometry().bounding_rect();
```
michaelkirk, line 45: We compute the bounding_rect from scratch, and cache that for use in `Relate`, rather than try to reuse it from the RTree.

geo/src/algorithm/relate/geomgraph/index/prepared_geometry.rs
```diff
@@ -41,19 +41,8 @@ where
...
-
-    // geo and rstar have different conventions for how to represet an empty bounding boxes
-    let bounding_rect: Option<Rect<F>> = if envelope == rstar::AABB::new_empty() {
-        None
-    } else {
-        Some(Rect::new(envelope.lower(), envelope.upper()))
-    };
-    // They should be equal - but we can save the computation in the `--release` case
-    // by using the bounding_rect from the RTree
-    debug_assert_eq!(bounding_rect, geometry_graph.geometry().bounding_rect());
```
michaelkirk, line 54: This is why I love asserts - they protect me from myself.

PR #1328, opened 2025-03-16
Bump MSRV to 1.81

urschrei:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/main/CODE_OF_CONDUCT.md).
- [x] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

PR #1329, opened 2025-03-17
Update proj dependency to 0.29.0

urschrei:
---

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Bump `geo` MSRV to 1.81
+- BREAKING: Update proj dependency to 0.29.0
 - BREAKING: `FrechetDistance` is now defined on the metric space, rather than a method on a Linestring.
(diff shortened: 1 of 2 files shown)


urschrei:
Blocked on #1328

geo/CHANGES.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Update proj dependency to 0.29.0
```
michaelkirk, line 5:
```suggestion
- BREAKING: Update proj dependency to 0.29.0
```

I think we re-export a bunch of proj symbols, so technically breaking, right?

(We already have a bunch of breaking changes in this release, so no big deal)

michaelkirk: BTW, I like to do this as a way of accounting when doing the release if it needs to be a minor or major release. 

In this case it doesn't matter because we already have a bunch of other breaking changes, but I feel it's good hygiene, if not a bit pedantic. 🤓

urschrei: Ah, good catch!

PR #1330, opened 2025-03-18
Fix a math error in Rhumb distance calculations

apendleton:
---

When I implemented Rhumb geometry operations in #1090 , it turns out I made a mistake in porting over the formula for distance. The Rhumb distance operation special-cases east-west lines (there's a `tan` in the calculations for all other lines whose value is undefined for east-west lines), and the check to test if we're on an east-west line was missing an absolute-value in my implementation, which meant some non-EW lines were getting calculated with the EW-line formula, which produces slightly incorrect results for some, but not all, rhumb lines. See for reference [this line](https://github.com/Turfjs/turf/blob/212d1664368db93361f29f626172412e0babf87c/packages/turf-rhumb-distance/index.ts#L95) in Turf, which does the same check and does include the required `abs`.

This change also required updating some tests. I've confirmed against the Turf implementation that our values are now correct. I had noticed that they were slightly different from Turf's in the first pass, but incorrectly attributed the difference to a different earth-radius constant, but we're now pretty much bang-on. In hindsight, I probably should have noticed that in one of these test cases, we were saying a rhumb distance between two points was shorter than the haversine distance between those same points, which ought to have been a red flag. Also note that the analogous check in the `calculate_destination` function _did_ already have the `abs` as expected.

### DIFF
--- a/geo/src/algorithm/line_measures/length.rs
+++ b/geo/src/algorithm/line_measures/length.rs
@@ -102,7 +102,7 @@ mod tests {
             Geodesic.length(&line).round()
         );
         assert_eq!(
-            341_088., // meters
+            343_572., // meters
             Rhumb.length(&line).round()
         );
         assert_eq!(
@@ -136,7 +136,7 @@ mod tests {
             Geodesic.length(&line_string).round()
         );
         assert_eq!(
-            6_332_790., // meters
+            6_308_683., // meters
             Rhumb.length(&line_string).round()
         );
         assert_eq!(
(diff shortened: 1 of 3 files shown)


apendleton:
It seems like there's a CI failure unrelated to this change (the dependency `half` seems to have an MSRV greater than the current Rust version being run on CI). Let me know how we want to handle that.

michaelkirk:
I just merged #1328 - can you try to rebase or merge with main?

michaelkirk:
I've gone ahead and updated your branch with latest main myself.

apendleton:
Looks like you beat me to it :sweat_smile: . Thanks for the quick review!

Issue #1320, opened 2025-03-03
LineLocatePoint in the Euclidean space spends a lot of time calculating hypot


AndrewLipscomb:
![Image](https://github.com/user-attachments/assets/c6b892e9-3632-4596-9838-ebb303c9cbf7)

`LineLocatePoint` appears to spend a lot of time calculating a hypotenuse. I believe some of these calculations are unnecessary. 

At the very least https://github.com/georust/geo/blob/70b1c1f38d4efc36f47635a7aa4d32286bdbc053/geo/src/algorithm/line_locate_point.rs#L101 appears to only be used for "shortest seen so far", that could be the cheaper non-sqrt'ed distance. 

I believe the other values will all need to be sqrt'ed - but given there is one potential improvement in there, maybe worth a double check from the more knowledgable guys here.

michaelkirk:
That seems plausible. Is this something you're interested in working on?

michaelkirk:
I don't think we yet have something like `comparable_distance`. 

I agree it looks plausible, but I wouldn't recommend `rstar::PointDistance` as the basis for this feature. The existing implementations should be correct, but it's not implemented for all geometry types.

We've considered implementing it for all geometries, but are kind of at an impasse: https://github.com/georust/geo/pull/984 has some further discussion

I'd recommend something more targeted, even if it's slightly redundant with some existing trait definitions.

>  I have no idea what a "comparable distance" would be for the other MetricSpace's 

I don't know that you have to address this - LineLocatePoint is (currently anyway) strictly Euclidean. If you can speed up the Euclidean calculation (and only the euclidean calculation) that's still valuable.

AndrewLipscomb:
Gonna have a crack at this today. For what its worth, my last couple of weeks have had me reaching for a trait doing something like the Boost `comparable_distance` functions - so it seems like any work I do here could hit two things. 

I'll roll both into a PR and get some feedback

PR #1334, opened 2025-03-22
Optimise line locate point

AndrewLipscomb:
- [*] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/main/CODE_OF_CONDUCT.md).
---

See https://github.com/georust/geo/issues/1320

Two things here - first is optimising the LineLocatePoint algorithm. 

A new bench shows this was the original timing
```
LineString line_locate_point
                        time:   [259.34 µs 259.57 µs 259.83 µs]
                        change: [-1.6333% -1.0309% -0.5978%] (p = 0.00 < 0.05)
  5 (5.00%) high mild
  1 (1.00%) high severe
```

After updates
```
LineString line_locate_point
                        time:   [73.201 µs 73.534 µs 74.106 µs]
                        change: [-71.777% -71.684% -71.525%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
  1 (1.00%) high severe
```

Great success, very nice. 

I've put some of this work out publicly under a public `ComparableDistance` trait. However I don't think that's necessarily the best idea. 

I started off just working off of Boost Geometry's `comparable_distance` idea - but I think given
 - This will be at least strictly Euclidean to start, which implies all "comparable distances" will just be the non-sqrt'ed versions which is kind of handy to know 
 - I have no idea what a `ComparableDistance` trait would look like for non-Euclidean spaces 
I think it would be better exposed as an explicit `DistanceSquared` trait that only `Euclidean` implements - if I am writing my own algo's I'd like to have it reliably convertible to a real distance. 

Keen for some feedback here - I still have some clean up here before I make this as ready to go.

### DIFF
--- a/geo/src/algorithm/line_measures/comparable_distance.rs
+++ b/geo/src/algorithm/line_measures/comparable_distance.rs
@@ -1,0 +1,34 @@
+/// Calculate a minimum distance between two geometries in a way that is useful for sorting operations
+pub trait ComparableDistance<F, Origin, Destination> {
+    /// This trait differs from [Distance](geo::Distance) in that the value returned is not a true distance 
+    /// but a value that is indicative of the true distance that can be used for sorting. It is generally
+    /// faster to compute this value rather than true distance if all you need is a comparable figure between two 
+    /// geometries 
+    /// 
+    /// Note that not all implementations support all geometry combinations, but at least `Point` to `Point`
+    /// is supported.
+    /// See [specific implementations](#implementors) for details.
+    ///
+    /// # Units
+    ///
+    /// - `origin`, `destination`: geometry where the units of x/y depend on the trait implementation.
+    /// - returns: depends on the trait implementation.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use geo::{Euclidean, ComparableDistance, Point};
+    /// let p1: Point = Point::new(0.0, 0.0);
+    /// let p2: Point = Point::new(0.0, 2.0);
+    /// let p3: Point = Point::new(0.0, 5.0);
+    /// 
+    /// let comparable_1_2 = Euclidean.comparable_distance(p1, p2);
+    /// let comparable_1_3 = Euclidean.comparable_distance(p1, p3);
+    ///
+    /// assert_eq!(comparable_1_2, 4.0);
+    /// assert_eq!(comparable_1_3, 25.0);
+    /// assert_lt!(comparable_1_2, comparable_1_3);
+    ///
+    /// ```
+    fn comparable_distance(&self, origin: Origin, destination: Destination) -> F;
+}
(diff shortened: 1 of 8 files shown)


michaelkirk:
This seems great! No notes. 👏👏 👏 

> I think it would be better exposed as an explicit DistanceSquared trait that only Euclidean implements - if I am writing my own algo's I'd like to have it reliably convertible to a real distance.

That makes sense to me. Though I'm also not super bothered by having a ComparableDistance trait which is currently only implemented for Euclidean.

AndrewLipscomb:
Just FYI - I've gotten bogged down with work this week. Still aiming to fix this up but will need a weekend to do it

urschrei:
No rush whatsoever!

geo/src/algorithm/line_measures/comparable_distance.rs
```diff
@@ -0,0 +1,34 @@
+/// Calculate a minimum distance between two geometries in a way that is useful for sorting operations
+pub trait ComparableDistance<F, Origin, Destination> {
+    /// This trait differs from [Distance](geo::Distance) in that the value returned is not a true distance 
+    /// but a value that is indicative of the true distance that can be used for sorting. It is generally
+    /// faster to compute this value rather than true distance if all you need is a comparable figure between two 
+    /// geometries 
+    /// 
+    /// Note that not all implementations support all geometry combinations, but at least `Point` to `Point`
```
michaelkirk, line 8: This comment seems spurious (maybe it was copied from the `Distance` trait which was implemented on more than just the Euclidean metric space).

michaelkirk (changes requested):
This seems great and results in a nice speedup. We should merge it soon. 

I know you've expressed a concern about the naming of the trait - I don't particularly share that concern. I'm fine either way. 

If you have time to make those changes soon, I'll re-review, otherwise l'll fix up the docs and merge in a couple days.

bennjii:
I've made fixes for the CI errors in https://github.com/AndrewLipscomb/geo/pull/1 if that helps :)

geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -27,13 +27,23 @@ macro_rules! symmetric_distance_impl {
 
 impl<F: CoordFloat> Distance<F, Coord<F>, Coord<F>> for Euclidean {
     fn distance(&self, origin: Coord<F>, destination: Coord<F>) -> F {
+        Self.comparable_distance(origin, destination).sqrt()
```
michaelkirk, line 30: Usually this is the same, and it's DRY to use our new comparable distance method here, but the old impl, using `hypot` should be more stable.

PR #1340, opened 2025-03-31
Don't borrow epsilon in the Simplify* traits

### LINKED ISSUE #1339: API ergonomics: Simplify should take an owned, not borrowed float:
Not a big deal, but it's weird to need to borrow a float like this: `line_string.simplify(&0.5)`

urschrei:
---

Fixes #1339

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -1,5 +1,9 @@
 # Changes
 
+# Unreleased
+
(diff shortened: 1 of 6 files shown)


dabreegster (approved):
Looks like a simple refactor to me. I dunno if there are any special steps to making breaking API changes, besides the note in the changelog?

urschrei:
…to which I should add a "BREAKING" annotation. But generally no, we don't do anything else.

PR #1343, opened 2025-04-10
Proof of concept: Implement `CoordsIter` based on `geo-traits` traits

frewsxcv:
---

### DIFF
--- a/geo/src/algorithm/coords_iter.rs
+++ b/geo/src/algorithm/coords_iter.rs
@@ -1,21 +1,25 @@
-use std::fmt::Debug;
+use crate::{coord, Coord, CoordNum};
+use geo_traits::{
+    to_geo::ToGeoCoord, CoordTrait, GeometryCollectionTrait, GeometryTrait, GeometryType,
+    LineStringTrait, LineTrait, MultiLineStringTrait, MultiPointTrait, MultiPolygonTrait,
+    PointTrait, PolygonTrait, RectTrait, TriangleTrait,
+};
+use std::marker::PhantomData; // Added import
 
-use crate::geometry::*;
-use crate::{coord, CoordNum};
-
-use std::{fmt, iter, marker, slice};
-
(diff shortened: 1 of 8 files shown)


frewsxcv:
cc @kylebarron Maybe relevant to your interests

geo/src/algorithm/coords_iter.rs
```diff
@@ -95,476 +99,830 @@ pub trait CoordsIter {
...
-        = Self::Iter<'a>
-    where
-        T: 'a;
-    type Scalar = T;
-
-    fn coords_iter(&self) -> Self::Iter<'_> {
-        iter::once(self.start).chain(iter::once(self.end))
-    }
+/// Helper iterator for Line coordinates
+pub struct LineIter<'a, L: LineTrait + 'a>
```
frewsxcv, line 235: The implementations of the iterators for each geometry type are much more...manual...than the old ones. I couldn't get lifetimes to behave.

geo/src/algorithm/coords_iter.rs
```diff
@@ -95,476 +99,830 @@ pub trait CoordsIter {
...
     where
-        T: 'a;
-    type Scalar = T;
+        Self: Sized + 'a,
+        Self::T: 'a + CoordNum;
 
     fn coords_iter(&self) -> Self::Iter<'_> {
-        iter::once(self.0)
+        match self.as_type() {
+            GeometryType::Point(p) => Iter::Point(Some(PointTrait::coord(p).unwrap().to_coord())),
```
kylebarron, line 141: This begs the question of how `geo` should handle empty points.

kylebarron (commented):
This is really exciting

geo/src/algorithm/coords_iter.rs
frewsxcv: I was intending to return an empty iterator here. Let me know if that doesn't sound right

PR #1346, opened 2025-04-20
Update traits in `geo-traits` to extend `GeometryTrait`

frewsxcv:
---

By doing this, we'll be able to write algorithms generically for `geo-traits::Geometry` and have the algorithm directly called on geometry types, akin to:

```rust
trait Area {
    fn area(...);
}

impl<G: GeometryTrait> Area for G { ... }

(geo_types::Polygon(...)).area();
```

Prior to this pull request, you would have had to wrap `geo_types::Polygon` in `geo_types::Geometry` before running the algorithm.

### DIFF
--- a/geo-traits/src/multi_line_string.rs
+++ b/geo-traits/src/multi_line_string.rs
@@ -2,11 +2,7 @@ use std::marker::PhantomData;
 
 use crate::iterator::MultiLineStringIterator;
 use crate::line_string::UnimplementedLineString;
-use crate::{
-    Dimensions, GeometryTrait, GeometryType, LineStringTrait, UnimplementedGeometryCollection,
-    UnimplementedLine, UnimplementedMultiPoint, UnimplementedMultiPolygon, UnimplementedPoint,
-    UnimplementedPolygon, UnimplementedRect, UnimplementedTriangle,
-};
+use crate::{GeometryTrait, LineStringTrait};
 #[cfg(feature = "geo-types")]
 use geo_types::{CoordNum, LineString, MultiLineString};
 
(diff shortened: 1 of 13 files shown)


kylebarron:
> Prior to this pull request, you would have had to wrap geo_types::Polygon in geo_types::Geometry before running the algorithm.

On my phone right now and haven't looked at this PR but no, that is not currently necessary because Polygon implements GeometryTrait 

https://docs.rs/geo-traits/latest/geo_traits/trait.GeometryTrait.html#impl-GeometryTrait-for-Polygon%3CT%3E

frewsxcv:
Wow. Big 'ol brain fart. @kylebarron Do you think there's anything worth scrapping from this PR? Is the inheritance worth it? It is nice that you only have to define `dim` and `T` once, since they're now on the parent.

geo-traits/src/triangle.rs
```diff
@@ -115,3 +99,69 @@ impl<T> TriangleTrait for UnimplementedTriangle<T> {
...
+        Self::MultiLineStringType<'_>,
+        Self::MultiPolygonType<'_>,
+        Self::GeometryCollectionType<'_>,
+        Self::RectType<'_>,
+        Self::TriangleType<'_>,
+        Self::LineType<'_>,
+    > {
+        GeometryType::Triangle(self)
+    }
+}
```
kylebarron, line 167: We should be able to put this in a macro similar to https://github.com/georust/geo/blob/750ed14a2f01c6a047e2ced76ab12261ef46ba5a/geo-traits/src/geometry.rs#L291-L438

Maybe in that same file?
https://github.com/georust/geo/blob/750ed14a2f01c6a047e2ced76ab12261ef46ba5a/geo-traits/src/geometry.rs#L440-L447

geo-traits/src/multi_line_string.rs
```diff
@@ -11,22 +11,16 @@ use geo_types::{CoordNum, LineString, MultiLineString};
...
 ///
 /// Refer to [geo_types::MultiLineString] for information about semantics and validity.
-pub trait MultiLineStringTrait: Sized {
-    /// The coordinate type of this geometry
-    type T;
-
+pub trait MultiLineStringTrait: Sized + GeometryTrait {
     /// The type of each underlying LineString, which implements [LineStringTrait]
-    type LineStringType<'a>: 'a + LineStringTrait<T = Self::T>
+    type InnerLineStringType<'a>: 'a + LineStringTrait<T = Self::T>
```
kylebarron, line 16: Why this change? Is this just a rename?

geo-traits/src/polygon.rs
```diff
@@ -1,8 +1,9 @@
 use std::marker::PhantomData;
 
+use crate::geometry::GeometryTrait;
 use crate::iterator::PolygonInteriorIterator;
 use crate::line_string::UnimplementedLineString;
-use crate::{Dimensions, LineStringTrait};
+use crate::LineStringTrait;
```
kylebarron, line 6: Nit: combine these two as 
```rs
use crate::{GeometryTrait, LineStringTrait};
```
?

geo-traits/src/triangle.rs
kylebarron: Pushed this change to this branch in [`f834667` (#1346)](https://github.com/georust/geo/pull/1346/commits/f8346671ca3c37b2534f65010f7e9b4b5a77256f), and then realized maybe I should double check that's ok etiquette? Not sure when it's frowned upon to push to another's branch. It did minimize the diff by a lot, but maybe I should've created another PR into this PR?

kylebarron (approved):
Back at my computer now and realized this is actually quite a minor change.

Previously it was _convention_ that anyone with a `Point` struct should implement both `PointTrait` _and_ `GeometryTrait` on their object. This change _enforces_ that both are implemented, which I think is great.

I think it's helpful to enforce this, because then we won't get questions down the line as to "why can't I pass this point into your function accepting a geometry", and it should be low-overhead for implementors because we have our `Unimplemented*` types.

geo-traits/src/triangle.rs
urschrei: I wouldn't say it's a question of etiquette, but a practical issue: it can surprisingly change the source branch (until you run `log` or whatever). A PR to the source branch is probably better (even though it always feels more finicky to me) because then the decision about what to do with the change is in the hands of the original author; they can pull it, comment on it, ask for changes themselves etc.

urschrei:
> Back at my computer now and realized this is actually quite a minor change.
> 
> Previously it was _convention_ that anyone with a `Point` struct should implement both `PointTrait` _and_ `GeometryTrait` on their object. This change _enforces_ that both are implemented, which I think is great.
> 
> I think it's helpful to enforce this, because then we won't get questions down the line as to "why can't I pass this point into your function accepting a geometry", and it should be low-overhead for implementors because we have our `Unimplemented*` types.

Agreed – I like this. It adds some up-front complexity, but you get a lot in return.

geo-traits/src/triangle.rs
kylebarron: That makes a lot of sense; sorry if I inconvenienced you @frewsxcv !

geo-traits/src/multi_line_string.rs
frewsxcv: Since all these traits now extend `GeometryTrait`, they all have the associated types on `GeometryTrait`, including `LineStringType`. So option 1 is to reuse `GeometryTrait::LineStringType` for the inner type on a multi line string (and we can remove the associated type here), or option 2 is to do what I did here which is rename this occurrence and leave the parent associated types untouched and unused. What do you think?

kylebarron: Oh I see, thanks for the explanation. I could imagine instances where the type for `LineString` is not the same as the inner type of a `MultiLineString`, so probably better to rename and keep them distinct?

frewsxcv: Yeah that's what I am thinking too

geo-traits/src/triangle.rs
frewsxcv: I do prefer getting consent from the PR author before pushing to their branches, but in this case I really didn't mind so thanks for the commit! I was wondering if a macro would've been more appropriate for that

PR #1348, opened 2025-04-22
Fix the lifetime annotation of PointTrait implemented for geo_types::Point

Kontinuation:
---

I found this lifetime problem when building layers of generic abstractions over geo-traits and geo-types. The fix makes the lifetime of `CoordType` of `Point` consistent with other types in geo-types.

### DIFF
--- a/geo-traits/src/point.rs
+++ b/geo-traits/src/point.rs
@@ -34,11 +34,11 @@ impl<T: CoordNum> PointTrait for Point<T> {
 }
 
 #[cfg(feature = "geo-types")]
-impl<T: CoordNum> PointTrait for &Point<T> {
-    type CoordType<'a>
+impl<'a, T: CoordNum> PointTrait for &'a Point<T> {
+    type CoordType<'b>
         = &'a Coord<<Self as GeometryTrait>::T>
     where
-        Self: 'a;
+        Self: 'b;
 
     fn coord(&self) -> Option<Self::CoordType<'_>> {
         Some(&self.0)
--- a/geo-traits/src/point.rs
+++ b/geo-traits/src/point.rs
@@ -61,3 +61,29 @@ impl<T> PointTrait for UnimplementedPoint<T> {
         unimplemented!()
     }
 }
+
+#[test]
+fn test_point_trait_lifetime() {
+    // This is a regression test for https://github.com/georust/geo/pull/1348
+
+    // let's say point has lifetime 'a
+    let point = Point::new(1.0, 2.0);
+    // as long as coord_ref references a part of point, it should be valid.
+    let coord_ref: &Coord<f64>;
+
+    {
+        // point_ref has lifetime 'b, it references a point which has lifetime 'a
+        let point_ref: &Point<f64> = &point;
+        {
+            let point_ref_ref: &&Point<f64> = &point_ref;
+            // the lifetime of coord it references should be 'a, since it references
+            // a inner part of point which has lifetime 'a
+            coord_ref = point_ref_ref.coord().unwrap();
+        }
+    }
+
+    // Using coord_ref here, it should be valid. Without https://github.com/georust/geo/pull/1348
+    // there would be a compile error here.
+    assert_eq!(coord_ref.x(), 1.0);
+    assert_eq!(coord_ref.y(), 2.0);
+}
--- a/geo-traits/CHANGES.md
+++ b/geo-traits/CHANGES.md
@@ -4,6 +4,8 @@
 
 - BREAKING: All traits now extend `GeometryTrait`.
   - <https://github.com/georust/geo/pull/1346>
+- Fix the lifetime annotation of `PointTrait` implemented for `geo_types::Point`.
+  - <https://github.com/georust/geo/pull/1348>
 
 ## 0.2.0 - 2024.11.06
 


kylebarron (approved):
Thank you! I don't understand lifetimes as well as I should so I'd love to get another's review on this as well.

michaelkirk:
Can you give an example of some code that wouldn't compile before this change which will compile after this change?

kylebarron:
This related PR https://github.com/georust/wkb/pull/59#pullrequestreview-2784245504 has an example for the WKB point object; not sure if it's relevant here too

frewsxcv (approved):

frewsxcv:
Yep a regression test would be great for this @Kontinuation if you have one.

I think I made this change locally last week to work around an issue with lifetimes but I think I blew away that branch sadly.

frewsxcv:
Also we should consider doing the same for the other geo-types implementations, assuming it's also applicable for them and not just `geo_types::Point`

geo-traits/src/point.rs
```diff
@@ -34,11 +34,11 @@ impl<T: CoordNum> PointTrait for Point<T> {
 }
 
 #[cfg(feature = "geo-types")]
-impl<T: CoordNum> PointTrait for &Point<T> {
-    type CoordType<'a>
+impl<'a, T: CoordNum> PointTrait for &'a Point<T> {
```
michaelkirk, line 37: I confess, I'm not great with lifetimes, but I feel like this is a no-op change?

urschrei: Well if we're all admitting that we're "not that great at lifetimes" I'll chime in and admit it too. I assume that this change is intended to make it clearer that the lifetime of the returned coordinate matches the lifetime of the source reference. But does "clearer" in this case actually mean "fixes a `borrowck` error"? Great if it does, and we'd love to see it – I thought the existing lifetime arrangement already made the relationship clear.

Kontinuation:
> Yep a regression test would be great for this @Kontinuation if you have one.
> 
> I think I made this change locally last week to work around an issue with lifetimes but I think I blew away that branch sadly.

I have added a regression test in https://github.com/georust/geo/pull/1348/commits/8efd199d6f2f782c72a8abb7a7961d7bbb18f1a3, which is a minimal reproducer we can derive from the overly restrictive lifetime annotation. This is not exactly the problem I initially encountered though (it was a trait bound about lifetime constraint), but I think it should be good enough.

Without the fix, the newly added test won't compile:

```
error[E0597]: `point_ref` does not live long enough
  --> geo-traits/src/point.rs:61:47
   |
59 |         let point_ref: &Point<f64> = &point;
   |             --------- binding `point_ref` declared here
60 |         {
61 |             let point_ref_ref: &&Point<f64> = &point_ref;
   |                                               ^^^^^^^^^^ borrowed value does not live long enough
...
66 |     }
   |     - `point_ref` dropped here while still borrowed
...
70 |     assert_eq!(coord_ref.x(), 1.0);
   |                --------- borrow later used here
```

frewsxcv:
@Kontinuation How do you feel about making this change for the other geometry types and not just point? I can do it in a follow-up if you want, or just push to your branch directly

Kontinuation:
> @Kontinuation How do you feel about making this change for the other geometry types and not just point? I can do it in a follow-up if you want, or just push to your branch directly

The lifetime annotations for all other types seems to be correct, do you want me to add tests for all other types?

Another thing amuses me is that the `CoordType` of `Rect` is an owned type instead of reference type, I'm not sure if it is by design.

kylebarron:
> Another thing amuses me is that the `CoordType` of `Rect` is an owned type instead of reference type, I'm not sure if it is by design.

`Rect`, `LineString`, and `Point` all have owned coord types with a lifetime parameter
https://github.com/georust/geo/blob/b0d02de7814ea4ca48312bad4b7e8d11040cce8f/geo-traits/src/point.rs#L12-L16
https://github.com/georust/geo/blob/b0d02de7814ea4ca48312bad4b7e8d11040cce8f/geo-traits/src/line_string.rs#L15-L18
https://github.com/georust/geo/blob/b0d02de7814ea4ca48312bad4b7e8d11040cce8f/geo-traits/src/rect.rs#L13-L16

Or do you mean this distinction
https://github.com/georust/geo/blob/b0d02de7814ea4ca48312bad4b7e8d11040cce8f/geo-traits/src/rect.rs#L27-L28
https://github.com/georust/geo/blob/b0d02de7814ea4ca48312bad4b7e8d11040cce8f/geo-traits/src/point.rs#L26-L27

Given that a `Coord` is `Copy`, indeed it seems better to not have a reference.

geo-traits/src/point.rs
```diff
@@ -34,11 +34,11 @@ impl<T: CoordNum> PointTrait for Point<T> {
...
+    type CoordType<'b>
         = &'a Coord<<Self as GeometryTrait>::T>
```
michaelkirk, line 39: Ah! This is the line that I missed. The `&'a` here used to refer to the `CoordType<'a>`, but now references the lifetime of the `impl &'a Point`

I see now how this is now more relaxed. Thanks for the test!

michaelkirk (approved):

PR #1350, opened 2025-04-24
Simplify tested Rust and proj version spec

urschrei:
---

Draft as it's a slightly speculative, but this reduces the number of places we have to define our testable rustc and proj versions to 1, at the expense of some more complex variable shuffling in the form of the `rust-matrix` test, which only exists to make the correct variables available to all other workflows.

Also took the opportunity to update our checkout actions versions, and add `geo-traits` to the success/failure matrices.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,21 @@ on:
   schedule: [cron: "45 6 * * *"]
 
 name: Run tests
+
 jobs:
+  # Set configurations used throughout the workflow
+  set-matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -23,6 +37,7 @@ jobs:
       - lint
       - geo_types
       - geo
+      - geo_traits
```
michaelkirk, line 40: good catch!

michaelkirk (approved):
Looks great! Thanks for DRYing this up.

PR #1351, opened 2025-04-25
Return owned coords for PointTrait and LineStringTrait `geo-types` impls

kylebarron:
---

### DIFF
--- a/geo-traits/src/line_string.rs
+++ b/geo-traits/src/line_string.rs
@@ -47,7 +47,7 @@ pub trait LineStringTrait: Sized + GeometryTrait {
 #[cfg(feature = "geo-types")]
 impl<T: CoordNum> LineStringTrait for LineString<T> {
     type CoordType<'a>
-        = &'a Coord<Self::T>
+        = Coord<Self::T>
     where
         Self: 'a;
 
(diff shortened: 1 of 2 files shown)


frewsxcv (approved):

frewsxcv:
I broke the unwritten rules and pushed to your branch a conflict resolution 🤪

frewsxcv:
Cool I broke it I can fix it. Feel free to remove my commits

frewsxcv:
Ah right it's the test that was just added. Do we still need the `test_point_trait_lifetime` if there's not a lifetime for the coord?

geo-traits/CHANGES.md
```diff
@@ -4,8 +4,8 @@
 
 - BREAKING: All traits now extend `GeometryTrait`.
   - <https://github.com/georust/geo/pull/1346>
-- Fix the lifetime annotation of `PointTrait` implemented for `geo_types::Point`.
```
frewsxcv, line 7: I don't think we need this changelog entry anymore since this pull request undoes those changes?

kylebarron:
@KontinuatiKon can you review this?

Kontinuation (approved):
I have not benchmarked this (especially on LineStringTrait) but think it should be OK. It is not a breaking change so we can revise it if we find any performance problems in the future.

kylebarron:
Thank you!

kylebarron:
I don't think I realized when reviewing https://github.com/georust/geo/pull/1348 that this PR was directly contrary to that PR.

PR #1353, opened 2025-04-29
document x/y lon/lat conventions

### LINKED ISSUE #1352: ~~Haversine distance incorrect?~~ lon/lat ordering mixup
Hello I think I have a bug?

I'm looking at shipping distances and I have this code

```
let dist: f64;

let p1: geo::Point = geo::Point::new(-1.5668581211956627, -48.76584589975404);
let p2: geo::Point = geo::Point::new(64.04653073512222, -22.040472086697353);

dist = geo::Haversine.distance(p1, p2);

println!("{:?}", dist);
```

The distance I get is 6415 km
On gooogle maps it's about 7620 km
And on https://www.vcalc.com/wiki/vcalc/haversine-distance it's also about 7620 km

Now I tried making my own function for the haversine formula

```
pub fn haversine_dist_test(p1: geo::Point, p2: geo::Point) -> f64 {
    // Haversine formula
    let lat1 = p1.x().to_radians();
    let lon1 = p1.y().to_radians();
    let lat2 = p2.x().to_radians();
    let lon2 = p2.y().to_radians();

    let dlat = lat2 - lat1;
    let dlon = lon2 - lon1;

    let a = (dlat / 2.0).sin().powi(2) + lat1.cos() * lat2.cos() * (dlon / 2.0).sin().powi(2);
    let c = 2.0 * a.sqrt().asin();

    // Radius of the Earth in kilometers
    const R: f64 = 6371.0;

    // Calculate the distance
    let dist: f64 = R * c;
    return dist
}
```

That one yielded also around 7620km.

I made sure to  check the radius of the earth with `println!("My craycray radius is {:?}", geo::Haversine.radius());` and made sure the radius of the earth I was using in the code was correct, that yielded a distance of around 7628 km which is still significantly close thans 6415km from the built in formula.
What am I doing wrong? Is this a bug?

michaelkirk:
---

FIXES #1352

### DIFF
--- a/geo-types/src/geometry/coord.rs
+++ b/geo-types/src/geometry/coord.rs
@@ -23,7 +23,11 @@ use crate::{coord, CoordNum, Point};
 #[derive(Eq, PartialEq, Clone, Copy, Hash, Default)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct Coord<T: CoordNum = f64> {
+    /// Typically, `x` is the horizontal position, or longitude for geographic coordinates,
+    /// but its interpretation can vary across coordinate systems.
     pub x: T,
+    /// Typically, `y` is the vertical position, or latitude for geographic coordinates,
+    /// but its interpretation can vary across coordinate systems.
     pub y: T,
 }
 
--- a/geo-types/src/geometry/point.rs
+++ b/geo-types/src/geometry/point.rs
@@ -4,25 +4,45 @@ use core::ops::{Add, AddAssign, Div, DivAssign, Mul, MulAssign, Neg, Sub, SubAss
 
 /// A single point in 2D space.
 ///
-/// Points can be created using the [`Point::new`] constructor,
-/// the [`point!`] macro, or from a `Coord`, two-element
-/// tuples, or arrays – see the `From` impl section for a
-/// complete list.
-///
 /// # Semantics
 ///
 /// The _interior_ of the point is itself (a singleton set),
 /// and its _boundary_ is empty. A point is _valid_ if and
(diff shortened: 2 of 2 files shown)


michaelkirk:
That seems like a good idea. TBH I don't personally want to carve out the time to document every single geographic method with this caveat any time soon.

michaelkirk:
> we might also want to mention ESPG:4326 in the Point docs.

What specifically do you think we should mention?

lnicola:
That lon/lat is a very popular order, with the notable exception of EPSG:4326, which uses lat/lon.

Regarding the geographic functions, I think we should document it, especially since we say that `Point` is usually lon/lat. What it usually is has no relevance on what those functions might expect.

urschrei:
> That lon/lat is a very popular order, with the notable exception of EPSG:4326, which uses lat/lon.
> 
> 
> 
> Regarding the geographic functions, I think we should document it, especially since we say that `Point` is usually lon/lat. What it usually is has no relevance to what those functions might expect.

But our geographic functions also expect / assume lon, lat, right? We don't accept Lat, Lon anywhere in the geo ecosystem. We even expect lon lat for proj conversions from 4326.

lnicola:
They do, but how would the users know that?

lnicola:
Ideally yes, but a line in the `Point` docs saying that the geographic functions functions in this crate expect lon/lat would be a lot better than nothing.

michaelkirk:
Ok, good idea. I've added a line about the 4326 caveat and another line about how `geo` expects lon/lat.

urschrei (approved):

lnicola (approved):

PR #1357, opened 2025-05-08
perf(line_measures): optimise Fréchet distance to `O(min(m, n))` memory

grevgeny:
---
Replaces the full `m × n` matrix with a single two-row scratch buffer (2 × `min(m, n)` floats).
Peak allocation drops from “quadratic” to “linear” in the shorter line, saving orders of magnitude of RAM on large tracks.

### DIFF
--- a/geo/src/algorithm/line_measures/frechet_distance.rs
+++ b/geo/src/algorithm/line_measures/frechet_distance.rs
@@ -52,47 +52,54 @@ where
     MetricSpace: Distance<F, Point<F>, Point<F>>,
 {
     fn frechet_distance(&self, ls_1: &LineString<F>, ls_2: &LineString<F>) -> F {
-        if ls_1.coords_count() != 0 && ls_2.coords_count() != 0 {
-            Data {
-                cache: vec![F::zero(); ls_1.coords_count() * ls_2.coords_count()],
-                ls_a: ls_1,
-                ls_b: ls_2,
-            }
-            .compute_linear(self)
-        } else {
-            F::zero()
+        if ls_1.coords_count() == 0 || ls_2.coords_count() == 0 {
+            return F::zero();
         }
+
+        let (ls_short, ls_long) = if ls_1.coords_count() <= ls_2.coords_count() {
+            (ls_1, ls_2)
+        } else {
+            (ls_2, ls_1)
+        };
+
+        Data {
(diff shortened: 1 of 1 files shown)


urschrei:
Nice! As a matter of interest, how did you come up with the improvement?

grevgeny:
I encountered memory issues when using `FrechetDistance` with tens of thousands of points. So, I started looking for ways to optimize it and discovered that maintaining just two rows for calculations was sufficient 🙂

geo/src/algorithm/line_measures/frechet_distance.rs
```diff
@@ -52,47 +52,54 @@ where
...
+
+        let (ls_short, ls_long) = if ls_1.coords_count() <= ls_2.coords_count() {
+            (ls_1, ls_2)
         } else {
-            F::zero()
+            (ls_2, ls_1)
+        };
+
+        Data {
+            cache: vec![F::zero(); ls_short.coords_count() + ls_short.coords_count()],
```
lnicola, line 66: Maybe this, so it doesn't look like a type for `ls_long`?

```suggestion
            cache: vec![F::zero(); ls_short.coords_count() * 2],
```

geo/src/algorithm/line_measures/frechet_distance.rs
```diff
@@ -52,47 +52,54 @@ where
...
-                    (_, 0) => self.cache[(i - 1) * columns_count].max(dist),
-                    (0, _) => self.cache[j - 1].max(dist),
-                    (_, _) => self.cache[(i - 1) * columns_count + j]
-                        .min(self.cache[(i - 1) * columns_count + j - 1])
-                        .min(self.cache[i * columns_count + j - 1])
-                        .max(dist),
+        let row_length = self.ls_short.coords_count();
+        let (mut prev_row, mut cur_row) = self.cache.split_at_mut(row_length);
+
+        for (i, p_long) in self.ls_long.points().enumerate() {
```
lnicola, line 85: No strong opinion, but I think we could fill `prev_row` right at the start, then we wouldn't need `i`. We could also initialize `cur_row[0]` at the start, then drop the `match`.

And maybe use `zip` to avoid the indexing, but that might be too much.

grevgeny: Done!

lnicola (approved):
Oof, this is reads less nicely than I expected, but I guess it's fine otherwise?

Issue #948, opened 2022-12-04
Panic: could not compare segments


hallahan:
When processing OSM data, sometimes we have bad geometry, and using the library for `interior_point` causes a panic. Normally, feeding bad geometry to a geometry library should return an error, not a panic, as this is a standard thing to expect.

I have to recover from the panic with `panic::catch_unwind`, which is not ideal.

https://github.com/georust/geo/blob/3b0d5738f54bd8964f7d1f573bd63dc114587dc4/geo/src/algorithm/sweep/active.rs#L54

```
[2022-12-04T19:14:40.661064000Z geo::algorithm::sweep::active] could not compare segments:
        Active(Segment{ LPt((444882789.06824327, 332473624.5), (444882817.0, 332473624.5))
        of Line { start: Coordinate { x: 444881047.0, y: 332473624.5 }, end: Coordinate { x: 444882817.0, y: 332473624.5 } }
         [NON/NON] })
        Active(Segment{ LPt((444882200.0, 332474331.0), (444882789.06824327, 332473624.5))
        of Line { start: Coordinate { x: 444882817.0, y: 332473591.0 }, end: Coordinate { x: 444882200.0, y: 332474331.0 } }
        [1st] [NON/NON] })
thread '<unnamed>' panicked at 'unable to compare active segments!', /Users/n/.cargo/registry/src/github.com-1ecc6299db9ec823/geo-0.23.0/src/algorithm/sweep/active.rs:54:13
```

softdevca:
The comments on of `Eq` and `Ord` for `Active<T>` state that they "Assert total ordering" but they use `partial_cmp` and thus must do something if the `partial_cmp` is not total and returns `None`.  This is not ideal and the implementations should require `Ord` instead of `PartialOrd` on `T`.

michaelkirk:
I just merged #1234 which replaces our BoolOps implementation with one backed by the i_overlay crate. This should resolve the issue you're seeing. Let us know if it recurs. You can use it now if you use the unreleased `geo` crate, otherwise we expect it to be part of the upcoming v0.29.0 release.

PR #1358, opened 2025-05-12
WIP: new sweep-line implementation

urschrei:
---

This is a from-scratch re-implementation of the Bentley-Ottmann sweep line algorithm, based on the JTS version (subsequently also ported to GEOS), and is intended to replace the `https://docs.rs/geo/latest/geo/algorithm/sweep/struct.Intersections.html` method.

In terms of performance, it is a lot faster than the existing sweep line impl:

In most cases, the existing impl is 45-50x **slower** than brute force. This implementation is around 1.3x slower, but for use cases in which you want to find a small number of intersections for a large number of lines (2,000 and up), it is considerably faster. Benchmarks can be run using `cargo bench --bench sweep_line_intersection`.

I've generated a lot of tests. The methodology for most of these is:

1. Run the sweep-line algorithm and the brute-force algorithm (`line_intersection` over all pairs of lines), and compare the output: if they don't match, the test fails.
2. If we panic, the test also fails.

So far, I haven't been able to produce a panic (but this is unsurprising: there are no `unwrap`s or `expect`s…), and sweep line results match brute-force results.

# Discussion
This impl doesn't use `SweepPoint`, `Cross`, or `PointOrLine`: we only accept `Line<GeoFloat>` segments for checking. If we adopt this, we would either have to do more work to accept "Crossable" geometries etc, or deprecate these (and adapt our [`InteriorPoint`](https://docs.rs/geo/latest/geo/algorithm/interior_point/trait.InteriorPoint.html) algorithm (which uses `SweepPoint` internally). That is overwhelmingly my preference.

# Feedback wanted
Tests that produce incorrect results. I'm going to add https://github.com/georust/geo/issues/948 if I can, but it's looking good so far. Is there anything in our JTS test suite we can use?


urschrei:
Split impl and tests, so the actual algorithm is easier to review.

urschrei:
@dabreegster tagging you in case this is useful, both for your own uses and because you might have some big real-world data for testing.

urschrei:
Another small cause for optimism: the _existing_ sweep line algorithm fails on [`test_overlapping_segments`](https://github.com/georust/geo/pull/1358/files#diff-63bf58618ec625b17b73a27b647557b9c2b0306cecdaf7b95d9c3c156c7f7636R417-R449):

```shell
  stderr ───

    thread 'algorithm::new_sweep::tests::test_overlapping_segments' panicked at geo/src/algorithm/new_sweep/tests.rs:39:5:
    assertion `left == right` failed: Sweep algorithm found 54 intersections but brute force found 50
      left: 54
     right: 50
```

urschrei:
```shell
Large Datasets/isolated_brute_force_20000
                        time:   [709.20 ms 709.54 ms 709.93 ms]
                        change: [+0.0083% +0.0594% +0.1199%] (p = 0.07 > 0.05)
Large Datasets/isolated_new_sweep_20000
                        time:   [7.3726 ms 7.3972 ms 7.4239 ms]
                        change: [-96.138% -96.124% -96.105%] (p = 0.00 < 0.05)
                        Performance has improved.
```

We're now two orders of magnitude faster than brute force on large sets with sparse intersections.

michaelkirk:
This is exciting! I might not have time to review it this week, but I'll take a look as soon as I can.

urschrei:
> This is exciting! I might not have time to review it this week, but I'll take a look as soon as I can.

No rush really. I'm gonna stop tinkering with it now, and maybe some people will suggest evil test cases.

dabreegster:
This looks like an incredible improvement! I only seem to have one use of this today: detecting places where OSM road segments cross over each other without a graph connection, aka, finding bridge/tunnel crossing points, to force a planar graph. I have never seen a crash from this, and I've never noticed the speed being an issue. So no evil test cases for you, sorry.
https://github.com/a-b-street/ltn/blob/7f0178c34de81f5da1163025d3ebf915448b6222/backend/src/route_snapper.rs#L46

urschrei:
> This looks like an incredible improvement! I only seem to have one use of this today: detecting places where OSM road segments cross over each other without a graph connection, aka, finding bridge/tunnel crossing points, to force a planar graph. I have never seen a crash from this, and I've never noticed the speed being an issue. So no evil test cases for you, sorry. https://github.com/a-b-street/ltn/blob/7f0178c34de81f5da1163025d3ebf915448b6222/backend/src/route_snapper.rs#L46

Ah well, if you have a representative dataset you use for that, lmk!

michaelkirk:
> This impl doesn't use SweepPoint, Cross, or PointOrLine: we only accept Line<GeoFloat> segments for checking. If we adopt this, we would either have to do more work to accept "Crossable" geometries etc, or deprecate these (and adapt our [InteriorPoint](https://docs.rs/geo/latest/geo/algorithm/interior_point/trait.InteriorPoint.html) algorithm (which uses SweepPoint internally). That is overwhelmingly my preference.

If I understood you correctly, I think you meant applying your new sweep line implementation to interior_point. I had a go at that here: 25255423c7114d8b60888d9edc13d0f763b7d0a0

Since it seems like this new implementation might be less prone to failure, I am in favor of using it internally everywhere, unless there's some reason not to.

I'd love to see this move forward - anything I can do to help?

geo/src/algorithm/new_sweep/tests.rs
```diff
@@ -0,0 +1,583 @@
...
+        });
+
+        assert!(
+            matching_intersection.is_some(),
+            "Sweep algorithm missed an intersection between lines"
+        );
+
+        // Compare the actual intersection details
+        #[allow(clippy::assertions_on_constants)]
+        if let Some((_, _, sw_intersection)) = matching_intersection {
```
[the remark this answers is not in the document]
frewsxcv: I think `let x = y.expect(z)` would be identical to `let Some(x) = y else { panic!(z) }`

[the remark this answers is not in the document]
michaelkirk: lol great point.

geo/src/algorithm/new_sweep/mod.rs
```diff
@@ -0,0 +1,417 @@
...
+/// These indices are established in `build_index()`, which uses a HashMap to efficiently
+/// match events with the same interval index in O(1) time. This approach provides a simple
+/// way to navigate between related events without the complexity of ownership or borrowing.
+///
+/// # Implementation Note: Why No Binary Search Tree or Priority Queue?
+///
+/// The Bentley-Ottmann algorithm as described typically uses two tracking data structures:
+///
+/// 1. A **priority queue** to process events in coordinate order
+/// 2. A **balanced binary search tree** to track active segments sorted by their `y`-coordinate position
```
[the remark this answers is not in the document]
urschrei: They're sorted by their `y` coordinate at the current `x` position of the sweep line, which is vertical.

geo/src/algorithm/new_sweep/mod.rs
```diff
@@ -0,0 +1,417 @@
...
+            }
+        });
+    }
+}
+
+impl<T: GeoFloat> Iterator for Intersections<T> {
+    type Item = (Line<T>, Line<T>, LineIntersection<T>);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.current_index < self.intersection_pairs.len() {
```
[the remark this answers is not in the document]
urschrei: I am totally failing to do this atm. I just get typeck errors and then almost burst into tears, so I'm gonna leave as is unless you have a genuine concern about complexity, in which case I can revisit with a fresh brain.

geo/src/algorithm/new_sweep/mod.rs
```diff
@@ -0,0 +1,417 @@
...
+        }
+    }
+
+    fn add(&mut self, interval: SweepLineInterval<T>) {
+        let x_min = interval.min;
+        let x_max = interval.max;
+
+        // Ensure capacity
+        if self.events.capacity() <= self.events.len() + 1 {
+            self.events.reserve(64);
```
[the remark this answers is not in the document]
urschrei: I think this was a hangover from when I was trying to wring some more perf out of the algorithm. Removed!

geo/src/algorithm/new_sweep/mod.rs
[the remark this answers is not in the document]
michaelkirk: No, I'm not very concerned about this code. We can keep it.

geo/src/algorithm/new_sweep/mod.rs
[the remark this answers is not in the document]
michaelkirk: ah right. I misunderstood which part of the algorithm this was.

urschrei:
I've squashed 2525542 into my HEAD to have a look (you should be able to rebase against ee8b99a to produce the same result). The problem we have now is that our monotone module makes heavy use of the deprecated sweep line module, producing a raft of clippy failures. I don't know anything about that module, but I'm pretty sure we can't swap in the new sweep-line algorithm – the existing impl exposes event data and other things that mine doesn't (and can't easily)

michaelkirk:
We don't necessarily have to deprecate the old sweep algorithm at this point. I did that in my commit mostly just to identify where all it was being used while developing.

That said, my understanding is that this new sweep implementation is strictly better:  very similar from a correctness standpoint and it avoids a class of panics which is great. Addressing panics we know users are hitting is my main interest in moving this forward.

So in that case, it feels weird to offer users the "worse" legacy sweep algorithms. My instinct is that we should just replace the public interface of the old one with the new one. This would be a breaking change,  since the API's are slightly different, but that seems worthwhile.

As for the `monotone` functionality: how about maintaining the old sweep impl, but only as a private implementation detail of Monotone? 

That would let us move forward with fixing crashes without losing functionality. 

What do you think? Is it important to maintain the old sweep API for other reasons? Or any alternative proposals?

urschrei:
> Is it important to maintain the old sweep API for other reasons?

My concern was purely about deprecating a big module I didn't write!

> My instinct is that we should just replace the public interface of the old one with the new one. This would be a breaking change, since the API's are slightly different, but that seems worthwhile.

Good plan. Done.

> As for the monotone functionality: how about maintaining the old sweep impl, but only as a private implementation detail of Monotone?

Creatively renamed to `old_sweep`, annotated with a warning, and made `pub(crate)`. I can move it fully into the `monotone` module if it's desired.

Note that the new implementation no longer offers `SweepPoint` which was a `Point` with full lexicographic sort capabilities (with the baked-in assumption that the underlying float values would be sortable? dot dot dot etc)

In any case, I had a look at `interior_point` and I don't think we need a full lexicographic sort, [as per my comment](https://github.com/georust/geo/pull/1358/files#diff-6b4867cfa72dbc11e1d13ae58795c9bd5a51f5f41e7d5bb06484e07a6534cd4bR210-R214), so that should be fine.

urschrei:
The two new commits are meaningfully separated for (hopefully) ease of review.

urschrei:
…and finally, a less messy (IMO) benchmark setup, and I managed to replace the HashMap with a lookup, which buys us another 20 % speedup or so on average (sorry!).

geo/src/algorithm/interior_point.rs
```diff
@@ -11,7 +11,8 @@ use crate::algorithm::{
     relate::Relate,
 };
 use crate::geometry::*;
-use crate::sweep::{Intersections, SweepPoint};
+// use crate::old_sweep::{Intersections, SweepPoint};
```
michaelkirk, line 14: Delete this?

geo/benches/utils/crossings.rs
```diff
@@ -8,7 +8,8 @@ use geo::{line_intersection::line_intersection, Line};
 use rstar::{primitives::GeomWithData, RTree};
 
 pub fn count_bo(lines: &[Line<f64>]) -> usize {
-    Intersections::from_iter(lines.iter()).count()
+    let lines = lines.to_vec();
```
michaelkirk: Is this change significant? It's not a big deal, but it might affect the bench.

geo/src/algorithm/sweep/mod.rs
```diff
@@ -1,29 +1,364 @@
...
-use im_segment::IMSegment;
+    /// Compute all intersections
+    fn compute(&mut self) {
+        self.compute_sweep_intersections();
 
-mod vec_set;
-pub(crate) use vec_set::VecSet;
+        self.intersection_pairs.retain(|(line1, line2, _)| {
+            // Keep only intersections between different lines
+            line1 != line2
```
michaelkirk: I don't understand why we'd want to do this. Can you elaborate? Is it from JTS?

As an example, if I ran:
```
let lines = wkt!(
  MULTILINESTRING(
    (0.0 0.0,1.0 1.0),
    (0.0 0.0,1.0 1.0,2.0 1.0)
  )
);
Intersections::from_iter(&lines)
```

I think I'd expect to see an intersection for the overlapping lines: `(0.0 0.0,1.0 1.0)`, but I think because of this code, we'd report zero intersections. Am I understanding correctly?

geo/src/algorithm/sweep/mod.rs
```diff
@@ -1,29 +1,364 @@
...
+        // Create INSERT event
+        let insert_event = SweepLineEvent::new(x_min, EventType::Insert, interval);
+        self.insert_events.push(insert_event);
+
+        // Create DELETE event
+        let delete_event = SweepLineEvent::new(x_max, EventType::Delete, interval);
+        self.delete_events.push(delete_event);
+    }
+
+    fn compute_intersections<F>(&mut self, mut intersection_action: F)
```
michaelkirk: take it or leave it: This name confused me a bit. 

I assumed `intersection_action` was called when there was an intersection, e.g. synonymous with `on_intersection` or `handle_intersection`, but it's actually called when there is an *overlap* and it is responsible for determining when there is an intersection. 

Better names might be: `check_for_intersection` or (similar to JTS) `overlap_action`

geo/src/algorithm/sweep/mod.rs
```diff
@@ -1,29 +1,364 @@
...
+                // 2. Only examine segments that start BEFORE the current segment ends
+                //    (their INSERT x <= our DELETE x), as these are the only ones that
+                //    could potentially intersect with the current segment
+                // 3. Skip self-intersections by filtering out segments with the same index
+                // 4. Check if the x-intervals actually overlap using intervals_overlap
+                // 5. For each overlapping pair, carry out the actual geometric intersection test
+                self.insert_events[i + 1..]
+                    .iter()
+                    .take_while(|other_event| other_event.interval.min <= delete_x)
+                    .filter(|other_event| other_event.interval.index != current_segment_index)
```
michaelkirk: Small perf change: 

My understanding is that each `interval.index` only appears in `insert_events` once. 

Given that we're starting just *after* the insert_event for `current_segment_index`, isn't this filter redundant?

geo/src/algorithm/sweep/mod.rs
```diff
@@ -1,29 +1,364 @@
...
+                    .filter(|other_event| {
+                        intervals_overlap(current_interval, &other_event.interval)
```
michaelkirk, line 318: Small perf change: 

Given that `other_event.interval.min <= delete_x` isn't this check redundant?

geo/src/algorithm/sweep/tests.rs
```diff
@@ -0,0 +1,583 @@
+use super::*;
+use crate::algorithm::line_intersection::line_intersection;
+use std::iter::FromIterator;
+
+/// Helper function to compute all intersections using brute force
+fn compute_brute_force_intersections<T: GeoFloat>(
```
michaelkirk, line 5: This brute_force_intersection check appears slightly different from the one in the benchmarks. I don't know that it matters, but I wanted to note it in case it was oversight.

geo/src/algorithm/sweep/mod.rs
```diff
@@ -1,29 +1,364 @@
...
+}
+
+impl<T: GeoFloat> Ord for SweepLineEvent<T> {
+    fn cmp(&self, other: &Self) -> Ordering {
+        // Use total_cmp for more robust floating point comparison
+        // This is crucial for grid patterns where many x-values are exactly equal
+        match self.x_value.total_cmp(&other.x_value) {
+            Ordering::Equal => {
+                // When x values are equal, prioritize INSERT events before DELETE events
+                // This ensures proper handling of overlapping segments
```
michaelkirk: I can't see where we ever have a mixed collection of INSERT and DELETE events - so it seems like this detail of the sorting implementation has no effect. 

Unless I'm missing something we could delete the `event_type` field altogether - we already infer the event_type by if it lives in the `index.insert_events` field or the `index.delete_events` field.

geo/src/algorithm/sweep/tests.rs
```diff
@@ -0,0 +1,583 @@
...
+                        intersection: sw_point,
+                        is_proper: sw_proper,
+                    },
+                ) => {
+                    assert_eq!(
+                        *bf_proper, *sw_proper,
+                        "Proper intersection status doesn't match for the same line pairs"
+                    );
+
+                    assert_relative_eq!(
```
michaelkirk: As a rule, I'm pretty happy to accept a little slop in our floating point math.

Just for my own understanding, in this case we use the same algorithm for computing the value of the intersection - so it should actually match exactly. Right? I'm fine to leave the slop in, just making sure I understand.

geo/src/algorithm/sweep/tests.rs
```diff
@@ -0,0 +1,583 @@
...
+            "Sweep algorithm found an intersection not found by brute force"
+        );
+    }
+}
+
+#[test]
+fn should_not_panic() {
+    // from https://github.com/georust/geo/issues/1342
+    // these segments cause a panic in the existing sweep line impl:
+    // segment not found in active-vec-set: 4
```
michaelkirk, line 114: 🎉

geo/src/algorithm/sweep/tests.rs
```diff
@@ -0,0 +1,583 @@
...
+        Line::from([(0.1, 0.0), (10.0, 0.9)]), // Intersects baseline
+        Line::from([(5.0, 0.0), (5.0, 10.0)]), // Vertical intersector
+    ];
+
+    verify_intersections(&lines);
+}
+
+/// Test with lines having coordinates near powers of 2
+#[test]
+fn test_power_of_two_boundaries() {
```
michaelkirk, line 307: Thanks for all the tests!

michaelkirk (changes requested):
Thanks for all the changes so far! It feels very close and I'm excited to get it merged.

geo/src/algorithm/sweep/mod.rs
urschrei: I added it because the brute force comparison was failing without it for the test at https://github.com/georust/geo/pull/1358/commits/7f1d6cee0edee6a98f0f12caf0a8559b903781ba?diff=unified&w=1#diff-63bf58618ec625b17b73a27b647557b9c2b0306cecdaf7b95d9c3c156c7f7636R432-R463 (sweep returns 54 overlapping segments, brute force returns 50). But I've just run that test through PostGIS, and it returns 54 segments. So that means my brute force verification function is wrong?

```sql
  WITH segments AS (
    SELECT unnest(ARRAY[
      -- First polygon set
      ST_GeomFromText('LINESTRING(10.0 10.0, 15.0 10.0)'),
      ST_GeomFromText('LINESTRING(15.0 10.0, 15.0 20.0)'),
      ST_GeomFromText('LINESTRING(15.0 20.0, 10.0 10.0)'),
      ST_GeomFromText('LINESTRING(10.0 -10.0, 15.0 -10.0)'),
      ST_GeomFromText('LINESTRING(15.0 -10.0, 15.0 -20.0)'),
      ST_GeomFromText('LINESTRING(15.0 -20.0, 10.0 -10.0)'),
      ST_GeomFromText('LINESTRING(20.0 10.0, 25.0 10.0)'),
      ST_GeomFromText('LINESTRING(25.0 10.0, 25.0 20.0)'),
      ST_GeomFromText('LINESTRING(25.0 20.0, 20.0 10.0)'),
      ST_GeomFromText('LINESTRING(20.0 -10.0, 25.0 -10.0)'),
      ST_GeomFromText('LINESTRING(25.0 -10.0, 25.0 -20.0)'),
      ST_GeomFromText('LINESTRING(25.0 -20.0, 20.0 -10.0)'),
      -- Second polygon set
      ST_GeomFromText('LINESTRING(10.0 10.0, 15.0 10.0)'),
      ST_GeomFromText('LINESTRING(15.0 10.0, 15.0 15.0)'),
      ST_GeomFromText('LINESTRING(15.0 15.0, 10.0 10.0)'),
      ST_GeomFromText('LINESTRING(10.0 -10.0, 15.0 -10.0)'),
      ST_GeomFromText('LINESTRING(15.0 -10.0, 15.0 -15.0)'),
      ST_GeomFromText('LINESTRING(15.0 -15.0, 10.0 -10.0)'),
      ST_GeomFromText('LINESTRING(20.0 10.0, 25.0 10.0)'),
      ST_GeomFromText('LINESTRING(25.0 10.0, 25.0 5.0)'),
      ST_GeomFromText('LINESTRING(25.0 5.0, 20.0 10.0)'),
      ST_GeomFromText('LINESTRING(20.0 -10.0, 25.0 -10.0)'),
      ST_GeomFromText('LINESTRING(25.0 -10.0, 25.0 -5.0)'),
      ST_GeomFromText('LINESTRING(25.0 -5.0, 20.0 -10.0)')
    ]) AS geom
  ),
  numbered_segments AS (
    SELECT geom, row_number() OVER () AS id
    FROM segments
  ),
  intersections AS (
    SELECT
      s1.id AS id1,
      s2.id AS id2,
      ST_Intersects(s1.geom, s2.geom) AS intersects,
      ST_AsText(ST_Intersection(s1.geom, s2.geom)) AS intersection_wkt
    FROM numbered_segments s1
    JOIN numbered_segments s2 ON s1.id < s2.id
    WHERE ST_Intersects(s1.geom, s2.geom)
  )
  SELECT
    COUNT(*) AS total_intersections,
    COUNT(CASE WHEN id1 <= 12 AND id2 > 12 AND s1.geom = s2.geom THEN 1 END) AS identical_line_pairs
  FROM intersections
  JOIN numbered_segments s1 ON intersections.id1 = s1.id
  JOIN numbered_segments s2 ON intersections.id2 = s2.id;
```

geo/src/algorithm/sweep/mod.rs
urschrei: Oh good catch.

geo/src/algorithm/sweep/tests.rs
urschrei: Yeah, we can switch this to an exact comparison.

geo/src/algorithm/sweep/mod.rs
urschrei: Yep, we can absolutely get rid of this. Nice.

geo/src/algorithm/sweep/mod.rs
michaelkirk: This might be related to my other comment here: https://github.com/georust/geo/pull/1358/files/70afe68cceb6ea1954df2ae4096af76cfaa004ac#r2190797548

It looks like this brute force method in the unit tests has code to explicitly remove duplicates (though the brute force version in the benches does not).

urschrei: Yep I've used the same for both and removed the retain call, and the test passes.

geo/benches/utils/crossings.rs
urschrei: Not sure whether significant, but we now pass a `Vec` (that in itself shouldn't affect the timings) and that Vec is cloned in the benchmark setup phase (9f8f0b90)

urschrei:
Wondering whether it's sensible to document the presence of this module in the `line_intersection` module docs: `sweep` isn't very intuitive, and it's already slightly confusing to have `line_intersection` and `intersects` modules (nb I'm not proposing to change that, just wondering how best to point someone at the module that best fits their needs)

michaelkirk:
> Wondering whether it's sensible to document the presence of this module in the line_intersection module docs: sweep isn't very intuitive, and it's already slightly confusing to have line_intersection and intersects modules (nb I'm not proposing to change that, just wondering how best to point someone at the module that best fits their needs)

Great point. I think more people would be be looking for something like "intersection" rather than "sweep" (which is an implementation detail). Cross linking from these other modules as you suggested seems like a good start.

michaelkirk:
I'm not sure if you're still actively pushing code, so tag me when you're ready for re-review @urschrei

urschrei:
OK, have a (hopefully final) look @michaelkirk

michaelkirk (approved):

michaelkirk:
Looks good! I'll leave it to you to merge it @urschrei

dabreegster:
> This impl doesn't use SweepPoint, Cross, or PointOrLine: we only accept Line<GeoFloat> segments for checking. If we adopt this, we would either have to do more work to accept "Crossable" geometries etc

I'm adapting some code to use the new implementation, and hitting a slight snag here. It's sometimes useful to associate arbitrary data with the lines that intersect. As a workaround, I'm hashifying the input lines (by picking a suitable precision and turning f64 into isize), then doing the lookup with the output Lines.

michaelkirk:
Thanks for being a beta tester 😉 

I think we assumed no one was using those traits except us. Now that we know that's not the case, I can follow up to try to make it generic like the old implementation.

dabreegster:
Thanks for pushing through a newer implementation that doesn't crash!

The workaround is not bad: https://github.com/a-b-street/ltn/commit/6b233ee90b04727267143f32ad4b2edd9dc22cda#diff-94c2fcfb4730187542a1dec886f4b03d40929ac66fdf6f1735d5bb6df7c91a50

The simplest API might be to return the index of the matching lines, ie, `((Line, usize), (Line, usize), LineIntersection`. The user can then associate data with that index easily while building the input. It's a little less type-safe than plumbing through a generic associated data `T` with each `Line`, or letting people implement a trait, but it would work totally fine and might (totally speculating, haven't been following the new perf improvements PR) be cheaper memory-wise.

PR #1364, opened 2025-06-11
Apply `empty` constructor to other crates.

michaelkirk:
---

This is a draft because it depends on the review/release of https://github.com/georust/geo/pull/1363 first.

We could merge this before releasing geo-types, but it means that the other crates won't be releasable *until* we release geo-types and bump their minimum geo-types req.


michaelkirk:
Now that geo-types-0.7.17 is released, use the new `empty` constructor in our other crates.

PR #1365, opened 2025-06-16
Add geometry buffering

### LINKED ISSUE #641: Geometry buffering

michaelkirk:
---

Geometry buffering constructs a new geometry whose boundary is some offset from the input boundary. Intuitively, you can "grow" or "shrink" a geometry.

FIXES #641

Like our BoolOps trait, this is mostly just a wrapper for @Nailxsharipov's excellent https://github.com/iShape-Rust/iOverlay functionality.

Diverging from i_overlay: Because we are a more general purpose geometry library, I thought it made sense to offer a more "unified" API for buffering _all_ Geometry types, whereas i_overlay has distinct API's for (Multi)LineStrings vs (Multi)Polygons, and doesn't buffer Point's at all. 

One downside to this approach, is that some options don't makes sense - e.g. passing a LineCap style option to a Polygon buffer operation doesn't do anything. And trying to buffer a Point with `Butt` end caps outputs an empty geometry. Still - I prefer this approach in aggregate, and it's in line with how JTS/GEOS behave.

I made a little interactive debug tool if you'd like to test out this buffering functionality in your browser:  https://michaelkirk.github.io/geo-buffer-debug/


geo/src/algorithm/buffer.rs
```diff
@@ -0,0 +1,835 @@
...
+                let num_segments = (2.0 * std::f64::consts::PI / angle.to_f64()).ceil() as usize;
+                let center = self.0;
+                let radius = style.distance;
+
+                let mut coords = Vec::with_capacity(num_segments + 1);
+                for i in 0..num_segments {
+                    let angle =
+                        F::from_f64(2.0 * std::f64::consts::PI / num_segments as f64 * i as f64)
+                            .expect("valid float constant");
+                    let x = center.x + radius * angle.cos();
```
dabreegster, line 296: Huh, I kind of thought [sin_cos](https://doc.rust-lang.org/std/primitive.f64.html#method.sin_cos) would somehow be more efficient than the two individually, but apparently not

geo/src/algorithm/buffer.rs
```diff
@@ -0,0 +1,835 @@
...
+        };
+        // This one is pretty close... but not quite right. Maybe add a parameter? It should be the same as the point buffer though?
```
dabreegster: Stale comment?

jts-test-runner/src/lib.rs
```diff
@@ -17,24 +21,78 @@ pub fn assert_jts_tests_succeed(pattern: &str) {
...
+}
+
+pub fn check_buffer_test_case(
+    actual: &Geometry,
+    expected: &Geometry,
+) -> std::result::Result<(), String> {
+    // This error threshold is arbitrary. I ratcheted it down until tests started failing.
+    // Manually inspecting the output of the borderline cases appeared subjectively "reasonable".
+    //
+    // In particular, we seem to diverge the most when doing large subtractive (negative buffers)
```
dabreegster, line 58: Is there a particular test case that showed this difference? Could be useful to preserve somewhere

dabreegster (approved):
This is fantastic, such an important feature, with rigorous tests! Thank you!!

geo/src/algorithm/buffer.rs
michaelkirk: It would be, but it's a very common compiler optimization.

geo/src/algorithm/buffer.rs
michaelkirk: Oh, good catch! This actually represents a badly labeled TODO. I'm going to convert this to a draft.

jts-test-runner/src/lib.rs
michaelkirk: You can ratchet this tolerance down and see which tests start to fail.

Here's an example just below the current tolerance:

Screenshot 2025-06-17 at 12 13 51

The blue is the input, and the yellow is the JTS output of negative buffering. We get a similar output from our implementation, but there is a bit of divergence (just under the tolerance).

geo/src/algorithm/buffer.rs
michaelkirk: I meant to look into this earlier, but forgot, since the error was pretty small, so thanks for the reminder.

Essentially, the "approximated circular" buffer JTS creates for a Point was rotated slightly from ours. I've now offset our rotation to match a bit better in https://github.com/georust/geo/pull/1365/commits/27f754f9cd49573be749a3136c545f31ae3027e4 and was consequently able to tighten the error tolerance in a couple places.

jts-test-runner/src/lib.rs
michaelkirk: Zooming in, here's our output (red) vs JTS (yellow) - you can see it's pretty similar:

Screenshot 2025-06-17 at 12 26 42

Zooming in even further to accentuate the divergence:
Screenshot 2025-06-17 at 12 27 45

dabreegster: Seems very very negligible

dabreegster (approved):
Updates LGTM

michaelkirk:
Thanks for the reviews @dabreegster! 

Since it's a big PR for a fundamental feature, I'll plan to leave it open for at least a week to solicit feedback before merging.

michaelkirk:
Unless I hear otherwise, I'm planning to merge this tomorrow. If you'd like to review and want me to hold off on merging, let me know.

geo/src/algorithm/buffer.rs
```diff
@@ -0,0 +1,862 @@
...
+    /// let style = BufferStyle::new(2.0).line_join(LineJoin::Miter(1.0));
+    /// ```
+    pub fn line_join(mut self, line_join: LineJoin<T>) -> Self {
+        self.line_join = line_join;
+        self
+    }
+
+    /// Sets how the ends of linear geometries should be capped.
+    ///
+    /// This only affects `LineString`, `MultiLineString`, and `Line` geometries.
```
mthh: It seems to me that it also affects `Points` and `MultiPoint`.

geo/src/algorithm/buffer.rs
```diff
@@ -0,0 +1,862 @@
...
+    /// ```
+    pub fn new(distance: T) -> Self {
+        BufferStyle {
+            distance,
+            line_cap: LineCap::Round(Self::default_join_angle()),
+            line_join: LineJoin::Round(Self::default_join_angle()),
+        }
+    }
+
+    /// Sets how two edges of a geometry should meet at vertices.
```
mthh, line 178: Maybe we could state that it doesn't affect `Point`/`MultiPoint`? (or state for which cases it has an effect, cf. also the documentation of `line_cap` which refers to cases where the option has an effect).

mthh (approved):
LGTM!

Sorry for the last minute review! I just left a comment about `line_cap` documentation which might be slightly  improved (and possibly `line_join` too).

I've also played with it a bit, really cool and useful addition! Well done!

The API is a bit different from what users are probably used to with geos/JTS (notably specifying `LineCap::Round(some_angle_in_radian)` rather than a number of quadseg), but I don't think it's troublesome.

michaelkirk:
Good doc suggestions @mthh - I've updated them accordingly.

PR #1366, opened 2025-06-18
add: jarvis binary search for polygons

MrFantOlas:
---

First part (out of 2) to implement this [algorithm](https://www.sciencedirect.com/science/article/pii/S2212017313005331). This section of the code find the upper and lower tangent to a convex hull from a point strictly outside of it in O(N log(N)).

### DIFF
--- a/geo/src/algorithm/jarvis_binary_search.rs
+++ b/geo/src/algorithm/jarvis_binary_search.rs
@@ -1,0 +1,414 @@
+use std::ops::Div;
+
+use crate::{CoordNum, Distance, Euclidean, GeoFloat, GeoNum, Orientation, Point, Polygon};
+
+use crate::algorithm::kernels::Kernel;
+
+/// Describes the tangent from a point outside of a convex polygon
+///
+/// Lower tangent is when all points of the polygon are to the left of the tangent.
+/// Upper tangent is the opposite.
+#[derive(PartialEq, Eq, Debug, Clone, Copy)]
+pub enum Target {
+    UpperTangent,
+    LowerTangent,
+}
+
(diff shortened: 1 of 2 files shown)


michaelkirk:
Hello! Thanks for the contribution. 

I'm not familiar with this algorithm - does this exist in other popular toolkits (in other languages, e.g. jts/geos)?

urschrei:
I'm familiar with the Jarvis March (or gift-wrapping) algorithm for computing convex hulls, but this is intended for something different. I don't know what the best way to review this is: if it's a component of a larger feature (as you imply), then I think we'll probably hold off on reviewing and merging it for now, since it doesn't make much sense to include it in the public API as a standalone algorithm (but please correct me if I'm wrong). Whenever you're ready, open a PR based on this branch and we'll review the full proposed feature then?

MrFantOlas:
Hello,

I have no idea if this exists in other toolkit. This algorithm can be use to make Chan's algorithm for example. Since it is part of a another algorithm I want to implement I will mark this as a draft and reopen in a week or two.

michaelkirk:
>> I'm not familiar with this algorithm - does this exist in other popular toolkits (in other languages, e.g. jts/geos)?
>
> I have no idea if this exists in other toolkit.

I'm trying to discern if this is something that's likely to be useful to a relatively broad audience. If it is, then it might make sense to include it in geo. On the other hand, if it's a pretty niche application, it might make more sense to publish it as a separate crate.

MrFantOlas:
This is somethings that can be used in other algorithms such as Chan's algorithm. Now I don't know what is the audience for such algorithms.

I wanted to created random polygons from a set of points and had to solve that issue for example.

PR #1368, opened 2025-06-20
rename rectangle corners

cookiedan42:
---
rename variables to match the corners they represent


michaelkirk (approved):
Assuming "lb" is "left-bottom"...

I suppose it's ultimately arbitrary how your coordinate system maps to left/right/up/down, but I agree that your framing is a more common convention.

PR #1370, opened 2025-06-20
Add Bearing and Destination for Euclidean Space

cookiedan42:
---

Added implementations for `Bearing` and `Destination` Trait on `Euclidean`


geo/src/algorithm/line_measures/metric_spaces/euclidean/bearing.rs
```diff
@@ -0,0 +1,74 @@
...
+    /// # Examples
+    ///
+    /// ```
+    /// # use approx::assert_relative_eq;
+    /// use geo::{Euclidean, Point, Bearing};
+    ///
+    /// let origin = Point::new(0.0, 0.0);
+    /// let destination = Point::new(1.0, 1.0);
+    /// let bearing = Euclidean.bearing(origin, destination);
+    /// // North
```
dabreegster: Northeast? The comment is a bit misleading

geo/src/algorithm/line_measures/metric_spaces/euclidean/destination.rs
```diff
@@ -0,0 +1,71 @@
...
+    /// ```
+    /// # use approx::assert_relative_eq;
+    /// use geo::{Point, Destination, Euclidean};
+    ///
+    /// // Determine the point 100 km NE of JFK airport.
+    /// let origin = Point::new(0.0, 0.0);
+    /// let northeast_bearing = 45.0;
+    /// let distance = 100.0;
+    ///
+    /// let northeast_location= Euclidean.destination(origin, northeast_bearing, distance);
```
dabreegster:
```suggestion
    /// let northeast_location = Euclidean.destination(origin, northeast_bearing, distance);
```

nit: spacing in the docstring

geo/src/algorithm/line_measures/metric_spaces/euclidean/bearing.rs
```diff
@@ -0,0 +1,74 @@
...
+use crate::Point;
+use num_traits::FromPrimitive;
+
+impl<F: CoordFloat + FromPrimitive> Bearing<F> for Euclidean {
+    /// Returns the bearing from `origin` to `destination` in degrees along a straight line.
+    ///
+    /// # Units
+    ///
+    /// - `origin`, `destination`: Points where x/y are cartesian coordinates
+    /// - returns: degrees, where: North: 0°, East: 90°, South: 180°, West: 270°
```
michaelkirk: I think this will often (usually) be correct, but I wonder if there is a more neutral way to phrase this that isn't predicated on having coordinates where +Y is due north and +X is due east.

Maybe something like:

```
/// # Units
///
/// - `origin`, `destination`: Points where x/y are cartesian coordinates
/// - returns: degrees, measured clockwise from the positive Y-axis direction
///   - 0° = positive Y direction (typically North)
///   - 90° = positive X direction (typically East)
///   - 180° = negative Y direction (typically South)
///   - 270° = negative X direction (typically West)
```

I feel like @urschrei might have some input.

urschrei: ^^ This is clearer, IMO. Thanks @michaelkirk

PR #1371, opened 2025-06-20
add contains<Coord<T>> for MultiLineString

cookiedan42:
---

While working on Contains, I realised that MultiLineString.contains(Coord<T>) wasn't implemented  

Implementation based on MultiLineString.contains(Point<T>)


urschrei (approved):

michaelkirk:
I'm sort of surprised that Contains is implemented for Coord at all, but we are not very consistent about coord vs point (see #15).

I'm hoping to avoid a slippery slope where we start adding redundant implementations for Coord/Point everywhere, but I suppose in this case, we've already implemented Contains for Coords + everything else, so I guess it's probably fine here?

PR #1372, opened 2025-06-20
add shortcut for impossible contains

cookiedan42:
---

Conceptually, a lower dimension geometry cannot contain a higher dimension geometry  

use the `HasDimensions` trait to check for true dimension ( taking into account degenerate geometry) and allow for fast `False` response without dispatching to `Relate` Trait  

Slightly more complex implementation required when working with multi-part geometries because their 1d degenerate forms can be a pseudo-`GeometryCollection` of mixed types  

Currently the macroed implementation of the trait has to be checked manually, wondering if it might be useful to trait single/multi part geoms and 0d/1d/2d geoms so they can be distinguished at compile time using trait bounds

Currently has one benchmark in `Contains_dim.rs`, planning to add more

### DIFF
--- a/geo/src/algorithm/contains/mod.rs
+++ b/geo/src/algorithm/contains/mod.rs
@@ -45,8 +45,9 @@ mod polygon;
 mod rect;
 mod triangle;
 
-// self is a 0d geometry (Point, MultiPoint)
-// target can be any geometry
+/// Implement `Contains` for a 0d geometry (Point, MultiPoint) containing a geometry
+/// * `self` is a 0d geometry (Point, MultiPoint)   
+/// * `target` can be either a multi-part of single-part geometry
 macro_rules! impl_contains_from_relate_0d {
     ($for:ty,  [$($target:ty),*]) => {
         $(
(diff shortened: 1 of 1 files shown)


michaelkirk:
> Currently the macroed implementation of the trait has to be checked manually, wondering if it might be useful to trait single/multi part geoms and 0d/1d/2d geoms so they can be distinguished at compile time using trait bounds

(Sorry for the slow review process.)

I'm not sure what you're proposing above - can you clarify?

Also, can you clarify why you're using "draft" status here? What work is remaining for you to accomplish before you'd consider this mergable?

cookiedan42:
Currently a draft because I had second thoughts on whether this would be a useful feature.  
It adds the dimension check overhead to all calls, but is only useful for a rare scenario (most people probably won't be doing LineString.contains(Polygon) on a regular basis

michaelkirk:
Yeah good point - adding overhead to the common case so as to optimize handling mismatched dimensions or degenerate input seems unlikely to pay off in aggregate. The vast majority of input is not degenerate.

I'm going to close this for now, but happy to discuss it more if you have new ideas.

(edited to clarify this isn't *only* about degenerate input, but also about input with dimensions inverted from the typical like `linestring.contains(polygon)`)

PR #1373, opened 2025-06-27
update wkt

michaelkirk:
---

Just updating some internal deps.

Also, removes one unnecessary dep on wkt now that we have wkt debug representation for geo-types.


michaelkirk:
Whoops! Moving to draft because there's one more crate I want to update.

kylebarron (approved):

PR #1375, opened 2025-07-01
Derive clone, copy and eq on `TriangulationError`

TimTheBig:
---
### Purpose
I'm working on an error enum that includes `TriangulationError` and I want to derive clone, copy and eq for better user experience so in my use-case `TriangulationError` needs to implement them.


TimTheBig:
I did not add it to the change log since it's fairly insignificant

frewsxcv (approved):

TimTheBig:
Thanks

PR #1376, opened 2025-07-04
Tidy up Intersects directory

cookiedan42:
Moved implementations around so they are grouped by caller eg. impl Line.intersects(*) are all together  

add missing implementations for
- `Coord.intersects(MultiLineString)`
- `Coord.intersects(MultiPolygon)`

---


michaelkirk (approved):
Yeesh, what a jumble - thanks for cleaning this up!

PR #1377, opened 2025-07-04
Intersects op

cookiedan42:
---

1. Implement a faster `impl<T> Intersects<Polygon<T>> for Rect<T>`
2. Implement a faster `impl<T> Intersects<Triangle<T>> for Rect<T>`

by using the more optimized methods of `Rect` and `Triangle` instead of the general `Polygon` methods

built on top of changes in #1376

### DIFF
--- a/geo/src/algorithm/intersects/rect.rs
+++ b/geo/src/algorithm/intersects/rect.rs
@@ -1,4 +1,4 @@
-use super::Intersects;
+use super::{has_disjoint_bboxes, Intersects};
 use crate::*;
 
 impl<T> Intersects<Coord<T>> for Rect<T>
@@ -73,6 +73,17 @@ where
     T: GeoNum,
 {
     fn intersects(&self, rhs: &Triangle<T>) -> bool {
-        self.intersects(&rhs.to_polygon())
+        // sufficient to show that any of these are true:
+        // some corner of the triangle intersects the rectangle
+        // some corner of the rectangle intersects the triangle
+        // some edge of triangle intersects edge of rectangle
+
+        if has_disjoint_bboxes(self, rhs) {
+            return false;
+        }
+
+        rhs.coords_iter().any(|p| self.intersects(&p))
+            || self.coords_iter().any(|p| rhs.intersects(&p))
+            || rhs.lines_iter().any(|l| self.intersects(&l))
     }
 }
--- a/geo/src/algorithm/intersects/rect.rs
+++ b/geo/src/algorithm/intersects/rect.rs
@@ -106,7 +106,7 @@ where
         }
 
         rhs.coords_iter().any(|p| self.intersects(&p))
-            || self.coords_iter().any(|p| rhs.intersects(&p))
             || rhs.lines_iter().any(|l| self.intersects(&l))
+            || self.coords_iter().any(|p| rhs.intersects(&p))
     }
 }
--- a/geo/src/algorithm/intersects/rect.rs
+++ b/geo/src/algorithm/intersects/rect.rs
@@ -12,9 +12,36 @@ where
             && rhs.y <= self.max().y
     }
 }
-symmetric_intersects_impl!(Coord<T>, Rect<T>);
+
+symmetric_intersects_impl!(Rect<T>, LineString<T>);
+symmetric_intersects_impl!(Rect<T>, MultiLineString<T>);
+
+// Same logic as Polygon<T>: Intersects<Line<T>>, but avoid
+// an allocation.
+impl<T> Intersects<Line<T>> for Rect<T>
+where
+    T: GeoNum,
+{
+    fn intersects(&self, rhs: &Line<T>) -> bool {
+        let lb = self.min();
+        let rt = self.max();
+        let lt = Coord::from((lb.x, rt.y));
+        let rb = Coord::from((rt.x, lb.y));
+        // If either rhs.{start,end} lies inside Rect, then true
+        self.intersects(&rhs.start)
+            || self.intersects(&rhs.end)
+            || Line::new(lt, rt).intersects(rhs)
+            || Line::new(rt, rb).intersects(rhs)
+            || Line::new(lb, rb).intersects(rhs)
+            || Line::new(lt, lb).intersects(rhs)
+    }
+}
+
 symmetric_intersects_impl!(Rect<T>, Point<T>);
 symmetric_intersects_impl!(Rect<T>, MultiPoint<T>);
+
+symmetric_intersects_impl!(Rect<T>, Polygon<T>);
+symmetric_intersects_impl!(Rect<T>, MultiPolygon<T>);
 
 impl<T> Intersects<Rect<T>> for Rect<T>
 where
@@ -41,27 +68,6 @@ where
     }
 }
 
-// Same logic as Polygon<T>: Intersects<Line<T>>, but avoid
-// an allocation.
-impl<T> Intersects<Line<T>> for Rect<T>
-where
-    T: GeoNum,
-{
-    fn intersects(&self, rhs: &Line<T>) -> bool {
-        let lb = self.min();
-        let rt = self.max();
-        let lt = Coord::from((lb.x, rt.y));
-        let rb = Coord::from((rt.x, lb.y));
-        // If either rhs.{start,end} lies inside Rect, then true
-        self.intersects(&rhs.start)
-            || self.intersects(&rhs.end)
-            || Line::new(lt, rt).intersects(rhs)
-            || Line::new(rt, rb).intersects(rhs)
-            || Line::new(lb, rb).intersects(rhs)
-            || Line::new(lt, lb).intersects(rhs)
-    }
-}
-symmetric_intersects_impl!(Line<T>, Rect<T>);
 
 impl<T> Intersects<Triangle<T>> for Rect<T>
 where
@@ -71,4 +77,3 @@ where
         self.intersects(&rhs.to_polygon())
     }
 }
-symmetric_intersects_impl!(Triangle<T>, Rect<T>);
--- a/geo/src/algorithm/intersects/rect.rs
+++ b/geo/src/algorithm/intersects/rect.rs
@@ -68,7 +68,6 @@ where
     }
 }
 
-
 impl<T> Intersects<Triangle<T>> for Rect<T>
 where
     T: GeoNum,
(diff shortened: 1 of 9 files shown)


cookiedan42:
## `Triangle` and `Rect` `intersects`

use the faster point in polygon methods on `Rect` and `Triangle` instead of general `Polygon` implementation

```intersects triangle pt in rect
                        time:   [3.1047 ns 3.1124 ns 3.1201 ns]
                        change: [-83.557% -83.482% -83.402%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low mild
  5 (5.00%) high mild
  3 (3.00%) high severe

intersects rect pt in triangle
                        time:   [11.654 ns 11.692 ns 11.731 ns]
                        change: [-62.771% -62.632% -62.478%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 29 outliers among 100 measurements (29.00%)
  9 (9.00%) low severe
  9 (9.00%) low mild
  4 (4.00%) high mild
  7 (7.00%) high severe

intersects edge intersections
                        time:   [7.7149 ns 7.7223 ns 7.7306 ns]
                        change: [-71.259% -71.151% -71.051%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low severe
  7 (7.00%) high mild
  1 (1.00%) high severe

intersects edge intersections as polygon
                        time:   [8.0827 ns 8.1024 ns 8.1231 ns]
                        change: [-25.012% -24.472% -23.932%] (p = 0.00 < 0.05)
                        Performance has improved.
  5 (5.00%) high mild
  4 (4.00%) high severe

triangle within rect    time:   [3.1050 ns 3.1088 ns 3.1129 ns]
                        change: [-83.700% -83.644% -83.590%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) low mild
  1 (1.00%) high mild
  4 (4.00%) high severe

rect within triangle    time:   [48.959 ns 49.107 ns 49.253 ns]
                        change: [-30.127% -29.859% -29.606%] (p = 0.00 < 0.05)
                        Performance has improved.

disjoint                time:   [2.8817 ns 2.8865 ns 2.8919 ns]
                        change: [-83.820% -83.761% -83.700%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low severe
  1 (1.00%) low mild
  2 (2.00%) high mild
  2 (2.00%) high severe

```

## Rect intersects polygon

1. if any point of polygon within rectangle => true
2. if any edge of polygon intersect edge of rectangle => true
3. if any corner of rectangle within polygon => true


```
intersects polygon pt in rect
                        time:   [3.1855 ns 3.2036 ns 3.2193 ns]
                        change: [-90.145% -90.081% -90.023%] (p = 0.00 < 0.05)
                        Performance has improved.
  5 (5.00%) low severe
  3 (3.00%) low mild
  1 (1.00%) high mild
  3 (3.00%) high severe

intersects rect pt in polygon
                        time:   [38.422 ns 38.540 ns 38.661 ns]
                        change: [-30.608% -30.192% -29.785%] (p = 0.00 < 0.05)
                        Performance has improved.

intersects edge intersections
                        time:   [9.1640 ns 9.1680 ns 9.1724 ns]
                        change: [-71.905% -71.801% -71.700%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 16 outliers among 100 measurements (16.00%)
  3 (3.00%) low mild
  7 (7.00%) high mild
  6 (6.00%) high severe

not intersects disjoint time:   [2.9147 ns 2.9340 ns 2.9470 ns]
                        change: [-85.341% -85.239% -85.134%] (p = 0.00 < 0.05)
                        Performance has improved.
  7 (7.00%) low severe
  2 (2.00%) low mild
  4 (4.00%) high mild

intersects rect equals polygon gap
                        time:   [4.9172 ns 4.9656 ns 5.0218 ns]
                        change: [-93.833% -93.788% -93.743%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 23 outliers among 100 measurements (23.00%)
  12 (12.00%) low severe
  2 (2.00%) low mild
  1 (1.00%) high mild
  8 (8.00%) high severe

intersects rect within polygon gap
                        time:   [144.13 ns 144.81 ns 145.56 ns]
                        change: [-53.745% -53.470% -53.210%] (p = 0.00 < 0.05)
                        Performance has improved.
  6 (6.00%) high mild
  4 (4.00%) high severe

intersects rect within polygon
                        time:   [60.401 ns 60.830 ns 61.348 ns]
                        change: [-35.319% -34.934% -34.541%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low mild
  3 (3.00%) high mild
  10 (10.00%) high severe


```

cookiedan42:
converting to draft because LineString intersects Polygon will affect the benchmarks
#1379

geo/src/algorithm/intersects/rect.rs
```diff
@@ -40,7 +40,33 @@ where
...
+        }
+
+        // rhs.intersects(pt) is the most expensive, so check it last
+        // only required if rectangle sits entirely within polygon
+        rhs.coords_iter().any(|p| self.intersects(&p))
+            || rhs.lines_iter().any(|l| {
+                self.lines_iter()
+                    .any(|other_line| l.intersects(&other_line))
+            })
+            || self.min().intersects(rhs)
```
michaelkirk: What's this `self.min` check about? 

I guess we've shown above that none of the lines intersect, and that the polygon isn't within the rect, so if there __is__ an intersection, the rect must be **fully** within the polygon. In which case, there's no need to check *all* the rect coords, if any one of them intersects, that's sufficient to show that the rect is entirely within the polygon. Is that right?

I got there eventually, but it's pretty dense. You've commented the cases nicely at the top of the function, but I suspect it'd be easier to follow if you re-organized the code down into a case by case analysis with the comments in line.

geo/src/algorithm/intersects/triangle.rs
```diff
@@ -47,7 +47,14 @@ symmetric_intersects_impl!(Triangle<T>, Line<T>);
 symmetric_intersects_impl!(Triangle<T>, Point<T>);
 symmetric_intersects_impl!(Triangle<T>, MultiPoint<T>);
 
-symmetric_intersects_impl!(Triangle<T>, Polygon<T>);
+impl<T> Intersects<Polygon<T>> for Triangle<T>
+where
+    T: GeoNum,
+{
+    fn intersects(&self, rhs: &Polygon<T>) -> bool {
+        self.to_polygon().intersects(rhs)
```
michaelkirk: Why this change?

geo/src/algorithm/intersects/rect.rs
```diff
@@ -40,7 +40,33 @@ where
...
+        some corner of the polygon intersects the rectangle
+        some corner of the rectangle intersects the polygon
+        some edge of polygon (interior or exterior) intersects edge of rectangle
+        */
+
+        if has_disjoint_bboxes(self, rhs) {
+            return false;
+        }
+
+        // rhs.intersects(pt) is the most expensive, so check it last
```
michaelkirk:
```suggestion
        // pt.intersects(polygon) is the most expensive, so check it last
```

This comment confused me for a minute because I didn't see where we're calling `rhs.intersects(pt)` anywhere, but I think this is what you mean.

michaelkirk (changes requested):
Nice perf improvements for the rect/triangle intersection checks!

You've added some tests in the benches, which is great, but as we continue to add more and more benches, it becomes harder to run them all in a reasonable amount of time. So, could you also add some unit tests for these new implementations?

Topology can be tricky. I confess I'm nervous about multiplying our implementations rather than delegating to a smaller number of better tested implementations. That's why I'm being a stickler about the clarity of code comments. But it is a big speed boost if there are folks using these implementations!

cookiedan42:
> Nice perf improvements for the rect/triangle intersection checks!
> 
> You've added some tests in the benches, which is great, but as we continue to add more and more benches, it becomes harder to run them all in a reasonable amount of time. So, could you also add some unit tests for these new implementations?
> 
> Topology can be tricky. I confess I'm nervous about multiplying our implementations rather than delegating to a smaller number of better tested implementations. That's why I'm being a stickler about the clarity of code comments. But it is a big speed boost if there are folks using these implementations!

These implementations were from the first phase of working through the concept, where reasoning through simple convex polygons with no holes are easier.  

Looking through your comments, I agree with not exploding implementations where possible. 
I have a more general polygon-polygon intersection implementation in #1378 based on further development of this intersection concept.   
Perhaps using the general implementation in #1378 , but using the faster shape specific `contains(pt)` implementations and interiors.iter() as a no-op would help?

michaelkirk:
To be clear, I'm not opposed to merging Rect and Triangle specific implementations when there's significant benefits. I think you've shown an impressive speedup here.

To pontificate a bit... 🧐...  There's a tension between Triangle/Rect vs Polygon. On the one hand, they are topologically redundant. So it's "annoying" to have to code for Triangle/Rect, which may introduce their own bugs, when we can accomplish the same things with general Polygon logic.

On the other hand, and I think you've shown that here in this PR, Triangle/Rect geometries often allow for significantly faster implementations. 

So, I think we should merge this, but I'd like to hedge against the increase in code from specialization, by adding some thorough documentation and some unit tests.

cookiedan42:
Reordered the order of operations for Rect and Triangle slightly to follow polygon intersects polygon. I think it's a good rule of thumb that the specialized operation should be at least as fast as the general operation (eg. converting to_polygon() and running intersect) in all scenarios, and the old sequence was slower when self was fully contained within rhs polygon

geo/src/algorithm/intersects/triangle.rs
```diff
@@ -60,3 +84,63 @@ where
...
+        let polygon = Polygon::new(
+            bound.exterior_coords_iter().collect(),
+            vec![hole.exterior_coords_iter().collect()],
+        );
+
+        assert!(!triangle.intersects(&polygon));
+    }
+
+    #[test]
+    fn test_rect_equals_polygon_hole() {
```
michaelkirk:
```suggestion
    fn test_triangle_equals_polygon_hole() {
```

michaelkirk:
squashed in https://github.com/georust/geo/commit/e8419735b5986f120ddf1de65ac68c1779c3df30

I mangled the PR while squashing and you lost credit for the commits. Sorry about that @cookiedan42  =/

cookiedan42:
@michaelkirk don't worry about it, git is truly the toughest part of development

PR #1378, opened 2025-07-04
Intersects Polygon Polygon

cookiedan42:
---

built on top of #1376 

Concept: if there are no line intersections between two polygons, then either one fully contains the other or they are disjoint.  
So after checking for line crossing, we only need to check one point within the other to check for intersection  

Not totally sure how performance characteristics might compare on significantly more complex geometries, but I don't think it would change much>?
 
Additionally, I noticed while implementing this that the recursive `LineString instersects LineString` implementation from blanket implementation recalculates the bounding box of the LineString on every recursion. Eg. `LS1.intersects(LS2)` calculates bounding box of LS2 (1+no. of segments in LS1) times due to `LS1.iter().any(LS2)` calling `LineString intersects Line` blanket implementation  

In the benchmark, using `LineString intersect LineString` as the polygon boundary check is ~33% slower than decomposing and running `Line intersects Line` The current blanket implementation might be a bit over eager in the bounding box checks

### DIFF
--- a/geo/src/algorithm/intersects/polygon.rs
+++ b/geo/src/algorithm/intersects/polygon.rs
@@ -1,6 +1,6 @@
 use super::{has_disjoint_bboxes, Intersects};
 use crate::coordinate_position::CoordPos;
-use crate::{BoundingRect, CoordinatePosition};
+use crate::{BoundingRect, CoordinatePosition, CoordsIter, LinesIter};
 use crate::{
     Coord, CoordNum, GeoNum, Line, LineString, MultiLineString, MultiPoint, MultiPolygon, Point,
     Polygon, Rect, Triangle,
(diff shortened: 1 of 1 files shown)


cookiedan42:
```
MultiPolygon intersects time:   [89.291 ms 89.457 ms 89.558 ms]
                        change: [-57.660% -57.257% -56.928%] (p = 0.00 < 0.05)
                        Performance has improved.
```

urschrei:
A drive-by comment, but a new faster (for large collections of Lines with assumed sparse intersections) and more robust Intersections iterator for slices of lines is about to land in https://github.com/georust/geo/pull/1358, if you wanted to try it.

cookiedan42:
@urschrei Briefly looked at the implementation, an approach along the lines of `|Intersections(a.Lines())|` + `|Intersections(b.Lines())|` < `|Intersections(a.Lines() + b.lines())|` would give correct results, but I'll have to figure out the cases when the performance curves overlap

cookiedan42:
Converting to Draft because #1379 might ease some of the performance hit here

urschrei:
@cookiedan42 Yep that's exactly what I had in mind. It _should_ be a lot faster in cases where there are few (or no) intersections, which I imagine is the common case, but let's see…

cookiedan42:
Additional speedup from skipping unnecessary interior-interior intersection checks
```
MultiPolygon intersects time:   [81.715 ms 81.862 ms 82.168 ms]
                        change: [-59.893% -59.786% -59.663%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Through multipolygon benchmark, using linestring intersect linestring for interior-exterior checks seem to be the fastest over line-line or linestring-line checks

geo/src/algorithm/intersects/polygon.rs
```diff
@@ -42,11 +42,43 @@ where
...
-        // self intersects (or contains) any line in polygon
-        self.intersects(polygon.exterior()) ||
-            polygon.interiors().iter().any(|inner_line_string| self.intersects(inner_line_string)) ||
-            // self is contained inside polygon
-            polygon.intersects(self.exterior())
+        // if there are no line intersections among exteriors and interiors,
+        // then either one fully contains the other
+        // or they are disjoint
+
+        // check 1 point of each polygon being within the other
```
michaelkirk, line 49:
```suggestion
        // check 1 point of each polygon intersects the other
```

**within** has a special meaning in topology.

geo/src/algorithm/intersects/polygon.rs
```diff
@@ -42,11 +42,43 @@ where
...
+        self.exterior().coords_iter().take(1).any(|p|polygon.intersects(&p))
+        || polygon.exterior().coords_iter().take(1).any(|p|self.intersects(&p))
+        // exterior exterior
+        || self.exterior().lines_iter().any(|self_line| polygon.exterior().lines_iter().any(|poly_line| self_line.intersects(&poly_line)))
+        // exterior interior
+        ||self.interiors().iter().any(|inner_line_string| polygon.exterior().intersects(inner_line_string))
+        ||polygon.interiors().iter().any(|inner_line_string| self.exterior().intersects(inner_line_string))
+
+        // interior interior (not needed)
+        /*
```
michaelkirk, line 59: Thanks for these notes!

PR #1379, opened 2025-07-06
Faster Intersects linestring 2d-shape

cookiedan42:
---

Faster Implementations for `LineString intersects 2d-shape` (`Polygon`, `MultiPolygon`, `Triangle`, `Rect`)

Let `n` be number of segments in the `LineString`  
Let `m` be number of segments in the `2d-shape`

Current blanket implementation of `LineString` intersects divides the operation into n * |`Line intersects 2d-shape`| checks  
This results in 2n Point in `2d-shape` checks and `n` * `m` `Line Intersect Line` checks  

Concept based off principle:
- If no segments intersect, then
  - either `LineString` is within `2d-shape`
  - or `LineString` is disjoint from `2d-shape`
- Hence when no edges cross, then 
  - either all `Point` of `LineString` intersects `2d-shape` 
  - or no `Point` of `LineString` intersects `2d-shape`

It is therefore sufficient to check that 
1. any one `Point` of the `LineString` intersects `2d-shape` 
2. `n` * `m` `Line intersect Line` checks  

to conclude if `LineString` intersects `2d-shape`, cutting out 2n-1 `Point` in `2d-shape` checks

Builds off tidying in #1376, but does not strictly require it

### DIFF
--- a/geo/benches/intersection.rs
+++ b/geo/benches/intersection.rs
@@ -1,6 +1,6 @@
 use criterion::{criterion_group, criterion_main, Criterion};
 use geo::intersects::Intersects;
-use geo::MultiPolygon;
+use geo::{MultiPolygon, Triangle};
 
 fn multi_polygon_intersection(c: &mut Criterion) {
     let plot_polygons: MultiPolygon = geo_test_fixtures::nl_plots_wgs84();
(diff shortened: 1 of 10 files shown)


michaelkirk:
Now that #1376 is merged, can you update this and any other PR's you'd like a review on, and I'll take a look.

cookiedan42:
rebased onto main

## Benchmarks 
```
LineString above Polygon
                        time:   [29.408 ns 29.518 ns 29.646 ns]
                        change: [-30.108% -29.705% -29.234%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe

LineString above Triangle
                        time:   [24.897 ns 24.963 ns 25.039 ns]
                        change: [-66.472% -66.294% -66.119%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) high mild
  3 (3.00%) high severe

LineString around Rectangle
                        time:   [24.610 ns 24.646 ns 24.686 ns]
                        change: [-11.991% -11.648% -11.343%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low mild
  7 (7.00%) high mild
  1 (1.00%) high severe

long disjoint           time:   [2.8547 ms 2.8601 ms 2.8668 ms]
                        change: [-40.333% -40.103% -39.883%] (p = 0.00 < 0.05)
                        Performance has improved.
  4 (4.00%) high mild
  6 (6.00%) high severe

ls within poly          time:   [8.2251 ns 8.2774 ns 8.3371 ns]
                        change: [-60.606% -60.351% -60.075%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
  1 (1.00%) high severe

ls within rect          time:   [2.2294 ns 2.2314 ns 2.2334 ns]
                        change: [-24.047% -23.813% -23.559%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low severe
  3 (3.00%) low mild
  3 (3.00%) high mild
  6 (6.00%) high severe
```

geo/src/algorithm/intersects/line_string.rs
```diff
@@ -3,31 +3,141 @@ use crate::BoundingRect;
 use crate::*;
 
 // Blanket implementation using self.lines().any().
```
michaelkirk, line 5: To make sure I'm understanding: you removed the generic ("blanket") implementation because you wanted to specialize the implementations for some types (Polygon, MultiPolygon, Rect, Triangle), right?

That makes sense. 

It __is__ a breaking change for anyone who has implemented these traits for their own types, so it should be documented as BREAKING in the change log. (I think it's extremely unlikely that anyone was taking advantage of this, but still...)

Also, since it's no longer a blanket implementation, can you remove "blanket" from the name? Usually we do something like: `intersects_line_string_impl!`

geo/src/algorithm/intersects/line_string.rs
```diff
@@ -3,31 +3,141 @@ use crate::BoundingRect;
...
-impl<T, G> Intersects<G> for LineString<T>
+macro_rules! blanket_intersects_linestring {
+    ($t:ty) => {
+        impl<T> $crate::Intersects<$t> for LineString<T>
+        where
+            T: GeoNum,
+            Line<T>: Intersects<$t>,
```
michaelkirk: We don't really need these constraints now that this isn't a real blanket implementation, right?

geo/src/algorithm/intersects/line_string.rs
```diff
@@ -3,31 +3,141 @@ use crate::BoundingRect;
...
+                .lines()
+                .any(|l| rhs.lines_iter().any(|other| l.intersects(&other)))
+    }
+}
+
+//
+// MultiLineString Implementations
+//
+
+macro_rules! blanket_intersects_multilinestring {
```
michaelkirk: Unlike the LineString case, it doesn't look like there's any specialization going on here for MultiLineString. So can we keep the previous blanket implementation as it was?

geo/src/algorithm/intersects/line_string.rs
```diff
@@ -3,31 +3,141 @@ use crate::BoundingRect;
...
+blanket_intersects_linestring!(MultiPoint<T>);
+
+blanket_intersects_linestring!(Line<T>);
+blanket_intersects_linestring!(LineString<T>);
+symmetric_intersects_impl!(LineString<T>, MultiLineString<T>);
+
+impl<T> Intersects<Polygon<T>> for LineString<T>
+where
+    T: GeoNum,
+    Line<T>: Intersects<Line<T>>,
```
michaelkirk: Aren't these `Intersects` constraints a (redundant) artifact from the blanket impl?

geo/benches/intersection.rs
```diff
@@ -144,6 +144,97 @@ fn point_triangle_intersection(c: &mut Criterion) {
     });
 }
 
+fn linestring_polygon_intersection(c: &mut Criterion) {
+    use geo::{coord, line_string, LineString, Polygon, Rect};
+    c.bench_function("LineString above Polygon", |bencher| {
+        let ls = line_string![
```
michaelkirk, line 150: This is just a minor personal preference, so take it or leave it, but if you use the `geo::wkt!` macro for constructing geometries in tests, not only will it be shorter, but also, I can then easily copy/paste it into JTSTestBuilder or some other tool to visually inspect the test cases. 

~~(It doesn't work for non-standard types like Rect/Triangle, though maybe it should?)~~ Update: It does now! #1389

geo/src/algorithm/intersects/line_string.rs
cookiedan42: renamed

geo/src/algorithm/intersects/line_string.rs
cookiedan42: reverted the change

michaelkirk:
Rather than go through another round of feedback, I pushed up some changes. Benches look unaffected, but it reduces some copy/paste. 

Can you take a look @cookiedan42?

cookiedan42:
Thanks! The changes look good to me, didn't think to use Some() as catch but it seems so obviously the correct way to do things like this :)

michaelkirk (approved):

PR #1380, opened 2025-07-09
Convex contains speedups

cookiedan42:
---
1. Handle degenerate cases in Rect contains Rect
1. Exploiting the properties of simple convex polygons to speed up the execution time of `Rect.contains` and `Triangle.contains` methods  

benchmarks and casual proofs in the committed markdown files

Draft as its still only partially complete, looking for challenging integer test cases because these implementations should work for them

### DIFF
--- a/geo/Rect-contains-multipoint.md
+++ b/geo/Rect-contains-multipoint.md
@@ -1,0 +1,51 @@
+
+all points must intersect rectangle, at least one must be within
+
+rect contains multipoint 100 (Contains Trait)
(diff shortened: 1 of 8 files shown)


cookiedan42:
Draft because this impl requires a Line covers Line to handle degenerate cases that become Line covers Point

cookiedan42:
Gonna close this because working through degenerate cases is messy, will reopen/ start a new PR if/when I think of a more systematic way of going about it

Issue #1381, opened 2025-07-12
Proposal: Make iOverlay's `FillRule` configurable in `BooleanOps::*`


cuppachino:
Current `BooleanOps` functions are [hardcoded to use `FillRule::EvenOdd`](https://github.com/georust/geo/blob/b6aa8d3b94fa3016edd8f17426fc68c05198c960/geo/src/algorithm/bool_ops/mod.rs#L63C17-L63C17), but `i_overlay` [supports multiple fill rules](https://docs.rs/i_overlay/latest/i_overlay/core/fill_rule/enum.FillRule.html), including `NonZero`. I'd love to be able to use `NonZero` in some cases for my project.

Would you be open to making the fill rule configurable in `BooleanOps`? This could be done by introducing new methods with an additional file rule parameter (non-breaking), or by changing `boolean_op`, `union`, etc to accept a fill rule argument directly (breaking). I'd be happy to open a PR if this is something you'd consider. Personally, I think the breaking change is preferable for Geo long-term.

michaelkirk:
This sounds worthwhile. 

My understanding is that the even/odd rule is what most people will expect as default, so I'd advocate for keeping the existing API and adding a new more verbose one for people who want to fine tune.

PR #1382, opened 2025-07-12
Add `FillRule`-configurable boolean operations to `BooleanOps` trait

cuppachino:
- [😁] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/main/CODE_OF_CONDUCT.md).
- [😎] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---
Resolves https://github.com/georust/geo/issues/1381 using @michaelkirk's preferred solution.


michaelkirk:
I see the original implementation is now using the new code path, so I guess any further tests would mostly be testing i_overlay, not `geo`. 

Still, it'd be nice to have a couple tests showing examples of an alternative fill rule.

Also, can you document that the un-parameterized method uses the EvenOdd fill rule.

geo/src/algorithm/bool_ops/mod.rs
```diff
@@ -57,10 +57,20 @@ pub trait BooleanOps {
...
+    ) -> MultiPolygon<Self::Scalar> {
+        self.boolean_op_with_fill_rule(other, op, FillRule::EvenOdd)
+    }
+
+    /// Performs a boolean operation with the specified fill rule.
+    fn boolean_op_with_fill_rule(
+        &self,
+        other: &impl BooleanOps<Scalar = Self::Scalar>,
+        op: OpType,
+        fill_rule: FillRule,
```
michaelkirk, line 95: Can you write a doc test verifying this works? I don't see where we're re-exporting the FillRule type. We don't want to require users to have to explicitly add i_overlay to their dependencies.

cuppachino: Good catch

cuppachino:
Let me know if you would like me to add more examples of other fill rules in the tests.

I would also like to drop these images here as visual aid to anyone looking. 

These are operations between a self-intersecting star and square, and the same shapes used in the test cases.

<table>
  <tr>
    <td align="center">
      Polygon union with fill rule: EvenOdd example
	  <br/>
      <sub><b>Union EvenOdd</b></sub>
    </td>
    <td align="center">
      Polygon union with fill rule: NonZero example
	  <br/>
      <sub><b>Union NonZero</b></sub>
    </td>
  </tr>
  <tr>
    <td align="center">
      Polygon difference with fill rule: EvenOdd example
	  <br/>
      <sub><b>Difference EvenOdd</b></sub>
    </td>
    <td align="center">
      Polygon difference with fill rule: NonZero example
	  <br/>
      <sub><b>Difference NonZero</b></sub>
    </td>
  </tr>
</table>

geo/src/algorithm/bool_ops/tests.rs
```diff
@@ -1,8 +1,44 @@
...
+    let square: Polygon = wkt!(POLYGON((50.0 50.0, 94.0 50.0, 94.0 8.0, 50.0 8.0)));
+
+    let multi1 = self_intersecting_star.difference(&square);
+    assert_eq!(multi1.0.len(), 6);
+
+    let multi2 = self_intersecting_star.difference_with_fill_rule(&square, FillRule::EvenOdd);
+    assert_eq!(multi2.0.len(), 6);
+    assert_eq!(multi1, multi2);
+
+    println!("multi1: {}", multi1.wkt_string());
```
michaelkirk: Please remove these `println!`

michaelkirk (approved):
Thanks for the quick work! 

Other than removing the debug `println`, this looks good to me! I'll leave it open for another couple days in case anyone else wants to weight in.

geo/src/algorithm/bool_ops/tests.rs
cuppachino: Ah good catch again. I was using that to generate the images.

PR #1386, opened 2025-07-17
More consistent sweep benches

michaelkirk:
---

Inspired by #1358, I had some ideas to further improve the intersection calculations.

I was seeing some pretty large oscillations in the benchmarks, and so wanted to start by first making these particular benchmarks a little more robust.

### DIFF
--- a/geo/benches/sweep_line_intersection.rs
+++ b/geo/benches/sweep_line_intersection.rs
@@ -59,11 +59,11 @@ fn bench_performance_comparison(c: &mut Criterion) {
     let mut rng = StdRng::seed_from_u64(42);
 
     // Test key sizes: crossover point, medium, and large datasets
-    for (n, sample_size) in [
-        (10, None),
-        (100, None),
-        (1_000, Some(100)),
-        (10_000, Some(10)),
+    for (n, sample_size, expected_intersections) in [
+        (10, None, 7),
+        (100, None, 1328),
+        (1_000, Some(100), 130_050),
+        (10_000, Some(10), 11_689_383),
     ] {
         let mut group = c.benchmark_group(&format!("Performance Comparison ({n} lines)"));
         if let Some(sample_size) = sample_size {
@@ -75,7 +75,8 @@ fn bench_performance_comparison(c: &mut Criterion) {
         // Brute force approach
         group.bench_function("brute_force", |b| {
             b.iter(|| {
-                black_box(brute_force_intersections(&lines));
+                let intersections = black_box(brute_force_intersections(&lines));
+                assert_eq!(intersections.len(), expected_intersections);
             });
         });
 
@@ -84,7 +85,7 @@ fn bench_performance_comparison(c: &mut Criterion) {
             b.iter(|| {
                 let intersections: Vec<_> =
                     NewSweepIntersections::<_>::from_iter(lines.iter().cloned()).collect();
-                black_box(intersections);
+                assert_eq!(intersections.len(), expected_intersections);
             });
         });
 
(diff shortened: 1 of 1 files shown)


geo/benches/sweep_line_intersection.rs
```diff
@@ -54,116 +63,78 @@ fn brute_force_intersections(lines: &[Line<f64>]) -> Vec<(Line<f64>, Line<f64>)>
     result
 }
 
-// Benchmark performance comparison across different dataset sizes
-fn bench_performance_comparison(c: &mut Criterion) {
```
michaelkirk, line 58: Sweep line performance is a factor of two main things: how many lines and how many intersections (how "dense") 

So, I was curious how dense this particular test case was. It turns out, it's quite dense! 5c646465a94c1e35d037487e26dc9da4ecbc8480

So this test was somewhat redundant with the existing `bench_dense_line_intersections`.

However unlike `bench_dense_line_intersections` and `bench_sparse_line_intersections`, this test was nice because it tested various line_count. 

Now I've reorganized the "perf,dense,sparse" tests into two tests: dense and sparse, and each of those runs on a variety of line_counts.

Said differently, we're now testing a variety of line_counts for sparse intersections (which is where the BO sweep really shines!)

geo/benches/sweep_line_intersection.rs
```diff
@@ -54,116 +63,78 @@ fn brute_force_intersections(lines: &[Line<f64>]) -> Vec<(Line<f64>, Line<f64>)>
...
+// When intersections are dense, the sweep algorithm has less of an advantage vs. brute force.
+fn bench_dense_line_intersections(c: &mut Criterion) {
     let mut rng = StdRng::seed_from_u64(42);
-
-    // Test key sizes: crossover point, medium, and large datasets
-    for &n in &[1000, 5000, 10000] {
-        let lines = generate_random_lines(n, &mut rng);
+    for (n, sample_size, expected_intersections) in [
+        (10, None, 7),
+        (100, None, 861),
```
michaelkirk, line 72: I added some smaller test sizes. I think they'll be used for datasets smaller than 1,000 - so it's worth considering if we improve/degrade those use cases.

That said, anecdotally I've noticed that the small test sizes seem to be noisier than the rest.

geo/benches/sweep_line_intersection.rs
```diff
@@ -177,7 +148,6 @@ fn generate_random_line(rng: &mut impl Rng) -> Line<f64> {
 // Benchmark edge cases
 fn bench_essential_edge_cases(c: &mut Criterion) {
     let mut group = c.benchmark_group("Essential Edge Cases");
-    group.sample_size(10);
```
michaelkirk, line 180: Using the default (larger) sample size made these benches a little more stable. These particular tests were fast enough to get a good sampling in the 5s warmup. 

There indeed were some slower ones - I've left the custom sample size for those.

geo/benches/sweep_line_intersection.rs
```diff
@@ -54,116 +63,78 @@ fn brute_force_intersections(lines: &[Line<f64>]) -> Vec<(Line<f64>, Line<f64>)>
...
 
-// Benchmark with "dense" case - many lines in small area
-// The new algorithm should be faster here
-fn bench_dense_line_intersections(c: &mut Criterion) {
-    let mut group = c.benchmark_group("Dense Line Intersections");
-    group.sample_size(10);
-
-    // Create a grid of lines that will have many intersections
-    // expected intersections = (n / 2) * (n / 2);
-    let n = 1000;
```
michaelkirk, line 96: Note - I replaced the previous "dense" and "sparse" logic with something tunable that can be used for both sparse and dense use cases.

However, if you think there's value in this previous approach - e.g. like maybe it's more realistic than the random lines I'm now generating, I could be convinced to revert that part.

urschrei:
Great, happy to see some more stable tests. The existing suite was kinda scattergun in places.

michaelkirk:
> The existing suite was kinda scattergun in places.

Rest assured, they're still pretty scatter gun (unfortunately). 

When experimenting with various perf changes, I'm regularly seeing 5-10% swings, even in the "brute force" tests which shouldn't be affected.

PR #1387, opened 2025-07-17
sweep line perf

michaelkirk:
  -  Since we haven't released the new Sweep, it seems redundant to add another changelog entry. 
---

While studying the new BO implementation, I kept finding little pieces that I could strip off, which lead to multiple rounds of simplification until what remains is (subjectively) essential, and (less subjectively) faster.

First the benches: I'm finding them pretty noisy, even after #1386, but there are clearly some improvements. I'd be curious to see if others get similar results when running the bench vs. main.

## benches

**Random dense lines**

Sweep does best with sparse intersections, but it's decent for reasonably dense ones too! Only on the very small line sets did Sweep perform worse than brute-force, and on the very large sets it did significantly better than brute-force.

```
$ cargo bench --bench="sweep_line_intersection" -- --baseline=main
   Compiling geo v0.30.0 (/Users/mkirk/src/georust/geo/geo)
   Compiling jts-test-runner v0.1.0 (/Users/mkirk/src/georust/geo/jts-test-runner)
    Finished `bench` profile [optimized] target(s) in 5.81s
     Running benches/sweep_line_intersection.rs (target/release/deps/sweep_line_intersection-05f43c36d3715a21)
Gnuplot not found, using plotters backend
Random dense lines (10 lines)/brute_force
                        time:   [362.70 ns 364.29 ns 365.95 ns]
                        change: [-0.7102% -0.2557% +0.1699%] (p = 0.25 > 0.05)
Random dense lines (10 lines)/sweep
                        time:   [528.17 ns 528.89 ns 529.69 ns]
                        change: [-32.666% -31.880% -31.208%] (p = 0.00 < 0.05)👍👍👍
                        Performance has improved.

Random dense lines (100 lines)/brute_force
                        time:   [56.225 µs 57.503 µs 58.815 µs]
                        change: [-3.1780% -0.9382% +1.3174%] (p = 0.43 > 0.05)
  4 (4.00%) high mild
Random dense lines (100 lines)/sweep
                        time:   [53.443 µs 54.006 µs 54.617 µs]
                        change: [-1.1526% +0.5026% +2.2079%] (p = 0.54 > 0.05) 🤷
  4 (4.00%) high mild

Random dense lines (1000 lines)/brute_force
                        time:   [9.7531 ms 9.8292 ms 9.9045 ms]
                        change: [+0.7552% +1.8500% +2.8884%] (p = 0.00 < 0.05)
Found 1 outliers among 50 measurements (2.00%)
  1 (2.00%) high mild
Random dense lines (1000 lines)/sweep
                        time:   [8.3380 ms 8.4277 ms 8.5149 ms]
                        change: [-14.630% -13.278% -11.909%] (p = 0.00 < 0.05)👍
                        Performance has improved.

Benchmarking Random dense lines (10000 lines)/brute_force: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 9.2s.
Random dense lines (10000 lines)/brute_force
                        time:   [928.95 ms 931.58 ms 934.19 ms]
                        change: [+0.9622% +1.3817% +1.7846%] (p = 0.00 < 0.05)
Benchmarking Random dense lines (10000 lines)/sweep: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 7.9s.
Random dense lines (10000 lines)/sweep
                        time:   [783.34 ms 783.93 ms 784.55 ms]
                        change: [-10.387% -9.7930% -9.2555%] (p = 0.00 < 0.05)👍
                        Performance has improved.
Found 3 outliers among 10 measurements (30.00%)
  1 (10.00%) low mild
  2 (20.00%) high mild
```

**random sparse lines**

Sweep does best with sparse intersections. These benches saw similar improvement (maybe even slightly better) than the dense case, further widening the gap between Sweep and brute-force.

```
Random sparse lines (10 lines)/brute_force
                        time:   [178.51 ns 178.60 ns 178.69 ns]
                        change: [-0.4702% -0.3432% -0.2204%] (p = 0.00 < 0.05)
  3 (3.00%) low severe
  4 (4.00%) low mild
  6 (6.00%) high mild
Random sparse lines (10 lines)/sweep
                        time:   [160.37 ns 160.76 ns 161.19 ns]
                        change: [-52.347% -52.190% -52.039%] (p = 0.00 < 0.05)👍👍👍
                        Performance has improved.
  10 (10.00%) high mild
  1 (1.00%) high severe

Random sparse lines (100 lines)/brute_force
                        time:   [19.197 µs 19.212 µs 19.231 µs]
                        change: [-0.1152% -0.0119% +0.0959%] (p = 0.83 > 0.05)
  2 (2.00%) low mild
  3 (3.00%) high mild
  5 (5.00%) high severe
Random sparse lines (100 lines)/sweep
                        time:   [1.6400 µs 1.6449 µs 1.6501 µs]
                        change: [-54.479% -54.130% -53.832%] (p = 0.00 < 0.05)👍👍👍
                        Performance has improved.
  6 (6.00%) high mild

Random sparse lines (1000 lines)/brute_force
                        time:   [1.9031 ms 1.9051 ms 1.9076 ms]
                        change: [-0.2159% +0.0024% +0.2208%] (p = 0.99 > 0.05)
Found 8 outliers among 50 measurements (16.00%)
  5 (10.00%) high mild
  3 (6.00%) high severe
Random sparse lines (1000 lines)/sweep
                        time:   [63.467 µs 65.359 µs 67.189 µs]
                        change: [-37.712% -35.861% -33.825%] (p = 0.00 < 0.05)👍👍👍
                        Performance has improved.

Random sparse lines (10000 lines)/brute_force
                        time:   [190.13 ms 190.35 ms 190.60 ms]
                        change: [-0.0380% +0.0959% +0.2384%] (p = 0.22 > 0.05)
Random sparse lines (10000 lines)/sweep
                        time:   [2.0398 ms 2.0425 ms 2.0456 ms]
                        change: [-13.212% -12.363% -11.619%] (p = 0.00 < 0.05)👍
                        Performance has improved.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
```

The "edge cases and realistic" benches saw mixed changes, but mostly positive. 

```
Essential Edge Cases/collinear_segments_sweep
                        time:   [195.72 µs 198.49 µs 201.20 µs]
                        change: [-5.7581% -4.6444% -3.5904%] (p = 0.00 < 0.05)
                        Performance has improved.
  6 (6.00%) low mild
Benchmarking Essential Edge Cases/collinear_segments_brute_force: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 9.6s, enable flat sampling, or reduce sample count to 50.
Essential Edge Cases/collinear_segments_brute_force
                        time:   [1.9052 ms 1.9104 ms 1.9196 ms]
                        change: [-0.1599% +0.0469% +0.2813%] (p = 0.69 > 0.05)
  1 (1.00%) low mild
  1 (1.00%) high mild
  1 (1.00%) high severe
Essential Edge Cases/numerical_precision_sweep
                        time:   [270.04 µs 272.16 µs 274.30 µs]
                        change: [-19.604% -18.756% -17.879%] (p = 0.00 < 0.05)👍👍
                        Performance has improved.
Benchmarking Essential Edge Cases/numerical_precision_brute_force: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 7.0s, enable flat sampling, or reduce sample count to 50.
Essential Edge Cases/numerical_precision_brute_force
                        time:   [1.3782 ms 1.3813 ms 1.3846 ms]
                        change: [-3.0719% -2.4662% -1.8839%] (p = 0.00 < 0.05)
                        Performance has improved.
  4 (4.00%) high mild
  1 (1.00%) high severe

Realistic Patterns/road_network_sweep
                        time:   [2.1498 ms 2.1618 ms 2.1769 ms]
                        change: [+1.7014% +2.5660% +3.4966%] (p = 0.00 < 0.05) 🤷
                        Performance has regressed.
  2 (2.00%) high mild
  1 (1.00%) high severe
Realistic Patterns/road_network_brute_force
                        time:   [47.600 ms 48.501 ms 49.572 ms]
                        change: [+3.0720% +4.7734% +7.2299%] (p = 0.00 < 0.05)
                        Performance has regressed.
  4 (4.00%) high mild
  10 (10.00%) high severe
Realistic Patterns/polygon_boundaries_sweep
                        time:   [257.92 µs 265.27 µs 272.30 µs]
                        change: [-20.730% -18.010% -15.211%] (p = 0.00 < 0.05) 👍
                        Performance has improved.
Realistic Patterns/polygon_boundaries_brute_force
                        time:   [5.2464 ms 5.3689 ms 5.5107 ms]
                        change: [+4.5549% +7.3895% +10.453%] (p = 0.00 < 0.05) 🤷
                        Performance has regressed.  
  1 (1.00%) high mild
  11 (11.00%) high severe
```

## Draft?

If we are going to merge this - I'd like to overhaul the docs to better match the implementation.

Even though the core algorithm flow is essentially the same old BO sweep, I've simplified it to the point that I'm not sure how recognizable it is - in particular I feel like talking about "events" sort of muddies the waters. It no longer coincides with the implementation, even though they are a part of the "spiritual" predecessor of this implementation. 

I'm currently leaning towards a top level doc, something like:

> If two segments intersect, their x values must overlap. Thus, to efficiently find all intersections between segments, we first narrow down the candidates to check for intersection by considering only segment pairs whose x-intersections overlap.
>
> To find overlapping x segments:
> 1. Take all segments and sort them by their minimum_x.
> 2. For each segment (the "current segment"), take all subsequent segments (those that start after the current segment), stopping once you find a segment that starts after the end of the current segment. All segments until then, are overlapping with the current segment.
> 3. These overlapping segments can then be checked for intersection

What do you think @urschrei? I imagine you spent a fair amount of effort writing up the current documentation, which explains the more classical translation of BO. Do you think it would be helpful to retain that language, even if only in the documentation?

### DIFF
--- a/geo/benches/sweep_line_intersection.rs
+++ b/geo/benches/sweep_line_intersection.rs
@@ -11,20 +11,18 @@
 //!
 //! To run specific benchmark groups:
 //! ```
-//! cargo bench --bench sweep_line_intersection "Performance Comparison"
-//! cargo bench --bench sweep_line_intersection "Dense Line Intersections"
-//! cargo bench --bench sweep_line_intersection "Sparse Large Dataset"
+//! cargo bench --bench sweep_line_intersection "Random Dense Lines"
+//! cargo bench --bench sweep_line_intersection "Random Sparse Lines"
 //! cargo bench --bench sweep_line_intersection "Essential Edge Cases"
 //! cargo bench --bench sweep_line_intersection "Realistic Patterns"
 //! ```
 
 use criterion::{black_box, criterion_group, criterion_main, Criterion};
 use geo::algorithm::line_intersection::line_intersection;
-use geo::algorithm::sweep::Intersections as NewSweepIntersections;
+use geo::algorithm::sweep::Intersections;
 use geo::{Destination, Euclidean, Line};
 use geo_types::Point;
 use rand::prelude::*;
-use std::iter::FromIterator;
 
 /// Generate a set of random lines
 ///
@@ -66,14 +64,14 @@ fn brute_force_intersections(lines: &[Line<f64>]) -> Vec<(Line<f64>, Line<f64>)>
 // Benchmark with "dense" case - lines with many intersections.
 // When intersections are dense, the sweep algorithm has less of an advantage vs. brute force.
 fn bench_dense_line_intersections(c: &mut Criterion) {
-    let mut rng = StdRng::seed_from_u64(42);
     for (n, sample_size, expected_intersections) in [
         (10, None, 7),
-        (100, None, 861),
-        (1_000, Some(50), 91_898),
-        (10_000, Some(10), 8_570_900),
+        (100, None, 827),
+        (1_000, Some(50), 90_438),
+        (10_000, Some(10), 8_604_894),
     ] {
-        let mut group = c.benchmark_group(format!("Random dense lines ({n} lines)"));
+        let mut rng = StdRng::seed_from_u64(42);
+        let mut group = c.benchmark_group(format!("Random Dense Lines ({n} lines)"));
         if let Some(sample_size) = sample_size {
             group.sample_size(sample_size);
         }
@@ -90,8 +88,9 @@ fn bench_dense_line_intersections(c: &mut Criterion) {
         // Sweep line algorithm
         group.bench_function("sweep", |b| {
             b.iter(|| {
-                let intersections: Vec<_> =
-                    NewSweepIntersections::<_>::from_iter(lines.iter().cloned()).collect();
+                let intersections: Vec<_> = Intersections::from_iter(lines.iter().cloned())
+                    .iter()
+                    .collect();
                 assert_eq!(intersections.len(), expected_intersections);
             });
         });
@@ -103,14 +102,14 @@ fn bench_dense_line_intersections(c: &mut Criterion) {
 // Benchmark with "sparse" case - lines with few intersections.
 // When intersections are sparse, the sweep algorithm tends to perform much better than brute force.
 fn bench_sparse_line_intersections(c: &mut Criterion) {
-    let mut rng = StdRng::seed_from_u64(42);
     for (n, sample_size, expected_intersections) in [
         (10, None, 0),
         (100, None, 0),
-        (1_000, Some(50), 10),
-        (10_000, Some(10), 796),
+        (1_000, Some(50), 11),
+        (10_000, Some(10), 798),
     ] {
-        let mut group = c.benchmark_group(format!("Random sparse lines ({n} lines)"));
+        let mut rng = StdRng::seed_from_u64(42);
+        let mut group = c.benchmark_group(format!("Random Sparse Lines ({n} lines)"));
         if let Some(sample_size) = sample_size {
             group.sample_size(sample_size);
         }
@@ -127,8 +126,9 @@ fn bench_sparse_line_intersections(c: &mut Criterion) {
         // Sweep line algorithm
         group.bench_function("sweep", |b| {
             b.iter(|| {
-                let intersections: Vec<_> =
-                    NewSweepIntersections::<_>::from_iter(lines.iter().cloned()).collect();
+                let intersections: Vec<_> = Intersections::from_iter(lines.iter().cloned())
+                    .iter()
+                    .collect();
                 assert_eq!(intersections.len(), expected_intersections);
             });
         });
@@ -174,8 +174,9 @@ fn bench_essential_edge_cases(c: &mut Criterion) {
 
         group.bench_function("collinear_segments_sweep", |b| {
             b.iter(|| {
-                let intersections: Vec<_> =
-                    NewSweepIntersections::<_>::from_iter(lines.iter().cloned()).collect();
+                let intersections: Vec<_> = Intersections::from_iter(lines.iter().cloned())
+                    .iter()
+                    .collect();
                 black_box(intersections);
             });
         });
@@ -252,8 +253,9 @@ fn bench_essential_edge_cases(c: &mut Criterion) {
 
         group.bench_function("numerical_precision_sweep", |b| {
             b.iter(|| {
-                let intersections: Vec<_> =
-                    NewSweepIntersections::<_>::from_iter(lines.iter().cloned()).collect();
+                let intersections: Vec<_> = Intersections::from_iter(lines.iter().cloned())
+                    .iter()
+                    .collect();
                 black_box(intersections);
             });
         });
@@ -311,8 +313,9 @@ fn bench_realistic_patterns(c: &mut Criterion) {
 
         group.bench_function("road_network_sweep", |b| {
             b.iter(|| {
-                let intersections: Vec<_> =
-                    NewSweepIntersections::<_>::from_iter(lines.iter().cloned()).collect();
+                let intersections: Vec<_> = Intersections::from_iter(lines.iter().cloned())
+                    .iter()
+                    .collect();
                 black_box(intersections);
             });
         });
@@ -352,8 +355,9 @@ fn bench_realistic_patterns(c: &mut Criterion) {
 
         group.bench_function("polygon_boundaries_sweep", |b| {
             b.iter(|| {
-                let intersections: Vec<_> =
-                    NewSweepIntersections::<_>::from_iter(lines.iter().cloned()).collect();
+                let intersections: Vec<_> = Intersections::from_iter(lines.iter().cloned())
+                    .iter()
+                    .collect();
                 black_box(intersections);
             });
         });
--- a/geo/src/algorithm/sweep/mod.rs
+++ b/geo/src/algorithm/sweep/mod.rs
@@ -1,52 +1,99 @@
-// This module implements the Bentley-Ottmann sweep line algorithm for efficiently finding
-// line segment intersections, exposed through the [`Intersections`] iterator.
+// This module implements a simplified variant of the [Bentley-Ottmann] sweep line algorithm
+// for efficiently finding line segment intersections, exposed through the [`Intersections`] iterator.
 //
-// Algorithm Flow
+// ## Relationship to Classical Bentley-Ottmann Algorithm
 //
-// 1. Collection:
-//    - Line segments are collected and stored in `Intersections.segments`
-//    - The `compute` method calls `compute_sweep_intersections` to process all segments.
+// The classical Bentley-Ottmann algorithm uses three key data structures:
+// 1. **Event queue**: A priority queue of sweep events (segment start/end points and intersections)
+// 2. **Status structure**: A balanced binary search tree tracking active segments sorted by y-coordinate
+// 3. **Event handling**: Dynamic insertion of newly discovered intersection events
 //
-// 2. Sweep Setup:
-//    - Creates a new `SweepLineIndex` to manage the sweep process
-//    - For each segment, creates a `SweepLineInterval` with min/max x-values
-//    - Each interval generates two events: `INSERT` at left end, `DELETE` at right end.
+// This implementation simplifies the classical approach in several ways:
 //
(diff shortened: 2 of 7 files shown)


geo/benches/sweep_line_intersection.rs
```diff
@@ -90,8 +89,9 @@ fn bench_dense_line_intersections(c: &mut Criterion) {
         // Sweep line algorithm
         group.bench_function("sweep", |b| {
             b.iter(|| {
-                let intersections: Vec<_> =
-                    NewSweepIntersections::<_>::from_iter(lines.iter().cloned()).collect();
+                let intersections: Vec<_> = Intersections::from_iter(lines.iter().cloned())
```
michaelkirk, line 91: Oh also very notable: I broke the API. 

Intersection no longer impl's iter, but you can **get** an iter.

The old API eagerly built up all the intersections into a Vec, and then its own iter implementation was based on that. The new API is lazy - it only computes the next intersection when asked. 

So if for some reason you want to do: `Intersections::new(segments).iter().take(3)` you don't waste any work.

If the old approach is preferred, I could eagerly evaluate the lazy iter, but I think I'd still want to somehow make the lazy iter available for those who prefer it.

urschrei:
> in particular I feel like talking about "events" sort of muddies the waters.

Well, they're still present and referred to by name in the implementation…

The docs as written were as much a way for me to understand how the algorithm works as anything else, and the B-O algorithm specifically uses "event" as way of discussing a specific step in the algorithm. I'm in no way wedded to that term, but it's – IMO – important to be able to explain how this implementation's Rust-specific moving parts relate to the "classic" algorithm, because it provides a way for people who may not have much Rust knowledge, but are familiar with the algorithm to find a way in (this difficulty is essentially what happens every time I look at JTS, and even more so for GEOS).

So as written, your current draft description makes sense, but I think it'd be even better if it could point at B-O where possible. How do we get from your explanation to "sweep line" and "event"?

geo/src/algorithm/sweep/mod.rs
```diff
@@ -191,7 +132,7 @@ struct SweepLineInterval<T: GeoFloat> {
 ///
 /// This implementation avoids both:
 ///
-/// - Instead of a priority queue, the **event vecs are sorted once** and process events in order.
+/// - Instead of a priority queue, the **event vecs are sorted once**, and then the events processed in order.
```
urschrei:
```suggestion
/// - Instead of a priority queue, the **event vecs are sorted once**, and then the events are processed in order.
```

michaelkirk:
Ok, I've have a go at reconciling the documentation with BO. Please take another look @urschrei

geo/src/algorithm/sweep/mod.rs
```diff
@@ -1,52 +1,78 @@
...
-/// lines that intersect or overlap. This is a drop-in
-/// replacement for computing [`LineIntersection`] over all pairs of
-/// the collection, with the perf caveat noted above.
+/// This implementation is most effective when you have many line segments with sparse intersections,
+/// but even quite dense intersections are competitive in all but extreme cases.
 ///
-/// The implementation uses the [Bentley-Ottmann] sweep line algorithm,
-/// which runs in `O((n + k) log n)` time, where `n` is the number of line segments
-/// and `k` is the number of intersections.
+/// As a rule of thumb, if each segment intersects fewer than 10% of the other segments, you can be
```
urschrei: NBD, but `total` is more intuitive to me than `other` (unless you avoided `total` because it's not strictly accurate)
```suggestion
/// As a rule of thumb, if each segment intersects fewer than 10% of the total segments, you can be
```

geo/src/algorithm/sweep/mod.rs
```diff
@@ -56,271 +82,140 @@ mod tests;
...
+                //    As these are the only segments that overlap, they are the only candidates
+                //    for intersection with the current segment
+                self.inserted_intervals[i + 1..]
                     .iter()
-                    .take_while(|other_event| other_event.interval.min <= delete_x)
-                    .filter(|other_event| {
-                        intervals_overlap(current_interval, &other_event.interval)
+                    .take_while(move |interval| interval.inserted_x <= current_interval.deleted_x)
+                    .map(move |overlapping_interval| {
+                        // overlap should be true by construction at this point
```
urschrei, line 235: 👍

geo/src/algorithm/sweep/mod.rs
```diff
@@ -1,52 +1,78 @@
...
+//    we use pre-sorted iteration with early termination based on x-coordinate bounds.
+//
+// 3. **No intersection events**: The classical algorithm discovers intersections during the sweep
+//    and adds them as new events. This implementation instead relies on iteration, and yields
+//    intersections immediately as they are discovered.
+//
+// ### Algorithm Flow
+//
+// 1. **Preprocessing**:
+//    - Convert line segments to `SweepLineInterval`s with normalized x-coordinates
```
urschrei: What's meant by "normalized" in this context?

urschrei (approved):
Great work winnowing the algorithm down even further @michaelkirk. The updated explanation of how it maps onto the core ideas behind B-O is extremely clear and straightforward to comprehend. Only a couple of minor comments inline.

geo/src/algorithm/sweep/mod.rs
michaelkirk: I've adjusted this copy to be (hopefully) be more clear. https://github.com/georust/geo/pull/1387/commits/59852fb3a518b6264c77bc61abc9236d2afc2d2f

geo/src/algorithm/sweep/mod.rs
```diff
@@ -1,52 +1,77 @@
...
+// 1. **No explicit event queue**: Instead of a priority queue, events are stored in vectors
+//    and sorted once. This has the same O(n log n) complexity but a simpler implementation.
+//    This is similar to [Chen & Chan, 2003].
+//    Further, **Only INSERT (segment start) events are sorted**. We only need the DELETE event
+//    (segment end) of the "current" event, which is already available to use by being at the
+//    front of the INSERT events. Rather than sort DELETE events separately, they are instead stored
+//    alongside their INSERT.
+//
+// 2. **No binary search tree**: Rather than maintaining a balanced BST of active segments,
+//    we use pre-sorted iteration with early termination based on x-coordinate bounds.
```
michaelkirk, line 24: I think we should call this out more. I think this (paired with point 3. below) is a pretty key difference between our implementation and BO, and it affects our runtime.

In traditional BO, the BST allows us to track the segments "above and below" the active segment. Those are the only ones which could intersect with the current segment. (I think this assumes there are no intersections crossed by more than 2 intersections. I'm not sure how/if there's some BO++ that handles the more general case.)

Our simplified sweep checks **all** overlapping segments for intersection. This means we'll find intersections regardless of if there are multiple segment intersections at the same point.

Importantly, this changes the asymptotic run time from `O((n+k) log n)` where `k` is the number of **intersections**, to `O((n + m) log n)` where `m` is the number of **overlaps** (and remember, `m >= k`).

In practice, I think this is OK. This algorithm seems robust and I believe gains some speed ups from its simplicity. I think the major downside is more susceptibility to pathological data - specifically, if most of your lines are overlapping without intersecting, I expect our simplified sweep would perform worse than traditional BO.

michaelkirk: Ok, I've documented this in https://github.com/georust/geo/pull/1387/commits/20d4ff989d5c82dd326b48a63c10d48965ac825d

Could you take another look @urschrei ?

urschrei: Revised docs look good to me. I think that in practice, this won't be an issue for most use cases, as mentioned.

Issue #1383, opened 2025-07-12
Bug: graham_hull contains extra point


cookiedan42:
Two parts to this  
1. Bug with `graham_hull` where there is an additional interior point in the hull
2. Bug with `is_ccw_convex` not catching this error in `graham_hull` tests

I was playing around with a impl for polygon convexity checks, and realized that the current impl (wrongly?) returns true for [star polygons] (https://en.wikipedia.org/wiki/Star_polygon). While running tests, these `graham_hull` tests failed, and when visualizing the results, I realized that the returned hulls seem to be wrong too  

`algorithm::convex_hull::graham::test::quick_hull_test_complex_2` produces hull of 
```
[...]

PR #1388, opened 2025-07-18
dedupe graham hull and convex

cookiedan42:
---

Fix for #1383 by de-duplicating adjacent points in the inputs  

Is there a better way to write the dedupe in `is_convex` without using itertools?

### DIFF
--- a/geo/src/algorithm/convex_hull/graham.rs
+++ b/geo/src/algorithm/convex_hull/graham.rs
@@ -5,7 +5,7 @@ use crate::{Coord, GeoNum, LineString};
 /// The [Graham's scan] algorithm to compute the convex hull
 /// of a collection of points. This algorithm is less
 /// performant than the quick hull, but allows computing all
-/// the points on the convex hull, as opposed to a strict
+/// the unique points on the convex hull, as opposed to a strict
 /// convex hull that does not include collinear points.
 ///
 /// # References
--- a/geo/src/algorithm/is_convex.rs
+++ b/geo/src/algorithm/is_convex.rs
@@ -289,7 +289,7 @@ mod tests {
     #[test]
     fn test_single_point() {
         // single point is closed
-        // will panic if is_empyt check in `is_convex_shaped` is removed
+        // will panic if is_empty check in `is_convex_shaped` is removed
         let ls: LineString<f64> = wkt! (LINESTRING (0 0)).convert();
         assert!(ls.is_strictly_convex());
     }
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -27,6 +27,11 @@
   - <https://github.com/georust/geo/pull/1387>
   - <https://github.com/georust/geo/pull/1359>
 
+- Fix `is_convex` to correctly handle duplicate points
+- Fix`graham_hull` to correctly handle duplicate points when `on_hull` is set to true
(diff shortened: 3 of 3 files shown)


geo/src/algorithm/is_convex.rs
```diff
@@ -142,6 +144,19 @@ fn is_convex_shaped<T>(
 where
     T: GeoNum,
 {
+    if coords.is_empty() {
```
michaelkirk, line 147: If I comment out this change, all the tests continue to pass. See https://github.com/georust/geo/compare/main...mkirk%2Ffix-hull?body=&expand=1

Can you add a test that exemplifies the bug that this code is fixing?

michaelkirk (changes requested):

michaelkirk:
> Is there a better way to write the dedupe in is_convex without using itertools?

See https://docs.rs/geo/latest/geo/algorithm/remove_repeated_points/trait.RemoveRepeatedPoints.html

I think currently it's only implemented on geometries. You could maybe copy/paste the implementation (it's very short) or maybe we could refactor the existing implementations to delegate to a method that operates on a `Vec` of points.

geo/src/algorithm/is_convex.rs
cookiedan42: added a test with a degenerate linestring containing a single point

cookiedan42:
> > Is there a better way to write the dedupe in is_convex without using itertools?
> 
> See https://docs.rs/geo/latest/geo/algorithm/remove_repeated_points/trait.RemoveRepeatedPoints.html
> 
> I think currently it's only implemented on geometries. You could maybe copy/paste the implementation (it's very short) or maybe we could refactor the existing implementations to delegate to a method that operates on a `Vec` of points.

these implementations use Vec::dedup, which won't work for the slice here

geo/src/algorithm/is_convex.rs
```diff
@@ -245,4 +262,35 @@ mod tests {
         assert!(!two.is_strictly_ccw_convex());
         assert!(!two.is_strictly_cw_convex());
     }
+
+    #[test]
+    fn test_duplicate_pt() {
+        let ls_unclosed: LineString<f64> = wkt! (LINESTRING (0 0, 1 1, 2 0)).convert();
```
michaelkirk, line 268: Nice use of `convert`! I've always added all the decimal points to every coord, but it's annoying. I'm going to steal this for sure.

geo/src/algorithm/is_convex.rs
michaelkirk: There still doesn't seem to be a test for the changes in this method. If I comment it out, all the tests still pass.

geo/src/algorithm/convex_hull/graham.rs
```diff
@@ -54,6 +54,10 @@ where
     for pt in points.iter() {
         while output.len() > 1 {
             let len = output.len();
+            if output[len - 1] == *pt {
```
michaelkirk, line 57: I'll be honest, I don't have a use case for `include_on_hull`, but based on the documentation it seems reasonable to expect that all the points in the input which are on the hull will be in the output - including **repeated** points.

Still, this change seems an improvement over the currently obviously broken behavior of returning a non-convex hull. 

Could you please document in the method signature and in the change log that repeated points will be removed?

PR #1389, opened 2025-07-18
`wkt!` macro for creating geometries gets support for non-standard variants: LINE/RECT/TRIANGLE

michaelkirk:
---

Note these are non-standard wkt - they wouldn't be parsable by an external parser. Maybe this is controversial...

I think they'll be easier to transform into valid wkt than the current representations, and it's a nice symmetry with the Debug representation, which already prints this format for Line, Rect, Triangle.

### DIFF
--- a/geo-types/CHANGES.md
+++ b/geo-types/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


kylebarron (approved):

michaelkirk:
I'll plan to merge this tomorrow unless I hear otherwise.

PR #1392, opened 2025-07-27
fix: Fix panic in `simplify::compute_rdp` with one point

sjackman:

---

Fix a panic in `simplify::compute_rdp` with one point.
The assertion `debug_assert_ne!(farthest_index, 0);` fails.


geo/src/algorithm/simplify.rs
```diff
@@ -79,15 +79,12 @@ fn compute_rdp<T, const INITIAL_MIN: usize>(
...
-    let first = rdp_indices[0];
-    let last = rdp_indices[rdp_indices.len() - 1];
-    if rdp_indices.len() == 2 {
-        return vec![first, last];
-    }
+    let (first, last) = match rdp_indices {
+        [] => return vec![],
+        &[only] => return vec![only],
+        &[first, last] => return vec![first, last],
+        &[first, .., last] => (first, last),
```
michaelkirk, line 86: I didn't realize you could do this with a pattern match. Nice.

michaelkirk:
I'll let this sit another day before merging in case anyone wants to weigh in, but I don't think it's controversial.

urschrei:
Yep, very happy to merge.

geo/src/algorithm/simplify.rs
sjackman: Yeah it's great! The best hidden feature of Rust. I find that it really helps with both readability and catching corner case bugs like this one, particularly if one avoids the temptation of ending with a catch all case. For example, this code hides the bug:
```rust
[] => return vec![],
&[first, last] => return vec![first, last],
_ => (),
```
whereas this code allows the compiler to catch the missing case of a single element, and complain that the `match` is non-exhaustive:
```rust
[] => return vec![],
// The compiler will complain that this case is missing:
// &[only] => return vec![only],
&[first, last] => return vec![first, last],
&[first, .., last] => (first, last),
```

PR #1395, opened 2025-08-03
Fix star polygon convexity check

### LINKED ISSUE #1394: is_convex false positive on star polygon boundary
`is_convex` returns a false positive for the boundary of [star polygons](https://en.wikipedia.org/wiki/Star_polygon), ie closed linestrings with same handedness in each corner, and where self intersection is allowed

allowing for star polygons also does not align with the definition of a [convex curve](https://en.wikipedia.org/wiki/Convex_curve)

minimal reproduction
```rust
fn test_self_intersection() {
    let ls: LineString<f64> = wkt! (LINESTRING (0 0, 10 0, 7 -1, 4 5, 6 5, 3 -1, 0 0)).convert();
    assert!(!ls.is_convex());
}
```

Image

urschrei:
The fix adds a self-intersection detection

Fixes #1394

---

Draft for now as the self-intersection check is quadratic and we can't use our sweep algorithm as it's `GeoNum`, so I'm open to feedback.


cookiedan42:
Two possible linear approaches I see to this 
1. calculating total internal angle = (n-2) * 180 (less preferred)
3. adding a [sign flip check](https://math.stackexchange.com/a/1745427)

urschrei:
Sign flip looks promising. Let me see if I can make it work.

geo/src/algorithm/is_convex.rs
```diff
@@ -128,6 +135,196 @@ impl<T: GeoNum> IsConvex for LineString<T> {
...
+    for next in vertices.iter().cycle().skip(1).take(vertex_count) {
+        // Compute edge vectors for sign flip tracking
+        let ax = next.x - curr.x; // "after" edge: curr → next
+        let ay = next.y - curr.y;
+
+        // Track sign flips in x direction
+        // The sign flip algorithm principle: a convex polygon traversed in order
+        // will have edge vectors that change sign exactly twice per axis
+
+        match ax.partial_cmp(&T::zero()) {
```
urschrei, line 231: Clippy failed the previous iteration of this logic (and the `y_sign` logic that follows) because it was an `if`-chain, and this `match` version is exhaustive, but it's also a lot harder to follow.

geo/src/algorithm/is_convex.rs
```diff
@@ -119,7 +120,13 @@ impl<T: GeoNum> IsConvex for LineString<T> {
         if !self.is_closed() || self.0.is_empty() {
             None
         } else {
-            is_convex_shaped(&self.0[1..], allow_collinear, specific_orientation)
+            // Use sign flip algorithm which detects both non-convexity and self-intersections
+            if is_convex_sign_flips(&self.0) {
+                // If sign flips indicate convex, determine orientation
+                is_convex_shaped(&self.0[1..], allow_collinear, specific_orientation)
```
michaelkirk, line 126: To make sure I'm understanding, `is_convex_sign_flips` is "forgiving" in that it allows convex shapes to have collinear points, and can be convex regardless of its orientation.

That's why, to maintain our existing API, we still have to fall back to our old impl which accepts `allow_collinear, specific_orientation` as parameters in order to be more "strict" in its convexity check.

Is that the idea?

I haven't dug in too deeply, but it seems like the internals of `is_convex_sign_flips` could be tweaked to fully replace `is_convex_shaped` - it already has a branch where it inspects collinear points. Presumably that branch could depend on whether `allow_collinear`.

It also seems to be tracking the `current_orientation`, so it seems like we could add the `specific_orientation` check there too. But again, I haven't fully studied your implementation yet.

Adding those two checks to `is_convex_sign_flips` seems like it would allow us to remove the call to `is_convex_shaped` all together.

As usual though, I think correctness is probably more important here than performance, so I'm happy to merge more or less as is, and follow up later with potential performance improvements.

If not the larger changes mentioned above, one small change that I'd expect to be easy:

```diff
/// Test if the shape is convex.
fn is_convex(&self) -> bool {
-    self.convex_orientation(true, None).is_some()
+    is_convex_sign_flips(self.0)
}
```

I think that should behave identically and avoid the redundant work in `is_convex_shaped`, right?

urschrei: > I think that should behave identically and avoid the redundant work in `is_convex_shaped`, right?

This fails in two tests:

1. `test_corner_cases`, because we treat `empty` as convex
2. `test_duplicate_points` because we treat unclosed LineStrings as convex

wrt 1, we diverged from PostGIS / GEOS before our change:

```sql
SELECT
    geom,
    ST_AsText(geom) AS wkt,
    ST_IsEmpty(geom) AS is_empty,
    ST_GeometryType(geom) AS geom_type,
    ST_Equals(geom, ST_ConvexHull(geom)) AS is_convex
FROM (
    SELECT ST_GeomFromText('LINESTRING EMPTY') AS geom
) AS t;
```

returns `true` for `is_convex`. We now match that.

The second failure relates to "open" geometries. These aren't considered convex, but the algorithm was assuming closed geometries. Fixed with an `is_closed` check.

I'll have a look at the larger issue today if I have time…I agree it'd be preferable to have a single function for all the checks.

geo/src/algorithm/is_convex.rs
michaelkirk: Ah clippy... I personally make pretty liberal usage of `#[allow()]`. I'm fine either way here.

geo/src/algorithm/is_convex.rs
```diff
@@ -119,14 +120,216 @@ impl<T: GeoNum> IsConvex for LineString<T> {
...
     }
 
     fn is_collinear(&self) -> bool {
         self.0.is_empty()
             || is_convex_shaped(&self.0[1..], true, Some(Orientation::Collinear)).is_some()
     }
+
+    /// Test if the shape is convex.
+    fn is_convex(&self) -> bool {
+        // Convexity only makes sense for closed shapes
```
michaelkirk: It might make sense to put this check in `is_convex_sign_flips`. 

It's a cheap check and we don't allow "convexity" without it, right?

geo/src/algorithm/is_convex.rs
```diff
@@ -119,14 +120,216 @@ impl<T: GeoNum> IsConvex for LineString<T> {
...
+        if !self.is_closed() {
+            return false;
```
michaelkirk: Just to make sure I'm understanding, we're returning here because only 2-dimensional shapes can be convex. Getting the "convexity" of a less than 2-d geometry is moreso "undefined" than it is false. Is that right?

It seems inconsistent to me that we return `false` here, while returning `true` for Points and other degenerate linestrings further down: https://github.com/georust/geo/pull/1395/files#diff-6ac00620efb09b6b79d22282f17a9aef5d9fb517f521fd61eab8f99895d7e45bR190

Personally I can't think of a reason to prefer one over the other, but it seems like we should be consistent? Unless we're just matching JTS quirk for quirk, which is a good enough reason for me.

michaelkirk (approved):
LGTM, I have only quibbles, that you can ignore if they don't resonate.

If you want to further investigating fully replacing the old method, I'm happy to wait, but otherwise, I think it's fine to merge this as is.

geo/src/algorithm/is_convex.rs
urschrei: Good point. `is_convex_sign_flips` accepts a slice: `[Coord<T>]`, but we can check the equality of the first and last slice elements.

geo/src/algorithm/is_convex.rs
urschrei: The single point (`one`) and duplicate line (`one_rep`) test cases were pre-existing, whereas the third special case (degenerate line) is new: the point set is mathematically convex, but that does raise the question of whether we should be treating it as such:

> Getting the "convexity" of a less than 2-d geometry is moreso "undefined"

It's not about dimensionality (nb: we're at the limit of my topology understanding here); convexity is a property of point sets. The empty set (an empty geometry) is convex according to the definition. So is a single point, and a degenerate line `(0 0, 1 1, 0 0)`: "For all pairs of points p, q ∈ S, the line segment between them is in S"

But even though the degenerate line is convex by that definition, it doesn't "feel" like it should be for our uses: it's unexpected to have a backtracking line be convex. So I'm gonna remove that.

michaelkirk: I wasn't thinking about the proper definition of convexity - thank you for correcting me!

I'm good to merge this as is.

Although you've also convinced me that it'd be defensible to consider unclosed line strings as convex if all their points fall on a single line. 

In practice I'm expecting it to be a distinction without much difference for our users, but time may tell differently!

Go ahead and make the call and merge this when you're ready.

PR #1399, opened 2025-08-11
Edit top-level intro docs

urschrei:
---

Draft for bikeshedding purposes.

- Do we want to add (or remove anything)?
- Should the items be links?
- Are we happy to retain the type listing further down?

### DIFF
--- a/geo/src/algorithm/bool_ops/mod.rs
+++ b/geo/src/algorithm/bool_ops/mod.rs
@@ -118,7 +118,7 @@ pub trait BooleanOps {
     }
 
     /// Combines the regions of both `self` and `other` into a single geometry, removing
-    /// overlaps and merging boundaries.
+    /// overlaps and merging boundaries. Consider using [`unary_union`] for efficiently combining several adjacent / overlapping geometries.
     fn union(&self, other: &impl BooleanOps<Scalar = Self::Scalar>) -> MultiPolygon<Self::Scalar> {
         self.boolean_op(other, OpType::Union)
     }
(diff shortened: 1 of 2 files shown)


geo/src/algorithm/bool_ops/mod.rs
```diff
@@ -118,7 +118,7 @@ pub trait BooleanOps {
     }
 
     /// Combines the regions of both `self` and `other` into a single geometry, removing
-    /// overlaps and merging boundaries.
+    /// overlaps and merging boundaries. Consider using [unary_union] for efficiently combining several adjacent / overlaping geometries.
```
urschrei: I snuck this in, and I'd do it again.

geo/src/lib.rs
```diff
@@ -1,6 +1,18 @@
 #![doc(html_logo_url = "https://raw.githubusercontent.com/georust/meta/master/logo/logo.png")]
 
-//! The `geo` crate provides geospatial primitive types and algorithms.
+//! The `geo` crate provides planar geospatial geometries and algorithms.
```
urschrei, line 3: I know we go back and forth on this, but I think it's important to mention planar geometries up front – I call out the availability of non-planar length and distance ops further down.

geo/src/algorithm/bool_ops/mod.rs
michaelkirk:
```suggestion
    /// overlaps and merging boundaries. Consider using [`unary_union`] for efficiently combining several adjacent / overlapping geometries.
```

urschrei:
I'll leave this open for a couple of days in case people want to make suggestions.

frewsxcv (approved):

kylebarron (approved):

PR #1401, opened 2025-08-24
WIP: Fast point in MultiPolygon using interval tree

### LINKED ISSUE #1184: performance for multipolygon contains much worse than for geos

urschrei:
---

fixes #1184

Draft because:

1. I couldn't make this work with `GeoNum` so we need to think about how we'd like to publicly expose the API if it turns out to be actually fast
2. It needs a benchmark. I don't have time to re-implement the benchmark from https://github.com/georust/geo/issues/1184#issuecomment-2119167047. But someone should. Sorry!


urschrei:
Benchmarks:

1. `cargo bench --bench contains "MultiPolygon contains MultiPoint"` (real-world benchmark, around 77 ms to check 1e6 points against 27k vertices)
2. `cargo bench --bench contains "MultiPolygon contains Point"` (more of a microbenchmark; 51.292 µs indexed vs 7.1592 ms unindexed)

PaulWagener:
I would recommend to keep the implementation of `impl<T> Contains<Point<T>> for MultiPolygon<T>` the same. It is slower to create an index for every point and use that once, then it is to just slog through it brute force.

urschrei:
<s>Hmm why do we have these [bench](https://github.com/georust/geo/blob/main/geo/benches/contains.rs#L197) regressions

```shell
MultiPolygon not contains MultiPoint (Contains trait)
                        time:   [8.1562 µs 8.2061 µs 8.2571 µs]
                        change: [+580.14% +583.66% +587.05%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

```shell
MultiPolygon contains MultiPoint (Contains trait)
                        time:   [8.0225 µs 8.0446 µs 8.0677 µs]
                        change: [+562.19% +563.51% +564.95%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 20 outliers among 100 measurements (20.00%)
  6 (6.00%) low mild
  6 (6.00%) high mild
  8 (8.00%) high severe
```
</s>

Solved (see below)

PaulWagener:
> Hmm why do we have these [bench](https://github.com/georust/geo/blob/main/geo/benches/contains.rs#L197) regressions

This is caused by the extra time required to build the index. In the data used in the benchmark 90% of the time is spent constructing the index and 10% is spent looking up the points.

However, this ratio depends on the number of points in the MultiPoint being tested. For a given MultiPolygon the index creation time is fixed, but for each individual point it is faster to test against the index than against the unindexed shape. At some threshold number of points it becomes faster to create the index.

Try changing the benchmark to use a MultiPoint with 1000 points and you will see the opposite result with this PR outperforming the main branch.

urschrei:
This benchmark (in latest commit) is ~2 % _slower_ than running under `main`:

```rust
#[path = "utils/random.rs"]
mod random;
use rand::thread_rng;
use random::*;

const NUMPOINTS: i32 = 1_000_000;

// ...

    let p_1: Polygon<f64> = Polygon::new(geo_test_fixtures::poly1(), vec![]);
    let p_2: Polygon<f64> = Polygon::new(geo_test_fixtures::poly2(), vec![]);
    let multi_poly = MultiPolygon(vec![p_1, p_2]);
    // these points are guaranteed to fall within the intersection of poly1 and poly2
    let bounds = Rect::new(Coord { x: -65.0, y: -32.0 }, Coord { x: 75.0, y: 82.0 });
    let points: Vec<Point<f64>> = (0..NUMPOINTS)
        .map(|_| Point(uniform_point(&mut thread_rng(), bounds)))
        .collect();
    let multi_point = MultiPoint::new(points);

    c.bench_function(
        "MultiPolygon contains MultiPoint (Contains trait)",
        |bencher| {
            bencher.iter(|| {
                assert!(multi_poly.contains(&multi_point));
            });
        },
    );
```

urschrei:
The big contributing factor isn't just points in the `MultiPoint` but Points + MultiPolygon vertices. Checking a million points against a 200-vertex `MultiPolygon` (as above) is basically the same speed whether we use the index or not. Checking a million points against the `nl_zones` `MultiPolygon` (27k+ vertices) without an index is projected to take 23 minutes on my machine. With an index it's ~75 ms (see updated benchmark)

geo/src/algorithm/contains/mod.rs
```diff
@@ -315,6 +325,82 @@ mod test {
...
+
+    #[test]
+    fn empty_multipolygon_fast_test() {
+        let multipoly = MultiPolygon::<f64>::new(Vec::new());
+        assert!(!multipoly.contains(&Point::new(2., 1.)));
+    }
+
+    // See https://github.com/georust/geo/issues/1184#issuecomment-3221772752
+    #[test]
+    fn contains_bug() {
```
michaelkirk: Which bug is this test title referring to? I followed the above link, but it wasn't clear.

My guess is we're testing against expected GEOS output? And that this test would have failed before this PR?

geo/src/algorithm/contains/polygon.rs
```diff
@@ -147,6 +157,171 @@ where
...
+            y_min..y_max,
+            YIntervalSegment {
+                segment: (p1, p2),
+                x_max,
+                is_exterior,
+            },
+        )
+    }
+
+    #[inline]
```
michaelkirk: I tend to be pretty conservative with inline hints. Unless it's a very short, very hot method, I just let the compiler do its thing, since inlining can be a net negative even if it helps microbenchmarks (code space is zero sum!)

I can see how this method is hot - but this is a pretty long method for inlining. How much does it speed up the benchmarks?

geo/src/algorithm/contains/polygon.rs
```diff
@@ -147,6 +157,171 @@ where
...
+struct YIntervalSegment<T: GeoNum> {
+    segment: (Coord<T>, Coord<T>),
+    /// Pre-computed maximum x-coordinate of the segment endpoints.
+    /// Used for early rejection of segments that don't extend far enough right.
+    /// This avoids computing max(seg.0.x, seg.1.x) for every point query.
+    x_max: T,
+    is_exterior: bool,
+}
+
+pub struct IndexedMultiPolygon<T: GeoNum> {
```
michaelkirk: Is this actually intended to be `pub`? If so, we should add some docs.

It would be helpful to distinguish its use case from PreparedGeometry. A name that hints at its purpose, rather than its structure, might be helpful. Maybe something like `struct PolygonContainsPoint` since this is a more specific index than the name currently implies.

If this is not actually intended to be public, then forget I said anything. =)

Also, just musing, but I wonder if longer term we can fold this optimization into PreparedGeometry/Relate somehow.

geo/src/algorithm/contains/polygon.rs
```diff
@@ -147,6 +157,171 @@ where
...
+                        if o == Orientation::CounterClockwise && seg.1.y != point.y {
+                            contribution = 1;
+                        } else if o == Orientation::Collinear
+                            && value_in_between(point.x, seg.0.x, seg.1.x)
+                        {
+                            // Point on boundary!
+                            return ControlFlow::Break(false);
+                        }
+                    }
+                } else if seg.1.y <= point.y {
```
michaelkirk: We got all these segments from `y_interval_tree.query`, so this could be a simple `else` (without the `if seg.1.y <= point.y`) right? 

It's not so much that the redundant `if` check is expensive, but it had me wondering if there was *another* unhandled case I needed to think about.

michaelkirk (changes requested):
`IndexedMultipolygon` should either be demoted from `pub` or documented, but everything else is just minor ignorable quibbles.

michaelkirk:
Oh, and also, this is a great speedup - Thank you!

geo/src/algorithm/contains/polygon.rs
urschrei: Removing had no significant impact.

geo/src/algorithm/contains/mod.rs
urschrei: Ah, the bug was a transitory logic bug in `contains_point`, tested against GEOS output. Renamed.

geo/src/algorithm/contains/polygon.rs
urschrei: It was `pub` so the benchmark could test it, but I rewrote the benchmark to use `contains` instead. That having been said, a better name is a good idea and if we do want to re-use it, the `contains::polygon` module is now `pub(crate)` – we could move it to wherever `PreparedGeometry` lives as an alternative? WDYT?

urschrei: New name: `IntervalTreeMultiPolygon`. Catchy!

PaulWagener:
I think it's better for the IntervalTreeMultiPolygon to be public. Currently the only way this index is exposed to users is in the MultiPoint in MultiPolygon check which only gives if *all* points are in the MultiPolygon. But I think it's a common use case to check wether each point in a user collection is in a MultiPolygon.

PaulWagener:
There is currently still a bug that gives an invalid answer when a point has the same Y coordinate as a boundary vertex.

Failing test case:
````rust
#[test]
fn test_point_inside_indexed_polygon() {
    // Square around the origin with an extra vertex on the right side
    let square = MultiPolygon::new(vec![polygon!(
        exterior: [
            (x: -1., y: 1.),
            (x:  1., y: 1.),
            (x:  1., y: 0.),
            (x:  1., y: -1.),
            (x: -1., y: -1.),
        ],
        interiors: [],
    )]);

    let square_index = IntervalTreeMultiPolygon::new(&square);

    assert!(square_index.contains_point(Coord { x: 0.0, y: 0.0 }));
}
````

**Comment by urschrei** [comment]
> There is currently still a bug that gives an invalid answer when a point has the same Y coordinate as a boundary vertex.
> 
> Failing test case:
> 
> ```rust
> #[test]
> fn test_point_inside_indexed_polygon() {
>     // Square around the origin with an extra vertex on the right side
>     let square = MultiPolygon::new(vec![polygon!(
>         exterior: [
>             (x: -1., y: 1.),
>             (x:  1., y: 1.),
>             (x:  1., y: 0.),
>             (x:  1., y: -1.),
>             (x: -1., y: -1.),
>         ],
>         interiors: [],
>     )]);
> 
>     let square_index = IntervalTreeMultiPolygon::new(&square);
> 
>     assert!(square_index.contains_point(Coord { x: 0.0, y: 0.0 }));
> }
> ```

Added as a failing test.

**Comment by michaelkirk** [comment]
Some notes:

input: `MULTIPOLYGON (((-1 1, 1 1, 1 0, 1 -1, -1 -1, -1 1)))`

Screenshot 2025-09-15 at 10 30 10

`self.y_interval_tree.query(0..0)` only returns the left segment of the square, which crosses `y = 0`.
The two segments making up the right side of the square (with an endpoint at `y = 0`) are not returned.

```
    let result = self
            .y_interval_tree
            .query(point_y..point_y, |(_, segment)| {
                dbg!(segment);
                // Early rejection: segment must extend to the right of point
                if segment.x_max < point.x {
                    return ControlFlow::Continue(());
                }
                ...
```


Trying to understand why by looking briefly at the sif-itree crate:
```
fn query<'a, K, V, H, R>(args: &mut QueryArgs<K, H>, mut nodes: &'a [Node<K, V>]) -> ControlFlow<R>
        ...
        # NOTE: Does this _LESS THAN_ check (rather than `<=`) preclude our segments? 
        if args.interval.start < mid.1 {
            if !left.is_empty() {
                go_left = true;
            }

            if args.interval.end > (mid.0).0.start {
                if !right.is_empty() {
                    go_right = true;
                }

                if args.interval.start < (mid.0).0.end {
                    (args.handler)(&mid.0)?;
                }
            }
        }
```

I'm not 100%, but it seems like the checks in `query` are *exclusive* (they do not include their endpoints), so a segment ending __at__ `y=0` is not considered __in__ the range. 

I'm not sure if that reflects a bug in itree or a bug in our usage. Given that the argument to query is a Range, I think I'd expect it to include the lower bound and exclude the upper bound (a..b) should include `a` and up to but not including `b`.

sif_itree's own documentation says (emphasis mine):
> Interval tree mapping __half-open__ intervals

Can you verify @adamreichold - are we holding it wrong?

**Comment by urschrei** [comment]
I think it's a bug in our usage…am pondering.

**Comment by adamreichold** [comment]
> Can you verify @adamreichold - are we holding it wrong?

To be honest, whenever I touch stuff like this (let's call it one-dimension-less problems), I feel like the chance to make a mistake is at least 110%.

So the model that the `query` method is tested against is

https://github.com/adamreichold/sif-itree/blob/d3b4b3497a0f09640c47c7e85f426f07ff8baed5/src/query.rs#L176

which should handle all overlap situations via the additional assumptions `query_start <= query_end` and `range.start <= range.end`.

That is not yet an answer to your question but possibly helpful. Will try to figure out as well...

**Comment by adamreichold** [comment]
So from a purely definitional point of view, the half-open interval [a,a) is necessarily empty for all a and hence cannot overlap anything...

**Comment by adamreichold** [comment]
I think, what you actually need is a `query_point` method on `ITree`. I never implemented it because I did not need it yet. It should be easy enough to add though...

(My personal use case for `sif-itree` is for time intervals over discrete data types, so the half-open interval containing only a single point was easily constructed as [a, a+1). You could try to emulate that via `next_up/down`, but that seems like a hack at best.)

**Comment by adamreichold** [comment]
Ah, but you still got trouble using half-open intervals as segments having a fixed y-coordinate value would still need to be represented using empty intervals [a,a).

Hhhmmm, so either open this up to support closed intervals or play tricks using next_up/down... 🤔

**Comment by adamreichold** [comment]
So maybe https://github.com/adamreichold/sif-itree/pull/2 would be sufficient for your use case? I think this does not affect tree construction at all and we can just reinterpret the ranges as closed instead of half-open at query time.

(I am not really sure about the API, i.e. just using `Range` and `RangeInclusive` as input types via a sealed trait to choose the interpretation... At least this should not be affected by the performance gotchas of `RangeInclusive`...)

**Comment by michaelkirk** [comment]
I've come across some more failing test cases dealing with points on the boundary.

Recall that `contains` means the point must be in the *interior* of the Polygon.  Being on the border is an `Intersects`, but not a `Contains`.

```rust
    #[test]
    fn test_point_on_boundary_indexed_polygon() {
        // Square around the origin with an extra vertex on the right side
        let square: MultiPolygon =
            wkt!(MULTIPOLYGON(((-1 1, 1 1, 1 0, 1 -1, -1 -1, -1 1)))).convert();
        let square_index = IntervalTreeMultiPolygon::new(&square);

        // PASSES: top left corner
        assert!(!square_index.contains_point(Coord { x: -1.0, y: 1.0 }));

        // FAILS: on left bondary
        assert!(!square_index.contains_point(Coord { x: -1.0, y: 0.5 }));

        // FAILS: on left boundary, directly left of extra point
        assert!(!square_index.contains_point(Coord { x: -1.0, y: 0.0 }));

        // PASSES: Outside (left) of square, directly left of extra point
        assert!(!square_index.contains_point(Coord { x: -2.0, y: 0.0 }));
    }
```

Note these fail with and without the changes in https://github.com/adamreichold/sif-itree/pull/2

The non-indexed counterparts pass — e.g.:
```rust
assert!(!square.contains(&Coord { x: -1.0, y: 0.5 }));
assert!(!square.contains(&Coord { x: -1.0, y: 0.0 }));
```

michaelkirk:
> I've come across some more failing test cases dealing with points on the boundary.

This seems like a simple fix! See https://github.com/georust/geo/tree/mkirk/fast_point_in_multipolygon

urschrei:
> > I've come across some more failing test cases dealing with points on the boundary.
> 
> This seems like a simple fix! See https://github.com/georust/geo/tree/mkirk/fast_point_in_multipolygon

Incorporated!

michaelkirk:
Excellent. I've pushed up some additional (passing) tests (EDIT: to https://github.com/georust/geo/tree/mkirk/fast_point_in_multipolygon) which you may or may not want to include.

To restate my understanding: 

There's currently a bug, which can be fixed by https://github.com/adamreichold/sif-itree/pull/2

So assuming that gets merged, we'll wait for a release, and incorporate an InclusiveRange query like this: https://github.com/georust/geo/commit/0c228ca7cb0ecf68e4450bc43fc70b12459f81d6

The only other thing, after working with it a bit, it itches me that `contains_POINT` takes a `Coord` (not a `Point`). But given that I didn't notice it until now, it's probably not a big deal. Feel free to ignore unless the spirit moves you.

urschrei:
> Excellent. I've pushed up some additional (passing) tests (EDIT: to https://github.com/georust/geo/tree/mkirk/fast_point_in_multipolygon) which you may or may not want to include.
> 
> To restate my understanding:
> 
> There's currently a bug, which can be fixed by [adamreichold/sif-itree#2](https://github.com/adamreichold/sif-itree/pull/2)
> 
> So assuming that gets merged, we'll wait for a release, and incorporate an InclusiveRange query like this: [0c228ca](https://github.com/georust/geo/commit/0c228ca7cb0ecf68e4450bc43fc70b12459f81d6)
> 
> The only other thing, after working with it a bit, it itches me that `contains_POINT` takes a `Coord` (not a `Point`). But given that I didn't notice it until now, it's probably not a big deal. Feel free to ignore unless the spirit moves you.

Sounds good. You make a good…`Point` (😎), though. And also: https://github.com/georust/geo/pull/1401#issuecomment-3279086428, which I think I agree with. But that means finding a new permanent (because `pub`) home for `IntervalTreeMultiPolygon`, and `relate` (where e.g. `PreparedGeometry` lives) also doesn't seem quite right.

michaelkirk:
And another passing thought, rather than `contains_point`, how would you feel about `impl Contains<Rhs=Coord<T>> for IntervalTreeMultiPolygon<T>` (and `impl Contains<Rhs=Point<T>> for IntervalTreeMultiPolygon<T>`)

michaelkirk:
eep! I keep forgetting that IntervalTreeMultiPolygon isn't public.

Feel free to disregard. Or if you want to make it public, I'm happy to move PreparedGeometry somewhere else too so that ITMP doesn't get lonely. 

Maybe:
```rust
use geometry::indexed::{IntervalTreeMultiPolygon, PreparedGeometry}
```

urschrei:
indexed seems like a good home. I might land that first and then rebase this.

adamreichold:
> So assuming that gets merged, we'll wait for a release, and incorporate an InclusiveRange query like this: https://github.com/georust/geo/commit/0c228ca7cb0ecf68e4450bc43fc70b12459f81d6

Released as version 0.4.0 on crates.io.

urschrei:
> And another passing thought, rather than `contains_point`, how would you feel about `impl Contains<Rhs=Coord<T>> for IntervalTreeMultiPolygon<T>` (and `impl Contains<Rhs=Point<T>> for IntervalTreeMultiPolygon<T>`)

These now both refer to a `pub(crate)` `containment()` method on `IntervalTree`.

geo/src/algorithm/contains/polygon.rs
```diff
@@ -164,6 +182,62 @@ mod test {
...
+    #[test]
+    fn test_point_on_boundary_indexed_polygon() {
+        // Square around the origin with an extra vertex on the right side
+        let square: MultiPolygon = wkt!(MULTIPOLYGON(((-1 1, 1 1, 1 -1, -1 -1, -1 1)))).convert();
+        let square_index = IntervalTreeMultiPolygon::new(&square);
+
+        // PASSES: top left corner
+        assert!(!square_index.contains(&Coord { x: -1.0, y: 1.0 }));
+
+        // FAILS: on left bondary
```
michaelkirk: You can remove my comments now that everything is copacetic.

michaelkirk (approved):
Wonderful. Love it.

urschrei:
@PaulWagener Thanks for all your work on this. We wouldn't have gotten here without you!
@gauteh Hope this is the improvement you needed – please let us know if it ends up being a fix for #1184.

gauteh:
Great! I'll test as soon as possible, and report back! Do I need to construct multipoint or can I test many points? I currently run the points in parallel.

adamreichold:
> Great! I'll test as soon as possible, and report back! Do I need to construct multipoint or can I test many points? I currently run the points in parallel.

Since the index is reused between the points of a multi-points, it would probably be ideal if you would batch your points and process the batches in parallel if that is possible.

PR #1403, opened 2025-08-27
replace custom point sort with lex_cmp from utils

cookiedan42:
---

Realised that the point sort which I added for MultiPoint contains can be replaced by the coord sort

### DIFF
--- a/geo/src/algorithm/contains/point.rs
+++ b/geo/src/algorithm/contains/point.rs
@@ -1,6 +1,7 @@
 use super::{impl_contains_from_relate, impl_contains_geometry_for, Contains};
 use crate::algorithm::{CoordsIter, HasDimensions};
 use crate::geometry::*;
+use crate::utils::lex_cmp;
 use crate::{CoordNum, GeoFloat};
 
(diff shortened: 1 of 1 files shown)


cookiedan42:
closing because using lex_cmp is significantly slower in benchmarks 🙃
```
Multipoint contains multipoint 10000 (Contains trait)
                        time:   [70.942 µs 71.304 µs 71.692 µs]
                        change: [+46.269% +47.304% +48.279%] (p = 0.00 < 0.05)
                        Performance has regressed.

Multipoint not contains multipoint 10000 (Contains trait)
                        time:   [73.829 µs 74.535 µs 75.257 µs]
                        change: [+51.497% +53.011% +54.456%] (p = 0.00 < 0.05)
                        Performance has regressed.

Multipoint not contains multipoint 10000 best case (Contains trait)
                        time:   [38.534 µs 39.490 µs 40.473 µs]
                        change: [+140.20% +145.79% +151.31%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

Issue #1129, opened 2023-12-26
Missing `contains` implementations


kylebarron:
Just making a stub here to keep track of a couple missing `contains` implementations:
```
the trait bound `geo::MultiPoint: geo::Contains<geo::Geometry>` is not satisfied
the trait bound `geo::MultiPolygon: geo::Contains<geo::Geometry>` is not satisfied
```

michaelkirk:
Fixed by #1404 I believe.

PR #1404, opened 2025-08-27
add missing contains implementations

cookiedan42:
---

Add tests cases for permutations of contains

Add derived implementations for #1129 

Add implementations for Coord.contains based on existing Point.contains implementations  
(Not fully sure how desirable this is, included for coverage)

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -45,7 +45,7 @@
 - BREAKING: Break up blanket implementation of `Intersects<LineString>` into specific traits
   - faster implementations for `Rect`, `Triangle`, `MultiPolygon`, `Polygon` intersects `LineString`
   - <https://github.com/georust/geo/pull/1379>
-
+- Added `Contains` implementation for all remaining geometries.  
 
 ## 0.30.0 - 2025-03-24
 
--- a/geo/src/algorithm/contains/coordinate.rs
+++ b/geo/src/algorithm/contains/coordinate.rs
@@ -1,0 +1,142 @@
+use super::{ impl_contains_geometry_for, Contains};
+use crate::algorithm::{CoordsIter, HasDimensions};
+use crate::geometry::*;
+use crate::{CoordNum, GeoFloat};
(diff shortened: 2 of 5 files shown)


michaelkirk:
I don't see any defects in this PR, but I'm conflicted about it. 

We have this weird distinction between point and coord, but they are semantically equivalent. See #15 

We are inconsistent about universally implementing for Coord everything that we implement for Point.

We could continue to increase our "coverage" by gradually copying all Point implementations and then `s/Point/Coord`, but is it actually helpful for people? And what's the cost of duplicating this code? e.g. if there's a bug in one of these impls we have to update it in two places. Is it worth it?

Making the point impl delegate to the coord impl might be one way to ameliorate this.
Introducing a Point/Coord trait might be another way to solve this. 
Making a big (yuge) breaking change in geo-types and getting rid of `Coord` might be another way.

Or we could just merge this, and I could stop making it the straw that broke the camels back.

I'm not mad that you opened this pull request. I'm just trying to explain why it's still sitting here. It's not being ignored, I just have a (small) spiritual crisis every time I look at it. 😆

urschrei:
Might it be helpful to create an issue for discussion of the "larger issue"? As you say, it's a good PR that just happens to raise a lot of questions as an unintended side effect. If we did that, we _could_ "just" merge this and pinky-swear to discuss the bigger picture. I don't think there's a rush btw, just that we should try to decouple things. We could also open a new discussion issue and continue to let this simmer for a while – that's also progress!

michaelkirk:
> Might it be helpful to create an issue for discussion of the "larger issue"?

I think that's #15

frewsxcv (approved):
Looks good to me. I think as long as we have both Point and Coord, it's fine to add implementations like this

frewsxcv:
@michaelkirk @urschrei Any blockers or concerns here?

michaelkirk:
No blockers! I'm just looking for hills to die on 😂

PR #1406, opened 2025-08-29
Update geo to rust 2024

michaelkirk:
---

Almost all the line diffs are changes from `cargo fmt`. I broke the commits apart for easier review.

### DIFF
--- a/geo-bool-ops-benches/Cargo.toml
+++ b/geo-bool-ops-benches/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "geo-bool-ops-benches"
 version = "0.1.0"
-edition = "2021"
+edition = "2024"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
--- a/geo-bool-ops-benches/benches/boolean_ops.rs
+++ b/geo-bool-ops-benches/benches/boolean_ops.rs
@@ -3,7 +3,7 @@ use std::f64::consts::PI;
 use criterion::{measurement::Measurement, *};
 use geo::algorithm::{BooleanOps, Rotate};
 
-use rand::{thread_rng, Rng};
+use rand::{Rng, thread_rng};
 use rand_distr::Standard;
 
 #[path = "../../geo/benches/utils/random.rs"]
--- a/geo-bool-ops-benches/src/tests.rs
+++ b/geo-bool-ops-benches/src/tests.rs
@@ -1,4 +1,4 @@
-use anyhow::{bail, Context, Result};
+use anyhow::{Context, Result, bail};
 use geo::prelude::*;
 use geo_booleanop::boolean::BooleanOp as OtherBOp;
 use geo_types::*;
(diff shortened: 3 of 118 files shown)


urschrei:
Is this purely a result of the recommended "here's how to update to the 2024 edition" process, followed by the "gen" changes, then fmt? I.e. two big "mechanical" commits?

michaelkirk:
CI is failing because:

```
$ cargo fmt --all -- --check
Diff in /Users/mkirk/src/georust/geo/geo-bool-ops-benches/benches/../../geo/benches/utils/random.rs:4:
 use geo::algorithm::{BoundingRect, ConcaveHull, ConvexHull, MapCoords, Rotate};
 use geo::geometry::*;
 
-use rand::{Rng, thread_rng};
+use rand::{thread_rng, Rng};
```

...and yet `cargo fmt --all` makes no changes, so I'm confused.

I'm not sure if it's related, but it might be complicating things that we have multiple rust editions across the workspace. It should be fine - `cargo fmt --all` respects the per-crate rust edition, but maybe adding the `--check` breaks this?

urschrei:
There you go!

urschrei (approved):

michaelkirk:
My approach was:

- bump edition to 2024
- fix compiler errors (For us, that was only the `gen` keyword collision)
- run `cargo fmt` and `cargo test`
- look over https://doc.rust-lang.org/edition-guide/rust-2024/index.html for anything that would "silently" break, but didn't see anything relevant.

Then I remembered about `cargo fix --edition`, so I went back and went through the migration again. The changes suggested by `cargo fix --edition` update your old code in an attempt to maintain the legacy semantics of the code. In our case, they were all superfluous, so I didn't commit them. For example, we're not relying on Drop implementations in any of our `if let` statements, so any changes to the lifetimes therein are irrelevant.

michaelkirk:
> There you go!

Just for my own notes in case this comes back to confuse me later: I also updated the bool-ops-benchs to rust2024. 

TBH, I don't understand why this fixed the issue, because if the problem **was** due to having a mix of editions, there are still several other crates that are rust2021. `cargo fmt --all -- --check` seems happy with them, so who am I to argue. 🤷

urschrei:
> My approach was:
> 
> 
> 
> - bump edition to 2024
> 
> - fix compiler errors (For us, that was only the `gen` keyword collision)
> 
> - run `cargo fmt` and `cargo test`
> 
> - look over https://doc.rust-lang.org/edition-guide/rust-2024/index.html for anything that would "silently" break, but didn't see anything relevant.
> 
> 
> 
> Then I remembered about `cargo fix --edition`, so I went back and went through the migration again. The changes suggested by `cargo fix --edition` update your old code in an attempt to maintain the legacy semantics of the code. In our case, they were all superfluous, so I didn't commit them. For example, we're not relying on Drop implementations in any of our `if let` statements, so any changes to the lifetimes therein are irrelevant.

Ah, the fix --edition switch is what I was wondering about. Makes sense – queue when you're ready. How do you feel about squashing?

Issue #1402, opened 2025-08-27
Edge case for LineString contains Line


cookiedan42:
The current implementation of `LineString contains Line` assumes that we encounter the covering segments of LineString in a "sorted" order which covers the Line starting from one end, enabling a "trimming" of the Line to begin from one vertex and progress toward the other vertex. It iterates over the segments twice to ensure that it 

Initial ideas:
1. Iterate over the collinear intersecting segments more times
   - Terminating on a full loop with no changes should be no slower than current implementation
1. Orient and Sort the collinear intersecting segments before running the "trimming " algorithm
   - Only one loop needed on the sorted segments

Reproduction test case:

LineString which first follows the blue path, then follows the red path should contain `LINE(0 0, 10 0)`  
<img width="546" height="143" alt="Image" src="https://github.com/user-attachments/assets/41c3a613-83ff-447b-8750-4baf20fec489" /><br/>

```rust
#[cfg(test)]
mod test {
    use crate::{Line,LineString, Validation};
    use crate::{wkt,Convert};
    use crate::{Contains,Relate};

    #[test]
    fn triangles(){
        let ln: Line<f64> = wkt! {LINE(0 0, 10 0)}.convert();
        let ls: LineString<f64> = wkt! {LINESTRING(0 0, 1 1, 2 0, 4 0, 5 1, 6 0, 8 0, 9 1,10 0, 8 0, 7 -1, 6 0, 4 0, 3 -1, 2 0 , 0 0 )}.convert();

        // ln and ls are valid
        assert!(ln.is_valid());
        assert!(ls.is_valid());

        assert_eq!(
            ls.relate(&ln).is_contains(), // true
            ls.contains(&ln) // false
        );

    }
}
```

PR #1409, opened 2025-09-04
Linestring contains line bugfix

cookiedan42:
---

add test case and fix bug #1402 

1. would more bench cases be needed?
1. thoughts on adding dedicated multilinestring.contains(line) and multilinestring.contains(linestring) implementations

### DIFF
--- a/geo/src/algorithm/contains/line_string.rs
+++ b/geo/src/algorithm/contains/line_string.rs
@@ -2,7 +2,6 @@ use super::{Contains, impl_contains_from_relate, impl_contains_geometry_for};
 use crate::Orientation;
 use crate::algorithm::kernels::Kernel;
 use crate::geometry::*;
-use crate::intersects::value_in_range;
 use crate::{CoordNum, GeoFloat, GeoNum, HasDimensions};
 
 // ┌────────────────────────────────┐
@@ -46,36 +45,44 @@ where
             return self.contains(&line.start);
         }
 
-        // orient the other segment
-        let mut line = if line.start.x > line.end.x {
-            Line::new(line.end, line.start)
-        } else {
-            *line
-        };
-
         let is_vertical = line.start.x == line.end.x;
 
-        let candidates: Vec<_> = if is_vertical {
+        // pre-order the line so that we can use the faster overlap check
+        let line = if is_vertical {
+            if line.start.y > line.end.y {
+                Line::new(line.end, line.start)
+            } else {
+                *line
+            }
+        } else {
+            if line.start.x > line.end.x {
+                Line::new(line.end, line.start)
+            } else {
+                *line
+            }
+        };
+
+        let candidates: Vec<(T, T)> = if is_vertical {
             self.lines()
-                .filter(|segment| y_overlap(&line, segment))
+                .filter(|segment| overlap::y_overlap(&line, segment))
                 .filter(|segment| is_collinear(&line, segment))
                 .map(|segment| {
                     if segment.start.y < segment.end.y {
-                        segment
+                        (segment.start.y, segment.end.y)
                     } else {
-                        Line::new(segment.end, segment.start)
+                        (segment.end.y, segment.start.y)
                     }
                 })
                 .collect()
         } else {
             self.lines()
-                .filter(|segment| x_overlap(&line, segment))
+                .filter(|segment| overlap::x_overlap(&line, segment))
                 .filter(|segment| is_collinear(&line, segment))
                 .map(|segment| {
                     if segment.start.x < segment.end.x {
-                        segment
+                        (segment.start.x, segment.end.x)
                     } else {
-                        Line::new(segment.end, segment.start)
+                        (segment.end.x, segment.start.x)
                     }
                 })
                 .collect()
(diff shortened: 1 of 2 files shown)


cookiedan42:
downgrade to draft because I figured out how robust/naive kernel works and need to rework this

cookiedan42:
performance slowdown when using robust predicate 😢
```
LineString not contains LineString (Contains trait)
                        time:   [109.83 ns 110.27 ns 110.79 ns]
                        change: [-46.732% -46.549% -46.360%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) low severe
  1 (1.00%) low mild
  5 (5.00%) high mild
  3 (3.00%) high severe

LineString contains LineString (Contains trait)
                        time:   [13.250 µs 13.271 µs 13.295 µs]
                        change: [+6.6774% +7.0203% +7.3883%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (2.00%) low mild
  2 (2.00%) high mild
  2 (2.00%) high severe
```

geo/src/algorithm/contains/line_string.rs
```diff
@@ -44,60 +45,53 @@ where
...
+            // faster method using knowledge that segments are collinear to line
+            .filter(|segment| {
+                greater_than(&line.end, &segment.start) && greater_than(&segment.end, &line.start)
+            })
+            .collect();
+
+        let mut changed = true;
+        while changed {
+            changed = false;
+            for candidate in candidates.iter() {
```
michaelkirk: I'm working on understanding this algorithm. I think I'm almost there. Can you verify my understanding?

Up until this point in the code, we've built up our `candidates`.  Every segment in `candidates` overlaps `line` to some extent.

We know candidates overlap `line` because we've verified that candidate segments are collinear with `line`, and we've used `greater_than` to verify that they overlap in the same x domain[^1]. 

Note: because every `candidate` segment is collinear with `line`, we also know that each `candidate` is collinear with each other, including adjacent segments.

We might have something like:

candidates = LINESTRING(1 1,3 3,5 5)
line = LINE(2 2,4 4)

The logic after my comment here takes care of "chomping" off pieces of `line` with the overlapping `candidate` segments.

Is that right?


[^1] Question 1: If we know the `candidates` are collinear, why do we even need to check the `y` value in `greater_than`? Isn't it sufficient to check only that they overlap in the `x` domain? By contradiction: show me an example where two collinear segments overlap in the x domain but not the y.

Question 2: If indeed we can filter by x_overlap, I'd expect that to be cheaper than checking orientation. So would it be faster to filter by x_overlap first *before* doing the collinear filter?

A couple good test cases would be input that starts and/or ends on candidate segment boundaries. To satisfy *contains* recall that the `line` must not touch the boundary of the linestring (`line` cannot touch the first point nor the last point).

michaelkirk (changes requested):
This looks promising, but I have some questions!

geo/src/algorithm/contains/line_string.rs
cookiedan42: Yep you got the logic behind it

(Question 1) Two vertical collinear segments would have equal x value but not overlap.  
(Question 2) Inspired a modified implementation which performs slightly better (but still a slowdown over the original)

Clarification on the test cases, to my knowledge, not touching the boundary is a constraint of `contains_properly` and not `contains`?

cookiedan42:
I broke my filtering 😞

cookiedan42:
reworked filtering, considers only x unless vertical line, in which case we consider only y

benches better 
```
LineString not contains LineString (Contains trait)
                        time:   [102.75 ns 103.24 ns 103.74 ns]
                        change: [-49.171% -48.887% -48.598%] (p = 0.00 < 0.05)
                        Performance has improved.
  6 (6.00%) low mild
  6 (6.00%) high mild
  2 (2.00%) high severe

LineString contains LineString (Contains trait)
                        time:   [12.461 µs 12.506 µs 12.555 µs]
                        change: [+1.0187% +1.5318% +2.0135%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) low mild
  4 (4.00%) high mild
```

geo/src/algorithm/contains/line_string.rs
michaelkirk: > (Question 1) Two vertical collinear segments would have equal x value but not overlap.

Of course! Thanks for the correction.

> Clarification on the test cases, to my knowledge, not touching the boundary is a constraint of contains_properly and not contains?

>> A couple good test cases would be input that starts and/or ends on candidate segment boundaries. To satisfy contains recall that the line must not touch the boundary of the linestring (line cannot touch the first point nor the last point).
> I hadn't heard of `contains_properly` before. It looks like this method?
https://shapely.readthedocs.io/en/2.0.7/reference/shapely.contains_properly.html

Sorry, I blew it again. 🤦 

Our `Contains` trait is intended to have the exact same semantics as [DE-9IM contains](https://en.wikipedia.org/wiki/DE-9IM). 

So, for `a.contains(b)`, touching the boundary is irrelevant, but their interiors must overlap (e.g. `b` cannot lay **only** within the boundary of `a`. 

As an example, a Polygon does not `contain` its own border. A LineString does not `contain` its start and end points (unless it is a closed linestring?)

So maybe the only relevant test case would be a degenerate line, like `Line(1 1,1 1)`, and how that interacts with the start/end of the LineString. But that logic is delegated to another method: `line_string.contains(point)`

geo/src/algorithm/contains/line_string.rs
```diff
@@ -146,3 +177,76 @@ where
...
+        && T::Ker::orient2d(l1.start, l1.end, l2.end) == Orientation::Collinear
+}
+
+#[inline]
+fn x_overlap<T: GeoNum>(l1: &Line<T>, l2: &Line<T>) -> bool {
+    let (p1, p2) = if l1.start.x < l1.end.x {
+        (l1.start.x, l1.end.x)
+    } else {
+        (l1.end.x, l1.start.x)
+    };
```
michaelkirk: Just some minor code golf... if you added this here:

```suggestion
    };
    value_in_range(l2.start, p1, p2) || value_in_range(l2.end, p1, p2) 
```

Could you then delete the rest of this method? 
We don't need to care about the ordering of l2, if we know either it's start or end is between l1.

cookiedan42:
put back double ended chomping which was in original algo, it's now tiny bit faster than pre-bugfix  
The branch for x and y path is kinda ugly, but I'm, not too sure what can be done about it


```
LineString not contains LineString (Contains trait)
                        time:   [101.10 ns 101.36 ns 101.70 ns]
                        change: [-50.156% -49.910% -49.687%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) high mild
  1 (1.00%) high severe

LineString contains LineString (Contains trait)
                        time:   [12.057 µs 12.089 µs 12.124 µs]
                        change: [-2.1180% -1.6717% -1.2249%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) low mild
  2 (2.00%) high mild
```

cookiedan42:
## Even more aggressive performance tweaks

1. expand the concept of interval based intersection checking to the whole process of chomping off end bits
    - halves space used in storing the values
    - unifies the very chomping of x and y routes

2. speed up filtering by orienting data before comparison
    - if we have intervals (p1,p2) and (q1,q2) 
    - where p1 < p2 and q1 < q2
    - if p1 < q2 and q1 < p2 
    - then the two intervals must intersect
    - Since (p1,p2) is the already oriented Line, we can execute this interval check using 3 comparisons (1 for ordering `q`, and 2 more for p1 < q2 & q1 < p2
    - compared to 3 * 4 comparisons for naive `value_in_between` with unordered interval bounds

## Benches
```
LineString not contains LineString (Contains trait)
                        time:   [53.802 ns 53.901 ns 53.998 ns]
                        change: [-74.027% -73.948% -73.866%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low severe
  5 (5.00%) low mild
  1 (1.00%) high severe

LineString contains LineString (Contains trait)
                        time:   [7.4358 µs 7.4902 µs 7.5464 µs]
                        change: [-40.983% -40.640% -40.289%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low mild
  2 (2.00%) high mild
```

michaelkirk (approved):
Thanks, looks great!

geo/src/algorithm/contains/line_string.rs
michaelkirk: Whoops, disregard this comment. I made it days ago on an old version of the code. You've since addressed this (and more!) in https://github.com/georust/geo/pull/1409/commits/36b4219336851db7e3c606246bd5144ea472d5d9

PR #1414, opened 2025-09-17
Add 'indexed' module for index-backed geometries

urschrei:
---

This sets the stage for other publicly available index-backed geometries, see #1401

### DIFF
--- a/geo/src/algorithm/relate/geomgraph/index/mod.rs
+++ b/geo/src/algorithm/relate/geomgraph/index/mod.rs
@@ -1,13 +1,20 @@
 mod edge_set_intersector;
-mod prepared_geometry;
 mod rstar_edge_set_intersector;
 mod segment;
 mod segment_intersector;
 mod simple_edge_set_intersector;
-
+#[deprecated(
+    since = "0.31.1",
+    note = "PreparedGeometry has moved to geo::indexed::PreparedGeometry"
+)]
+pub use crate::indexed::PreparedGeometry;
 pub(crate) use edge_set_intersector::EdgeSetIntersector;
-pub use prepared_geometry::PreparedGeometry;
-pub(crate) use prepared_geometry::prepare_geometry;
+
+#[deprecated(
+    since = "0.31.1",
+    note = "prepare_geometry has moved to geo::indexed::prepared_geometry::prepare_geometry"
+)]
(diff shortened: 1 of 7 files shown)


geo/src/algorithm/relate/geomgraph/index/mod.rs
```diff
@@ -1,13 +1,10 @@
 mod edge_set_intersector;
-mod prepared_geometry;
 mod rstar_edge_set_intersector;
 mod segment;
 mod segment_intersector;
 mod simple_edge_set_intersector;
 
 pub(crate) use edge_set_intersector::EdgeSetIntersector;
-pub use prepared_geometry::PreparedGeometry;
```
michaelkirk, line 9: How would you feel about leaving a deprecated re-export to make this non-breaking, and also to point people at the new location?

urschrei: I had the very same thought as I stared into space on the tram a couple of hours ago, feeling vaguely annoyed about having to ship a breaking change for no good reason. So: very positive!

geo/src/algorithm/relate/geomgraph/index/mod.rs
```diff
@@ -1,13 +1,20 @@
 mod edge_set_intersector;
-mod prepared_geometry;
 mod rstar_edge_set_intersector;
 mod segment;
 mod segment_intersector;
 mod simple_edge_set_intersector;
-
+#[deprecated(
+    since = "0.31.1",
```
urschrei, line 7: We don't yet know whether we'll cut 0.31.1 or 0.32.0 next (probably 0.31.1?), so not sure what the best practice is here.

geo/src/algorithm/relate/geomgraph/index/mod.rs
urschrei: Done!

geo/src/algorithm/relate/geomgraph/index/mod.rs
michaelkirk: I think what you've done is the most reasonable.

Adding the key word "deprecated" into the release notes might be a helpful hint to look for such updates, if it in fact becomes a breaking release.

michaelkirk (approved):

PR #1417, opened 2025-09-18
Add conservative caching strategy to CI workflows

urschrei:
---

This is (AFAIK) the simplest caching setup that could possibly work: only dependency builds are cached, and only on successful runs. This _should_ cut build times by ~50 %.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -71,7 +71,8 @@ jobs:
       image: "ghcr.io/georust/geo-ci:proj-${{ needs.set-matrix.outputs.proj-version }}-rust-${{ fromJson(needs.set-matrix.outputs.rust-versions)[2] }}"
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
+      - uses: Swatinem/rust-cache@v2
       - run: rustup component add rustfmt clippy
       - run: cargo fmt --all -- --check
       - run: cargo clippy --all-features --all-targets -- -Dwarnings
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -71,7 +71,11 @@ jobs:
       image: "ghcr.io/georust/geo-ci:proj-${{ needs.set-matrix.outputs.proj-version }}-rust-${{ fromJson(needs.set-matrix.outputs.rust-versions)[2] }}"
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
+      - uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: ". -> target"
```
michaelkirk: Based on my reading of https://github.com/Swatinem/rust-cache it seems like the values you are setting here are redundant with the defaults. It's not a big deal, just redundant, as far as I can tell.

michaelkirk:
Screenshot 2025-09-18 at 10 40 01

I re-ran the job to see how the compared before/after the cache was warmed. It's a nice speed bump. 

The check/build portion is much faster. The actual test running of course isn't sped up nor are the non-cargo steps, plus there's a little overhead managing the cache itself, but it seems like a nice 15%-20% overall speedup.

urschrei:
I only set them in order to be explicit (in case a newer version changes them) but maybe that's being overly defensive 🤔

urschrei:
> Screenshot 2025-09-18 at 10 40 01
> 
> 
> 
> I re-ran the job to see how the compared before/after the cache was warmed. It's a nice speed bump. 
> 
> 
> 
> The check/build portion is much faster. The actual test running of course isn't sped up nor are the non-cargo steps, plus there's a little overhead managing the cache itself, but it seems like a nice 15%-20% overall speedup.

I was hoping for more, but this is a low-complexity change so I'll take it. I'm also experimenting with separate nextest + doctest steps…

michaelkirk:
> I was hoping for more, but this is a low-complexity change so I'll take it. 

I'll take it as a positive sign that the rustc compiler speeds aren't the bottleneck they once were.

PR #1418, opened 2025-09-18
Switch to cargo nextest for geo and geo_types tests

urschrei:
with separate doctest runs.

---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,6 +11,9 @@ on:
 
 name: Run tests
 
+env:
+  CARGO_TERM_COLOR: always
+
 jobs:
   # Set configurations used throughout the workflow
   set-matrix:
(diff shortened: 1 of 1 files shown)


michaelkirk:
I wasn't aware of nextest. I just tried it locally and I prefer the output over vanilla `cargo test`. Though it's a shame it doesn't support doctest.

It seems like it's not any faster given that we still need to run the doctests in CI, plus the overhead of installing the plugin.

without nextest:
https://github.com/georust/geo/actions/runs/17839206956/job/50724040430

with nexttest
https://github.com/georust/geo/actions/runs/17839224445/job/50724248210

Was there some other motivation for including it in CI?

I'm going to incorporate it into my own local workflow for a while to see if I like it, since I don't run doc tests as often. For CI I think there's value in keeping things vanilla, unless there's a substantial win that I'm not seeing.

urschrei:
Yeah, it seems like the time is being dominated by the doctests, bah. I much prefer it and have been running it locally for the same reason (rarely run doctests).

urschrei:
Closing until something changes (doctest support)

PR #1419, opened 2025-09-18
Speed up CI by skipping bench build

michaelkirk:
---

The `bench` build is a "no run" build. So the bench program is built,
but not run. Presumably this was done just to be sure we didn't break
the build.

However this is already largely, and more quickly, accomplished by the
`check --all-targets`.

For context I added the bench build to CI in https://github.com/georust/geo/pull/585


urschrei:
Excellent.

PR #1421, opened 2025-09-19
Fix winding number logic in interval tree multipolygon

urschrei:
---


urschrei:
And apparently not trivially wrong 🫣

PR #1422, opened 2025-09-24
Add weekly grouped dependabot checks for actions

urschrei:
---

This is a weekly check and updates are grouped so it's neither noisy nor annoying to merge them (assuming that we always want to update actions, which is reasonable for our set IMO)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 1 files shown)


michaelkirk:
Oh wait, I misunderstood. 🤦  This is enabling dependabot to update our gh-actions, not the more ambitious change to use dependabot for cargo deps...

Sure, good idea.

urschrei:
This PR only applies dependabot checks to GitHub actions (checkout@v5 etc), in order to stop deprecations breaking the build, as we've now added things like caching actions – it doesn't apply to cargo (though we should revisit that…)

PR #1423, opened 2025-09-25
Add projection_1d module for projecting polygons from 2D to 1D space

urschrei:
---

NB: this is supporting work for a new fast minimum polygon distance impl. It's in a separate module to avoid bloating `distance` with yet more helpers, and it may also have uses beyond the new algorithm.


urschrei:
Don't think we're going to need this, so closing.

PR #1424, opened 2025-09-25
Faster minimum polygon-polygon distance for disjoint polygons

urschrei:
---

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -7,6 +7,7 @@
 - LOF algorithm efficiency improvements due to caching kth distance
 - Add DBSCAN clustering algorithm implementation
 - Add `distance_within` method with default impl for any geometry that implements `Distance`, with similar semantics to the PostGIS [ST_DWithin](https://postgis.net/docs/ST_DWithin.html) function
+- Add fast minimum 1D and 2D Euclidean distance algorithm for linearly separable geometries (#1424)
 
 ## 0.31.0 - 2025-09-01
 
--- a/geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
+++ b/geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
@@ -1,4 +1,7 @@
 use super::{Distance, Euclidean};
+use crate::Centroid;
+use crate::HasDimensions;
+use crate::algorithm::BoundingRect;
 use crate::algorithm::Intersects;
 use crate::coordinate_position::{CoordPos, coord_pos_relative_to_ring};
 use crate::geometry::*;
@@ -6,6 +9,7 @@ use crate::{CoordFloat, GeoFloat, GeoNum};
 use num_traits::{Bounded, Float};
 use rstar::RTree;
 use rstar::primitives::CachedEnvelope;
+use std::ops::ControlFlow;
 
 // Distance is a symmetric operation, so we can implement it once for both
 macro_rules! symmetric_distance_impl {
@@ -175,18 +179,67 @@ symmetric_distance_impl!(GeoFloat, &LineString<F>, &Line<F>);
 impl<F: GeoFloat> Distance<F, &LineString<F>, &LineString<F>> for Euclidean {
     fn distance(&self, line_string_a: &LineString<F>, line_string_b: &LineString<F>) -> F {
         if line_string_a.intersects(line_string_b) {
-            F::zero()
-        } else {
-            nearest_neighbour_distance(line_string_a, line_string_b)
+            return F::zero();
         }
+        if line_string_a.0.is_empty() || line_string_b.0.is_empty() {
+            return F::zero();
+        }
+
+        // Check if bounding boxes are non-overlapping: if so, use project-and-prune optimisation
+        // Safety: both linestrings have been checked for emptiness, so bounding_rect() will return Some
+        let rect_a = line_string_a.bounding_rect().unwrap();
+        let rect_b = line_string_b.bounding_rect().unwrap();
+
+        // Check for bbox separation along both axes
+        let x_separated = rect_a.max().x < rect_b.min().x || rect_b.max().x < rect_a.min().x;
+        let y_separated = rect_a.max().y < rect_b.min().y || rect_b.max().y < rect_a.min().y;
+
+        // Ensure geometries meet minimum size requirements for the fast algorithm
+        // LineStrings need at least 2 coordinates to form valid segments
+        let has_min_coords = line_string_a.0.len() >= 2 && line_string_b.0.len() >= 2;
+
+        if (x_separated || y_separated) && has_min_coords {
+            return separable_geometry_distance_fast(line_string_a, line_string_b, rect_a, rect_b);
+        }
+
+        nearest_neighbour_distance(line_string_a, line_string_b)
     }
 }
 
 impl<F: GeoFloat> Distance<F, &LineString<F>, &Polygon<F>> for Euclidean {
     fn distance(&self, line_string: &LineString<F>, polygon: &Polygon<F>) -> F {
         if line_string.intersects(polygon) {
-            F::zero()
-        } else if !polygon.interiors().is_empty()
+            return F::zero();
+        }
+        if line_string.0.is_empty() || polygon.exterior().0.is_empty() {
+            return F::zero();
+        }
+
+        // Check if bounding boxes are non-overlapping: if so, use project-and-prune optimisation
+        // Safety: both geometries have been checked for emptiness, so bounding_rect() will return Some
+        let rect_ls = line_string.bounding_rect().unwrap();
+        let rect_poly = polygon.bounding_rect().unwrap();
+
+        // Check for bbox separation along both axes
+        let x_separated =
+            rect_ls.max().x < rect_poly.min().x || rect_poly.max().x < rect_ls.min().x;
+        let y_separated =
+            rect_ls.max().y < rect_poly.min().y || rect_poly.max().y < rect_ls.min().y;
+
+        // Ensure geometries meet minimum size requirements for the fast algorithm
+        // LineStrings need at least 2 coordinates; Polygons need at least 4 (triangle + closing vertex)
+        let has_min_coords = line_string.0.len() >= 2 && polygon.exterior().0.len() >= 4;
+
+        if (x_separated || y_separated) && has_min_coords {
+            return separable_geometry_distance_fast(
+                line_string,
+                polygon.exterior(),
+                rect_ls,
+                rect_poly,
+            );
+        }
+
+        if !polygon.interiors().is_empty()
             // FIXME: Explodes on empty line_string
             && ring_contains_coord(polygon.exterior(), line_string.0[0])
         {
@@ -214,6 +267,33 @@ impl<F: GeoFloat> Distance<F, &Polygon<F>, &Polygon<F>> for Euclidean {
         if polygon_a.intersects(polygon_b) {
             return F::zero();
         }
+        if polygon_a.is_empty() || polygon_b.is_empty() {
+            return F::zero();
+        }
+
+        // Check if bounding boxes are non-overlapping: if so, use project-and-sort optimisation
+        // Safety: both polygons have been checked for emptiness, so bounding_rect() will return Some
+        let rect_a = polygon_a.bounding_rect().unwrap();
+        let rect_b = polygon_b.bounding_rect().unwrap();
+
+        // Check for bbox separation along both axes
+        // TODO: do we have anything built-in that does this cheaply?
+        let x_separated = rect_a.max().x < rect_b.min().x || rect_b.max().x < rect_a.min().x;
+        let y_separated = rect_a.max().y < rect_b.min().y || rect_b.max().y < rect_a.min().y;
+
+        // Ensure geometries meet minimum size requirements for the fast algorithm
+        // Polygons need at least 4 coordinates (triangle + closing vertex)
+        let has_min_coords = polygon_a.exterior().0.len() >= 4 && polygon_b.exterior().0.len() >= 4;
+
+        if (x_separated || y_separated) && has_min_coords {
+            return separable_geometry_distance_fast(
+                polygon_a.exterior(),
+                polygon_b.exterior(),
+                rect_a,
+                rect_b,
+            );
+        }
+
         // FIXME: explodes when polygon_b.exterior() is empty
         // Containment check
         if !polygon_a.interiors().is_empty()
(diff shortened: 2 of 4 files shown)


geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -214,6 +215,22 @@ impl<F: GeoFloat> Distance<F, &Polygon<F>, &Polygon<F>> for Euclidean {
         if polygon_a.intersects(polygon_b) {
             return F::zero();
         }
+
+        // Check if bounding boxes intersect - if not, use minimum_separable_distance
```
michaelkirk: This comment confuses me.

We're interested in __x overlap__ not in bbox __intersection__, right?

The phrasing on line 226 makes sense to me, but not this one.

geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -384,6 +401,105 @@ fn ring_contains_coord<T: GeoNum>(ring: &LineString<T>, c: Coord<T>) -> bool {
...
+/// The minimum separable distance between the two polygons along the x-axis
+pub(crate) fn minimum_separable_distance<F: GeoFloat>(poly1: &Polygon<F>, poly2: &Polygon<F>) -> F {
+    use crate::algorithm::projection_1d::Project1D;
+
+    // Project both polygons onto the x-axis while preserving indices
+    let mut projected1 = poly1.project_x();
+    let mut projected2 = poly2.project_x();
+
+    // Sort by x-coordinate (first element of tuple)
+    projected1.sort_unstable_by(|a, b| a.0.partial_cmp(&b.0).unwrap());
```
michaelkirk: Can we avoid these crashes? 

Maybe use total_cmp? I'm somewhat ok with "garbage in garbage out", but I want to try harder to avoid crashing.

geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -384,6 +401,105 @@ fn ring_contains_coord<T: GeoNum>(ring: &LineString<T>, c: Coord<T>) -> bool {
...
+    projected2.sort_unstable_by(|a, b| a.0.partial_cmp(&b.0).unwrap());
+
+    // Calculate midpoints using projected x-coordinates to determine left/right
+    let midpoint1 = if projected1.is_empty() {
+        F::zero()
```
michaelkirk: It's hard for me to reason about `0` as the default here. It could be either left or right depending on the value of midpoint2. Maybe it doesn't matter in the end, but there's a non-trivial amount of code to reason through. Instead, is there some way we could short-circuit the algorithm if the polygons are empty? 

Or guard and use `unreachable` to document that this isn't actually a possible state of execution.

geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -384,6 +401,105 @@ fn ring_contains_coord<T: GeoNum>(ring: &LineString<T>, c: Coord<T>) -> bool {
...
+    Euclidean.distance(&left_linestring, &right_linestring)
+}
+
+/// Creates a 3-vertex LineString from a Polygon index, with the given index as the middle vertex
+fn create_context_linestring<F: GeoFloat>(poly: &Polygon<F>, center_idx: usize) -> LineString<F> {
+    // Only consider the exterior ring since interior rings cannot be closest to external polygons
+    let exterior_coords = &poly.exterior().0;
+    let num_coords = exterior_coords.len();
+
+    if num_coords < 3 || center_idx >= num_coords {
```
urschrei: Not totally sure whether we need this? I think validity checks happen further up the chain.

urschrei:
<s># Benchmarks:

**Polygon Euclidean distance RTree f64**: 5.6049 µs --> 550.00 ns (~ -90.0 %)
**Polygon Euclidean distance rotating calipers f64**: 5.3861 µs --> 503.00 ns (~ -91.0 %)

Now a 10x speedup.</s>

5.6049 µs --> 1.2 µs (5x faster, 80 % reduction)

geo/src/algorithm/projection_1d.rs
```diff
@@ -0,0 +1,285 @@
...
+    #[allow(dead_code)]
+    fn project_y(&self) -> Vec<(T, usize)>;
+}
+
+impl<T> Project1D<T> for Polygon<T>
+where
+    T: CoordFloat,
+{
+    fn project_x(&self) -> Vec<(T, usize)> {
+        let mut projections = Vec::new();
```
frewsxcv: Should we preallocate here with `Vec::with_capacity`?

urschrei: Yep, and that buys us quite a bit of speedup…

geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -384,6 +401,105 @@ fn ring_contains_coord<T: GeoNum>(ring: &LineString<T>, c: Coord<T>) -> bool {
...
+    // Get the indices of the candidate closest points
+    let left_rightmost_idx = left_proj.last().unwrap().1; // Index of rightmost point of left polygon
+    let right_leftmost_idx = right_proj.first().unwrap().1; // Index of leftmost point of right polygon
+
+    // Create 3-vertex LineStrings around each candidate point
+    let left_linestring = create_context_linestring(left_poly, left_rightmost_idx);
+    let right_linestring = create_context_linestring(right_poly, right_leftmost_idx);
+
+    // Calculate the distance between these LineStrings
+    Euclidean.distance(&left_linestring, &right_linestring)
```
michaelkirk: I'm surprised that this is sufficient! I'm going to see if I can break it for a while as a learning exercise. 😈

Is this a novel approach or is this an approach you've seen described/used elsewhere?

geo/CHANGES.md
```diff
@@ -4,6 +4,7 @@
 
 - Move `PreparedGeometry` into a new `indexed` module intended to provide index-backed geometries. `relate::PreparedGeometry` has been deprecated.
 - Use an interval tree for faster (Multi)Point in MultiPolygon checks
+- Add `Project1D trait` to allow projection of Polygons from 2D to 1D along x or y axes
```
michaelkirk: Take it or leave it: It seems confusing to document a trait in the changelog that's not public. Advertising the speedup in euclidean distance seems more relevant.

geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
urschrei: Yes good point.

geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
urschrei: I've just replaced the sort step with an even simpler (and extensively commented) alternative…

geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
urschrei: Yep, we definitely can. That should maybe happen a bit further up (in `distance` maybe) since it applies to other algorithms too.

geo/src/algorithm/projection_1d.rs
```diff
@@ -0,0 +1,285 @@
...
+    /// Project all polygon points onto the Y-axis.
+    ///
+    /// Returns a vector of tuples where:
+    /// - First element: y-coordinate (projected value)
+    /// - Second element: index into the original polygon's vertex array
+    #[allow(dead_code)]
+    fn project_y(&self) -> Vec<(T, usize)>;
+}
+
+impl<T> Project1D<T> for Polygon<T>
```
frewsxcv: Low priority, non-blocking: Rather than having separate implementations for `Polygon` and `MultiPolygon`, could it be a generic implementation on `G` where `G: CoordsIter`? https://docs.rs/geo/latest/geo/algorithm/coords_iter/trait.CoordsIter.html

michaelkirk:
I think I found a failure (see test case).

Screenshot 2025-09-25 at 13 17 26

I'm not sure how dropping the y component could lead to solving a distance calculation like this. The [crunchy data blogpost](https://www.crunchydata.com/blog/inside-postgis-calculating-distance) talks about projecting onto a line between the geometry centers, not necessarily the x-axis. That solution sounds more plausible, but TBH I still haven't understood how that could always be correct.

```rust
 #[test]
        fn test_separable_distance() {
            // Sanity check: `a` is a triangle pointing right. `b` is a rectangle strictly right of `a`.
            //             _
            // a->  |>    |_| <-b
            let a : Polygon = wkt!{ POLYGON ((160 70, 160 320, 400 200, 160 70)) }.convert();
            let b : Polygon = wkt!{ POLYGON ((500 100, 500 300, 700 300, 700 100, 500 100)) }.convert();
            assert_eq!(100.0, Euclidean.distance(&a, &b));

            // Leave the base of `b`, but extend it's upper area to trigger a pathological case in the overlap heuristic.
            // Intuitively, since the closest edge (the left edge of `b`) is still there, it can't be farther than
            // the previous test case.
            //          ____
            //         [_   | <-b
            //           \  |
            //            | |
            //            | |
            // a->  |>    |_|
            let a : Polygon = wkt!{ POLYGON ((160 70, 160 320, 400 200, 160 70)) }.convert();
            let b : Polygon = wkt!{ POLYGON ((450 500, 485 449, 500 400, 500 200, 900 500, 640 520, 450 500)) }.convert();
            // FAILS!
            // assertion `left == right` failed
            // left: 100.0
            // right: 260.72552617647546
            assert_eq!(100.0, Euclidean.distance(&a, &b));
        }
```

urschrei:
Oh thank god. Let me have a think about projecting onto a line between the centroids – only way to find out if it works is to see whether it can be done easily.

urschrei:
Regrettably, this is an almost complete re-write, and the diff is now ~400 LoC. The good news is that probably 50 % of those are comments, and that I understand exactly how this thing works now. The big "concept" that the blog post hand-waves about is the projection step, e.g. going from 2D coords to 1D values. This mystical process is quite simple:

1. It draws a line (let's call it `connector`) between the two polygon bbox centroids, then calculates the slope of a perpendicular to `connector`
2. Given the slope (`k`) and a polygon vertex (`v`), its projected value is the _intercept_ of a line with slope `k` passing through `v` and hitting the x or y axis (we choose one based on `connector`'s slope to avoid division by 0, fp instability etc.)
3. Once we've projected all vertices in both polygons, we sort both lists
4. We arrange one polygon as "left" and one as "right" based on their bbox centroid projection value, then set the Euclidean distance between the _last_ left polygon intercept value's original point and the right polygon's _first_ intercept value's original point as the initial minimum distance.
5. Using this, we perform a clever calculation that derives a maximum projected value delta.
6. This delta is fast to calculate for an intercept pair (it's just subtraction)
7. We start looping through both polygons in intercept-sorted order: the left polygon backwards from the highest value, the right polygon forwards from the lowest.
8. If we had to check each pair of points, this algorithm would be quadratic, but every time the delta between a point from the left polygon and a point from the right polygon is larger than our max delta, we break early and move on to the next point pair. That's what gets us back to `O(n log n)`
9. If it's smaller, we have candidate vertices. We retrieve the line segments associated with each vertex, and do an "expensive" `Euclidean.distance` check, updating our min distance and max delta if it's smaller.
10. The final min distance is the minimum polygon distance, which can (of course) be vertex-vertex, edge-vertex, or edge-edge, which is why we need the "expensive" check in step 9.

The image (yes I ported the projection logic to Python so I could start drawing examples) shows the connector and the points are annotated with their projected values (which are their `x`-axis intercepts along lines perpendicular to `connector`, as can be seen).

polydist_example

urschrei:
I added a fast-path specific test suite, with results verified against PostGIS (repro is inline). NB: these are partially machine-generated.

There's nothing weird or outlandish in the test geometries, alas. I'm trying to come up with something really fiendish that's still linearly separable but exercises the pruning logic a bit more; maybe something with a "left" geometry with a "tentacle" that wraps around its top from behind its bbox centroid and contains the closest point. Hmm.

geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -384,6 +445,371 @@ fn ring_contains_coord<T: GeoNum>(ring: &LineString<T>, c: Coord<T>) -> bool {
...
+/// original geometry and its 1D projection value
+#[derive(Clone, Copy, Debug)]
+struct ProjectedVertex<F: GeoFloat> {
+    /// The 1D projection value of this vertex
+    intercept: F,
+    /// Index into original geometry
+    vertex_idx: usize,
+}
+
+/// Optimized minimum distance calculation for linearly-separable 2D geometries
```
michaelkirk: Documentation nit: Looks like we're using this for 1-D geometries too, right?

Relatedly, I guess polygon_polygon_distance is a bad name, but it's internal so no biggie.

geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -175,18 +179,56 @@ symmetric_distance_impl!(GeoFloat, &LineString<F>, &Line<F>);
...
+        }
+        if line_string_a.0.is_empty() || line_string_b.0.is_empty() {
+            return F::zero();
         }
+
+        // Check if bounding boxes are non-overlapping: if so, use project-and-prune optimisation
+        let bbox_a = line_string_a.bounding_rect();
+        let bbox_b = line_string_b.bounding_rect();
+
+        if let (Some(rect_a), Some(rect_b)) = (bbox_a, bbox_b) {
```
michaelkirk: take it or leave it: Checking for a None bounding box is redundant with the previous `is_empty` checks. (right?) 

This could be slightly re-written to remove the redundant is_empty check, but really not a big deal.

geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -214,6 +256,25 @@ impl<F: GeoFloat> Distance<F, &Polygon<F>, &Polygon<F>> for Euclidean {
...
+            return F::zero();
+        }
+
+        // Check if bounding boxes are non-overlapping: if so, use project-and-sort optimisation
+        let bbox_a = polygon_a.bounding_rect();
+        let bbox_b = polygon_b.bounding_rect();
+
+        if let (Some(rect_a), Some(rect_b)) = (bbox_a, bbox_b) {
+            // Check for bbox separation along both axes
+            // TODO: do we have anything built-in that does this cheaply?
```
michaelkirk: For future work maybe... prepared geometries have a cached bbox, so implementing distance between them could take advantage.

geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -384,6 +445,371 @@ fn ring_contains_coord<T: GeoNum>(ring: &LineString<T>, c: Coord<T>) -> bool {
...
+/// # References
+/// https://www.crunchydata.com/blog/inside-postgis-calculating-distance
+fn distance_polygon_polygon_fast<F: GeoFloat>(
+    linestring_p: &LineString<F>,
+    linestring_q: &LineString<F>,
+) -> F {
+    // Calculate bounding box centroids.
+    // Safety: both polygons have already been checked for emptiness by the caller,
+    // so bounding_rect() will return Some.
+    let bbox_p = linestring_p.bounding_rect().unwrap();
```
michaelkirk: I think in all cases, we've just computed these same bboxes to check for eligibility.

It might be worth passing it in rather than re-computing. It's `O(n)`, so it's unlikely to make a huge difference.

(You could also delete an `unwrap`!)

geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -384,6 +445,371 @@ fn ring_contains_coord<T: GeoNum>(ring: &LineString<T>, c: Coord<T>) -> bool {
...
+    // to the polygon bbox centroid connector.
+    let (slope, use_x_projection) = if delta_x.abs() < delta_y.abs() {
+        // Midpoint connection is more vertical → use horizontal-favouring projection
+        (-delta_x / delta_y, false)
+    } else {
+        // Midpoint connection is more horizontal → use vertical-favouring projection
+        (-delta_y / delta_x, true)
+    };
+
+    // Convenient access to the Polygon coordinate slices
```
michaelkirk:
```suggestion
    // Convenient access to the coordinate slices
```

geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -384,6 +445,371 @@ fn ring_contains_coord<T: GeoNum>(ring: &LineString<T>, c: Coord<T>) -> bool {
...
+    let poly_p_coords = &linestring_p.0;
```
michaelkirk:
```suggestion
    let p_coords = &linestring_p.0;
```

michaelkirk: There's a few more outdated references to `polygon`, I'll stop calling them all out, but please update.

geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -384,6 +445,371 @@ fn ring_contains_coord<T: GeoNum>(ring: &LineString<T>, c: Coord<T>) -> bool {
...
+    // Step 1: Project all vertices into 1D space
+    // This gives us intercepts + index of original vertex
+    let mut projected_vertices_p =
+        calculate_vertex_intercepts(poly_p_coords, slope, use_x_projection);
+    let mut projected_vertices_q =
+        calculate_vertex_intercepts(poly_q_coords, slope, use_x_projection);
+
+    // Step 2: Sort vertices by intercepts for spatial locality
+    // Intercepts computed from valid polygon coordinates are never NaN, so unwrap is safe
+    projected_vertices_p.sort_unstable_by(|a, b| a.intercept.partial_cmp(&b.intercept).unwrap());
```
michaelkirk: I beg you to `total_cmp` instead of exploding. 🥺

geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -384,6 +445,371 @@ fn ring_contains_coord<T: GeoNum>(ring: &LineString<T>, c: Coord<T>) -> bool {
...
+        };
+
+    // Get segment indices for both geometries
+    let (prev_idx_p, next_idx_p) = get_segment_indices(coords_p, vertex_idx_p, is_closed_p);
+    let (prev_idx_q, next_idx_q) = get_segment_indices(coords_q, vertex_idx_q, is_closed_q);
+
+    // Build segment arrays
+    // separate arrays means it's easy to compute the cross product, and we won't have to alter
+    // the iterator if we ever expand the number of segments
+    let segments_p = [
```
michaelkirk, line 807: TODO: verify that there are degenerate checks for polygons with 1, 2, and 3 points and linestrings with 1 and 2 points.

geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
urschrei: Got rid of the `if let` (and in the two other impls). I can't see how to re-write to remove the `is_empty` check, but happy to fix in a follow-up.

PR #1426, opened 2025-10-01
Add DistanceWithin trait for distance threshold checks

### LINKED ISSUE #1425: Inconsistency with LineString closest_point and LineString contains. Bad precision with coordinates.
Hello, I experienced this issue while working on geo positions, in the algorithm I was implementing.

```rs
#[test]
    fn test_closest_line_contains_point() {
        let location = Point::new(14.25671219, 57.77645104);
        let line_string = LineString::from(vec![(14.2564459, 57.776875), (14.2563051, 57.7767632)]);

        // get the closest point on the line to the location
        if let Closest::SinglePoint(point_on_line) = line_string.closest_point(&location) {
            // the closest point should be on the line
            // fails due to precision issues
            assert!(line_string.contains(&point_on_line));
        }   
    }
```
The above fails due to precision, if I use simpler values, it works perfectly.
I am not confident enough about floating points to say there is a good solution around this for the closest_point/contains algorithms.

If anyone else comes across this, here is a simple workaround to check if a linestring contains a point, using an epsilon of 1 cm.
```rs
// due to bad precision of floats, we need to check if the point is very close to the line
fn line_contains_point(line: &LineString<f64>, point: &Point<f64>) -> bool {
    match line.closest_point(point) {
        Closest::SinglePoint(p) | Closest::Intersection(p) => {
            // due to bad precision of floats, we need to check if the point is very close to the line
            p.haversine_distance(point) < 0.01 // within 1 cm
        }
        Closest::Indeterminate => true,
    }
}
```
Perhaps something similar could be implemented?

urschrei:
---

Closes #1425


urschrei:
Oh good idea. What do you mean by default impl? I am being dense.

michaelkirk:
https://doc.rust-lang.org/book/ch10-02-traits.html#default-implementations

michaelkirk:
Essentially, I'm suggesting something like:

```diff
 trait Distance {
    /// ... existing method signatures here...


+    /// Returns true if the minimum distance between `origin` and `destination` is less than
+    /// or equal to `distance`
+    fn distance_within(&self, origin: Origin, destination: Destination, distance: F) -> bool {
+        self.distance(origin, destination) <= distance
+    }
 }
```

michaelkirk:
Note I had an egregious typo (now fixed) in my example above. 🤦

urschrei:
How's this?

PR #1427, opened 2025-10-04
Add ContainsProperly relate predicate + custom polygon operation

cookiedan42:
---  

# Draft for copywriting and tests

## Context
I run lots of Polygon \~in\~ Polygon checks. The tighter boundary intersection of `ContainsProperly` allows for faster algorithms since other geometry which intersects the boundary in any way can immediately be rejected

## Implementation
1. Added the is_contains_properly method to `Relate` using `[T**FF*FF*]`
2. Added base `ContainsProperly` trait, mostly using the `Relate` method
3. Added custom `ContainsProperly` implementations for combinations of `Polygon` and `MultiPolygon` 

## Documentation
Will be written before undrafting

## Tests
Currently not very comprehensive, but covers some of the edge cases I've encountered  
ContainsProperly also isn't in the JTS `*Relate*.xml` files, which is rather inconvenient

## References:
[JTS](https://locationtech.github.io/jts/javadoc/org/locationtech/jts/geom/prep/PreparedGeometry.html#containsProperly-org.locationtech.jts.geom.Geometry-)
[PostGIS](https://postgis.net/docs/ST_ContainsProperly.html)

### DIFF
--- a/geo/src/algorithm/contains_properly/coordinate.rs
+++ b/geo/src/algorithm/contains_properly/coordinate.rs
@@ -1,0 +1,2 @@
+// use super::impl_contains_properly_from_relate;
+
--- a/geo/src/algorithm/contains_properly/coordinate.rs
+++ b/geo/src/algorithm/contains_properly/coordinate.rs
@@ -1,2 +1,1 @@
 // use super::impl_contains_properly_from_relate;
-
--- a/geo/src/algorithm/contains_properly/mod.rs
+++ b/geo/src/algorithm/contains_properly/mod.rs
@@ -29,7 +29,6 @@ macro_rules! impl_contains_properly_from_relate {
 }
 use impl_contains_properly_from_relate;
 
-
 macro_rules! impl_contains_properly_geometry_for {
     ($geom_type: ty) => {
         impl<T> ContainsProperly<Geometry<T>> for $geom_type
--- a/geo/src/algorithm/contains_properly/mod.rs
+++ b/geo/src/algorithm/contains_properly/mod.rs
@@ -1,0 +1,239 @@
+mod coordinate;
+mod geometry;
+mod geometry_collection;
+mod line;
+mod line_string;
+mod point;
+mod polygon;
+mod rect;
+mod triangle;
+
+pub trait ContainsProperly<Rhs = Self> {
+    fn contains_properly(&self, rhs: &Rhs) -> bool;
+}
+
+macro_rules! impl_contains_properly_from_relate {
+    ($for:ty,  [$($target:ty),*]) => {
+        $(
+            impl<T> ContainsProperly<$target> for $for
+            where
+                T: GeoFloat
+            {
+                fn contains_properly(&self, target: &$target) -> bool {
+                    use $crate::algorithm::Relate;
+                    self.relate(target).is_contains_properly()
+                }
+            }
+        )*
+    };
+}
+use impl_contains_properly_from_relate;
+
+
+macro_rules! impl_contains_properly_geometry_for {
+    ($geom_type: ty) => {
+        impl<T> ContainsProperly<Geometry<T>> for $geom_type
+        where
+            T: GeoFloat,
+        {
+            fn contains_properly(&self, geometry: &Geometry<T>) -> bool {
+                match geometry {
+                    Geometry::Point(g) => self.contains_properly(g),
+                    Geometry::Line(g) => self.contains_properly(g),
+                    Geometry::LineString(g) => self.contains_properly(g),
+                    Geometry::Polygon(g) => self.contains_properly(g),
+                    Geometry::MultiPoint(g) => self.contains_properly(g),
+                    Geometry::MultiLineString(g) => self.contains_properly(g),
+                    Geometry::MultiPolygon(g) => self.contains_properly(g),
+                    Geometry::GeometryCollection(g) => self.contains_properly(g),
+                    Geometry::Rect(g) => self.contains_properly(g),
+                    Geometry::Triangle(g) => self.contains_properly(g),
+                }
+            }
+        }
+    };
+}
+use impl_contains_properly_geometry_for;
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn exhaustive_compile_test() {
+        use geo_types::*;
+        // let c = Coord { x: 0., y: 0. };
+        let pt: Point = Point::new(0., 0.);
+        let ls = line_string![(0., 0.).into(), (1., 1.).into()];
+        let multi_ls = MultiLineString::new(vec![ls.clone()]);
+        let ln: Line = Line::new((0., 0.), (1., 1.));
+
+        let poly = Polygon::new(LineString::from(vec![(0., 0.), (1., 1.), (1., 0.)]), vec![]);
+        let rect = Rect::new(coord! { x: 10., y: 20. }, coord! { x: 30., y: 10. });
+        let tri = Triangle::new(
+            coord! { x: 0., y: 0. },
+            coord! { x: 10., y: 20. },
+            coord! { x: 20., y: -10. },
+        );
+        let geom = Geometry::Point(pt);
+        let gc = GeometryCollection::new_from(vec![geom.clone()]);
+        let multi_point = MultiPoint::new(vec![pt]);
+        let multi_poly = MultiPolygon::new(vec![poly.clone()]);
+
+        // let _ = c.contains_properly(&c);
+        // let _ = c.contains_properly(&pt);
+        // let _ = c.contains_properly(&ln);
+        // let _ = c.contains_properly(&ls);
+        // let _ = c.contains_properly(&poly);
+        // let _ = c.contains_properly(&rect);
+        // let _ = c.contains_properly(&tri);
+        // let _ = c.contains_properly(&geom);
+        // let _ = c.contains_properly(&gc);
+        // let _ = c.contains_properly(&multi_point);
+        // let _ = c.contains_properly(&multi_ls);
+        // let _ = c.contains_properly(&multi_poly);
+
+        // let _ = pt.contains_properly(&c);
+        let _ = pt.contains_properly(&pt);
+        let _ = pt.contains_properly(&ln);
+        let _ = pt.contains_properly(&ls);
+        let _ = pt.contains_properly(&poly);
+        let _ = pt.contains_properly(&rect);
+        let _ = pt.contains_properly(&tri);
+        let _ = pt.contains_properly(&geom);
+        let _ = pt.contains_properly(&gc);
+        let _ = pt.contains_properly(&multi_point);
+        let _ = pt.contains_properly(&multi_ls);
+        let _ = pt.contains_properly(&multi_poly);
+
+        // let _ = ln.contains_properly(&c);
+        let _ = ln.contains_properly(&pt);
+        let _ = ln.contains_properly(&ln);
+        let _ = ln.contains_properly(&ls);
+        let _ = ln.contains_properly(&poly);
+        let _ = ln.contains_properly(&rect);
+        let _ = ln.contains_properly(&tri);
+        let _ = ln.contains_properly(&geom);
+        let _ = ln.contains_properly(&gc);
+        let _ = ln.contains_properly(&multi_point);
+        let _ = ln.contains_properly(&multi_ls);
+        let _ = ln.contains_properly(&multi_poly);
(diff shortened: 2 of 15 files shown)


cookiedan42:
## Concept behind Implementation for Polygon

TLDR: `ContainsProperly` lets us fast fail any scenario where the RHS element intersects Self's boundary, this is a very useful precondition we can use for faster polygon in polygon checks

1. In a valid polygon, holes can only touch each other at points and rings cannot self intersect

2. There are no intersections between rings (enforced by `boundary_intersects`)
2.1. if intersection --> return false
2.2 therefore, with pairwise relation between rings, all rings must be either concentric or disjoint with any other ring

3. For each hole in `Self` which is inside `RHS`'s exterior, if there is no hole in `RHS` which `ContainsProperly` this self_hole, then there must exist some point of `RHS`'s interior which lies outside of `Self`'s interior
3.1. Hence, for `ContainsProperly` to hold, every hole in `Self` must either be inside some hole in `RHS` or outside of `RHS`

## Expanding to MultiPolygon
1. Polygons in a valid MultiPolygon can only touch at a point, therefore there is no case where the a Polygon requires two components of a MultiPolygon to `ContainsProperly`
1.1 if it touches two components of the MultiPolygon, then there it must touch the boundary of these two components and hence cannot be `ContainsProperly`ed by the multipolygon
2. Hence, MultiPolygon operation can be handled by appropriate `any` and `all` iteration

## Fast checking operation
1. for two rings which do not overlap `a` and `b` (ie they are either concentric or disjoint)
4. if any point of `a` is Inside ring `b`, then all of `a` must be inside `b`
2.1 Symmetrical for `b` and `a`
5. if neither of (2) are true then they must be disjoint


## Benchmark 

### Complex input

Complex Polygon (Louisiana)
```
complex polygon contains_properly polygon (Trait)
                        time:   [262.51 µs 264.08 µs 265.80 µs]

complex polygon contains_properly polygon (Relate)
                        time:   [309.64 µs 311.57 µs 313.85 µs]

complex polygon contains_properly polygon (prepared Relate)
                        time:   [5.5367 µs 5.5578 µs 5.5778 µs]
```

### Simple input

~ 20x of prepared Relate
~ 50x of unprepared Relate

Polygon x Polygon
```
contains_properly poly poly (Trait)
                        time:   [117.11 ns 119.50 ns 122.02 ns]

relate prepared poly poly
                        time:   [2.7458 µs 2.7528 µs 2.7601 µs]

contains poly poly (Trait)
                        time:   [6.2445 µs 6.2662 µs 6.2899 µs]

relate poly poly        time:   [6.2607 µs 6.2823 µs 6.3044 µs]
```

MultiPolygon x MultiPolygon
```
contains_properly multipoly multipoly (Trait)
                        time:   [85.860 ns 86.333 ns 86.816 ns]

relate prepared multipoly multipoly
                        time:   [2.7557 µs 2.7657 µs 2.7756 µs]

contains multipoly multipoly (Trait)
                        time:   [5.3473 µs 5.4202 µs 5.4994 µs]

relate multipoly multipoly
                        time:   [5.2378 µs 5.2668 µs 5.2967 µs]
```

However, it has a +700% runtime vs `Relate` on the louisiana contains east_baton_rouge benchmark, so this approach is not universally the better choice

geo/src/algorithm/contains_properly/mod.rs
```diff
@@ -0,0 +1,279 @@
+/// Checks if `rhs` is completely contained within `self`.
```
michaelkirk:
```suggestion
/// Checks if `rhs` is completely contained within the interior of `self`.
```

"contained" and "within" already have meanings slightly at odds with this, so let's be _painfully_ clear.

geo/src/algorithm/contains_properly/polygon.rs
```diff
@@ -0,0 +1,546 @@
...
+        self.coordinate_position(rhs) == CoordPos::Inside
+    }
+}
+
+impl<T> ContainsProperly<Point<T>> for Polygon<T>
+where
+    T: GeoNum,
+{
+    fn contains_properly(&self, rhs: &Point<T>) -> bool {
+        if self.is_empty() || rhs.is_empty() {
```
michaelkirk: geo points can't be empty.

geo/src/algorithm/contains_properly/polygon.rs
```diff
@@ -0,0 +1,546 @@
...
+impl<T> ContainsProperly<MultiPoint<T>> for Polygon<T>
+where
+    T: GeoNum,
+{
+    fn contains_properly(&self, rhs: &MultiPoint<T>) -> bool {
+        if self.is_empty() || rhs.is_empty() {
+            return false;
+        }
+
+        rhs.coords_iter().all(|p| self.contains_properly(&p))
```
michaelkirk, line 41: Probably not a big deal, but you'll be rechecking if polygon.is_empty every time through this loop.

edit to be clear: I think it's fine to leave it. It's a cheap check.

cookiedan42:
Louisiana bench improved to 264.08 µs  
Faster than base Relates for this case too

geo/src/algorithm/contains_properly/coordinate.rs
```diff
@@ -0,0 +1 @@
+// use super::impl_contains_properly_from_relate;
```
michaelkirk, line 1: I see a couple spots in the PR where you've commented out the Coordinate functionality. Can you clarify what your intention is for having this commented out code?

geo/src/algorithm/contains_properly/polygon.rs
```diff
@@ -0,0 +1,563 @@
...
+    fn contains_properly(&self, rhs: &Polygon<T>) -> bool {
+        if self.is_empty() || rhs.is_empty() {
+            return false;
+        }
+
+        // no boundary intersection
+        if boundary_intersects::<T, Polygon<T>, Polygon<T>>(self, rhs) {
+            return false;
+        }
+        // established that pairwise relation betwwen any two rings is either concentric or disjoint
```
michaelkirk:
```suggestion
        // established that pairwise relation between any two rings is either concentric or disjoint
```

geo/src/algorithm/contains_properly/polygon.rs
```diff
@@ -0,0 +1,563 @@
...
+    let Some(rhs_ext_coord) = rhs_poly.exterior().0.first() else {
+        return false;
+    };
+
+    if !self_poly.contains_properly(rhs_ext_coord) {
+        // is disjoint
+        return false;
+    }
+
+    //  let Some(lhs_ext_coord) = self_poly.exterior().0.first() else {
```
michaelkirk: What's the idea with this commented out code?

geo/src/algorithm/contains_properly/polygon.rs
```diff
@@ -0,0 +1,563 @@
...
+    // and hence self does not contains_properly rhs
+    for self_hole in self_poly.interiors() {
+        // empty hole is always covered
+        let Some(self_hole_first_coord) = self_hole.0.first() else {
+            continue;
+        };
+
+        // hole outside of RHS does not affect intersection
+        if coord_pos_relative_to_ring(*self_hole_first_coord, rhs_poly.exterior())
+            != CoordPos::Inside
```
michaelkirk, line 259: Mostly for my own understanding:

Since we've verified the boundaries are disjoint, we know `CoordPos::OnBoundary` is impossible. So this is equally written:


```
if coord_pos_relative_to_ring(*self_hole_first_coord, rhs_poly.exterior()) == CoordPos::Outside {

}
```

geo/src/algorithm/contains_properly/polygon.rs
```diff
@@ -0,0 +1,563 @@
...
+    // basic pairwise test
+    #[test]
+    fn test_poly_for_poly() {
+        let base: Polygon<f64> =
+            wkt! {POLYGON((90 0,90 90,0 90,0 0,90 0),(60 30,60 60,30 60,30 30,60 30))}.convert();
+
+        let donut_inside: Polygon<f64> =
+            wkt! {POLYGON((80 10,80 80,10 80,10 10,80 10),(70 20,70 70,20 70,20 20,70 20))}
+                .convert();
+        let fully_inside: Polygon<f64> = wkt! {POLYGON((20 10,20 20,10 20,10 10,20 10))}.convert();
```
michaelkirk, line 293: Thank you again for writing your tests with WKT! It's so helpful to be able to pop these into JTSTestBuilder.

Image

geo/src/algorithm/contains_properly/polygon.rs
```diff
@@ -0,0 +1,563 @@
...
+            wkt! {POLYGON((59 31,59 59,31 59,31 31,59 31),(55 35,55 55,35 55,35 35,55 35))}
+                .convert();
+        // two concentric donuts
+        let mp_base = MultiPolygon::new(vec![base.clone(), base_2.clone()]);
+
+        let fully_inside: Polygon<f64> = wkt! {POLYGON((20 10,20 20,10 20,10 10,20 10))}.convert();
+        let disjoint: Polygon<f64> =
+            wkt! {POLYGON((150 140,150 150,140 150,140 140,150 140))}.convert();
+
+        let empty_poly = Polygon::empty();
```
michaelkirk, line 513: yay, thanks for using the new `empty` constructors. ❤️

geo/src/algorithm/contains_properly/polygon.rs
```diff
@@ -0,0 +1,563 @@
...
+        assert!(base.contains_properly(&mp1));
+        assert!(mp_base.contains_properly(&mp1));
+        assert!(!base.contains_properly(&mp2));
+        assert!(!mp_base.contains_properly(&mp2));
+    }
+
+    // degenerate polygon should always return false
+    // because it would be intersecting the boundary of the degenerate polygon
+    #[test]
+    fn test_degenerate_self() {
```
michaelkirk, line 539: Thank you for these edge case tests!

geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
```diff
@@ -419,6 +419,23 @@ impl IntersectionMatrix {
         && self.0[CoordPos::Outside][CoordPos::OnBoundary] == Dimensions::Empty
     }
 
+    /// Returns `true` if every point in Geometry `b` is a point of Geometry `a` interior.
+    ///
+    /// # Notes
+    /// - If Geometry `b` has any interaction with the boundary of Geometry `a`, then the result is `false`.
+    /// - Geometry`a` will never contains_properly itself.
+    /// - Matches `[T**FF*FF*]`
+    /// - This predicate is **transitive** but not **reflexive**
```
michaelkirk, line 428: Could you please cross reference the two implementations in their corresponding docs.

Also, it'd be good to mention something about how performance characteristics might change between them.

If you have an intuition for when it's preferable to use `ContainsProperly` vs `Relate`, please include that in the docs too.  As you mentioned, the baton rouge test is a lot slower when `ContainsProperly`. Is it that generally `Relate` works better for larger inputs?

geo/src/algorithm/contains_properly/polygon.rs
cookiedan42: yes

geo/src/algorithm/contains_properly/polygon.rs
cookiedan42: this was leftover reference code from a refactor, will be removed

geo/src/algorithm/contains_properly/coordinate.rs
cookiedan42: The file for coordinate is there as a placeholder for when I have time to resolve the Point impl together. The initial idea was to use Relate as a reference impl for all but that didn't work for coordinate

geo/src/algorithm/relate/geomgraph/intersection_matrix.rs
cookiedan42: Which part of the docs should I put that in?
1. Trait (top level of doc)
2. Trait impl (in the docs on a specific impl)
3. fn in trait (not in docs but shows up in tooling when using the fn)

michaelkirk: Whatever you think is best. We have two different ways to solve the same problem which have different performance characteristics, so it'd be good to alert users of one as to the existence of the other.

michaelkirk:
In the interest of less back and forth, I went ahead and added the cross references in the docs and merged this. Thanks for the PR!

PR #1428, opened 2025-10-06
add Trait operation tests into Relate

cookiedan42:
---

Add a way to more generally test `Trait` implementation against `Relate` equivalents for test data in jts-test-runner  

while working on #1427 , I realized that the jts-test-runner xml files are limited in that they only run the predicates explicitly spelled out in the xml file (and ContainsProperly isn't one of these). 

I'm thinking we could add these as a baseline test set which checks that the Trait implementation matches the Relate results for test data in jts-test-runner

Thoughts and ideas?
1. it does end up running the test case which are explicitly stated in the xml file twice 
1. should be it bi-directional? ie test both A.predicate(B) and B.predicate(A)?

### DIFF
--- a/jts-test-runner/src/lib.rs
+++ b/jts-test-runner/src/lib.rs
@@ -136,7 +136,7 @@ mod tests {
         //
         // We'll need to increase this number as more tests are added, but it should never be
         // decreased.
-        let expected_test_count: usize = 3911;
+        let expected_test_count: usize = 5720;
         let actual_test_count = runner.unexpected_failures().len() + runner.successes().len();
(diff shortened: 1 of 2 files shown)


urschrei:
The wall-clock time difference for `cargo nextest r` is very small, FYI:

~2.9 s --> 3.15 s

cookiedan42:
I've been using basic `cargo test` for all my projects,,, you've just saved me hours/weeks of my life going forward

cookiedan42:
I'll put in the reverse test cases

PR #1429, opened 2025-10-07
Add DBSCAN clustering algorithm implementation

urschrei:
---

https://en.wikipedia.org/wiki/DBSCAN


geo/benches/dbscan.rs
```diff
@@ -0,0 +1,122 @@
...
+
+    // Shuffle points to mix the two semicircles
+    use rand::seq::SliceRandom;
+    points.shuffle(&mut rng);
+
+    points
+}
+
+/// Standardise points to have zero mean and unit variance (like sklearn's StandardScaler)
+fn standardise_points(points: &mut [Point<f64>]) {
```
lnicola, line 52: Nit: I think most Rust code prefers the American spelling

urschrei: Neverrrr (we use UK spelling in a lot of places, e.g. `neighbour` so I think it's OK)

geo/src/algorithm/dbscan.rs
```diff
@@ -0,0 +1,503 @@
...
+///
+/// # Returns
+///
+/// A vector of cluster labels, one for each input point, in the same order as the input points:
+/// - `Some(cluster_id)`: The point belongs to cluster with ID `cluster_id` (starting from 0)
+/// - `None`: The point is considered noise (doesn't belong to any cluster)
+///
+/// # Algorithm
+///
+/// The algorithm works as follows:
```
lnicola, line 31: Oh, this is a lovely explanation of DBSCAN.

geo/src/algorithm/dbscan.rs
```diff
@@ -0,0 +1,503 @@
...
+    // Process each point
+    for point_idx in 0..n {
+        if states[point_idx] != PointState::Unvisited {
+            continue;
+        }
+
+        // Reuse queue for each new cluster
+        queue.clear();
+        queue.extend(
+            tree.locate_within_distance(points[point_idx], epsilon * epsilon)
```
lnicola, line 174: Will this also return `points[point_idx]`, overestimating the neighbour count by 1?

urschrei: I believe so, but this is correct behaviour according to the paper: `min_samples` should include the point in its neighbourhood count (see lines 20 and 33). We also test for for it (`test_dbscan_min_samples_includes_self`)

PR #1432, opened 2025-10-09
Add k-means clustering algorithm

urschrei:
---

This PR adds the k-means clustering algorithm. No point in pretending it won't be a bear to review: the algorithm is complex and has a lot of moving parts, and the optimisations add to that complexity. That having been said, there's a lot of inline documentation and (IMO) robust error-handling, and the implementation has a tunable max iteration setting with a sensible limit; I'm pretty confident it's not going to break or blow up, and the test cases cover the obvious edge cases well.

You have questions? I have (some) answers. In particular, I've chosen to return an `Err` when the algorithm fails to converge, along with data that allows the final output to be evaluated, which should allow the algorithm to be easily re-run with new parameters (that's fairly common, since k-means isn't optimal, so non-convergence is to be expected for certain data and cluster-size combinations. I'm open to iterating on the API design, though.

### DIFF
--- a/geo/Cargo.toml
+++ b/geo/Cargo.toml
@@ -33,6 +33,7 @@ rstar = "0.12.0"
 serde = { version = "1.0", optional = true, features = ["derive"] }
 i_overlay = { version = "4.0.0, < 4.1.0", default-features = false }
 sif-itree = "0.4.0"
+rand = "0.8.0"
 
 [dev-dependencies]
 approx = ">= 0.4.0, < 0.6.0"
@@ -40,7 +41,6 @@ criterion = { version = "0.4", features = ["html_reports"] }
 geo-test-fixtures = { path = "../geo-test-fixtures" }
 jts-test-runner = { path = "../jts-test-runner" }
 pretty_env_logger = "0.4"
-rand = "0.8.0"
 rand_distr = "0.4.3"
 wkt = "0.14.0"
 
@@ -138,4 +138,8 @@ harness = false
 
 [[bench]]
 name = "dbscan"
+harness = false
+
+[[bench]]
+name = "kmeans"
 harness = false
--- a/geo/src/algorithm/kmeans/error.rs
+++ b/geo/src/algorithm/kmeans/error.rs
@@ -1,0 +1,174 @@
+use crate::GeoFloat;
+use rand::distributions::WeightedError;
+
+/// Errors that can occur during k-means++ initialisation
+#[derive(Debug, Clone, PartialEq)]
+pub enum KMeansInitError {
+    /// Input data contains `NaN` coordinates
+    ///
+    /// This typically indicates invalid or corrupted input data. Check your point coordinates
+    /// for NaN values before running k-means.
+    NaNCoordinate,
+
+    /// Input data contains infinite coordinates
+    ///
+    /// This typically indicates invalid or corrupted input data. Check your point coordinates
+    /// for infinite values before running k-means.
+    InfiniteCoordinate,
+
+    /// All points are at identical locations, making clustering impossible
+    ///
+    /// When all points have the same coordinates, k-means++ cannot select diverse initial
+    /// centroids because all distances are zero. This makes the weighted sampling distribution
+    /// degenerate.
+    ///
+    /// Consider:
+    /// - Checking if your data has been properly loaded
+    /// - Adding small random perturbations to break ties if appropriate.
+    DegenerateData,
+
+    /// Failed to create weighted distribution for k-means++ sampling
+    ///
+    /// This error occurs when the random weighted sampling distribution cannot be constructed,
+    /// typically because all weights sum to zero or contain invalid values.
+    WeightedDistributionFailed {
+        /// The underlying error from rand's WeightedIndex
+        error: WeightedError,
+    },
+}
+
+impl std::fmt::Display for KMeansInitError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            KMeansInitError::NaNCoordinate => {
+                write!(f, "Input data contains NaN coordinates")
+            }
+            KMeansInitError::InfiniteCoordinate => {
+                write!(f, "Input data contains infinite coordinates")
+            }
+            KMeansInitError::DegenerateData => {
+                write!(
+                    f,
+                    "All points are at identical locations, making clustering impossible"
+                )
+            }
+            KMeansInitError::WeightedDistributionFailed { error } => {
+                write!(
+                    f,
+                    "Failed to create weighted distribution for k-means++ sampling: {}",
+                    error
+                )
+            }
+        }
+    }
+}
+
+/// Errors that can occur during k-means clustering
+#[derive(Debug, Clone, PartialEq)]
+pub enum KMeansError<T: GeoFloat> {
+    /// Invalid number of clusters requested
+    InvalidK {
+        /// The requested number of clusters
+        k: usize,
+        /// The number of points in the dataset
+        n: usize,
+    },
+
+    /// An empty cluster was encountered and could not be recovered
+    EmptyCluster {
+        /// The iteration number when the error occurred
+        iteration: usize,
+        /// The cluster ID that became empty
+        cluster_id: usize,
+    },
+
+    /// Initialisation failed due to degenerate or invalid data
+    ///
+    /// Contains specific information about the initialisation failure mode.
+    /// See [`KMeansInitError`] for details on each failure type.
+    InitializationFailed(KMeansInitError),
+
+    /// Maximum iterations reached without convergence
+    ///
+    /// The algorithm completed `iterations` iterations without converging (no assignment
+    /// changes and centroid movement below tolerance). The cluster assignments from the
+    /// final iteration are included and may still be useful.
+    ///
+    /// ## Interpreting the Results
+    ///
+    /// Consider using the partial result if:
+    /// - `max_centroid_shift` is relatively small compared to your data scale
+    /// - `changed_assignments` is a small fraction of total points
+    /// - Visual inspection or validation metrics are acceptable
+    ///
+    /// Consider increasing `max_iter` if:
+    /// - `max_centroid_shift` is large relative to tolerance
+    /// - A significant number of points are still changing clusters
+    /// - You require guaranteed convergence
+    MaxIterationsReached {
+        /// Cluster assignments from the final iteration.
+        ///
+        /// These are valid assignments (each point assigned to nearest centroid)
+        /// but may improve with additional iterations.
+        assignments: Vec<usize>,
+        /// Number of iterations completed (equal to max_iter)
+        iterations: usize,
+        /// Maximum distance any centroid shifted in the final iteration
+        max_centroid_shift: T,
+        /// The tolerance threshold for convergence
+        tolerance: T,
+        /// Number of points that changed clusters in the final iteration
+        changed_assignments: usize,
+    },
+}
+
+impl<T: GeoFloat> std::fmt::Display for KMeansError<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            KMeansError::InvalidK { k, n } => {
+                write!(
+                    f,
+                    "Invalid k={}: must be > 0 and <= {} (number of points)",
+                    k, n
+                )
+            }
+            KMeansError::EmptyCluster {
+                iteration,
+                cluster_id,
+            } => {
+                write!(
+                    f,
+                    "Empty cluster {} at iteration {} could not be recovered",
+                    cluster_id, iteration
+                )
+            }
+            KMeansError::InitializationFailed(init_error) => {
+                write!(f, "K-means initialisation failed: {}", init_error)
+            }
+            KMeansError::MaxIterationsReached {
+                iterations,
+                max_centroid_shift,
+                tolerance,
+                changed_assignments,
+                ..
+            } => {
+                let shift_f64 = max_centroid_shift
+                    .to_f64()
+                    .expect("max_centroid_shift must be representable as f64");
+                let tol_f64 = tolerance
+                    .to_f64()
+                    .expect("tolerance must be representable as f64");
+                write!(
+                    f,
+                    "k-means did not converge within {} iterations. \
+                     Final centroid shift: {:.6} (tolerance: {:.6}), \
+                     {} points changed clusters in final iteration. \
+                     Consider increasing max_iter or using the partial result.",
+                    iterations, shift_f64, tol_f64, changed_assignments
+                )
+            }
+        }
+    }
+}
+
+impl<T: GeoFloat> std::error::Error for KMeansError<T> {}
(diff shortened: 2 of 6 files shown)


geo/src/algorithm/kmeans.rs
```diff
@@ -0,0 +1,1235 @@
+//!
+//! Uses Lloyd's algorithm with `k-means++` initialisation for improved cluster quality, accelerated
+//! using [Hamerly's triangle inequality bounds (2010)](https://cs.baylor.edu/~hamerly/papers/2014_pca_chapter_hamerly_drake.pdf). This implementation is optimised for
+//! 2D points with moderate cluster counts (k=2-50).
+//!
+//! _k_-means clustering partitions n observations into _k_ clusters in which each observation belongs
+//! to the cluster with the nearest mean (centroid). Unlike DBSCAN, _k_-means requires specifying the
+//! number of clusters in advance.
```
frewsxcv, line 4: What do you think about moving this paragraph first? I think it's a better opener for describing what this algorithm is rather than the current opener which is implementation details.

geo/src/algorithm/kmeans.rs
```diff
@@ -0,0 +1,1235 @@
...
+        .iter()
+        .enumerate()
+        .map(|(i, centroid_i)| {
+            centroids
+                .iter()
+                .enumerate()
+                .filter(|(j, _)| *j != i)
+                .map(|(_, centroid_j)| Euclidean.distance(*centroid_i, *centroid_j))
+                .fold(T::infinity(), T::min)
+        })
```
frewsxcv: I see this is iterating over each centroid and calculating the distance to every other centroid. As currently written, is it doing some redundant calculations? For example, it'll calculate `Euclidean.distance(centroids[4], centroids[9])` and then later on `Euclidean.distance(centroids[9], centroids[4])`. Does that make sense?

urschrei: Good catch, swapped out for an iterator that exploits the fact that the distances are symmetric.

geo/src/algorithm/kmeans.rs
```diff
@@ -0,0 +1,1300 @@
...
+
+        // Convert to f64 for WeightedIndex, checking for invalid values
+        let distances_f64: Vec<f64> = distances
+            .iter()
+            .map(|&d| {
+                let f = d.to_f64();
+                match f {
+                    Some(val) if val.is_finite() => Ok(val),
+                    Some(_) => Err(KMeansError::InitializationFailed {
+                        reason: "infinite distance encountered during initialization".to_string(),
```
frewsxcv: Non-blocking: Rather than these string-y reasons, what do you think about having a sub-enum here with specific `InitializationFailed` reasons so the user can match on them if they wish to.

geo/Cargo.toml
```diff
@@ -33,6 +33,8 @@ rstar = "0.12.0"
 serde = { version = "1.0", optional = true, features = ["derive"] }
 i_overlay = { version = "4.0.0, < 4.1.0", default-features = false }
 sif-itree = "0.4.0"
+rand = "0.8.0"
+rand_distr = "0.4.3"
```
frewsxcv: From what I can tell this is only used in the benchmarks. If so could you move it to `dev-dependencies`?

geo/src/algorithm/kmeans.rs
urschrei: Refactored errors into their own module, now with sub-errors for things that can go wrong during the initialisation step.

geo/Cargo.toml
urschrei: Good catch: we need `rand`, but not `rand_distr`.

frewsxcv:
@urschrei Can you also add an entry for this in the algorithm index in the `geo` root crate docs?

urschrei:
> @urschrei Can you also add an entry for this in the algorithm index in the `geo` root crate docs?

Done!

urschrei:
Happy to leave it open for another while. It's a pain to review all at once.

geo/src/algorithm/kmeans/mod.rs
```diff
@@ -0,0 +1,1213 @@
...
+//!
+//! For larger _k_ (> 50), the O(k²) centroid-to-centroid distance computation can dominate.
+//!
+//! # Parameters
+//!
+//! - `k`: The number of clusters to form. Must be > 0 and <= number of points.
+//! - `max_radius` (optional): Maximum radius for clusters. If specified, clusters exceeding this radius will be subdivided,
+//!   potentially resulting in more than `k` clusters.
+//! - `max_iter` (optional): Maximum number of iterations. Default: 300.
+//! - `tolerance` (optional): Relative tolerance for convergence based on centroid movement. Default: 0.0001.
```
michaelkirk: relative to what?

geo/src/algorithm/kmeans/mod.rs
```diff
@@ -0,0 +1,1213 @@
...
+//! - `Ok(labels)`: A vector of cluster labels (0-based indices), one for each input point
+//! - `Err(KMeansError::InvalidK)`: If `k` is 0 or greater than the number of points
+//! - `Err(KMeansError::EmptyCluster)`: If an empty cluster cannot be recovered during iteration
+//! - `Err(KMeansError::InitializationFailed)`: If k-means++ initialisation fails due to degenerate data (e.g., all points at same location, NaN/infinite coordinates)
+//! - `Err(KMeansError::MaxIterationsReached)`: If max_iter is reached before convergence. The error contains the partial result with convergence metadata.
+//!
+//! # Algorithm
+//!
+//! The algorithm works as follows:
+//! 1. **Initialisation**: Uses k-means++ for better initial centroids
```
michaelkirk: You've documented it elsewhere in a private method, but could you add a reference to k++ to this public doc also/instead?

```
/// Reference: Arthur & Vassilvitskii (2007), ["k-means++: the Advantages of Careful Seeding"](https://theory.stanford.edu/~sergei/papers/kMeansPP-soda.pdf)
```

geo/src/algorithm/kmeans/mod.rs
```diff
@@ -0,0 +1,1213 @@
...
+//!
+//! - **Well-separated clusters**: Greater separation → tighter bounds → more pruning
+//! - **More iterations**: Bounds tighten progressively, so longer convergence provides more savings
+//! - **Moderate k**: Provides pruning opportunities whilst keeping O(k²) centroid distance overhead manageable
+//!
+//! For larger _k_ (> 50), the O(k²) centroid-to-centroid distance computation can dominate.
+//!
+//! # Parameters
+//!
+//! - `k`: The number of clusters to form. Must be > 0 and <= number of points.
```
michaelkirk: Worth repeating?

```
More than `k` clusters may be returned if `max_radius` is specified.
```

geo/src/algorithm/kmeans/mod.rs
```diff
@@ -0,0 +1,1213 @@
...
+//!     point!(x: 11.0, y: 10.0),
+//!     point!(x: 10.0, y: 11.0),
+//! ]);
+//!
+//! let labels = points.kmeans(2).unwrap();
+//!
+//! // Two clusters should be found
+//! let cluster_0_count = labels.iter().filter(|&l| *l == 0).count();
+//! let cluster_1_count = labels.iter().filter(|&l| *l == 1).count();
+//! assert_eq!(cluster_0_count + cluster_1_count, 6);
```
michaelkirk, line 91: I wonder if it'd be useful to add some kind of helper that iterates through points by cluster (as followup work, not in this PR!)

geo/src/algorithm/kmeans/mod.rs
```diff
@@ -0,0 +1,1213 @@
...
+    ///     point!(x: 0.0, y: 0.0),
+    ///     point!(x: 1.0, y: 1.0),
+    ///     point!(x: 10.0, y: 10.0),
+    ///     point!(x: 11.0, y: 11.0),
+    /// ];
+    ///
+    /// let params = KMeansParams::new(2)
+    ///     .seed(42)           // Reproducible results
+    ///     .max_iter(500)      // More iterations
+    ///     .tolerance(0.001);  // Stricter convergence
```
michaelkirk:
```suggestion
    ///     .tolerance(0.001);  // Less strict convergence
```

The default is 0.0001

geo/src/algorithm/kmeans/mod.rs
```diff
@@ -0,0 +1,1213 @@
...
+    max_radius: Option<T>,
+    max_split_depth: usize,
+    /// Optional seed for random number generator. None = random, Some(seed) = deterministic
+    seed: Option<u64>,
+}
+
+impl<T: GeoFloat> Default for KMeansConfig<T> {
+    fn default() -> Self {
+        Self {
+            k: 0,
```
michaelkirk: Minor take it or leave it: I think you could use composition to avoid repeating some of these defaults:

```
struct KMeansConfig<T: GeoFloat> {
   max_split_depth: usize,
   params: KMeansParams<T>
}
```

geo/src/algorithm/kmeans/mod.rs
```diff
@@ -0,0 +1,1213 @@
...
+impl<T> KMeans<T> for &MultiPoint<T>
+where
+    T: GeoFloat,
+{
+    fn kmeans_with_params(&self, params: KMeansParams<T>) -> Result<Vec<usize>, KMeansError<T>> {
+        kmeans_impl(&self.0, params)
+    }
+}
+
+impl<T> KMeans<T> for [Point<T>]
```
michaelkirk, line 884: How about for a slice too?

geo/src/algorithm/kmeans/mod.rs
```diff
@@ -0,0 +1,1213 @@
...
+        let (nearest_idx, nearest_sq_dist, second_nearest_sq_dist) =
+            find_nearest_and_second_nearest(
+                *point,
+                &centroids,
+                point_sq_norms[i],
+                &centroid_sq_norms,
+            );
+        *assignment = nearest_idx;
+        // Store actual distances (take sqrt of squared distances)
+        *upper = nearest_sq_dist.sqrt();
```
michaelkirk, line 378: To investigate: can we just keep dist_squared?

My only thought is it might be faster. We compare against a user supplied distance tolerance, so we could square that tolerance one time (k_means_config_tolerance_squared), rather than `n` times.

It might not be straight forward - I'd be happy to leave that as potential followup work since it shouldn't change the API in any case.

geo/src/algorithm/kmeans/mod.rs
```diff
@@ -0,0 +1,1213 @@
...
+
+    // Track convergence state
+    let mut iterations_completed = 0;
+    let mut final_max_delta = T::zero();
+    let mut final_changed_count = 0;
+    let mut converged = false;
+
+    // Iterate until convergence
+    for iter in 0..config.max_iter {
+        iterations_completed = iter + 1;
```
michaelkirk: nit: seems like we could trivially remove this var

geo/src/algorithm/kmeans/mod.rs
```diff
@@ -0,0 +1,1213 @@
...
+        let mut changed_count = 0;
+
+        // Compute distance from each centroid to its nearest other centroid
+        let s = compute_centroid_distances(&centroids);
```
michaelkirk:
```suggestion
        let centroid_distances = compute_centroid_distances(&centroids);
```

If the `s` naming is helpful for comparing to a reference implementation, then feel free to leave it (but curse those paper authors and their one letter variables!).

geo/src/algorithm/kmeans/mod.rs
```diff
@@ -0,0 +1,1213 @@
...
+        }
+
+        centroids = new_centroids;
+    }
+
+    // Check if we hit max_iter without converging
+    if !converged {
+        return Err(KMeansError::MaxIterationsReached {
+            assignments,
+            iterations: iterations_completed,
```
michaelkirk: nit: As I understand it, this will always be `config.max_iter`, right? 

So it seems superfluous to track `iterations_completed`, or is there a reason?

geo/src/algorithm/kmeans/error.rs
```diff
@@ -0,0 +1,173 @@
...
+    /// - Checking if your data has been properly loaded
+    /// - Adding small random perturbations to break ties if appropriate.
+    DegenerateData,
+
+    /// Failed to create weighted distribution for k-means++ sampling
+    ///
+    /// This error occurs when the random weighted sampling distribution cannot be constructed,
+    /// typically because all weights sum to zero or contain invalid values.
+    WeightedDistributionFailed {
+        /// Description of the underlying failure from rand's WeightedIndex
```
michaelkirk: Alternatively, I think this could be a WeightedError instance rather than a String.

...I doubt it matters much in practice though, so take it or leave it.

geo/src/algorithm/kmeans/mod.rs
```diff
@@ -0,0 +1,1213 @@
...
+
+        let labels = points.kmeans_with_params(params).unwrap();
+
+        // Should successfully cluster into 2 groups
+        let unique_labels: std::collections::HashSet<_> = labels.iter().copied().collect();
+        assert_eq!(unique_labels.len(), 2);
+    }
+
+    #[test]
+    fn test_kmeans_converges_with_sufficient_iterations() {
```
michaelkirk, line 1189: Based on the other tests, we already know that kmeans *can* converge. A more interesting test would see it **fail** to converge. I briefly tried such a test, but it didn't fail (presumably convergence happened quickly).


```
        // Use low max_iter which is not enough to converge
        let params = KMeansParams::new(2).seed(42).max_iter(1);
        let result = points.kmeans_with_params(params);

        // Should converge successfully
        assert!(result.is_err());
```

Similarly, a test that actually uses the max_radius parameter to produce interesting results would be nice — the current max_radius test is basically a "it doesn't explode" test.

michaelkirk (approved):
Sorry for the slooooow review. It was a joy to read. 

The underlying algorithm makes sense based on your very nice documentation. The implementation is complex enough that it's a little tricky for me to analyze. 

None of my feedback is blocking, but some of the doc changes should be easy enough to incorporate.

geo/src/algorithm/kmeans/mod.rs
urschrei: Going to mark for follow-up: I suspect you're correct and I think it's a straightforward change, but want to be sure.

geo/src/algorithm/kmeans/mod.rs
urschrei: Nah you're right, it's a terrible habit.

geo/src/algorithm/kmeans/mod.rs
urschrei: This should be "absolute" 🤦‍♂️

geo/src/algorithm/kmeans/mod.rs
urschrei: We now have a good test for `max_radius`, but I'm really struggling to write one that actually fails to converge. As you say, most runs converge within very few iterations. I'm going to have to go looking for something pathological, but I've noted the difficulty in the tests module for now.

PR #1433, opened 2025-10-10
Covers base

cookiedan42:
---

Simplification of `Contains` in the opposite direction of #1427 , `Covers` trait allows for simpler Geom in Geom operations because we don't need to be concerned about rhs being inside self's boundary

## Parts implemented
1. New `Covers` Trait 
2. Blanket implementation based on `Relate`
3. Results which can be derived from `intersects`
4. Some results which are equivalent to `Contains`
5. enable jts-tests-runner case for `Covers`

lacking in the test department, but these are the basic derivable implementations

### DIFF
--- a/geo-benches/Cargo.toml
+++ b/geo-benches/Cargo.toml
@@ -56,6 +56,11 @@ path = "src/contains_properly.rs"
 harness = false
 
 [[bench]]
+name = "covers"
+path = "src/covers.rs"
+harness = false
+
+[[bench]]
 name = "convex_hull"
 path = "src/convex_hull.rs"
 harness = false
--- a/geo-benches/src/covers.rs
+++ b/geo-benches/src/covers.rs
@@ -1,0 +1,89 @@
+use criterion::{BenchmarkId, Criterion, Throughput, criterion_group, criterion_main};
+use geo::algorithm::{Contains, Covers};
+use geo::{Convert, wkt};
+use geo::{coord, geometry::*};
+
+fn rect_covers(c: &mut Criterion) {
+    c.bench_function("rect covers line", |bencher| {
+        let rect = Rect::new(coord! { x: 0., y: 0. }, coord! { x: 10., y: 10. });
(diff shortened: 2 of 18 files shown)


cookiedan42:
Bench

Big leap over `Contains` delegated to `Relate`
```
rect covers line        time:   [949.68 ps 951.09 ps 952.49 ps]
                        change: [-1.8060% -1.4820% -1.1652%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) low severe
  5 (5.00%) low mild
  2 (2.00%) high severe

rect contains line      time:   [2.4912 µs 2.5091 µs 2.5262 µs]
                        change: [-1.7329% -1.1366% -0.5625%] (p = 0.00 < 0.05)
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
```

michaelkirk:
Is there a reason that this is a draft? I haven't looked at it closely yet, but directionally it makes sense.

Sorry again for the slow review cycle. I appreciate the new functionality and speedups! I've been busy elsewhere, but trying to steadily get through these geo PRs.

cookiedan42:
A few reasons 
1. I'm pretty confident in the implementations here, but I do want to add tests for anything not relate-derived
1. just generally being busy in cloud/infra headspace 

I probably could clean this up and add tests for the non relate and intersects-derived parts,   
Or I could pare the PR back to just the base trait and relate-derived implementations, and add back the others when I have the time to work on them properly

michaelkirk:
Ok, no rush! I just wanted to be sure you weren't waiting on the repository maintainers for something.

cookiedan42:
Added tests for the non-derived linestring functions to ensure they're correctly different with the contains functions

frewsxcv:
Please add an entry for this somewhere in the `geo` algorithm index:

https://github.com/georust/geo/blob/87b70cc53c3fd7a9750f40460cd427e656a0f4fa/geo/src/lib.rs#L45-L187

frewsxcv (approved):

urschrei:
@cookiedan42 Please squash your commits!

PR #1436, opened 2025-10-15
Remove the `geo-postgis` crate, which has been moved to a separate repository.

frewsxcv:
---

As mentioned in Discord.

New repository: https://github.com/georust/geo-postgis

I have a hunch that nobody is using this crate anymore, especially since it's on such an old version of the `postgis` crate. So if we find that nobody is using it in the future we can consider archiving this repository.


urschrei (approved):

michaelkirk:
Thanks for doing this. I always felt a little weird having it around given that it's not really maintained.

PR #1440, opened 2025-10-24
improve linestring contain coord perf

cookiedan42:
---

Removed necessary enumerate and check in the LineString.contains(Coord) implementation  

benched using boundary of `point in simple polygon` bench
```
point in simple linestring time:   [2.9182 ns 2.9279 ns 2.9396 ns]
                        change: [-28.259% -27.484% -26.754%] (p = 0.00 < 0.05)
                        Performance has improved.
```


michaelkirk:
I'll merge this tomorrow in case someone else wants to weigh in.

lnicola (approved):

PR #1441, opened 2025-10-25
feat(geo-traits): Provide basic structs implementing geo-traits' traits

yutannihilation:
---

This pull request adds structs like `Coord`, `Point`, `LineString`, etc to geo-traits. These structs implement the corresponding trait (i.e. `CoordTrait`, `PointTrait`, `LineStringTrait`, etc) and the conversion from the objects implementing these traits so that we can use these structs to store the geometry data.

The code is mostly derived from [the wkt crate](https://github.com/georust/wkt/tree/main/src/types). I don't know how to add attribution properly, sorry..., but I'm not sure if I can finish this anyway. The main purpose of this pull request is to sho what the implementation would be.

The main differences from wkt are:

- Removed `Display` and conversion from text.
- Added conversion from geo-traits (e.g. `LineString::from_linestring(linestring: impl LineStringTrait<T = T>)`)
- Added more constructors for `Coord` and `Point`
- Added more `From` conversions for `Coord` and `Point`

### DIFF
--- a/geo-traits/src/structs/coord.rs
+++ b/geo-traits/src/structs/coord.rs
@@ -26,11 +26,11 @@ impl<T: Copy> Coord<T> {
         let y = coord.y();
 
         match coord.dim() {
-            Dimensions::Xyzm | Dimensions::Unknown(4) => Self {
+            Dimensions::Xy | Dimensions::Unknown(2) => Self {
                 x,
                 y,
-                z: coord.nth(2),
-                m: coord.nth(3),
+                z: None,
+                m: None,
             },
             Dimensions::Xyz | Dimensions::Unknown(3) => Self {
                 x,
@@ -44,13 +44,14 @@ impl<T: Copy> Coord<T> {
                 z: None,
                 m: coord.nth(2),
             },
-            Dimensions::Xy | Dimensions::Unknown(2) => Self {
+            // For >4 dimension of Unknown, we don't know how to handle it;
+            // we simply discard these extra dimensions.
+            Dimensions::Xyzm | Dimensions::Unknown(_) => Self {
                 x,
                 y,
-                z: None,
-                m: None,
+                z: coord.nth(2),
+                m: coord.nth(3),
             },
-            Dimensions::Unknown(_) => todo!(),
         }
     }
 


geo-traits/src/structs/coord.rs
```diff
@@ -0,0 +1,313 @@
...
+        Self { x, y, z, m }
+    }
+
+    /// Creates a new coordinate from a coordinate trait.
+    pub fn from_coord(coord: &impl CoordTrait<T = T>) -> Self {
+        let x = coord.x();
+        let y = coord.y();
+
+        match coord.dim() {
+            Dimensions::Xyzm | Dimensions::Unknown(_) => Self {
```
kylebarron: I don't know how people intend to use `Dimensions::Unknown`; I usually pass them into `XY` if 2 dimensions; into `XYZ` if 3 dimensions; into `XYZM` if 4 dimensions

geo-traits/src/structs/coord.rs
```diff
@@ -0,0 +1,313 @@
...
+    }
+}
+
+/// Convert from a tuple of (X, Y, Z). If you want to create a `Coord` of
+/// `Dimensions::Xym`, use `Coord::from_xym`.
+impl<T: Copy> From<(T, T, T)> for Coord<T> {
+    fn from((x, y, z): (T, T, T)) -> Self {
+        Self::new(x, y, Some(z), None)
+    }
+}
```
kylebarron: I'm inclined to remove this and only have `from_xyz` because it's non-obvious statically that this passes the third dimension as z

geo-traits/src/structs/geometry.rs
```diff
@@ -0,0 +1,378 @@
...
+            Self::Polygon(g) => g.dimension(),
+            Self::MultiPoint(g) => g.dimension(),
+            Self::MultiLineString(g) => g.dimension(),
+            Self::MultiPolygon(g) => g.dimension(),
+            Self::GeometryCollection(g) => g.dimension(),
+        }
+    }
+
+    /// Create a new Geometry from an objects implementing [GeometryTrait]. The
+    /// result is `None` if the geometry is `Line`, `Rect`, or `Triangle`.
```
kylebarron: I think _either_ we could convert Line to LineString, Rect to Polygon, Triangle to Polygon, or potentially it would make sense to also have those concepts here

geo-traits/src/structs/line_string.rs
```diff
@@ -0,0 +1,196 @@
...
+    /// Create a new LineString from a non-empty sequence of objects implementing [CoordTrait].
+    ///
+    /// This will infer the dimension from the first coordinate, and will not validate that all
+    /// coordinates have the same dimension.
+    ///
+    /// Returns `None` if the input iterator is empty.
+    ///
+    /// To handle empty input iterators, consider calling `unwrap_or` on the result and defaulting
+    /// to an [empty][Self::empty] geometry with specified dimension.
+    pub fn from_coords(coords: impl IntoIterator<Item = impl CoordTrait<T = T>>) -> Option<Self> {
```
kylebarron, line 51: This doesn't need to be `impl IntoIterator`; it can just be `impl Iterator`, right?

geo-traits/src/structs/linestring.rs
```diff
@@ -0,0 +1,196 @@
...
+            None
+        } else {
+            let dim = coords[0].dim();
+            Some(Self::new(coords, dim))
+        }
+    }
+
+    /// Create a new LineString from an objects implementing [LineStringTrait].
+    pub fn from_linestring(linestring: &impl LineStringTrait<T = T>) -> Self {
+        Self::from_coords(linestring.coords()).unwrap()
```
kylebarron: I don't think this unwrap is ok, because empty linestrings are valid

geo-traits/src/structs/multilinestring.rs
```diff
@@ -0,0 +1,220 @@
...
+        }
+    }
+
+    /// Create a new MultiLineString from an objects implementing [MultiLineStringTrait].
+    pub fn from_multilinestring(multilinestring: &impl MultiLineStringTrait<T = T>) -> Self {
+        let line_strings = multilinestring
+            .line_strings()
+            .map(|l| LineString::from_linestring(&l))
+            .collect::<Vec<_>>();
+        let dim = line_strings[0].dimension();
```
kylebarron: ditto, we can't assume input objects are non-empty

geo-traits/src/structs/coord.rs
yutannihilation: I see. Actually I was wondering if I should add this or not. Let's remove.

geo-traits/src/structs/line_string.rs
yutannihilation: I think you are right. I couldn't make it due to some compiler error, but I'll try again...

geo-traits/src/structs/linestring.rs
yutannihilation: Thanks for catching!

geo-traits/src/structs/coord.rs
yutannihilation: Thanks, I'll follow the convention. How do you handle other irregular numbers? Should this function return `Option<Self>` to choose `None` for such cases?

geo-traits/src/structs/line_string.rs
yutannihilation: Sorry, probably I too was confused, but probably you meant we want `.iter()` rather than `.into_iter()` so that we don't need to consume the object, right? If so, it's not about `IntoIterator` vs `Iterator`, and currently it should already be possible because ` &Coord<T>` implements `CoordTrait`.

geo-traits/src/structs/coord.rs
kylebarron: I don't know... I find it really annoying that there is a `Dimensions::Unknown` at all, but then again I only work with `xy`/`xyz`/`xym`/`xyzm` data

geo-traits/src/structs/line_string.rs
kylebarron: Yes, we don't need to consume the object. But taking in `impl IntoIterator` means that the input needs to be able to be _consumed into_ an iterator. If we take in an `impl Iterator` instead, then the input doesn't need to be consumed

yutannihilation: Sorry, I still don't understand your point. For example, if we have `Vec<(f64, f64)>`, I think we typically pass `&[(f64, 64)]` to `impl IntoIterator` argument because the slice implements `IntoIterator`. I think it's fine to have the slice consumed because we still hold the original `Vec`, no? Do you mean changing it to `impl Iterator` might save us from accidentally passing `Vec` itself...?

This might be because I'm not familiar with designing such interface. Do you have some good example of a signature that takes `Iterator` instead of `IntoIterator`?

kylebarron:
I think the main question here is what the scope of `geo-traits` should be, and whether we should ship structs like this inside of `geo-traits`. As it is `geo-traits` only has an _optional_ dependency on `geo-types`. This would create another optional feature that partially overlaps with `geo-types`.

Another possible worry is that if we add more structs to the `geo-traits` crate, then people might start writing APIs in terms of the _structs_ thinking that those are part of the public consumption API of geo-traits, while this PR would be only intended to simplify life for the _producers_ of geo-traits data.

Another option is to create and publish a separate crate with just these structs. That doesn't expand the maintenance burden of crates in the `geo` repo.

In my own work I don't find it that burdensome to use the `wkt` crate to manage these structs.

I'd like to hear from other maintainers too

Issue #1124, opened 2023-12-09
concave_hull returns nonsense results for simple polygons


wlinna:
I'm testing concave_hull with a  simple polygon, and concave_hull returns very weird results. Here are the screenshots

Original polygon
![OriginalPolygon](https://github.com/georust/geo/assets/1473454/882b8dcb-911c-4d62-a3c7-fec73bde6b12)

Concave hull
![FailedConcaveHull](https://github.com/georust/geo/assets/1473454/3307ce7e-a53a-43d9-9678-e6a668859c13)

Here's the test code
```
       let exterior = LineString(vec![
            Coord { x: 5.1734996, y: 1.4128553 },
            Coord { x: 5.1734996, y: 1.5378553 },
            Coord { x: -4.9266067, y: 1.5378562 },
            Coord { x: -4.9266057, y: -8.45025 },
            Coord { x: -4.8016057, y: -8.45025 },
            Coord { x: -4.8016057, y: 1.4128553 },
            Coord { x: 5.1734996, y: 1.4128553 },
        ]);
    
        let polygon = Polygon::new(exterior.clone(), vec![]);
        let hull = polygon.concave_hull(0.5); // Whether the concavity is 0.5 or 4.0 doesn't seem to matter
```

I'm using `geo 0.27.0` with Rustc 1.72

wlinna:
I tried this again with both  `geo 0.27` and `geo 0.28` , and it looks like nothing changed with concave hull itself. Both versions result in  following coordinates

-4.9266057 -8.45025
-4.8016057 -8.45025
5.1734996 1.4128553
5.1734996 1.5378553
-4.9266067 1.5378562
-4.9266057 -8.45025

However, I'm not sure anymore how I produced the original 3D model of the concave hull. When I generate it now, it looks like this

![imagen](https://github.com/georust/geo/assets/1473454/1575ddbc-f819-4570-9e65-1fd7a6016af6)

I checked the printed coordinates visually one by one, and they correspond with the generated triangle mesh.

So at least the algorithm produces a convex hull. But I can't make it produce the original shape.

kirillolenev-dm:
Any updates about that? I have a similar experience when `concave_hull` provides something not close to from expected result, `concavity` param doesn't change much.

michaelkirk:
@kirillolenev-dm - can you provide a minimal reproduction? 

e.g. the input you gave and the output you're expecting?

[...]

Issue #792, opened 2022-03-24
concave_hull appears to be returning convex_hull in at least one case


caewok:
Related: #607 

I was testing `concave_hull` and ran across a potential bug. The following set of points are basically two polygons combined: a square, and a four-sided polygon that overlaps the square, forming something near to a diamond. The concave hull should close in on the intersection points, but instead geo's `concave_hull` is not doing that and instead returning a convex hull. This differs from the result in concaveman. 

```rust
let mp: MultiPoint<f64> = vec![
	(450., 100.),
	(611.1111111111111, 100.),
	(100., 450.),
	(100., 611.1111111111111),
	(275., 1000.),
	(1000., 1000.),
	(1000., 275.),
	(1000., 1000.),
	(100., 100.),
	(1000., 100.),
	(1000., 1000.),
	(100., 1000.),
	(100., 100.),
	(50., 500.),
	(500., 50.),
	(1500., 500.),
	(500., 1500.),
	(50., 500.),
].into();

let expected: Polygon<f64> = Polygon::new(
	LineString::from(vec![
		(100., 100.),
		(100., 450.),
		(50., 500.),
		(100., 611.1111111111111),
		(100., 1000.),
		(275., 1000.),
		(500., 1500.),
		(1000., 1000.),
                 (1500., 500.),
		(1000., 275.),
		(1000., 100.),
		(611.1111111111111, 100.),
		(500., 50.),
		(450., 100.),
		(100., 100.),
	]),
	vec![]);

assert_eq!(mp.concave_hull(1.), expected); // will fail
```

wlinna:
My project also suffers from this.  My `MultiPolygon` looks like this

Issue #607, opened 2021-01-16
Rewrite concave hull algorithm


RestitutorOrbis:
As noted in #541, there is an interest in making the concave hull algorithm converge with the implementation in concaveman. Since, the current implementation diverges by avoiding the use of a linked list and preferring use of RTree over an implementation closer to Concaveman's, it would probably be necessary to re-write the algorithm.

michaelkirk:
Note that, in the meanwhile, https://github.com/georust/geo/pull/635 (k nearest neighbors concave hull) has been merged.

The backstory on this issue, as I remember it...

Our first concave hull algorithm was introduced in https://github.com/georust/geo/pull/480

It was based on https://github.com/mapbox/concaveman, which, I believe, is widely used and often faster than the knn approach.

However, our implementation behavior diverges from the mapbox reference implementation in some cases.

I'm going to rename the issue to clarify what I understand this issue to be about.

frewsxcv:
This would be resolved by https://github.com/georust/geo/pull/1442

PR #1442, opened 2025-11-01
Use concaveman method for concave hull

wwaltersdavis:
---
An overhaul for the concave hull algorithm to be inline with [mapbox/concaveman](https://github.com/mapbox/concaveman/) as raised in [#607](https://github.com/georust/geo/issues/607). Apologies it is quite a large change to the algorithm but have tried to add inline documentation to help.

### Background
The current method differs from concaveman's approach in two key ways:
- First, it selects candidate points for drilling the hull down by proximity to the centroid from a given hull line, doing this by exploring points within a given radius from the centroid. Concaveman however, selects the nearest point to the line as a whole by a depth first search through an R-tree.
- Secondly, once a candidate point is successful, the previous hull line is replaced with the two new hull lines connected to the candidate, these two lines are currently pushed to the front of the queue, whereas concaveman's approach pushes to the back.

These differences in methods cause significantly different results which can be visualised below made from [driving.json](https://github.com/user-attachments/files/23287033/driving.json) (a potential use for this algorithm representing a catchment area from a given point).
driving_2 0_0 0_old

### Solution
Move to an implementation of the concaveman method.

### Differences between the proposed method and concaveman
When porting concaveman across a few minor changes were made.
- In the proposed approach, duplicated interior points are remove from the interior points R-tree when they are accepted as hull points. This stops the hull doubling back on itself and "cutting off" an area.
- Used a vector based approach to log the order of the hull points rather than using a linked list
- Use Geo's inbuilt Euclidean distance functions rather than concavemans. 

### Result
Similar to concaveman
driving_2 0_0 0

### Performance
Using the Norway Mainland benchmark found a +422.59% increase. However this is using a concavity of 2 which means different things for each method and causes different hulls as seen below. The current method produces a hull of length 413 with the proposed method having a length of 1956. 
norway_main_points_2 0_0 0_comparison

## Future concave hull improvements
This method goes part of the way to address [#792](https://github.com/georust/geo/issues/792) and [#1124](https://github.com/georust/geo/issues/1124) however it still does not respect interior geometry boundaries. This is something I am working on also.


geo/src/algorithm/concave_hull.rs
```diff
@@ -87,363 +130,536 @@ where
...
-    max_dist: T,
-    edge_length: T,
-    concavity: T,
-    line_tree: &RTree<Line<T>>,
+    i: usize,
+    prev_i: usize,
+    next_i: usize,
+}
+
+fn line_to_bbox_distance<T>(line: &Line<T>, aabb: &AABB<Coord<T>>) -> T
```
urschrei: It's not immediately obvious what this function is for (I can take a pretty good guess, but I'd prefer not to). Could you document its purpose?

geo/src/algorithm/concave_hull.rs
```diff
@@ -87,363 +130,536 @@ where
...
+        return d3;
+    }
+    let d4 = Euclidean.distance(line, &Line::new(c2, c3));
+    if d4 == T::zero() {
+        return d4;
+    }
+    partial_min(partial_min(d1, d2), partial_min(d3, d4))
+}
+
+fn no_hull_intersections<T>(line: &Line<T>, current_hull_tree: &RTree<Line<T>>) -> bool
```
urschrei, line 354: See previous comment!

geo/src/algorithm/concave_hull.rs
```diff
@@ -1,59 +1,93 @@
...
 
 /// Returns a polygon which covers a geometry. Unlike convex hulls, which also cover
 /// their geometry, a concave hull does so while trying to further minimize its area by
 /// constructing edges such that the exterior of the polygon incorporates points that would
 /// be interior points in a convex hull.
 ///
-/// This implementation is inspired by <https://github.com/mapbox/concaveman>
-/// and also uses ideas from the following paper:
-/// www.iis.sinica.edu.tw/page/jise/2012/201205_10.pdf
+/// This implementation is a port of <https://github.com/mapbox/concaveman>
```
urschrei: It would be extremely helpful to reviewers and to future maintainers if you could provide a brief explanation of how the concaveman algorithm works.

geo/src/algorithm/concave_hull.rs
```diff
@@ -87,363 +130,536 @@ where
...
+    tree_node: RTreeNodeRef<'a, T>,
+    distance: T,
+}
+
+impl<'a, T> Ord for NodeQueueItem<'a, T>
+where
+    T: GeoFloat + RTreeNum,
+{
+    fn cmp(&self, other: &Self) -> Ordering {
+        other.distance.partial_cmp(&self.distance).unwrap()
```
urschrei:
```suggestion
        other.distance.total_cmp(&self.distance)
```

wwaltersdavis: Thanks, have updated

geo/src/algorithm/concave_hull.rs
wwaltersdavis: Yeah 100%, have added

geo/src/algorithm/concave_hull.rs
wwaltersdavis: Thanks, yes have added a description for this

wwaltersdavis:
> Thanks for taking a new run at this! I didn't review the original, so I'm not terribly familiar with the details of this implementation (hence my comments).
> 
> The perf question is interesting: there's nothing obviously inefficient in this implementation at first glance, although the appearance of a `HashMap` is often a source of suspicion. It would be good to get a more thorough sense of the performance compared to the existing implementation, especially with realistic use cases. As I've said elsewhere correctness and elimination of corner cases is more important than perf in the first instance, so I wouldn't regard a single benchmark regression as a block to merging, all else being equal.

Thanks for having a look through! All your comments make sense, hopefully clearer now with an explanation of the algorithm and additional inline docs.

I am currently exploring if there are any overlooked corner cases and will update tests to reflect these.

On the performance side, I have removed the HashMap as it was unnecessary. I am going to have a look further into comparing like-for-like performances as the current benchmark gives too different a result to be that useful as a comparison.

dabreegster:
Thanks a ton for doing this work! I'm using it in an internal project and both the results/perf have been fine so far. I haven't done any detailed testing, but if I hit any weird edge cases, I'll pass them along as tests.

wwaltersdavis:
> Thanks a ton for doing this work! I'm using it in an internal project and both the results/perf have been fine so far. I haven't done any detailed testing, but if I hit any weird edge cases, I'll pass them along as tests.

Thanks mate! Yeah that would be really useful. I have recently added a slight divergence from the concaveman method to fix what I believe is a [potential edge case](https://github.com/mapbox/concaveman/issues/31) and keen to make sure the results still look right.

geo/src/algorithm/concave_hull.rs
```diff
@@ -1,59 +1,114 @@
...
-/// assert_eq!(res.exterior(), &correct_hull);
+/// let hull = points.concave_hull(1.0, 0.0);
+/// assert_eq!(hull, correct_hull);
 /// ```
 pub trait ConcaveHull {
     type Scalar: CoordNum;
-    fn concave_hull(&self, concavity: Self::Scalar) -> Polygon<Self::Scalar>;
+    /// Create a concave hull around the geometry set.
+    ///
+    /// See the [module-level documentation](self) for details on the algorithm and parameters.
```
michaelkirk, line 115: The javascript method has the signature:
```
concaveman(points[, concavity = 2, lengthThreshold = 0])
```
I like how there are reasonable defaults and you don't have to specify all the parameters.

Unfortunately rust doesn't have nice syntax for default arguments. Consider these alternatives:

```rust
// Option 1 - as phrased
polygon.convave_hull(2.0, 0.0)

// Option 2 - Leverage `Option::None` for default values
polygon.convave_hull(Some(2.0), Some(0.0))
polygon.convave_hull(None, None) // same as above
polygon.convave_hull(Some(1.0), None)  // concavity 1.0 with default length_threshold

// Option 3.a: long method names
polygon.convave_hull() // default params
polygon.convave_hull_with_concavity(1.0) // default length_threshold
polygon.convave_hull_with_length_threshold(2.0) // default concavity
polygon.convave_hull_with_concavity_and_length_threshold(1.0, 2.0) // 😢 so long

// Option 3.b - a new option type 
polygon.convave_hull() // defaults params
polygon.convave_hull_with_options(ConcaveHullOptions { concavity: 2.1, length_threshold: 0.1 })
polygon.convave_hull_with_options(ConcaveHullOptions::default().length_threshold(0.1))

// Note: Option 3.a and 3.b are compatible - we could do either or both (my preference) 

// Option 4 - builder pattern?
polygon.convave_hull() // default params
polygon.convave_hull_operation().compute() // default  params

polygon.convave_hull_operation()
  .concavity(2.0)
  .compute()

polygon.convave_hull_operation()
  .concavity(2.0)
  .length_threshold(3.0)
  .compute()

```

Any preferences out there?

For completeness, I'll say we **could** maintain the existing API and expose length_threshold through a new method, but I think we should seize the opportunity to improve the API. It'll also give users a hint that the output of this method is about to change (for the better hopefully!)

urschrei: FWIW 3.b is my preference (I don't think we need both `3.a` **and** `3.b`, but don't feel _strongly_ about it). The "new options type" pattern works well for this kinda thing

wwaltersdavis:
> Thanks for taking a new run at this! I didn't review the original, so I'm not terribly familiar with the details of this implementation (hence my comments).
> 
> The perf question is interesting: there's nothing obviously inefficient in this implementation at first glance, although the appearance of a `HashMap` is often a source of suspicion. It would be good to get a more thorough sense of the performance compared to the existing implementation, especially with realistic use cases. As I've said elsewhere correctness and elimination of corner cases is more important than perf in the first instance, so I wouldn't regard a single benchmark regression as a block to merging, all else being equal.

Have had a proper look at this and feeling confident any edge cases have been caught. I created 100m real world isochrones of varying sizes and all met the criteria of no overlapping polygon edges and no point lying outside the hull.


The performance side was a little bit trickier to test as there is no true definition of what the output from the concave hull function should be. In this case a concavity of 2 causes a vastly different output between the current and proposed method. To address this can measure the complexity of each output hull as the total number of sides. Using this ran 10k real world examples for each could and measured the total complexity. Could then compare somewhat like-for-like with the results below showing a trend of the current method performing better with very low complexity results with less points, but becoming much slower when 1000s of points are being used.

image

geo/src/algorithm/concave_hull.rs
wwaltersdavis: I am happy to implement either or both, though for what its worth I agree and have a slight preference towards 3.b

michaelkirk: Sounds like there's a solid preference for 3b. Would you make those changes?

geo/CHANGES.md
```diff
@@ -15,6 +15,10 @@
 - Renamed features `use-proj`, `use-serde` to simply `proj` and `serde` (removing the `use-` prefix)
   and deprecated the old spelling.
   - <https://github.com/georust/geo/pull/1447>
+- Update `ConcaveHull` algorithm with implementation of [mapbox/concaveman](https://github.com/mapbox/concaveman).
```
michaelkirk, line 18: Great changelog entry, thank you!

geo/src/algorithm/concave_hull.rs
```diff
@@ -1,59 +1,156 @@
...
+        concave_hull_options: ConcaveHullOptions<Self::Scalar>,
+    ) -> Polygon<Self::Scalar>;
+}
+
+impl<T> ConcaveHull for MultiPoint<T>
+where
+    T: GeoFloat + RTreeNum,
+{
+    type Scalar = T;
+    fn concave_hull(&self) -> Polygon<T> {
```
michaelkirk: Not a big deal, but aren't all these per impl `concave_hull` definitions redundant since we have a default definition for this method in the trait?

geo/src/algorithm/concave_hull.rs
```diff
@@ -87,363 +198,703 @@ where
...
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
     }
 }
+impl<'a, T> PartialEq for NodeQueueItem<'a, T>
+where
+    T: GeoFloat + RTreeNum,
+{
+    fn eq(&self, other: &Self) -> bool {
+        self.distance == other.distance
```
michaelkirk: It's always a little bit dicey playing with NaN.

Beyond `PartialEq`, we're declaring `Eq` below, whose contract requires reflexivity, but we won't get that if self.distance = NaN.

The [`Eq` docs](https://doc.rust-lang.org/std/cmp/trait.Eq.html) say:

> The behavior resulting from a logic error is not specified, but users of the trait must ensure that such logic errors do not result in undefined behavior.

I don't expect to get useful output when distance is NaN, but I'm not sure what happens when you break the contract of Equal - for all I know, something like an infinite loop or a panic might be possible.

Anyway, how about `self.distance.cmp(other) == Ordering::Equal` so we can leverage total_cmp to fulfill the contract?

geo/src/algorithm/concave_hull.rs
```diff
@@ -87,363 +198,703 @@ where
...
 
+fn concave_hull_with_options<T>(
+    coords: &mut [Coord<T>],
+    concave_hull_options: ConcaveHullOptions<T>,
+) -> Polygon<T>
+where
+    T: GeoFloat + RTreeNum,
+{
+    // Ensure concavity is non-negative
+    let concavity: T = T::max(T::zero(), concave_hull_options.concavity);
```
michaelkirk, line 530: What would you think about adding a debug_assert so programmers can fix their error?

geo/src/algorithm/concave_hull.rs
```diff
@@ -87,363 +198,703 @@ where
...
 
-// This takes significant inspiration from:
-// https://github.com/mapbox/concaveman/blob/54838e1/index.js#L11
-fn concave_hull<T>(coords: &mut [Coord<T>], concavity: T) -> LineString<T>
+fn order_concave_hull<T>(current_hull_edges: Vec<CurrentHullEdge<T>>) -> LineString<T>
+where
+    T: GeoFloat,
+{
+    // Order the constituent concave hull edges and return as a `LineString`
+    let mut ordered_coords: Vec<Coord<T>> = vec![];
```
michaelkirk:
```suggestion
    let mut ordered_coords: Vec<Coord<T>> = Vec::with_capacity(current_hull_edges.len() + 1);
```

(performance nit)

geo/src/algorithm/concave_hull.rs
wwaltersdavis: Ah yes of course, thanks

urschrei:
(I think we'd like to see the commits squashed before we merge)

geo/src/algorithm/concave_hull.rs
wwaltersdavis: Thanks, yes makes sense, I have updated this

geo/src/algorithm/concave_hull.rs
wwaltersdavis: Yes I like it, have added

wwaltersdavis:
Thank you both for all your reviews and advise!

michaelkirk:
Thank you so much for the algorithm improvements! I hope to see more contributions in the future. =)

PR #1447, opened 2025-11-19
Rename features to take advantage of `dep:` syntax

### LINKED ISSUE #1284: deprecate `use-foo` style feature flags in favor of just `foo`
[Rust 1.60]( https://blog.rust-lang.org/2022/04/07/Rust-1.60.0.html) introduced the `dep:foo` syntax, which removed the ambiguity between a feature name and an optional crate name, so we no longer need the `use-` prefix.

Let's change it and deprecate the old ones.

michaelkirk:
---

A million years ago, rust introduced the `dep:` syntax to disambiguate between "the optional crate called serde" and "the feature called serde", we're finally taking advantage of it here.

I've also attempted to provide some kind of compiler warning to allow people to migrate to the new feature flag names, since this isn't a thing the rust compiler natively supports.

Example deprecation output of project using the old `use-rstar` feature:

```
warning: use of deprecated unit struct
`deprecated_feature_flags::UseRstar`: The `use-rstar` feature has been
renamed to simply `rstar`. Use the `rstar` feature instead.
   --> /Users/mkirk/src/georust/geo/geo-types/src/lib.rs:203:35
    |
203 | pub use deprecated_feature_flags::UseRstar;
    |                                   ^^^^^^^^
```

It's not ideal ("unit struct") but I since the compiler doesn't support deprecating features, this is the best I could come up with.

Note: I did not add an explicit `proj` feature for the deprecated `use-proj`. Instead we rely on the implicit `proj` feature from the optional proj crate. I assume the old `use-proj` feature was added as an act of forward thinking to avoid a breaking change in the case we needed to enable any accessory crates for the `proj` feature. We can safely do that now if/when we need to without causing a breaking change.

FIXES #1284

### DIFF
--- a/geo-types/Cargo.toml
+++ b/geo-types/Cargo.toml
@@ -29,6 +29,8 @@ use-rstar_0_9 = ["rstar_0_9"]
 use-rstar_0_10 = ["rstar_0_10"]
 use-rstar_0_11 = ["rstar_0_11"]
 use-rstar_0_12 = ["rstar_0_12"]
+# Only used to make `cargo clippy --all-features` happy
+__allow_deprecated_features = []
 
 [dependencies]
 rayon = { version = "1.10.0", optional = true }
--- a/geo-types/src/lib.rs
+++ b/geo-types/src/lib.rs
@@ -162,39 +162,57 @@ pub mod _alloc {
 // Some gymnastics to help migrate people off our old feature flag naming conventions
 #[allow(unused, non_camel_case_types, missing_debug_implementations)]
 mod deprecated_feature_flags {
-    #[deprecated(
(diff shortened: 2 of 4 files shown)


michaelkirk:
Converting to draft - I've got some CI shenanigans to work around.

urschrei:
This seems good and sensible (the CI shenanigans are awful…we'll be able to remove those eventually though, right?)

frewsxcv (approved):

michaelkirk:
> we'll be able to remove those eventually though, right?

Yep, once the deprecated features are removed.

urschrei (approved):

PR #1454, opened 2025-11-26
Handle an edge case in linestring contains linestring

cookiedan42:
if a segment is degenerate and sits on start or end, but there are other non-degenerate, contains will no longer false negative

also added test case for this scenario

---

### DIFF
--- a/geo/src/algorithm/contains/line_string.rs
+++ b/geo/src/algorithm/contains/line_string.rs
@@ -1,5 +1,6 @@
 use super::{Contains, impl_contains_from_relate, impl_contains_geometry_for};
 use crate::algorithm::kernels::Kernel;
+use crate::dimensions::Dimensions;
 use crate::geometry::*;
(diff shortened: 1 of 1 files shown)


geo/src/algorithm/contains/line_string.rs
```diff
@@ -212,6 +224,31 @@ mod test {
...
+        assert!(!degen_start.is_valid());
+        assert!(!degen_end.is_valid());
+
+        assert!(ls.relate(&ls_start).is_contains());
+        assert!(ls.contains(&ls_start));
+        assert!(ls.relate(&ls_end).is_contains());
+        assert!(ls.contains(&ls_end));
+
+        assert!(!ls.contains(&degen_start));
+        assert!(!ls.contains(&degen_end));
```
michaelkirk, line 250: Can you include the `relate` output for these degenerate cases?

PR #1457, opened 2025-11-28
polygon scaling bench

cookiedan42:
Custom implementation for Polygon contains_properly Polygon doesn't scale as well as Relate  
Add a criterion benchmark comparing Polygon contains_properly Polygon scaling, where both polygons have the same number of segments being compared  
Overlap occurs at just under 700 points in each polygon  
lines

---

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -21,6 +21,9 @@
   - BREAKING: The `concave_hull` method now has no `concavity` parameter.  
   - Add `concave_hull_with_options` method which requires `ConcaveHullOptions` as a parameter with `concavity` and `length_threshold` options.
   - <https://github.com/georust/geo/pull/1442>
+- Add `Covers` trait to relate and as a standalone operation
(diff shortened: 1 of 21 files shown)


geo/CHANGES.md
```diff
@@ -9,6 +9,8 @@
 - Add `distance_within` method with default impl for any geometry that implements `Distance`, with similar semantics to the PostGIS [ST_DWithin](https://postgis.net/docs/ST_DWithin.html) function
 - Add fast minimum 1D and 2D Euclidean distance algorithm for linearly separable geometries (#1424)
 - Add `ContainsProperly` trait to relate and as a standalone operation
+  - custom implementations for `Polygon` and `MultiPolygon`, faster than `Contains` until about 650 vertices in each polygon
```
michaelkirk: "custom implementations" is kind of vague

```suggestion
  - `ContainsProperly` is faster for `Polygon` and `MultiPolygon` when inputs are smaller than about 650 vertices, otherwise use `Relate.is_contains_properly`
```

And could you add a similar comment to the `ContaintsProperly` trait docs?

PR #1460, opened 2025-12-02
improve de-9im docs

michaelkirk:
---

Render some example geometries in SVG and show the output matrix rendered as a table.

This is the first time we've used SVG in an example. It's verbose to write, but I think it's helpful for understanding.

If it's a thing we want to do more of, maybe we can come up with some machinery to make it easier / DRY.

Rendered docs:

Screenshot 2025-12-02 at 15 50 37


michaelkirk:
I want to add one more section to the docs to show off predicates.

urschrei:
!!!

PR #1465, opened 2025-12-07
Fix CoordinatePosition for Triangle

wwaltersdavis:
---

See #1468 
### Fix
- Remove the x coordinate check (this was used to ensure if a corner point was used the `boundary_count` would only go to 1).
- Instead set `boundary_count` = 1 if one is present, so if multiple are found (if a corner coord is provided) still accepted as on boundary

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -1,5 +1,9 @@
 # Changes
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


michaelkirk:
Thanks for catching this bug! 

I made an alternative fix more in the spirit of the original boundary counting like we do elsewhere here:
https://github.com/georust/geo/pull/1475

If you want the commit, feel free to implement something similar in this PR, and I'd be happy to see that merged instead.

Issue #1466, opened 2025-12-08
Monotone Chain


cookiedan42:
# Overview
Montone Chains are a way of preprocessing an array of coordinates so that intersections can be efficiently calculated using binary search, and looks to be a primitive used in JTS and GEOS for some internal operations

I built a proof of concept #1467 which uses `&[Coord]` to an existing geometry's backing `Vec`(s) as the backing data and implemented `Intersects` for existing geometries.  
Overall there's a significant speedup on larger geometries in `Intersects` and `ContainsProperly` features where I replaced the brute force linestring-linestring intersections, but slowdown in smaller geoms. Haven't done the benching to find the crossing point, so the current threshold const is arbitrary

# Discussion points

### Option 1: Monotone Geometry analogous to Prepared Geometry (leaning toward this option)
```rust
struct MonotoneLineString{
  geom:LineString
  chain: MonotoneChain
}

struct MonotoneMultiLineString{
  geom:MultiLineString
  chain: Vec<MonotoneChain>

  // or maybe Vec<MonotoneLineString>?
}

struct MonotonePolygon{
  geom:Polygon
  exterior: MonotoneChain
  interiors:Vec<MonotoneChain>
}

```
- Make the choice to use Monotone-based geometries and algorithms explicit
- Possible to delegate to base geometry for anything not implemented 
- Easier to reuse already constructed Monotone-based geometry multiple times  

However
- Not all base geometries have a useful Monoton-based equivalent
- Adds cognitive overhead when choosing to convert into this type

### Option 2: Integrate into existing methods
- Requires tuning of thresholds to switch between the different approaches to not cause performance regression
- pay the conversion overhead every invocation
- most "discoverable"
- most compatible with `Geometry` and `GeometryCollection` based workflows

### Option 3: Independent Function
- least disruptive
- pay the conversion overhead every invocation

I'm personally leaning toward option 1, with selective use of option 2 when benefit is clear

## 2. Naming
1. How should this new class of structs be named? `MonotoneMultiLineString` is quite a mouthful
2. How to disambiguate `MonotoneChain` concept of monotone from the existing `MonoPoly` which uses a different concept of Monotonic Geometry

References:  
![GEOS docs](https://libgeos.org/doxygen/classgeos_1_1index_1_1chain_1_1MonotoneChain.html)  
![JTS docs](https://locationtech.github.io/jts/javadoc/org/locationtech/jts/index/chain/MonotoneChain.html)  
[...]

PR #1467, opened 2025-12-08
Implement MonotoneChain data structure

cookiedan42:
see #1466 

Features:
1. `MonotoneChain` and `MonotoneChainSegment` primitives
2. `MonotomeChain`-backed geometries: `MonotoneChainGeometry`, `MonotoneChainLineString`, `MonotoneChainMultiLineString`, `MonotoneChainPolygon`, `MonotoneChainMultiPolygon`
3. `Intersects` trait between MonotoneChain-backed geometries and all geometries
4. `ContainsProperly` trait between `MonotoneChainPolygon` and `MonotoneChainMultiPolygon`

---


geo/src/algorithm/monotone_chain/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+
+impl<'a, T: GeoNum> MonotoneChain<'a, T> {
+    pub(crate) fn segment_iter(&self) -> impl Iterator<Item = &MonotoneChainSegment<'a, T>> {
+        self.segments.iter()
+    }
+}
+
+impl<'a, T: GeoNum> From<&'a LineString<T>> for MonotoneChain<'a, T> {
+    fn from(linestring: &'a LineString<T>) -> Self {
+        // each segment is a a series of coordinates which are montonically increasing/decreasing
```
michaelkirk:
```suggestion
        // each segment is a series of coordinates which are montonically increasing/decreasing
```

geo/src/algorithm/monotone/chain/monotone_segment.rs
```diff
@@ -0,0 +1,118 @@
...
+use std::cmp::Ordering;
+
+#[derive(Debug, Clone)]
+pub struct MonotoneChainSegment<'a, T: CoordNum> {
+    pub(crate) ls: &'a [Coord<T>],
+    bounding_rect: Rect<T>,
+}
+
+impl<'a, T: GeoNum> MonotoneChainSegment<'a, T> {
+    /// inclusive, exclusive
```
michaelkirk: This doc comment is pretty cryptic. Can you elaborate?

geo/src/algorithm/monotone/chain/monotone_segment.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub fn new(linestring: &'a [Coord<T>]) -> Self {
+        debug_assert!(is_monotone(linestring));
+
+        Self {
+            ls: linestring,
+            bounding_rect: Rect::<T>::new(linestring[0], linestring[linestring.len() - 1]),
```
michaelkirk: If this is public, we should avoid panic, or at least document the panic condition in the case of `LINESTRING EMPTY`.

jts-test-runner/src/runner.rs
```diff
@@ -336,8 +337,26 @@ impl TestRunner {
...
                     let direct_actual = subject.intersects(clip);
                     let relate_actual = subject.relate(clip).is_intersects();
+                    let monotone_actual = {
+                        let subject_mono = TryInto::<MonotoneChainGeometry<f64>>::try_into(subject);
+                        let clip_mono = TryInto::<MonotoneChainGeometry<f64>>::try_into(clip);
+                        match (subject_mono, clip_mono) {
+                            (Ok(subject), Ok(clip)) => subject.intersects(&clip),
+                            (Ok(subject), Err(_)) => subject.intersects(clip),
+                            (Err(_), Ok(clip)) => subject.intersects(&clip),
+                            (Err(_), Err(_)) => subject.intersects(clip),
```
michaelkirk: In this case we're not actually doing anything with monotone chains - we're just doing the vanilla `intersects` check, right?

geo/src/lib.rs
```diff
@@ -36,6 +36,14 @@
 //! The preceding types are reexported from the [`geo-types`] crate. Consider using that crate
 //! if you only need access to these types and no other `geo` functionality.
 //!
+//! ## Monotone Chain backed geometries
```
michaelkirk: Don't get me wrong, this is a cool feature, but this strikes me as too prominent a place for this new feature documentation.

Maybe a new top level heading after the `# Algorithms` section?

```
# Algorithms

## Blah blah blah

...

(end of existing  algorithms documentation)

# Indexed Geometries

## Monotone Chain backed geometries
...

## RTree backed geometries (PreparedGeometry)
...

## Interval Tree
...
```

You don't have to add the sections about IntervalTree or RTree in this PR - they're missing from before. I can follow up with that (but feel free if you're feeling bold!).

michaelkirk (commented):
Sorry for the slow review as I unbury myself from the new year.

I haven't fully reviewed yet (this is a big PR!), but it's a nice feature and in my quick review looks reasonably implemented.

One concern though, it's a lot of code:
```
# before
-rw-r--r--  1 mkirk  staff   7.8M Jan  5 17:52 target/release/libgeo.rlib
# after
-rw-r--r--  1 mkirk  staff   8.6M Jan  5 17:51 target/release/libgeo.rlib
```

It appears that the compiled (release) library grows 10% from just this feature. Do you have any ideas for ameliorating that?

It'd be great to get it smaller. If not, we'll probably need to put it behind a feature flag, which I think would recommend some code re-arranging.

cookiedan42:
> Sorry for the slow review as I unbury myself from the new year.
> 
> I haven't fully reviewed yet (this is big PR!), but it's a nice feature and in my quick review looks reasonably implemented.
> 
> One concern though, it's a lot of code:
> 
> ```
> # before
> -rw-r--r--  1 mkirk  staff   7.8M Jan  5 17:52 target/release/libgeo.rlib
> # after
> -rw-r--r--  1 mkirk  staff   8.6M Jan  5 17:51 target/release/libgeo.rlib
> ```
> 
> It appears that the compiled (release) library grows 10% from just this feature. Do you have any ideas for ameliorating that?
> 
> It'd be great to get it smaller. If not, we'll probably need to put it behind a feature flag, which I think would recommend some code re-arranging.

By disabling parts of the new code and recompiling, the bloat comes from combinatorial explosion when adding algorithms for working with the new geometries  (`Chain`,`ChainSegment`,`LineString`,`MultiLineString`,`Polygon`,`MultiPolygon`,`Geometry`)
```
# neither algorithm
-rw-r--r--@    1 dan  staff   7.9M Jan  6 11:45 libgeo.rlib
# intersects only
-rw-r--r--@    1 dan  staff   8.2M Jan  6 11:46 libgeo.rlib
# intersects + contains properly
-rw-r--r--@    1 dan  staff   8.4M Jan  6 11:51 libgeo.rlib
```

Adding 7 new types to the existing 12 does add a lot of monomorphised generic code for the operations, I'll try removing the purely delegated implementations and see how much that helps.

michaelkirk:
Maybe using dyn would help? The overhead of method dispatch might not be appreciable. 

I’ll also apologize for picking on you. I don’t think I’ve ever done a before/after with code size before, but I saw so many macros and got nervous.

cookiedan42:
Don't worry about it, I really appreciate the feedback I've been receiving while working on geo 😄

geo/src/algorithm/contains_properly/monotone_chain/multipolygon.rs
```diff
@@ -0,0 +1,126 @@
...
+            if !rhs_poly_covered {
+                return false;
+            }
+        }
+
+        true
+    }
+}
+
+impl_contains_properly_target_monotone!(MonotoneChainPolygon<'a, T>, [
```
michaelkirk: Whoops - I think this is in the wrong file (and the MultiPolygon impls are symmetrically misplaced in  polygon.rs)

geo/src/algorithm/contains_properly/monotone_chain/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+            where
+                T: GeoNum,
+                $for:
+                    ContainsProperly<LineString<T>> +
+                    ContainsProperly<MultiLineString<T>> +
+                    ContainsProperly<Polygon<T>> +
+                    ContainsProperly<MultiPolygon<T>>
+            {
+                fn contains_properly(&self, rhs: &$target) -> bool {
+                    self.contains_properly(rhs.geometry())
```
michaelkirk: To make sure I'm understanding, this macro and the one below will delegate to the source geometry of the Monotone geometry.

So for example:
(pseudo code)
```
let poly_1 = todo!()
let poly_2 = todo!()
let mono_poly_1 = MontonePolygon::from(poly_1)
mono_poly_1.contains_properly(poly_2)
```

...offers literally zero performance benefit over 
```
let poly_1 = todo!()
let poly_2 = todo!()
poly_1.contains_properly(poly_2)
```

It is admittedly an ergonomic improvement if you already have the mono_poly, but not a performance one. Is that the idea?

cookiedan42: yeah these were the ones which I was thinking to remove in the earlier comment

michaelkirk: Gotcha. I'm curious to see how much goes away when you remove those implementations. 

It's arguably preferable that the restrictive API leads the user to apply the monotonic geometries only where they are useful. They can always get at the original geometry `monotonic_polygon.geometry()` to perform other operations, and I think it's more obvious to the user what the performance characteristics would be.

Compare this to PreparedGeometry and Relate:

```rust
// 1. slowest
poly_1.relate(poly_2)

// 2. fastest
let prepared_poly_1.relate(prepared_poly_2)

// 3. medium-fast
let prepared_poly_1.relate(poly_2)
```

So, in the case of Relate, there is benefit to using a PreparedPolygon even if for only one side of the operation. I don't think the MC impls have this property.

Also consider, we can add these additional APIs in the future if it seems worthwhile.

michaelkirk: **Addendum:** 

Though it looks like you can get the underlying geometry of the variants:

```
mc_polygon.geometry()
mc_line_string.geometry()
```

I don't see a method for the enum:
```
let mc_geom: MonotoneChainGeometry = todo()
// unimplemented
mc_geom.geometry()
```

I guess that's tricky because we only have a _reference_ to the geometry... we could reasonably return a GeometryCow, but it doesn't (yet?) have many algorithms implemented on it.

michaelkirk: **Addendum addendum:**

We don't have to address that in this PR. We could just say that you need to `match` an MCGeometry to get at the MC variant (e.g. `MCPolygon`), and then from there you can get its non-mc-geometry (e.g. `Polygon`).

(sorry for the stream of consciousness)

michaelkirk:
Just checking in @cookiedan42  - were you able to reduce the issue with increased code size?

cookiedan42:
> Just checking in @cookiedan42 - were you able to reduce the issue with increased code size?

The PR as it is sits at ~0.6mb  
scoping the `intersects` feature down to only handle intersections between `monotone_chain` geometries drops it down to ~0.4mb since the corresponding `intersect` with `monotone_chain` and `monotone_chain_segment` can also be cut out (https://github.com/cookiedan42/geo/pull/22)

geo/src/algorithm/monotone_chain/chain_iter.rs
```diff
@@ -0,0 +1,175 @@
...
+};
+use crate::GeoNum;
+use crate::geometry::*;
+use std::iter;
+
+/// An iterator over a compatible Geometry type which yields [`MonotoneChain`]s.
+///
+/// Similar to [`MonotoneChains`], but returns owned values instead of references
+/// and works on non-[`MonotoneChain`] backed geometries too
+pub trait MonotoneChainIter<'a, T: GeoNum + 'a> {
```
michaelkirk: This trait is not used within `geo`.

For my own understanding, can you give an example for what someone would want to use this trait for? It seems like sort of a hybrid constructor/iterator. "Give me the chains efficiently if I already have them (e.g. from `MonotonChainLineString`, otherwise compute them on the fly (e.g from `LineString`)."

What would this hypothetical use case look like with and without this trait?

To confess my motivations, this pull request is very large, and so I'm trying to trim off some of the bells and whistles. We should keep it if this trait results in writing code that's easier to understand, but I haven't seen it used yet.

michaelkirk (changes requested):
> the corresponding intersect with monotone_chain and monotone_chain_segment can also be cut out

My understanding is there's no performance benefit to allowing `MCPoly.intersects(regular_poly)` vs `MCPoly.geometry().intersects(regular_poly)`. As such, the benefit is only that you can write code more concisely. I like concise code, but I'd argue the latter is actually less surprising.

Also, as I mentioned elsewhere, this is already a very large PR, so anything we can reasonably due to trim it down is good. So, please remove these delegated trait impls for now.

geo/src/algorithm/monotone_chain/chain_iter.rs
cookiedan42: This trait was built for a prior design where it might be used for operations on non-monotone chain backed geometries, but in keeping this PR tight, I'll remove it too

geo/src/algorithm/intersects/monotone_chain/monotone_chain_segment.rs
```diff
@@ -0,0 +1,87 @@
...
+                match self.ls().len() {
+                    0 => false,
+                    1 => self.ls()[0].intersects(rhs),
+                    2 => {
+                        // PERF: One potential speed optimization here is to use specialized function
+                        // with pre-requisite of bounding box intersection.
+                        // e.g. Line intersect Line could become just the orientation checks
+                        Line::<T>::new(self.ls()[0], self.ls()[1]).intersects(rhs)
+                    }
+                    n if n > SEARCH_THRESHOLD => {
```
michaelkirk:
```suggestion
                   _ => {
```

At this point we know n > 2, based on the previous cases. And the subsequent case is unreachable and should be deleted. Right?

geo/src/algorithm/intersects/monotone_chain/monotone_chain_segment.rs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        // recurse with binary split on both sides
+        match (self.divide(), rhs.divide()) {
+            ((a, Some(b)), (c, Some(d))) => {
+                a.intersects(&c) || a.intersects(&d) || b.intersects(&c) || b.intersects(&d)
+            }
+            ((a, Some(b)), (_c, None)) => a.intersects(rhs) || b.intersects(rhs),
+            ((_a, None), (c, Some(d))) => self.intersects(&c) || self.intersects(&d),
+            ((_a, None), (_c, None)) => unreachable!("both sides are single segments"),
```
michaelkirk: nit: the `handle base case` above could be moved here, instead of using `unreachable`, right?

geo/src/algorithm/monotone_chain/segment.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    /// Split the segment into two segments at the halfway mark for binary search
+    pub fn divide(
+        &'caller self,
+    ) -> (
+        MonotoneChainSegment<'a, T>,
+        Option<MonotoneChainSegment<'a, T>>,
+    ) {
+        // if segment is a line, cannot be divided
+        if self.ls.len() == 2 {
```
michaelkirk:
```suggestion
        if self.ls.len() <= 2 {
```

I **think** it's not currently possible to call `divide` on `MonotoneChainSegment` with a shorter line string, so this is effectively a no-op, but it seems like we could reasonably be more cautious here.

michaelkirk (changes requested):
OK! My feedback is increasingly nitpicky, but I'm feeling like I understand the changes in this branch pretty well at this point. This is my last round of feedback unless you have further ideas.

One thing: it took me a while to understand the logic in MonotoneChainSegmentFactory. I rewrote parts of it mostly as an exercise to help me understand how the method works, but what I ended up with is a little faster and clearer (I think 😬). 

If you like it too, would you cherry-pick it into this branch? https://github.com/georust/geo/tree/mkirk/monotone-chain-3

If it's controversial, I can open a followup PR with it.

michaelkirk:
Hi @cookiedan42 - do you have time to see this through? Otherwise I might pick it up.

cookiedan42:
Oh whoops I must have missed the last notification, I'll make the changes in the next few days

michaelkirk (approved):

michaelkirk:
Thank you!

PR #1472, opened 2025-12-09
Add simply connected interior validation and test suite

bretttully:
---

Add validation to detect when a polygon's interior is not simply connected, which violates the OGC Simple Feature Specification requirement that "the interior of every Surface is a connected point set."

## What This Detects

The validation detects two types of interior disconnection:

1. **Multi-touch**: Two rings sharing 2+ touch points at different coordinates
2. **Cycle**: Rings forming a cycle through distinct single-touch coordinates

## Algorithm

Uses the `GeometryGraph` edge intersection data to find where rings touch:

1. **Collect intersections**: For each edge intersection, record the coordinate, which ring (edge) it belongs to, and whether it's a vertex of that ring
2. **Sort and group**: Sort intersections by `(x, y, edge_idx)` using `total_cmp` for consistent ordering, then group by coordinate
3. **Build touch map**: For coordinates where 2+ rings meet (and at least one has it as a vertex), record the touch for each ring pair
4. **Detect disconnection**:
   - If any ring pair has 2+ touch coordinates → disconnected
   - Build a graph of single-touch ring pairs and DFS for cycles through distinct coordinates

A "touch" requires at least one ring to have the point as an actual vertex (not a mid-segment crossing).

## Key Insight

Multiple rings meeting at a **single** coordinate does NOT disconnect the interior. For example, three triangular holes meeting at one point still have a connected interior around their perimeter. Disconnection only occurs when touches happen at **distinct** coordinates.

## Error Reporting

The error message includes the problematic coordinates to aid debugging:
```
polygon interior is not simply connected at coordinate(s): (2, 2), (3, 3)
```

## Performance

Uses `PreparedGeometry` to cache R-tree structures for interior/exterior containment checks. The sorting-based coordinate grouping provides better cache locality than hash-based approaches.


geo-test-fixtures/src/checkerboard.rs
```diff
@@ -0,0 +1,192 @@
...
+            (fi + 5.0) * sz,
+            (fi + 1.0) * sz,
+        ));
+    }
+
+    holes
+}
+
+/// Translate a ring by an offset.
+pub fn translate_ring(ring: &LineString<f64>, dx: f64, dy: f64) -> LineString<f64> {
```
michaelkirk: See the `Translate::translate` / `Translate::translate_mut`

geo-benches/src/validation.rs
```diff
@@ -0,0 +1,129 @@
...
+use geojson::GeoJson;
+use std::convert::TryInto;
+use std::fs;
+use std::path::PathBuf;
+
+/// Load benchmark geometries from the validate.geojson fixture.
+/// Returns a vector of (name, polygon) tuples.
+fn load_benchmark_geometries() -> Vec<(String, Polygon<f64>)> {
+    let mut path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    path.push("fixtures/rust-geo-booleanop-fixtures/benchmarks/validate.geojson");
```
michaelkirk: What's the providence of `fixtures/rust-geo-booleanop-fixtures/benchmarks/validate.geojson`

It's seems like ultimately we just want a vector of polygons. If it's an arbitrary test fixture generated for this test, can you simplify the test data rather than complicating the parsing code?

geo/src/algorithm/validation/polygon.rs
```diff
@@ -153,10 +196,212 @@ impl<F: GeoFloat> Validation for Polygon<F> {
...
+        .collect();
+
+    // Collect all intersection coordinates from each edge
+    let edge_intersections: Vec<Vec<(u64, u64)>> = edges
+        .iter()
+        .map(|edge: &std::rc::Rc<RefCell<Edge<F>>>| {
+            let edge = RefCell::borrow(edge);
+            edge.edge_intersections()
+                .iter()
+                .filter_map(|ei| coord_to_bits(&ei.coordinate()))
```
michaelkirk: re: `f64::to_bits`

Is this in pursuit of getting the floating points to fit into a HashSet? There's a reason you can't put floats in a HashSet, and it's not clear to me that this usage is valid.

Can you talk more about this? Maybe it's OK, but I've just never used floats this way. 

The only problematic examples I can think of off the top of my head are:
1. NaN coords which have many representations, but we don't make any guarantees for NaN coords anyway, so that's probably fine

2. -0.0 vs. +0.0  - topologically those are the same number, but they'd appear as distinct entries, which seems like it could confound this algorithm.

I'm not sure if there are any others.

geo/src/algorithm/validation/polygon.rs
```diff
@@ -234,35 +486,45 @@ mod tests {
...
             )
         );
 
-        assert_validation_errors!(
-            &polygon,
-            vec![InvalidPolygon::IntersectingRingsOnALine(
-                RingRole::Exterior,
-                RingRole::Interior(0)
-            )]
+        let errors = polygon.validation_errors();
```
michaelkirk: Please use the assert_validation_errors macro. If you're having trouble I can help, but I'm not sure if the issue is you tried and it didn't work or if this is just what the AI spit out.

geo/CHANGES.md
```diff
@@ -2,6 +2,8 @@
 
 ## 0.32.0 - 2025-12-05
 
+- Add simply connected interior validation for polygons. Polygons with holes that touch at vertices in ways that disconnect the interior (e.g., two holes sharing 2+ vertices, or cycles of holes each sharing a vertex) are now detected as invalid via `Validation::is_valid()`. This aligns with OGC Simple Features and matches PostGIS behavior.
+- Performance: Polygon validation now uses `PreparedGeometry` to cache R-tree structures, improving validation speed by 26-45% for polygons with many holes.
```
michaelkirk: `PreparedGeometry` is just used for this newly added validation right?  So there's no world in which people validating before would see a speedup, right?

michaelkirk (changes requested):
Hello! Thank you for the pull request. This would be a great feature to incorporate, but I've got some questions to work through.

geo-benches/src/validation.rs
bretttully: These are generated from the fuzzer I created in https://github.com/bretttully/iOverlay-py (which is where I found that these invalid geometries are missed by geo). For benchmarking though, I don't think we need them, so I will drop back to using the existing geojson files in the repo.

geo/CHANGES.md
bretttully: I also added a `let prepared_exterior = PreparedGeometry::from(&polygon_exterior);` for the exterior-interior relate calls that has a nice improvement in the benchmarks for polygons with lots of holes.

geo/src/algorithm/validation/polygon.rs
bretttully: Done

geo/src/algorithm/validation/polygon.rs
bretttully: Based on this comment, the algo is now using ordered points via `total_cmp` -- gave a performance boost too.

geo/src/algorithm/validation/polygon.rs
bretttully: >just what the AI spit out.

I have to say I spent a long time deliberating whether or not to make this PR. I've never written a line of rust before, and I _really_ didn't want to pollute the codebase with AI slop. Nor did I want to waste your time with overly verbose / poor code. 

I find the solution readable, and the tests make sense to me (and are validated using shapely), so I hope I'm not too far off track!

geo/src/algorithm/validation/polygon.rs
```diff
@@ -99,24 +120,44 @@ impl<F: GeoFloat> Validation for Polygon<F> {
...
+            // interior-interior comparisons
+            let interior_1_as_poly = Polygon::new(interior_1.clone(), vec![]);
+            let prepared_interior_1 = PreparedGeometry::from(&interior_1_as_poly);
+
+            // Check containment: interior ring must be properly inside exterior.
+            // We use polygon-to-polygon comparison and check both:
+            // - is_contains(): exterior must contain interior
+            // - is_equal_topo(): they must NOT be identical (an interior ring equal to
+            //   exterior is invalid)
+            let exterior_vs_interior = prepared_exterior.relate(&interior_1_as_poly);
```
michaelkirk:
```suggestion
            let exterior_vs_interior = prepared_exterior.relate(&prepared_interior_1);                   
```

Might as well use this since you've already prepared it, right?

geo/src/algorithm/bool_ops/tests.rs
```diff
@@ -187,6 +187,77 @@ fn jts_test_overlay_la_1() {
...
+        //   1 │       └───────┘   │
+        //     │                   │
+        //   0 └───────────────────┘
+        //
+        // OGC Simple Feature Specification (ISO 19125-1) states:
+        // "The interior of every Surface is a connected point set."
+
+        let exterior: Polygon<f64> = wkt!(POLYGON((0.0 0.0, 5.0 0.0, 5.0 5.0, 0.0 5.0, 0.0 0.0)));
+
+        // Combined L-shaped holes using union of boxes
```
michaelkirk: So when i_overlay is updated, we'll need to declaratively build this invalid MultiPolygon, right? Any reason to not do so now? I don't think we gain anything by constructing the MultiPolygon from i_overlay, since this new validation feature applies to any geometry with this characteristic, regardless of how it came to be.

geo/src/algorithm/validation/polygon.rs
```diff
@@ -27,6 +28,15 @@ pub enum InvalidPolygon {
...
     IntersectingRingsOnAnArea(RingRole, RingRole),
+    /// The polygon interior is not simply connected because rings touch in a way that
+    /// disconnects the interior into multiple regions.
+    ///
+    /// This can occur when:
+    /// - Two rings share two or more vertices (creating a "corridor" that cuts through the interior)
+    /// - Rings form a cycle of single-vertex touches that encloses part of the interior
+    ///
+    /// The coordinates indicate the problematic touch point(s) where the disconnection occurs.
+    InteriorNotSimplyConnected(Vec<Coord<f64>>),
```
michaelkirk: f64 is strange here, since we support generic geometry types. 

We could make this error type generic (`InvalidPolygon<T>`), but I think we should align with our other error cases in this enum and take an index (usize) based approach using `RingRole::Interior(idx)` and `CoordIndex` error.

I admit I prefer some things about your proposal (it's nice to see the actual coordinates), but I think aligning with the existing pattern outweighs that in this case.

geo/src/algorithm/validation/polygon.rs
```diff
@@ -99,24 +120,44 @@ impl<F: GeoFloat> Validation for Polygon<F> {
...
-            if !exterior_vs_interior.is_contains() {
+            // Build polygon from interior ring - used for both exterior check and
+            // interior-interior comparisons
+            let interior_1_as_poly = Polygon::new(interior_1.clone(), vec![]);
+            let prepared_interior_1 = PreparedGeometry::from(&interior_1_as_poly);
+
+            // Check containment: interior ring must be properly inside exterior.
+            // We use polygon-to-polygon comparison and check both:
+            // - is_contains(): exterior must contain interior
+            // - is_equal_topo(): they must NOT be identical (an interior ring equal to
```
michaelkirk: > // - is_equal_topo(): they must NOT be identical (an interior ring equal to exterior is invalid)

I suppose that's true, but isn't that wholly redundant with the more general "Rings may not touch other rings, except at a point." criteria that we enforce?

Did this change maybe because `interior` is now a Polygon rather than a LineString? Or what spurred this change?

geo/src/algorithm/validation/polygon.rs
```diff
@@ -153,14 +196,253 @@ impl<F: GeoFloat> Validation for Polygon<F> {
...
+
         Ok(())
     }
 }
 
+/// Check that the polygon interior is simply connected using the GeometryGraph.
+///
+/// This function extracts touch point information from a pre-computed GeometryGraph,
+/// which is more efficient than brute-force comparison because the graph's R-tree
+/// intersection detection runs in O((n + k) log n) instead of O(V²).
```
michaelkirk: Is this the right complexity? I'm also not sure what these variables mean. I think we should probably skip re-hashing the big-O notation here. If anywhere, it could live on GeometryGraph.

geo/src/algorithm/validation/polygon.rs
```diff
@@ -153,14 +196,253 @@ impl<F: GeoFloat> Validation for Polygon<F> {
...
+        let current_coord = all_intersections[i].coord;
+
+        // Find all entries with the same coordinate
+        let mut j = i + 1;
+        while j < all_intersections.len() && all_intersections[j].coord == current_coord {
+            j += 1;
+        }
+
+        // Group spans from i to j (exclusive), sorted by edge_idx within the group
+        let group = &all_intersections[i..j];
```
michaelkirk, line 297: A "group" means all the intersections at a single coordinate, right?

geo/src/algorithm/validation/polygon.rs
```diff
@@ -213,19 +495,26 @@ mod tests {
             )
         );
 
+        // When rings share a line segment, they share multiple vertices,
+        // which also disconnects the interior
```
michaelkirk: I don't understand why this comment would necessarily be true.

Here's the input:
```
POLYGON(
    (0. 0., 4. 0., 4. 4.,0. 4.,0. 0.),
    (1. 2., 2. 1., 3. 2., 2. 3., 1. 2.),
    (3. 2., 2. 1., 3.5 1., 3.75 2., 3.5 3., 3. 2.)
)
````

Image

I don't intuitively expect that to be a "disconnected interior".

Instead I'd expect disconnected interior to apply to something like this:

Image

```
POLYGON ((100 100, 100 0, 0 0, 0 100, 100 100), 
  (11 10, 50 10, 30 50, 50 90, 10 90, 11 10), 
  (50 10, 70 50, 50 90, 90 90, 90 10, 50 10))
```

Another example I'd expect to be **invalid** but **not** due to a disconnected interior.

```
POLYGON ((300 200, 194 200, 194 305, 300 305, 300 200), 
  (240 250, 270 250, 270 200, 240 200, 240 250))
```

Image


Am I thinking about it wrong? Are you using some reference for this behavior?

michaelkirk (changes requested):
Sorry for the slow turn around! I'd like to have this behavior, but I have some questions about the implementation. I haven't done a full review, but wanted to check in before dedicating more time to the review.

I think directionally it makes sense. Can I ask, did you use some kind of reference implementation for the graph algorithm, or did it spring mostly from you and the AI?

michaelkirk:
Hi @bretttully - are you able to respond to any of my feedback? As I said, I think this is useful functionality, but I'm not comfortable merging it without understanding more how it works.

bretttully:
@michaelkirk I haven't had a chance to carve out time just yet. I am keen to get the functionality in, so will have something together next week. Sorry for the delay!

bretttully:
@michaelkirk I am just picking this up again. I have started by getting in the OGC changes for iOverlay here: https://github.com/georust/geo/pull/1500

I'm going to rebase this branch off that one, and possibly split this PR into smaller ones to make review easier.

geo/CHANGES.md
bretttully: I have split the use of prepared geometry out into #1501

geo/src/algorithm/validation/polygon.rs
bretttully: Done in #1501

geo/src/algorithm/bool_ops/tests.rs
bretttully: i_overlay update introduced in #1500 with a couple of the i_overlay tests for the boolean ops.

geo/src/algorithm/validation/polygon.rs
bretttully: Changed to using RingRole

geo/src/algorithm/validation/polygon.rs
bretttully: removed comment

geo/src/algorithm/validation/polygon.rs
bretttully: Yes. Updated comment to make that clearer

geo/src/algorithm/validation/polygon.rs
bretttully: Yes, in contrast to GEOS/shapely that early exit on the first error, I was collecting all errors which means these cases report as having a disconnected interior as well as the other error that you see with shapely. 

I have changed the implementation to skip intersection points if they have already been reported as an error. If we didn't skip their coordinates then they would be found as disconnected interiors using the current algorithm. 

Given their degeneracy, I'm not sure if that is an issue or not -- more that they are degenerate in multiple ways?

bretttully:
Hi @michaelkirk 

> Can I ask, did you use some kind of reference implementation for the graph algorithm, or did it spring mostly from you and the AI?

This is has been a mostly iterative process with Claude; I have included the iOverlay repo in context and have found that referencing its implementation has guided the solution here. 

I have been using Claude to validate test cases against shapely as I go, but have shied away from using GEOS/JTS in context given licensing of those libraries. I have also been getting it to generate explanatory images such as this for the union find implementation that I think is easier to understand than what was there previously.

image

urschrei:
> validate test cases against shapely as I go, but have shied away from using GEOS/JTS in context given licensing of those libraries

Just to note that a) Shapely is GEOS under the hood, b) JTS is BSD licensed, and c) the use of test suites belonging to a library is not covered under that library's license AFAIK (that is, the test suite falls under fair use). NB: I am not a lawyer

michaelkirk:
Just a note to say that I've been playing around with this here #1522. Please take a look at my changes if you can @bretttully.

michaelkirk:
Unless someone cares to disagree, I'm going to say this is superseded by #1522

PR #1473, opened 2025-12-09
Fix k-means premature convergence bug

### LINKED ISSUE #1469: kmeans instability
I've been encountering sporadic test failures due to the `test_kmeans_three_clusters` test  
Running it in a loop seems to show instability in the results of kmeans  

@urschrei 

```rust

 #[test]
    fn test_kmeans_three_clusters_loop() {
        let points = [
            // Cluster 1
            point!(x: 0.0, y: 0.0),
            point!(x: 1.0, y: 0.0),
            // Cluster 2
            point!(x: 10.0, y: 10.0),
            point!(x: 11.0, y: 10.0),
            // Cluster 3
            point!(x: 20.0, y: 20.0),
            point!(x: 21.0, y: 20.0),
        ];
        
        for i in 0..100_000 {
            
            let labels = points.kmeans(3).unwrap();
            // Each pair should be in the same cluster
            assert_eq!(labels[0], labels[1],"failed at loop {}",i);
            assert_eq!(labels[2], labels[3],"failed at loop {}",i);
            assert_eq!(labels[4], labels[5],"failed at loop {}",i);

            // All three pairs should be in different clusters
            assert_ne!(labels[0], labels[2],"failed at loop {}",i);
            assert_ne!(labels[2], labels[4],"failed at loop {}",i);
            assert_ne!(labels[0], labels[4],"failed at loop {}",i);
        }
    }

```

urschrei:
When k-means++ initialisation selects data points as initial centroids, points at those locations have `upper_bound=0` in Hamerly's algorithm, causing them to be incorrectly pruned from reassignment checks. This can cause the algorithm to declare convergence on the first iteration without ever computing true cluster centroids.

This fix updates centroids to be cluster means immediately _after_ the initial assignment, before entering the main convergence loop. This ensures Hamerly bounds are computed against true centroids rather than the data points selected by k-means++.

Other improvements:

- Added docs explaining that k-means converges to local minima and may produce suboptimal results depending on initialisation
- Update `test_kmeans_three_clusters` to use a fixed seed (42) for deterministic testing which should avoid intermittent failures from unlucky k-means++ initialisation

---

Fixes #1469

### DIFF
--- a/geo/src/algorithm/kmeans/mod.rs
+++ b/geo/src/algorithm/kmeans/mod.rs
@@ -54,6 +54,17 @@
 //!
 //! # Notes
 //!
+//! ## Local Minima
+//!
+//! _k_-means converges to a local minimum, not necessarily the global optimum. The final clustering
+//! depends on the initial centroid placement. Although _k_-means++ provides good initialisation,
+//! it can still occasionally place multiple initial centroids within the same "natural" cluster,
+//! leading to suboptimal results. For critical applications, consider running _k_-means multiple
+//! times with different seeds and selecting the result with lowest inertia (sum of squared
+//! distances to centroids).
+//!
+//! ## Empty Clusters
+//!
 //! Empty clusters may rarely occur during iteration if all points are reassigned away from a centroid.
 //! When this happens, the algorithm attempts to recover by reassigning the farthest point from its
 //! current centroid to the empty cluster (following scikit-learn's approach). If recovery fails,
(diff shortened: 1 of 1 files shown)


geo/src/algorithm/kmeans/mod.rs
```diff
@@ -359,24 +370,51 @@ where
...
+            &centroids,
+            point_sq_norms[i],
+            &centroid_sq_norms,
+        );
+        *assignment = nearest_idx;
+    }
+
+    // Update centroids to be cluster means (initial centroids from k-means++ are data points,
+    // not true cluster means). This is required before the first convergence check because
+    // otherwise points at the initial centroid locations have upper_bound=0 and get pruned.
```
michaelkirk: Where/why would they get pruned?

(Sorry I'm not very familiar with this algo yet...)

michaelkirk (commented):
I've spent over an hour trying to understand this, and confess I still don't fully. So here's some notes:

Ultimately, kmeans/++ is known to be non-optimal, and susceptible to initial conditions, so these are somewhat expected.

The "bad" clustering looks like this:

Image

In 30b4916d I counted how many seeds lead to failure before your changes: 603 failures in 100_000.

And after your changes, this was reduced to 133 failures in 100_000

So that seems good. 

It seems like you're now pulling out part of the main loop to run initialization. I'm not sure why the difference between the initial-point-as-centroid and the "true" initial cluster centroid makes this difference - I'd assume it's just another loop of the iteration, but clearly something more is going on there.

At the end of the day, it seems like strictly better results, and the code is not unreasonable, so I'm happy to LGTM. But maybe wait a couple days before merging in case someone smarter or more dedicated can review.

PR #1474, opened 2025-12-11
quieter logging

michaelkirk:
---


geo/src/algorithm/monotone/builder.rs
```diff
@@ -89,7 +89,7 @@ impl<T: GeoNum> Builder<T> {
         incoming.sort_by(|a, b| a.partial_cmp(b).unwrap());
         outgoing.sort_by(|a, b| a.partial_cmp(b).unwrap());
 
-        info!(
+        debug!(
```
michaelkirk, line 92: This is the one that really got me. 

Log levels are very much a bike shedding topic, but generally if there's something normal happening in an `O(n)` loop (like here, once per coordinate, in a geometry) I'd generally make it `debug!`.

urschrei (approved):
No argument from me!

kylebarron (approved):

PR #1475, opened 2025-12-11
Fix coordinate position on vertical boundary

michaelkirk:
---

Alternative to #1465 that keeps in the spirit of boundary counting (and includes a test).


geo/src/algorithm/coordinate_position.rs
```diff
@@ -192,7 +192,12 @@ where
                 let orientation = T::Ker::orient2d(l.start, l.end, *coord);
                 if orientation == Orientation::Collinear
                     && point_in_rect(*coord, l.start, l.end)
-                    && coord.x != l.end.x
```
michaelkirk, line 195: Previously we were only checking the `x` value - but that broke down for vertical lines.

PR #1478, opened 2025-12-15
add wasm check

### LINKED ISSUE #1477: Add a wasm32 check to CI
We did this for flatgeobuf after inadvertently breaking the wasm build: https://github.com/flatgeobuf/flatgeobuf/blob/master/.github/workflows/test.yml#L115

I think we should do it here too.

Inspired by https://github.com/georust/geo/pull/1476#issuecomment-3655283374

michaelkirk:
---
FIXES #1477 

I'm assuming this will fail for starts...

### DIFF
--- a/geo/Cargo.toml
+++ b/geo/Cargo.toml
@@ -40,7 +40,20 @@ i_overlay = { version = "4.0.0, < 4.1.0", default-features = false }
 sif-itree = "0.4.0"
 rand = "0.9.2"
 
-[target.'cfg(target_arch = "wasm32")'.dependencies]
+# `rand` crate doesn't integrate with wasm-unknown-unknown directly
+# Instead they recommend that users create an explicit dependency on getrandom, and enable wasm-js support
+#
+# `rand` maintainers recommend that each application does this to fix their build;
+# However, if the version of getrandom ever rev's, all out users builds will break again.
+#
+# Instead, we require a specific version of getrandom, and use it to manually seed the RNG,
+# rather than relying on the suggested "back door/unstable" configuration.
+#
+# The failure mode here, if `random` revs their version of `getrandom` is a duplicate dependency, rather than
+# a broken build, until we update our getrandom dep.
+#
+# See https://github.com/georust/geo/pull/1478
+[target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
 getrandom = { version = "0.3.4", features = ["wasm_js"] }
 
 [dev-dependencies]
--- a/geo/src/algorithm/kmeans/error.rs
+++ b/geo/src/algorithm/kmeans/error.rs
@@ -120,6 +120,10 @@ pub enum KMeansError<T: GeoFloat> {
         /// Number of points that changed clusters in the final iteration
         changed_assignments: usize,
     },
+
+    #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
+    /// Error when initializing randomness
+    GetRandom(getrandom::Error),
 }
 
 impl<T: GeoFloat> std::fmt::Display for KMeansError<T> {
@@ -167,6 +171,8 @@ impl<T: GeoFloat> std::fmt::Display for KMeansError<T> {
                     iterations, shift_f64, tol_f64, changed_assignments
                 )
             }
+            #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
+            KMeansError::GetRandom(err) => err.fmt(f),
         }
     }
 }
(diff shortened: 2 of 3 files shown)


martinfrances107:
```
 name: Build geo for wasm32
     run: cargo check -p geo --lib --target wasm32-unknown-unknown --no-default-features
```

This will be insufficient to expose the problems I was seeing when I reported  #1476 

I am going to relate this to a concrete example from the rust_d3_geo crate

<https://github.com/martinfrances107/rust_d3_geo/tree/v3.x-dev/examples/globe/albers_usa_svg>

Here is the dependency tree 

albers_usa_svg -> rust_d3_geo -> geo

cargo checking for a "target wasm32-unknown-unknown" passed in that instance.

The albers_usa_svg npm module was failing at a further build step down the line

the example constructs a npm module via this invocation of the wasm-pack builder 
see [package.json](https://github.com/martinfrances107/rust_d3_geo/blob/v3.x-dev/examples/globe/albers_usa_svg/package.json)

scripts: {
   "build": "rimraf dist pkg && RUST_LOG=info webpack",
    /*... */
}

so 

```
npm run build
```

In the getrandom crate There are plenty of compile time errors we must not hit.
like this <https://github.com/rust-random/getrandom/blob/master/src/backends/wasm_js.rs>

```rust
#[cfg(not(all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none"))))]
compile_error!("`wasm_js` backend can be enabled only for OS-less WASM targets!");
```

I can criticise my  own fix ... just from the outline

1) In the test infrastrurcture pull in a dependency on node, npm, wasm-pack,  ( and the npm package @wasm-tool/wasm-pack-plugin ) 
2) Just like the albers_usa_svg example - Build a minimal test npm package that calls some function  from the geo/kmeans module to test the source of randomness.
3) Run the test-package .. (In a headless browser or  curl or something.)

This laundry list is very heavyweight ... there must be a simpler way.....

Again I need to sleep on this overnight... 

In terms of review .. 

If I develop a geo example first then use that to develop a small test case..later 

That  example would be a talking point .. where with input from others we get a golden reference or a test vehicle on how to do WASM projects

Hmm 

Would you be minded to veto my plan just on the cost of stage 1 alone?

michaelkirk:
I've run CI on the first commit so you can see that it fails: https://github.com/georust/geo/actions/runs/20282090351/job/58246521264

> ```
> error: The wasm32-unknown-unknown targets are not supported by default; you may need to enable the "wasm_js" > configuration flag. Note that enabling the `wasm_js` feature flag alone is insufficient. For more information see: > https://docs.rs/getrandom/0.3.4/#webassembly-support
> ```

I'd prefer to avoid building a wasm-pack'd demo application if we can do something simpler like this. It seems to be checking for what we want it to check, unless I'm missing something.

michaelkirk:
I'll also note that my second commit, which fixes CI, also fixes the problem, by having an explicit dependency on getrandom on the wasm target.

One (the only?) problem with this approach is that `getrandom` is an internal dependency of `rand`, and so they can reasonably bump their `getrandom` dependency a major version without needing to bump their own major version, at which point all the geo+wasm users would have their build break again.

See here: https://github.com/rust-random/rand/pull/1537

michaelkirk:
I opened an issue here to see if we can get any guidance: https://github.com/rust-random/rand/issues/1694

martinfrances107:
Here is a golden reference I was talking about #1480

martinfrances107:
> I'd prefer to avoid building a wasm-pack'd demo application if we can do something simpler like this. It seems to be checking for what we want it to check, unless I'm missing something.

I would like to stand up to the need for a golden reference
Everyone accepts that rust applications in a browser are a mess.. it is a sea of shifting sands.

But  that precisely is why keeping the reference design around is good idea

Yes I am placing a maintenance  debut  on to the community -  But I see getting it to compile as a wasm*-unknown-unknown target as a essential first step *ONLY*.

When a developer files a report that "My app" works on Chrome, but not on Safari I don't want to debug another's design. I want to see if the reference design has also come unstuck on the same shifting sand bank.

And as a fallback from that  I  want to update the reference design with all the known workarounds for the different browsers.

michaelkirk:
Good point. 

We have at least a few users using geo in the browser. 

Beyond CI purposes, having an up to date example of "here's how to use geo in the browser" seems useful on its own. I'd be in favor of trying it out. We can always revisit if it proves unduly onerous.

martinfrances107:
> We can always revisit if it proves unduly onerous.

"The first rule of reference design club is never roll your own!"

I spent about about 2 hours cutting down and adapting a previous reference design.

When wasmbindgen and webpack falls out of fashion - and others have defined new best practises then 
I hope it should be a similarly sized task to throw it all away and start over.

michaelkirk:
I think we should supersede this with #1480 (assuming the example app gets added to the GH actions build)

PR #1481, opened 2025-12-17
fix bug in linestring intersection when linestring is degenerate point

cookiedan42:
---

Embarrassing that this edge case slipped through my earlier work 😞

### DIFF
--- a/geo/src/algorithm/intersects/line_string.rs
+++ b/geo/src/algorithm/intersects/line_string.rs
@@ -13,6 +13,11 @@ macro_rules! intersects_line_string_impl {
                 if has_disjoint_bboxes(self, rhs) {
                     return false;
                 }
+                // degenerate linestring treated as point
+                if self.0.len() == 1 {
+                    return self.0[0].intersects(rhs);
+                }
+
                 self.lines().any(|l| l.intersects(rhs))
             }
         }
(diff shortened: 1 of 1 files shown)


geo/src/algorithm/intersects/line_string.rs
```diff
@@ -86,6 +91,21 @@ mod test {
...
+        let pt: Point<f64> = wkt! {POINT(1 1)}.convert();
+        let poly: Polygon<f64> = wkt! {RECT(0 0, 1 1)}.to_polygon().convert();
+        let ls1: LineString<f64> = wkt! {LINESTRING(1 1)}.convert();
+        let ls2: LineString<f64> = wkt! {LINESTRING(1 1, 1 1)}.convert();
+
+        assert_eq!(ls1.intersects(&pt), ls1.relate(&pt).is_intersects());
+        assert_eq!(ls2.intersects(&pt), ls2.relate(&pt).is_intersects());
+        assert!(ls1.intersects(&poly));
+        // TODO: uncomment when relate is fixed, this currently panics
+        // assert_eq!(ls1.intersects(&poly), ls1.relate(&poly).is_intersects());
```
michaelkirk, line 105: Looks like you opened #1482 to track this. Thank you!

Good find. This seems like a bug in CoordinatePosition. I'm taking a look now.

michaelkirk (approved):
Much ink and blood has been spilled over degenerate geometries. Thank you for adding more testing.

PR #1486, opened 2025-12-26
Split TriangulateDelaunay trait to support Point collections

urschrei:
New trait structure:

- `TriangulateDelaunayUnconstrained`: **new** trait with unconstrained triangulation methods only
- `TriangulateDelaunay`: now **extends** `TriangulateDelaunayUnconstrained`

New:
- `MultiPoint`, `Point`, and `&[Coord]` get TriangulateDelaunayUnconstrained via the `CoordsIter` blanket impl
- They do NOT get `TriangulateDelaunay` (constrained methods) because they don't implement `LinesIter + Contains`
- Existing types like `Polygon`, `Triangle`, `Rect` etc. get both traits

Why are you doing this?? Because I want to be able to generate Voronoi diagrams from `Point` collections (and, indeed triangulate them).

**NB** I don't believe this is breaking, as the new trait structure is additive.

---

### DIFF
--- a/geo-benches/src/triangulate.rs
+++ b/geo-benches/src/triangulate.rs
@@ -1,5 +1,5 @@
 use criterion::{criterion_group, criterion_main};
-use geo::algorithm::{TriangulateDelaunay, TriangulateEarcut};
+use geo::algorithm::{TriangulateDelaunay, TriangulateDelaunayUnconstrained, TriangulateEarcut};
 use geo::geometry::Polygon;
 use geo::triangulate_delaunay::DelaunayTriangulationConfig;
 
@@ -11,7 +11,8 @@ fn criterion_benchmark(c: &mut criterion::Criterion) {
             bencher.iter(|| {
                 for poly in &multi_poly {
                     let triangulation =
-                        TriangulateDelaunay::unconstrained_triangulation(poly).unwrap();
+                        TriangulateDelaunayUnconstrained::unconstrained_triangulation(poly)
+                            .unwrap();
                     assert!(triangulation.len() > 1);
                     criterion::black_box(triangulation);
                 }
@@ -48,7 +49,8 @@ fn criterion_benchmark(c: &mut criterion::Criterion) {
     c.bench_function("TriangulateSpade (unconstrained) - large_poly", |bencher| {
         let poly = Polygon::new(geo_test_fixtures::norway_main::<f64>(), vec![]);
         bencher.iter(|| {
-            let triangulation = TriangulateDelaunay::unconstrained_triangulation(&poly).unwrap();
+            let triangulation =
+                TriangulateDelaunayUnconstrained::unconstrained_triangulation(&poly).unwrap();
             assert!(triangulation.len() > 1);
             criterion::black_box(triangulation);
         });
--- a/geo/src/algorithm/triangulate_delaunay.rs
+++ b/geo/src/algorithm/triangulate_delaunay.rs
@@ -61,36 +61,43 @@ impl<T: GeoFloat + SpadeNum> SpadeTriangulationFloat for T {}
 
 pub type Triangles<T> = Vec<Triangle<T>>;
 
-// seal the trait that needs to be implemented for TriangulateDelaunay to be implemented. This is done
-// so that we don't leak these weird methods on the public interface.
+// Sealed traits for triangulation requirements. Split into two:
+// - UnconstrainedRequirementTrait: only needs coords (for point collections)
+// - ConstrainedRequirementTrait: needs coords + lines + contains (for polygons etc.)
 mod private {
     use super::*;
 
     pub(crate) type CoordsIter<'a, T> = Box<dyn Iterator<Item = Coord<T>> + 'a>;
 
-    pub trait TriangulationRequirementTrait<'a, T>
+    /// Requirement trait for unconstrained triangulation: only needs coordinates
+    pub trait UnconstrainedRequirementTrait<'a, T>
     where
         T: SpadeTriangulationFloat,
     {
-        /// collect all the lines that are relevant for triangulations from the geometric object that
-        /// should be triangulated.
-        ///
-        /// intersecting lines are allowed
-        fn lines(&'a self) -> Vec<Line<T>>;
-        /// collect all the coords that are relevant for triangulations from the geometric object that
-        /// should be triangulated
+        /// Collect all coords from the geometry
         fn coords(&'a self) -> CoordsIter<'a, T>;
-        /// define a predicate that decides if a point is inside of the object (used for constrained triangulation)
+    }
+
+    /// Requirement trait for constrained triangulation: needs coords, lines, and containment
+    pub trait ConstrainedRequirementTrait<'a, T>: UnconstrainedRequirementTrait<'a, T>
+    where
+        T: SpadeTriangulationFloat,
+    {
+        /// Collect all lines from the geometry (intersecting lines are allowed)
+        fn lines(&'a self) -> Vec<Line<T>>;
+
+        /// Check if a point is inside the geometry (used for constrained triangulation)
         fn contains_point(&'a self, p: Point<T>) -> bool;
 
-        // processing of the lines that prepare the lines for triangulation.
+        // Processing of the lines that prepare the lines for triangulation.
         //
-        // `spade` has the general limitation that constraint lines cannot intersect or else it
-        // will panic. This is why we need to manually split up the lines into smaller parts at the
-        // intersection point
+        // Spade has the limitation that constraint lines cannot intersect or else it
+        // will panic. This is why we need to manually split up the lines into smaller
+        // parts at the intersection point.
         //
-        // there's also a preprocessing step which tries to minimize the risk of failure of the algo
-        // through edge cases (thin/flat triangles are prevented as much as possible & lines are deduped, ...)
+        // There's also a preprocessing step which tries to minimize the risk of failure
+        // through edge cases (thin/flat triangles are prevented as much as possible,
+        // lines are deduped, etc.)
         fn cleanup_lines(lines: Vec<Line<T>>, snap_radius: T) -> TriangulationResult<Vec<Line<T>>> {
             let (known_coords, lines) = preprocess_lines(lines, snap_radius);
             prepare_intersection_contraint(lines, known_coords, snap_radius)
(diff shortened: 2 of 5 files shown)


geo-benches/src/triangulate.rs
```diff
@@ -11,7 +11,8 @@ fn criterion_benchmark(c: &mut criterion::Criterion) {
             bencher.iter(|| {
                 for poly in &multi_poly {
                     let triangulation =
-                        TriangulateDelaunay::unconstrained_triangulation(poly).unwrap();
+                        TriangulateDelaunayUnconstrained::unconstrained_triangulation(poly)
```
michaelkirk, line 14: Not a big deal, but aren't the changes in this file superfluous? The old code should still work the same, right?

geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -61,82 +61,82 @@ impl<T: GeoFloat + SpadeNum> SpadeTriangulationFloat for T {}
...
             prepare_intersection_contraint(lines, known_coords, snap_radius)
         }
     }
 }
 
-/// Triangulate polygons using a [Delaunay
-/// Triangulation](https://en.wikipedia.org/wiki/Delaunay_triangulation)
+/// Unconstrained Delaunay triangulation for geometries that provide coordinates.
+///
+/// This trait provides triangulation methods that only require point coordinates,
```
michaelkirk:
```suggestion
/// This trait provides triangulation methods based only on point coordinates,
```

What you had is clear in the context of this PR, but this change might make for clearer documentation.

geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -61,82 +61,82 @@ impl<T: GeoFloat + SpadeNum> SpadeTriangulationFloat for T {}
...
+/// making it suitable for point collections like `MultiPoint`, `Vec<Point>`, etc.
 ///
-/// This trait contains both constrained and unconstrained triangulation methods. To read more
-/// about the differences of these methods also consult [this
-/// page](https://en.wikipedia.org/wiki/Constrained_Delaunay_triangulation)
-pub trait TriangulateDelaunay<'a, T>: private::TriangulationRequirementTrait<'a, T>
+/// For constrained triangulation methods that respect geometry boundaries,
```
michaelkirk:
```suggestion
/// For constrained triangulation methods that respect the geometries existing boundaries,
```
clearer?

geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -331,36 +415,22 @@ where
     }
 }
 
-// it would be cool to impl the trait for GS: AsRef<[G]> but I wasn't able to get this to compile
-// (yet)
-
-impl<'a, T, G> private::TriangulationRequirementTrait<'a, T> for Vec<G>
+// Vec<G> where G has constrained triangulation
```
michaelkirk: comment typo - shouldn't that be **un**constrained?

Could probably just delete this comment.

geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -370,6 +440,13 @@ where
     }
 }
 
+// Note: &[G] impls are omitted because &[Coord] implements CoordsIter, which would
+// conflict with &[G] where G: TriangulateDelaunay. Users can use Vec<G> instead.
```
michaelkirk: Hm... I'm not sure I understand. It seems like previously you could triangulate a slice of coords, but now you can't?

michaelkirk (commented):
I have a clarifying question about whether we're losing functionality here WRT to coord slices.

My other comments should be considered non-blocking.

geo/src/algorithm/triangulate_delaunay.rs
urschrei: Previously, `TriangulationRequirementTrait` required `LinesIter + CoordsIter + Contains`. Because `&[Coord]` doesn't implement `LinesIter` or `Contains`, it couldn't be triangulated. But it can now! However, slices of _previously_ triangulate-able types (`&[Polygon]` and friends) now _can't_ be triangulated: the ability to triangulate them wasn't a blanket impl, it was done directly per geometry type (via auto-deref), so you could iterate over the slice and it would work.

Let me try to implement `LinesIter` and co for borrowed geometries…

urschrei: OK, ba89f4fb **adds** support for triangulating borrowed impls, but the PR still **removes** support for slices of geometries other than `&[Coord]` (which were not triangulate-able beforehand). This is breaking! Of course, `&Multi*` works fine.

So, it's not clear that anyone was specifically triangulating slices of geometries before, but if they were this PR will that.

urschrei: I think I've managed to work around this without creating a huge annoying mess of type tetris. Some new changes in 5726c701 (to be squashed eventually):

  | File                            | Implementation                                                                          |
  |---------------------------------|-----------------------------------------------------------------------------------------|
  | coords_iter.rs:557-579          | `impl<T: CoordsIter> CoordsIter for &T`                                                   |
  | lines_iter.rs:149-156           | `impl<'a, T: LinesIter<'a>> LinesIter<'a> for &T`                                         |
  | contains/mod.rs:46-53           | `impl<T, Rhs> Contains<Rhs> for &T where T: Contains<Rhs>`                                |
  | triangulate_delaunay.rs:442-468 | `impl UnconstrainedRequirementTrait for &[G] + impl ConstrainedRequirementTrait for &[G]` |


That gives us:

  | Type                   | Before dfde9073 | After dfde9073 | After  5726c701 |
  |------------------------|-----------------|----------------|-----------------|
  | `Polygon`        | Yes             | Yes                   | Yes            |
  | `&Polygon`    | No              | No                    | Yes            |
  | `&[Polygon]`     | Yes             | No                       | Yes            |
  | `Vec<Polygon>`           | Yes             | Yes                    | Yes            |
  | `&[Coord]` | No              | Yes                  | Yes            |


So we're now no longer breaking…

geo/src/algorithm/contains/mod.rs
```diff
@@ -43,6 +43,15 @@ pub trait Contains<Rhs = Self> {
     fn contains(&self, rhs: &Rhs) -> bool;
 }
 
+impl<T, Rhs> Contains<Rhs> for &T
```
michaelkirk: I commented this out and the branch still builds. Am I missing something?

geo/src/algorithm/coords_iter.rs
```diff
@@ -550,6 +550,34 @@ impl<'a, T: CoordNum> CoordsIter for &'a [Coord<T>] {
     }
 }
 
+// ┌─────────────────────────────────┐
+// │ Implementation for & references │
+// └─────────────────────────────────┘
+
+impl<T: CoordsIter> CoordsIter for &T {
```
michaelkirk: As mentioned elsewhere, this new impl seems unnecessary. 

(Same concern also for the new LinesIter impl). 

Conceptually it's not unreasonable, but why in this PR? It seems unrelated and unused as far as I can see.

geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -331,18 +415,21 @@ where
...
 }
 
-// it would be cool to impl the trait for GS: AsRef<[G]> but I wasn't able to get this to compile
-// (yet)
-
-impl<'a, T, G> private::TriangulationRequirementTrait<'a, T> for Vec<G>
+impl<'a, T, G> private::UnconstrainedRequirementTrait<'a, T> for Vec<G>
 where
     T: SpadeTriangulationFloat + 'a,
     G: TriangulateDelaunay<'a, T>,
```
michaelkirk:
```suggestion
    G: TriangulateDelaunayUnconstrained<'a, T>,
```

copy/paste error I think?

geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -352,15 +439,25 @@ where
...
 
-impl<'a, T, G> private::TriangulationRequirementTrait<'a, T> for &[G]
+// Note: We add LinesIter bound to disambiguate from &[Coord] which implements CoordsIter.
+// Coord doesn't implement LinesIter, so this prevents the coherence conflict.
+impl<'a, 'l, T, G> private::UnconstrainedRequirementTrait<'a, T> for &[G]
 where
+    'a: 'l,
     T: SpadeTriangulationFloat + 'a,
-    G: TriangulateDelaunay<'a, T>,
+    G: TriangulateDelaunay<'a, T> + LinesIter<'l, Scalar = T>,
```
michaelkirk:
```suggestion
    G: TriangulateDelaunayUnconstrained<'a, T> + LinesIter<'l, Scalar = T>,
```

copy/paste error I think?

geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -370,6 +467,10 @@ where
     }
 }
 
+// Note: Point collections like MultiPoint, Point, and &[Coord] get
+// TriangulateDelaunayUnconstrained automatically via the blanket impl for CoordsIter.
+// They don't implement LinesIter + Contains, so they don't get TriangulateDelaunay.
```
michaelkirk: Maybe this comment answers my other question about the new seemingly unnecessary blanket impls. If so, could you please add a test as an example of what breaks when those new impls are commented out?

geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -784,4 +885,102 @@ mod spade_triangulation {
...
+        let triangulation = coords.unconstrained_triangulation();
+        assert_num_triangles(&triangulation, 1);
+    }
+
+    #[test]
+    fn borrowed_polygon_triangulates() {
+        use super::{TriangulateDelaunay, TriangulateDelaunayUnconstrained};
+
+        let polygon = Polygon::new(
+            LineString::new(vec![
```
michaelkirk: I realize this is probably copy/paste from the pre-existing tests for consistency but I really prefer `wkt!` for test fixtures - less to read and (more importantly) I can easily examine the shapes in an external tool. 

Totally fine to keep it for consistency for now, but maybe someday converting all the test fixtures en-masse would be a straight forward LLM guided task.

geo/src/algorithm/triangulate_delaunay.rs
michaelkirk: Thanks for the helpful chart!

geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -149,10 +149,56 @@ where
             .map(triangulation_to_triangles)
     }
 
-    /// returns triangulation that's based on the points of the geometric object and also
-    /// incorporates the lines of the input geometry
+    /// Same as `unconstrained_triangulation`, but returns Spade's triangulation result directly.
+    fn unconstrained_triangulation_raw(
+        &'a self,
+    ) -> TriangulationResult<DelaunayTriangulation<Point2<T>>> {
```
michaelkirk: The PR description discusses splitting up the Delaney trait to allow Unsconstrained triangulations in a separate trait. But this is a new method right?

I'm not sure what this is used for yet — maybe it's required by your upcoming voronoi work, and can't be helped, but do we need to leak spade types in our own API? Could this go into the sealed trait? Or you could punt this change to the PR that actually requires it if that's an option.

Also since this method is "Same as `unconstrained_triangulation` but returns Spade's triangulation directly" could we DRY up `unconstrained_triangulation` as

```
    fn unconstrained_triangulation(
        &'a self,
    ) -> TriangulationResult<Triangles<T>> {
        self.unconstrained_triangulation_raw()
            .map(triangulation_to_triangles)
    }
```

I have symmetrical concerns with the new constrained flavors of the `_raw` methods.

geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -149,10 +149,56 @@ where
...
+            |mut tris, p| {
+                tris.insert(p).map_err(TriangulationError::SpadeError)?;
+                Ok(tris)
+            },
+        )
+    }
+
+    /// same as `unconstrained_triangulation_raw`, but with a tolerance parameter for snapping
+    /// nearby points together before triangulation
+    fn unconstrained_triangulation_raw_with_tolerance(
```
michaelkirk:
```suggestion
    fn unconstrained_triangulation_raw_with_snap_radius(
```

Seemingly everywhere else in the file we call this snap_radius - is there a reason to diverge here?

Or actually, maybe we should follow the `config: DelaunayTriangulationConfig<T>` pattern on the Constrained flavor?

I can see the argument that it's overly ceremonious to have a config object with one parameter, but the Constrained vs Unconstrained traits are symmetrical, and consistency seems natural to me. Alternatively we could get rid of the config (Breaking change!) in both places and pass snap_radius only.

What do you think?

geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -149,10 +149,56 @@ where
...
+        &'a self,
+        tolerance: T,
```
michaelkirk:
```suggestion
        snap_radius: T,
```

geo/src/algorithm/coords_iter.rs
urschrei: Hmm I swear they were needed for `&[G]`. But you're right, `polygon_slice_triangulates` still passes. Removing!

geo/src/algorithm/triangulate_delaunay.rs
urschrei: Updated for clarity: we now explain _why_ this is the case, rather than just pointing it out.

geo/src/algorithm/triangulate_delaunay.rs
urschrei: Done!

geo/src/algorithm/triangulate_delaunay.rs
urschrei: I've updated the name to `_with_config`. Seems better.

geo/src/algorithm/triangulate_delaunay.rs
urschrei: OK, new methods have been punted into #1487, DRYd up, and #1490 removes the `spade` types on the way there.

urschrei:
[muttering] the PR's practically empty now

michaelkirk:
Can you please add a changelog entry before merging?

urschrei:
> Can you please add a changelog entry before merging?

Yep, was just holding off til the scope was clear.

PR #1487, opened 2025-12-27
Add Voronoi trait

urschrei:
Adds the `Voronoi` trait with methods to compute Voronoi diagrams:

- `voronoi_edges()`: Returns line segments from unconstrained Delaunay triangulation
- `voronoi_edges_with_params()`: Same with configurable tolerance and clipping
- `voronoi_cells()`: Returns `Vec<Polygon<T>>` with cells clipped to padded bounding box
- `voronoi_cells_with_params()`: Same with configurable tolerance and clipping

Configuration is via VoronoiParams builder:
- `.tolerance(T)`: Snap nearby points before triangulation
- `.clip(VoronoiClip)`: Control clipping behaviour
  - Padded: 50% padding around input bbox (default, matches PostGIS)
  - Envelope: Clip to exact input bounding box
  - `Polygon(&Polygon<T>)`: Clip to custom boundary

Cells are constructed by iterating over Spade's `voronoi_faces()` and converting vertices (circumcenters for inner, bounding box intersections for outer) into polygon rings.

As part of this work, the point-snapping functionality in `triangulate_delaunay` has been updated to use an `rstar` tree.

---

Depends on #1486 and #1490, please review / merge those first.

This functionality requires a new geo-types release which incorporates the changes in #1490, and a corresponding bump to the geo-types version in geo


urschrei:
london_postcodes

Some example output. I haven't written a benchmark, but single-core perf on my M2 is around <s>5.5</s> 2.5 seconds to produce a clipped Voronoi diagram from 2693630 input vertices. That drops to around 1.3 seconds (including the snapping step) if you use a tolerance of `250.0` (points <= 100 metres apart are snapped together, resulting in fewer, larger Voronoi cells).

geo/src/algorithm/voronoi.rs
```diff
@@ -0,0 +1,1014 @@
...
+//! ```
+//!
+// TODO: This module manually computes bounding boxes from Spade triangulation vertices because
+// Spade's `Point2<T>` type is not compatible with geo-types. If future work converts Spade
+// vertices to geo-types (e.g. `Coord<T>`), the following geo traits could be used instead:
+//   - `BoundingRect` trait for bounds computation (replaces `compute_bounds_from_vertices`)
+//   - `CoordsIter` trait for vertex iteration
+// The `line_intersection` and `Euclidean::distance` functions from geo are already used here.
+//
+// TODO: Cell clipping currently uses BooleanOps intersection. Alternative approaches that
```
urschrei, line 77: These are possible TODOs for follow-up work; I'm fairly happy with the existing  clipping mechanisms, which took a long time to get into their current state. I suspect they could be further simplified if we could clip `LineString` geometries using `BooleanOps`, but I don't want to hold this up while we think through that.

geo/src/algorithm/voronoi.rs
```diff
@@ -0,0 +1,1034 @@
...
+        //
+        // Optimisation: skip intersection for cells entirely within the clip bounds.
+        // Interior cells (not on the convex hull) often fall entirely within the
+        // clipping region, so we can avoid the BooleanOps overhead for those.
+        let clip_rect = clip_poly.bounding_rect();
+
+        Ok(raw_cells
+            .into_iter()
+            .flat_map(|cell| {
+                // Skip intersection if cell is entirely within clip bounds
```
urschrei, line 548: This gets us a 40-45 % speedup (inspired by JTS: https://github.com/locationtech/jts/blob/7b79342eb04dc572a02b6520c7c2d2e58b54ae86/modules/core/src/main/java/org/locationtech/jts/triangulate/VoronoiDiagramBuilder.java#L184)

geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -38,6 +39,9 @@ pub enum TriangulationError {
     SpadeError(spade::InsertionError),
     LoopTrap,
     ConstraintFailure,
+    /// Input points are collinear; Voronoi cells cannot be computed.
+    /// Use `voronoi_edges()` instead to get perpendicular bisectors.
+    CollinearInput,
```
michaelkirk: This is an error returned by Voronoi, not triangulation, right? So how are users of the triangulation API supposed to handle it?

It seems like there should be a new Voronoi error enum, of which TriangulationError is a case.

If we *do* keep it here, because this is a public exhaustive enum, it's a breaking change, so we should document in the changelog.

geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -52,8 +56,8 @@ pub type TriangulationResult<T> = Result<T, TriangulationError>;
 
 // ======= Float trait ========
 
-pub trait SpadeTriangulationFloat: GeoFloat + SpadeNum {}
-impl<T: GeoFloat + SpadeNum> SpadeTriangulationFloat for T {}
+pub trait SpadeTriangulationFloat: GeoFloat + SpadeNum + RTreeNum {}
```
michaelkirk, line 56: Breaking change, but only for those mythical users using exotic numeric types. For basic types (i32,i64,f32,f64, etc) they should already impl RTreeNum.  Still we should note it in the changelog.

As an aside, there are several places in the codebase where we do `GeoNum + RTreeNum`. Since rtree is not an optional dependency, I wonder if we should simplify by constraining GeoNum like `GeoNum: RTreeNum + other existing reqs`

I wouldn't do that in this PR, but something to potentially cleanup separately. It would be breaking for geo (but again only for those using exotic numeric types). There might be downsides for exotic use cases. I'll create a discussion.

geo/src/algorithm/voronoi.rs
michaelkirk: > If we could clip LineString geometries using BooleanOps,

Is that something other than: 

https://docs.rs/geo/latest/geo/algorithm/bool_ops/trait.BooleanOps.html#method.clip

geo/src/algorithm/voronoi.rs
```diff
@@ -0,0 +1,1001 @@
...
+    Rect::new(
+        (base.min().x - padding, base.min().y - padding),
+        (base.max().x + padding, base.max().y + padding),
+    )
+}
+
+/// Return the four edges of a bounding box as [top, right, bottom, left].
+///
+/// Edge order does not affect correctness of intersection calculations.
+fn rect_edges<T: SpadeTriangulationFloat>(bounds: &Rect<T>) -> [Line<T>; 4] {
```
michaelkirk: see Rect::to_lines

geo/src/algorithm/voronoi.rs
urschrei: Oh, hmm. I think I looked at the signature and got confused and didn't think it applied to 1D geoms. I'll revisit.

geo/src/algorithm/triangulate_delaunay.rs
```diff
@@ -136,18 +140,44 @@ where
...
+                Ok(tris)
+            },
+        )
+    }
+
+    /// Same as `unconstrained_triangulation_raw`, but with configuration for snapping
+    /// nearby points together before triangulation.
+    fn unconstrained_triangulation_raw_with_config(
+        &'a self,
+        config: DelaunayTriangulationConfig<T>,
```
michaelkirk, line 162: This predates your PR, but I was surprised to learn that `DelaunayTriangulationConfig.snap_radius` defaults to 0.000_1. 

It's an admittedly small number, but a non-zero amount.

Specifically, this leads to the (IMO) surprising behavior:
```
// Does no snapping
unconstrained_triangulation_raw()
// snaps to 0.000_1
unconstrained_triangulation_raw_with_config(Default::default())
```

What's more, the `constrained_triangulation` **requires** a config object.

```
// snaps to 0.000_1
constrained_triangulation(Default::default())
```

So if you want to have a non-snapping constrained triangulation, you can do this:
```
// snaps to 0.000_1
constrained_triangulation(DelaunayTriangulationConfig { snap_radius: 0 })
```

It will still "snap" (dedupe) identical points, which, I think, is different from truly **no** snapping as in `unconstrained_triangulation_raw`.

So, firstly, I'm not sure why these API's aren't more parallel in naming an behavior, like this:
```
constrained_triangulation()
constrained_triangulation_with_config(config)
unconstrained_triangulation()
unconstrained_triangulation_with_config(config)
```

Secondly, I think (?) it doesn't affect your new code, because we only ever call `unconstrained_triangulation_raw_with_config` with the `DelaunayTriangulationConfig.snap_radius = Voronoi::params.tolerance` and `Voronoi::params.tolerance` *does* default to 0 (non-snapping).

Can you please confirm or deny my interpretation.

As long as I'm understanding correctly that it cannot affect this new voronoi impl, I think we should just leave it for now. But it might be worth revisiting for the triangulation APIs in a follow-up.

geo/src/algorithm/voronoi.rs
```diff
@@ -0,0 +1,1001 @@
...
+///     .tolerance(0.001)
+///     .clip(VoronoiClip::Envelope);
+/// ```
+#[derive(Debug, Clone)]
+pub struct VoronoiParams<'a, T: SpadeTriangulationFloat> {
+    /// Points within this distance are snapped together before triangulation.
+    ///
+    /// Similar to PostGIS `ST_VoronoiPolygons` tolerance parameter.
+    /// Default is 0.0 (no snapping).
+    pub tolerance: T,
```
michaelkirk, line 197: This is the same as  DelaunayTriangulationConfig.snap_radius and SpadeTriangulationConfig.snap_radius, can we consolidate on naming? 

Or maybe it's worth diverging from our existing naming to align with PostGIS's "tolerance" parameter?

geo/src/algorithm/voronoi.rs
```diff
@@ -0,0 +1,1001 @@
...
+    /// ]);
+    /// let edges = sites.voronoi_edges().unwrap();
+    /// assert_eq!(edges.0.len(), 3);
+    /// ```
+    fn voronoi_edges(&'a self) -> Result<MultiLineString<T>, TriangulationError>;
+
+    /// Compute Voronoi edges with custom parameters.
+    ///
+    /// See [`VoronoiParams`] for available configuration options.
+    fn voronoi_edges_with_params(
```
michaelkirk, line 289: Am I understanding that all the elements of `MultiLineString` returned have exactly 2 coords? If so, returning `Vec<Line>` would likely be faster.

On the other hand, it's arguably "more useful" to return a Geometry (like MultiPolygon) because we implement a lot of traits for `MultiPolygon` that we don't for `Vec<Line>`.

So I guess the question is:

Is the annoyance of having to convert a `Vec<Line>` to `MultiLineString` on the occasions that you actually need it, outweighed by the performance gains you get when you don't?

If we do return a `Vec<Line>`, and somone actually needs the `MultiLineString`, we could add an `impl From<Vec<Line>> for MultiLineString` to ameliorate some of that tedium. (__edit:__ I think this method already exists)

It's also arguably nicer that a `Line` communicates how many points it has, whereas working with LineStrings you have to rely on documentation and judicious use of `unwrap`.

TBH, I don't really know how this API will be used, so maybe I'm off base.

(Also, I might be overly sensitive to unnecessary heap allocations due to spending so much time in the geojson crate lately 😅)

geo/src/algorithm/voronoi.rs
```diff
@@ -0,0 +1,1001 @@
...
+        },
+    );
+    Rect::new((min_x, min_y), (max_x, max_y))
+}
+
+/// Compute a padded bounding box around the base bounds.
+///
+/// Padding is calculated as `padding_factor * max(width, height)` and applied
+/// uniformly on all sides. A factor of 0.5 matches PostGIS ST_VoronoiPolygons default.
+fn padded_bounds<T: SpadeTriangulationFloat>(base: Rect<T>, padding_factor: T) -> Rect<T> {
```
michaelkirk, line 734: I think this is equivalent to existing code:

```rust
padded_bounds(rect, 0.5)
// affine transform
// scale = 1 + (padding * 2)
rect.scale(2.0)
```

Right?

geo/src/algorithm/voronoi.rs
```diff
@@ -0,0 +1,1001 @@
...
+        })?
+    } else {
+        geom.unconstrained_triangulation_raw()?
+    };
+
+    let num_vertices = triangulation.num_vertices();
+    if num_vertices < 2 {
+        return Ok((
+            Vec::new(),
+            Rect::new((T::zero(), T::zero()), (T::zero(), T::zero())),
```
michaelkirk: Is this a reasonable default? Recall that 0 is just like any other coordinate. 

For example, would returning `Rect(6 6,6 6)` be OK? If not, then it's probably also not ok to return `Rect(0 0,0 0)`.

Would an Option or an Error type work here?

geo/src/algorithm/voronoi.rs
```diff
@@ -0,0 +1,1001 @@
...
+        assert_eq!(voronoi.0.len(), 3);
+
+        // The centroid should be at (0.5, sqrt(0.75) / 3)
+        let expected_centroid = Point::new(0.5, f64::sqrt(0.75) / 3.0);
+
+        // All lines should have one endpoint at the centroid
+        for line in voronoi {
+            let start: Point<_> = line.0[0].into();
+            // ugh
+            let end: Point<_> = Point(line.0.last().copied().unwrap());
```
michaelkirk:
```suggestion
            let end: Point<_> = line[1];
```

geo/src/algorithm/voronoi.rs
```diff
@@ -0,0 +1,1001 @@
...
+    fn voronoi_test_square() {
+        // Unit square
+        let square = wkt!(POLYGON((0.0 0.0, 1.0 0.0, 1.0 1.0, 0.0 1.0, 0.0 0.0)));
+
+        let voronoi = square.voronoi_edges();
+
+        // A square triangulated with a diagonal should have:
+        // 1. One finite edge (connecting the circumcenters of the two triangles)
+        // 2. Four infinite edges (perpendicular to each side)
+        assert_eq!(voronoi.unwrap().0.len(), 5);
```
michaelkirk: I think it would be more illustrative to compare against the actual expected output. 

Maybe not for real-world data where things get really verbose, but when we're using nice round numbers as our input, the results should usually be stable and readable.

e.g. 

```
let expected = wkt!(MULTILINESTRING((0.5 0.5,0.5 -0.5),(0.5 0.5,1.5 0.5),(0.5 0.5,0.5 0.5),(0.5 0.5,0.5 1.5),(0.5 0.5,-0.5 0.5)));
assert!(expected.relate(&voronoi).is_equal_topo());
```

This same comment applies to all of the tests in this module (except maybe the islington_post_boxes example. I assume that result would be too long to be useful at a glance.)

geo/src/algorithm/voronoi.rs
```diff
@@ -0,0 +1,1001 @@
...
+
+        // The bisector lines should be at x = -0.5 and x = 0.5
+        let expected_x_coords = vec![-0.5, 0.5];
+
+        for (line, expected_x) in voronoi.iter().zip(expected_x_coords) {
+            assert_relative_eq!(line[0].x, expected_x, epsilon = Float::epsilon());
+            assert_relative_eq!(
+                line.0.iter().last().unwrap().x,
+                expected_x,
+                epsilon = Float::epsilon()
```
michaelkirk: Why are you specifying the epsilon? 

But again, I think it would be better to compare to the actual expected output, it's predictable and stable as I understand it.

geo/src/algorithm/voronoi.rs
```diff
@@ -0,0 +1,1001 @@
...
+
+    #[test]
+    fn voronoi_cells_collinear_returns_error() {
+        use crate::algorithm::triangulate_delaunay::TriangulationError;
+
+        // Collinear points cannot produce Voronoi cells (only infinite strips)
+        let collinear = wkt!(LINESTRING(0.0 0.0, 1.0 0.0, 2.0 0.0));
+
+        let result = collinear.voronoi_cells();
+        match result {
```
michaelkirk: ignorable nit: Looks like a good spot for let/else

geo/src/algorithm/triangulate_delaunay.rs
urschrei: Good point – I'm going with a new `VoronoiError` enum with a member that wraps `TriangulationError` when appropriate.

geo/src/algorithm/triangulate_delaunay.rs
urschrei: > Can you please confirm or deny my interpretation.

I can confirm it!

The Voronoi code explicitly checks `params.tolerance > T::zero()` before calling the `_with_config` variant, and when it does, it passes `snap_radius:
  params.tolerance` directly rather than using `Default::default()`. Since `VoronoiParams::new()` sets `tolerance: T::zero()`, the default path calls
  `unconstrained_triangulation_raw()` with no snapping.

We should revisit this separately. Something like:

```rust
constrained_triangulation()
constrained_triangulation_with_config(config)
constrained_outer_triangulation()
constrained_outer_triangulation_with_config(config)
unconstrained_triangulation()
unconstrained_triangulation_with_config(config)
```

might be more clean?

geo/src/algorithm/voronoi.rs
urschrei: I went with the inspired-by-postGIS name because I thought it might be more familiar. Not tied to it, but it's a reasonable choice IMO.

geo/src/algorithm/voronoi.rs
urschrei: Hmm yes they're always two-element. `Line` is clearer. I'll update the API and add a hint about converting to `MultiLineString`.

geo/src/algorithm/voronoi.rs
urschrei: Yeah, the `Rect` is returned but then used in the caller to compute a clip polygon, which is then used to intersect with…an empty `Vec` of cells. So it "works" but it's sloppy. I'm going to return a new error enum member with a meaningful name instead.

geo/src/algorithm/voronoi.rs
urschrei: They're only equivalent for squares. For non-square rectangles, `padded_bounds` uses the max dimension to compute uniform padding, making the result "squarer" than a simple scale. This matches PostGIS behaviour where the padding ensures sufficient room for infinite rays regardless of the input's aspect ratio.

So I don't think we can replace it with scale because the semantics differ for non-square inputs.

geo/src/algorithm/voronoi.rs
urschrei: Resolved by switch to `Line` because it has an `Into`.

geo/src/algorithm/voronoi.rs
urschrei: Ha, fair.

geo/src/algorithm/triangulate_delaunay.rs
urschrei: We could do this, but when we'd need specific impls for

```rust
  impl SpadeTriangulationFloat for f32 {}
  impl SpadeTriangulationFloat for f64 {}
```

etc

geo/src/algorithm/voronoi.rs
urschrei: nm, it's always `Polygon`-`Polygon` now, so line clipping doesn't come into it.

geo/src/algorithm/triangulate_delaunay.rs
michaelkirk: Discussion moved to https://github.com/georust/geo/discussions/1497 since I don't think it affects this PR.

geo/src/algorithm/triangulate_delaunay.rs
michaelkirk: Sounds good to me!

geo/src/algorithm/voronoi.rs
michaelkirk: You're right, sorry!

geo/src/algorithm/voronoi.rs
```diff
@@ -0,0 +1,1047 @@
...
+    /// ```
+    /// use geo::{Voronoi, MultiPoint, MultiLineString, point};
+    ///
+    /// let sites = MultiPoint::from(vec![
+    ///     point!(x: 0.0, y: 0.0),
+    ///     point!(x: 1.0, y: 0.0),
+    ///     point!(x: 0.5, y: 1.0),
+    /// ]);
+    /// let edges = sites.voronoi_edges().unwrap();
+    /// let mls: MultiLineString<_> = edges.into_iter().collect();
```
michaelkirk:
```suggestion
    /// let mls = MultiLineString::from_iter(edges);
```

No big deal, but slightly more concise.

geo/src/algorithm/voronoi.rs
```diff
@@ -0,0 +1,1047 @@
...
+        params: VoronoiParams<'a, T>,
+    ) -> Result<Vec<Line<T>>, VoronoiError> {
+        // 1. Build Delaunay triangulation from input geometry vertices
+        // 2. Compute bounding box for clipping infinite edges
+        // 3. For each undirected Voronoi edge, handle three cases:
+        //    a) Inner-Inner: Both endpoints are circumcenters, emit finite edge directly
+        //    b) Inner-Outer: One endpoint is infinite, extend ray from circumcenter,
+        //       find closest intersection with bounding box, emit clipped edge
+        //    c) Outer-Outer: Both endpoints infinite (collinear points), create
+        //       vertical bisector line between sorted input points
```
michaelkirk:
```suggestion
        //       perpendicular bisector line between sorted input points
```

I think? Please verify, I'm just trying to be consistent with your other comments.

michaelkirk (approved):
I have some ignorable nits, but I think everything looks great! 

Merge when you're ready.

geo/src/algorithm/voronoi.rs
urschrei: Yep, bisectors are always perpendicular to the line connecting the collinear points.

urschrei:
> I have some ignorable nits, but I think everything looks great!
> 
> Merge when you're ready.

Thank you! These are solid improvements.

PR #1489, opened 2025-12-30
Amend the notes about MultiPolygon::unsigned_area

dabreegster:
---

I was confused by the note about signed and unsigned area for `MultiPolygon`, so clarify it.

### DIFF
--- a/geo/src/algorithm/area.rs
+++ b/geo/src/algorithm/area.rs
@@ -193,11 +193,15 @@ where
     T: CoordFloat,
 {
     fn signed_area(&self) -> T {
-        self.0.iter.map(|polygon| polygon.signed_area()).sum()
+        self.0
+            .iter()
+            .fold(T::zero(), |total, next| total + next.signed_area())
     }
 
     fn unsigned_area(&self) -> T {
-        self.0.iter.map(|polygon| polygon.unsigned_area()).sum()
+        self.0
+            .iter()
+            .fold(T::zero(), |total, next| total + next.unsigned_area())
     }
 }
 


geo/src/algorithm/area.rs
```diff
@@ -186,8 +186,8 @@ where
 }
 
 /// **Note.** The implementation is a straight-forward
-/// summation of the signed areas of the individual
-/// polygons. In particular, `unsigned_area` is not
+/// summation of the signed or unsigned areas of the individual
```
dabreegster, line 189: @rmanoka, you added this note in eb693154ee488639e3c46b73d258c16fdbefc8c6. Am I misunderstading something?

geo/src/algorithm/area.rs
```diff
@@ -204,7 +204,7 @@ where
     fn unsigned_area(&self) -> T {
         self.0
             .iter()
-            .fold(T::zero(), |total, next| total + next.signed_area().abs())
+            .fold(T::zero(), |total, next| total + next.unsigned_area())
```
dabreegster, line 204: These are equivalent, but maybe more clear -- it saves the caller from double-checking the Polygon implementation does this?

michaelkirk: Relevant: https://github.com/georust/geo/pull/487

In that PR the implementation of `signed_area` was updated so that valid Polygons would have a non-negative area, regardless of the winding of their exterior. Essentially the winding of the exterior becomes the "reference" direction for positive.

Also in that PR, this comment was added to unsigned_area, but the implementation of unsigned_area wasn't changed - it was only attempting to document existing behavior.

I'm totally guessing, but based on the previous comment, I wonder if the code was previously misread to be:

```diff
-           .fold(T::zero(), |total, next| total + next.signed_area().abs())
+           .fold(T::zero(), |total, next| total + next.signed_area()).abs()
```

That said, I _think_ the current behavior is correct, and it aligns with JTS et al

I think your suggestion for simplifying the code and correcting the doc comment are both good.

michaelkirk: `.map(|polygon| polygon.unsigned_area()).sum()` might be even clearer.

geo/src/algorithm/area.rs
```diff
@@ -186,8 +186,8 @@ where
 }
 
 /// **Note.** The implementation is a straight-forward
-/// summation of the signed areas of the individual
-/// polygons. In particular, `unsigned_area` is not
+/// summation of the signed or unsigned areas of the individual
+/// polygons. In particular, `signed_area` is not
```
michaelkirk: Actually this still reads strange. I think it'd be better to delete the last sentence entirely.

geo/src/algorithm/area.rs
dabreegster: Good point, done

geo/src/algorithm/area.rs
dabreegster: Done

geo/src/algorithm/area.rs
dabreegster: Apologies, I made some edits without running anything, hence the CI being so red! I've kept the old fold style, because otherwise we need to introduce a constraint on `std::iter::Sum`. Seemingly that trait requirement is then necessary on `Polygon` and more places. I'm not really sure why https://docs.rs/num-traits/0.2.19/num_traits/float/trait.Float.html doesn't implement `Sum` already.

OK to stick with the fold and avoid the larger change?

michaelkirk: Oh geeze, sorry for the bad advice.

Feature request to add `std::iter::Sum` to num_traits: https://github.com/rust-num/num-traits/issues/131

(It seems very unlikely to be adopted any time soon.)

rmanoka:
Agree that the comment is quite confusing;  best expln. I've is that I added the comment, before fixing the area calc, and forgot to remove it after.

PR #1490, opened 2025-12-31
Implement Spade HasPosition for Coord, use Coord as triangulation vertex type

urschrei:
- Add `spade_2` feature and dependency to `geo-types`
- Implement `spade::HasPosition` for `Coord<T`> in `geo-types`
- Update `geo/Cargo.toml` to enable `geo-types/spade_2` when spade is enabled
- Change triangulation to use `Coord<T>` as vertex type instead of `Point2<T>`
- Remove `to_spade_point` and `to_spade_line` helper functions
- Simplify `triangulation_to_triangles` since vertices are now `Coord` directly

---

This is follow-on work from #1486, and _precedes_ #1487. Rather than having to go to and fro between `Coord` and `Point2`, we implement `HasPosition` in a feature which is enabled by default in `geo`. The (mythical) `geo-types` users who don't use `geo` do not have the `spade_2` feature by default.

### DIFF
--- a/geo-types/Cargo.toml
+++ b/geo-types/Cargo.toml
@@ -44,6 +44,7 @@ rstar_0_10 = { package = "rstar", version = "0.10", optional = true }
 rstar_0_11 = { package = "rstar", version = "0.11", optional = true }
 rstar_0_12 = { package = "rstar", version = "0.12", optional = true }
 serde = { version = "1", optional = true, default-features = false, features = ["alloc", "derive"] }
+spade_2 = { package = "spade", version = "2", optional = true }
 
 [dev-dependencies]
 approx = ">= 0.4.0, < 0.6.0"
--- a/geo-types/CHANGES.md
+++ b/geo-types/CHANGES.md
@@ -1,5 +1,10 @@
 # Changes
 
+## Unreleased
+
+- Implement Spade HasPosition for Coord
+  - <https://github.com/georust/geo/pull/1490>
+
 ## 0.7.18 - 2025-12-01
 
 - Renamed features `use-rstar`, `use-rstar-0_8`, etc. to simply `rstar`, `rstar-0_8`, etc. (removing the `use-` prefix)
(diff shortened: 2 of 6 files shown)


geo-types/Cargo.toml
```diff
@@ -44,6 +46,7 @@ rstar_0_10 = { package = "rstar", version = "0.10", optional = true }
 rstar_0_11 = { package = "rstar", version = "0.11", optional = true }
 rstar_0_12 = { package = "rstar", version = "0.12", optional = true }
 serde = { version = "1", optional = true, default-features = false, features = ["alloc", "derive"] }
+spade_2 = { package = "spade", version = "2", optional = true }
```
michaelkirk, line 47: Please specify precise dependency versions in Cargo.toml

I know we are not consistent about this, but I don't think there's any value in consistency in this case.

geo-types/Cargo.toml
```diff
@@ -23,6 +23,8 @@ rstar_0_10 = ["dep:rstar_0_10", "approx"]
 rstar_0_11 = ["dep:rstar_0_11", "approx"]
 rstar_0_12 = ["dep:rstar_0_12", "approx"]
 
+spade_2 = ["dep:spade_2"]
```
michaelkirk: This is superfluous isn't it? We need to do this in the case of rstar because we are enabling other features alongside the `dep:`

geo-types/Cargo.toml
michaelkirk: We don't have an existing spade feature, so why `spade_2` and not just the default `spade`? 

My guess is this is laying the groundwork to be consistent in the case of a hypothetical spade_3 release? That seems fine, just making sure I understand.

michaelkirk (approved):
Love this great spade crate trait!

No blocking concerns, LGTM!

geo-types/Cargo.toml
urschrei: > My guess is this is laying the groundwork to be consistent in the case of a hypothetical spade_3 release

Yep, just being cautious.

PR #1492, opened 2026-01-06
Fix WASM build

weiznich:
This commit fixes the wasm build of geo by reducing the amount of enabled features for the rand crate to a subset that's wasm compatible by default. This requires putting some functionality of k-means behind feature gates.

This commit also adds a CI check for wasm-builds.

This is an alternative to #1476 

---

### DIFF
--- a/geo/src/algorithm/kmeans/mod.rs
+++ b/geo/src/algorithm/kmeans/mod.rs
@@ -181,10 +181,30 @@ impl<T: GeoFloat> KMeansParams<T> {
     /// - max_iter: 300
     /// - tolerance: 0.0001
     /// - No max_radius constraint
+    ///
+    /// This function is not available on WASM targets
+    #[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
     pub fn new(k: usize) -> Self {
+        let mut s = Self::new_with_seed(k, 42);
+        s.seed = None;
+        s
+    }
+
+    /// Create new k-means parameters with the specified number of clusters with an explicit random seed.
+    ///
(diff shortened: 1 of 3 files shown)


weiznich:
Hopefully that should fix the CI failures

geo/src/algorithm/kmeans/mod.rs
```diff
@@ -181,10 +181,30 @@ impl<T: GeoFloat> KMeansParams<T> {
     /// - max_iter: 300
     /// - tolerance: 0.0001
     /// - No max_radius constraint
+    ///
+    /// This function is not available on WASM targets
+    #[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
     pub fn new(k: usize) -> Self {
+        let mut s = Self::new_with_seed(k, 42);
```
michaelkirk, line 188: This is on non-wasm targets right? Shouldn't we be using the OSRng here?

michaelkirk (changes requested):

geo/src/algorithm/kmeans/mod.rs
michaelkirk: Oh! lol it gets zero'd out on the next line. My attention span is so short.

michaelkirk (approved):

geo/src/algorithm/kmeans/mod.rs
weiznich: I rather should have used a more explicit update syntax…

```
Self {
    seed: None,
    ..Self::new_with_seed(k, 42),
}
```

PR #1502, opened 2026-02-22
Add ball tree spatial index

urschrei:
---

Ball trees are a somewhat specialised spatial index that are more performant then e.g. KD trees or r-trees for certain arrangements of input data, for higher-dimensional data (not really relevant to us), and for clustering (highly relevant, as will become apparent in a follow-up PR).

I wrote this impl with Claude's assistance, but very happy to answer questions. It draws directly from license-compatible implementations in other crates (ball-tree, petal_cluster). One thing to note is that I've intentionally left it out of the `indexed` module as that's intended for existing geometries backed by a spatial index, and this is a bit more general-purpose.

### DIFF
--- a/geo/src/algorithm/ball_tree.rs
+++ b/geo/src/algorithm/ball_tree.rs
@@ -18,7 +18,7 @@
 //! # Construction
 //!
 //! Build a tree from any collection of point-like values via [`BallTree::new`].
-//! Bare [`Point`]s, [`Coord`]s, and [`MultiPoint`](crate::MultiPoint)s are all
+//! Bare [`Point`](crate::Point)s, [`Coord`](crate::Coord)s, and [`MultiPoint`](crate::MultiPoint)s are all
 //! accepted directly; to attach per-point data (labels, identifiers, etc.),
 //! wrap each point in a [`PointWithData`].
 //!
--- a/geo/src/algorithm/ball_tree.rs
+++ b/geo/src/algorithm/ball_tree.rs
@@ -17,60 +17,49 @@
 //!
 //! # Construction
 //!
-//! There are three ways to build a ball tree:
+//! Build a tree from any collection of point-like values via [`BallTree::new`].
+//! Bare [`Point`]s, [`Coord`]s, and [`MultiPoint`](crate::MultiPoint)s are all
+//! accepted directly; to attach per-point data (labels, identifiers, etc.),
+//! wrap each point in a [`PointWithData`].
 //!
-//! - **[`BuildBallTree::build_ball_tree`]**: a convenience trait implemented on
-//!   `[Point<T>]`, `Vec<Point<T>>`, and `MultiPoint<T>`. Use this when you just
-//!   need to index points without associated data.
+//! ```
+//! use geo::point;
+//! use geo::algorithm::ball_tree::BallTree;
 //!
-//!   ```
-//!   use geo::{MultiPoint, point};
-//!   use geo::algorithm::ball_tree::BuildBallTree;
+//! let tree = BallTree::new(vec![
+//!     point!(x: 0.0, y: 0.0),
+//!     point!(x: 1.0, y: 1.0),
+//!     point!(x: 2.0, y: 2.0),
+//! ]);
+//! let nearest = tree.nearest_neighbour(&point!(x: 0.1, y: 0.1)).unwrap();
+//! assert_eq!(nearest.point, point!(x: 0.0, y: 0.0));
+//! ```
 //!
-//!   let points = MultiPoint::new(vec![
-//!       point!(x: 0.0, y: 0.0),
-//!       point!(x: 1.0, y: 1.0),
-//!       point!(x: 2.0, y: 2.0),
-//!   ]);
-//!   let tree = points.build_ball_tree();
-//!   ```
+//! ```
+//! use geo::point;
+//! use geo::algorithm::ball_tree::{BallTree, PointWithData};
 //!
-//! - **[`BallTree::new`]**: takes a `Vec<Point<T>>` and a parallel `Vec<D>` of
-//!   associated data. Use this when each point carries a label, identifier, or
-//!   other payload.
+//! let tree = BallTree::new(vec![
+//!     PointWithData::new(point!(x: 0.0, y: 0.0), "origin"),
+//!     PointWithData::new(point!(x: 1.0, y: 1.0), "middle"),
+//!     PointWithData::new(point!(x: 2.0, y: 2.0), "far"),
+//! ]);
+//! let nearest = tree.nearest_neighbour(&point!(x: 0.1, y: 0.1)).unwrap();
+//! assert_eq!(*nearest.data, "origin");
+//! ```
 //!
-//!   ```
-//!   use geo::point;
-//!   use geo::algorithm::ball_tree::BallTree;
+//! To tune the leaf size, use [`BallTreeBuilder`]:
 //!
-//!   let points = vec![
-//!       point!(x: 0.0, y: 0.0),
-//!       point!(x: 1.0, y: 1.0),
-//!       point!(x: 2.0, y: 2.0),
-//!   ];
-//!   let labels = vec!["origin", "middle", "far"];
-//!   let tree = BallTree::new(points, labels);
-//!   let nearest = tree.nearest_neighbour(&point!(x: 0.1, y: 0.1)).unwrap();
-//!   assert_eq!(*nearest.data, "origin");
-//!   ```
+//! ```
+//! use geo::point;
+//! use geo::algorithm::ball_tree::BallTreeBuilder;
 //!
-//! - **[`BallTreeBuilder`]**: a builder that allows configuring parameters such
-//!   as the leaf size before construction. Use this when you need to tune tree
-//!   structure for your workload.
-//!
-//!   ```
-//!   use geo::point;
-//!   use geo::algorithm::ball_tree::BallTreeBuilder;
-//!
-//!   let points = vec![
-//!       point!(x: 0.0, y: 0.0),
-//!       point!(x: 1.0, y: 1.0),
-//!       point!(x: 2.0, y: 2.0),
-//!   ];
-//!   let tree = BallTreeBuilder::new()
-//!       .leaf_size(8)
-//!       .build_pointset(points);
-//!   ```
+//! let tree = BallTreeBuilder::with_leaf_size(16).build(vec![
+//!     point!(x: 0.0, y: 0.0),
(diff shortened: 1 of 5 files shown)


geo/src/algorithm/ball_tree.rs
```diff
@@ -0,0 +1,1253 @@
+//! # Construction
```
michaelkirk: Maybe include first a line about what a balltree is / is for.

geo/src/algorithm/ball_tree.rs
```diff
@@ -0,0 +1,1253 @@
...
+
+    /// Helper: brute-force nearest neighbour for verification.
+    fn brute_force_nn(points: &[Point<f64>], query: &Point<f64>) -> (usize, f64) {
+        points
+            .iter()
+            .enumerate()
+            .map(|(i, p)| {
+                let dx = p.x() - query.x();
+                let dy = p.y() - query.y();
+                (i, (dx * dx + dy * dy).sqrt())
```
michaelkirk: How about `Euclidean.distance(&p, &query)` here and the several other places doing distance calculations?

geo/src/algorithm/ball_tree.rs
```diff
@@ -0,0 +1,1253 @@
...
+                .leaf_size(leaf_size)
+                .build_pointset(points.clone());
+
+            for query in &queries {
+                // NN -- compare distances, not indices, because ties are
+                // broken arbitrarily depending on tree structure.
+                let nn = tree.nearest_neighbour(query).unwrap();
+                let (_, bf_dist) = brute_force_nn(&points, query);
+                assert!(
+                    (nn.distance - bf_dist).abs() < 1e-10,
```
michaelkirk: Can you update this and the other tests to use approx for these kinds of "almost" checks?

geo/src/algorithm/ball_tree.rs
```diff
@@ -0,0 +1,1253 @@
...
+    let dist_to_center = coord_distance_sq(query, center).sqrt();
+    let gap = dist_to_center - radius;
+    if gap > T::zero() {
+        gap * gap
+    } else {
+        T::zero()
+    }
+}
+
+/// Compare two `GeoFloat` values, treating NaN as equal (for ordering purposes).
```
michaelkirk: Elsewhere we use total_cmp  - is there a reason not to do so here? 

I'd prefer to have one way of doing things, rather than reinvent this wheel.

geo/src/algorithm/ball_tree.rs
```diff
@@ -0,0 +1,1253 @@
...
+    debug_assert!(count > 0);
+
+    // Single pass: compute centroid (sum) and axis dispersion (min/max)
+    let mut sum_x = T::zero();
+    let mut sum_y = T::zero();
+    let mut min_x = T::infinity();
+    let mut max_x = T::neg_infinity();
+    let mut min_y = T::infinity();
+    let mut max_y = T::neg_infinity();
+    for &idx in &indices[start..end] {
```
michaelkirk: A little cleaner:

```
for point in points[start..end] {
...
}
```

geo/src/algorithm/ball_tree.rs
```diff
@@ -0,0 +1,1253 @@
...
+        max_y = max_y.max(y);
+    }
+    let n = T::from(count).unwrap();
+    let center = Coord {
+        x: sum_x / n,
+        y: sum_y / n,
+    };
+
+    // Compute radius as max distance from centroid
+    let radius = indices[start..end]
```
michaelkirk:
```suggestion
    let radius = points[start..end]
```

geo/src/algorithm/ball_tree.rs
```diff
@@ -0,0 +1,1253 @@
...
+    }
+}
+
+// -- Trait: BuildBallTree -----------------------------------------------------
+
+/// Construct a [`BallTree`] from a collection of points (without associated data).
+///
+/// This is the convenience route for building a `BallTree<T, ()>`. To attach
+/// per-point data, use [`BallTree::new`] directly.
+pub trait BuildBallTree<T: GeoFloat> {
```
michaelkirk: Currently if you just want to index Points, you choose between:

- `BuildBallTree::build_ball_tree(points)` 
- `BallTreeBuilder::new().leaf_size(foo).build_pointset(points)` 

If you want to index points with Data you ensure points/data have same length, and then choose from:

- `BallTree::build(points, data)`
- `BallTreeBuilder::new().leaf_size(foo).build(points, data)`

Alternatively, we could follow RTree's design and introduce a new trait `BallTreePoint` (name TBD), implemented on Coord and/or Point, and (potentially) PointWithData<T>.

```
trait BallTreePoint {
  type T: GeoNum;
  /// used to compute distance to other BallTreePoint or Node.center
  fn coord(&self, other: Self) -> Coord<T>;
}

BallTree::new(points)
BallTree::new(points_with_data)
BallTreeBuilder::new().leaf_size(foo).build(points)
BallTreeBuilder::new().leaf_size(foo).build(points_with_data)
```

Then we could get rid of the BuildBallTree trait and the `build_pointset` method.

What do you think? I'm willing to be convinced otherwise.

michaelkirk: Regardless of the above, a nice discoverability boost for the builder might be something like:

```
let ball_tree = BallTree::with_leaf_size(6).build(points)

// requiring some impls like:
impl BallTree {
  fn builder() -> BallTreeBuilder  { BallTreeBuilder::new() }
  fn with_leaf_size(size: usize) -> BallTreeBuilder {
     BallTreeBuilder::leaf_size(size)
  }
}
```

It's not critical though - take it or leave it.

geo/src/algorithm/ball_tree.rs
```diff
@@ -0,0 +1,1253 @@
...
+                    (nn.distance - bf_dist).abs() < 1e-10,
+                    "NN distance mismatch with leaf_size={leaf_size} for query {query:?}"
+                );
+
+                // k-NN
+                let k = 5;
+                let knn = tree.nearest_neighbours(query, k);
+                let bf_knn = brute_force_knn(&points, query, k);
+                assert_eq!(knn.len(), bf_knn.len());
+                for (tree_r, (_, bf_d)) in knn.iter().zip(bf_knn.iter()) {
```
michaelkirk: Kind awkward how you destructure the bf_knn tuple but not the radius tree tuple.
```suggestion
                for ((_,radius_tree_d), (_, bf_d)) in knn.iter().zip(bf_knn.iter()) {
```

Similarly throughout the tests. It's not a big deal but makes it just a little harder to read the tests. I admit I'm being picky, but at the rate LLMs are generating PR lines, I've gotta defend my sanity.

geo/src/algorithm/ball_tree.rs
```diff
@@ -0,0 +1,1253 @@
...
+//! - [`BallTree::nearest_neighbour`] -- single nearest neighbour
+//! - [`BallTree::nearest_neighbours`] -- k nearest neighbours, sorted by distance
+//! - [`BallTree::within_radius`] -- all points within a given radius
+
+use crate::{Coord, GeoFloat, MultiPoint, Point};
+use std::cmp::Ordering;
+use std::collections::BinaryHeap;
+
+/// The default minimum number of points in a leaf node before we stop splitting.
+const DEFAULT_LEAF_SIZE: usize = 4;
```
michaelkirk: Completely naively, I'd expect a higher number here. I'm not sure if you tuned this and I'm wrong (v plausible!), and I know users can adjust this with the builder, but we should endeavor to have a good default.

michaelkirk (changes requested):
This was pretty understandable - nice work!

I don't have any big conceptual blockers, but would like to do a round of clarification.

As an addendum (read: not for this PR), and bearing in mind I've never used a ball tree index before, it seems like they would be equally well suited for geographic coordinates if we made the distance calculation pluggable. Does that seem like a plausible future avenue to you?

geo/src/algorithm/ball_tree.rs
urschrei: I tuned it for the HDBSCAN case, _then_ added the builder – let me see whether there's any consensus about defaults, since we can now use 4 specifically for the HDBSCAN case if need be.

geo/src/algorithm/ball_tree.rs
urschrei: minor: `for point in points[start..end]` can't apply as written, because `points` is immutable and stays in original insertion order; each recursion level owns `indices[start..end]`, which is the slice that gets partitioned in place. I've left a short comment at the top of `build_recursive` explaining this, and pushed the iteration-through-indices logic into a single named closure so it only appears once.

geo/src/algorithm/ball_tree.rs
urschrei: This is done, in a separate commit for now. I chose to put `with_leaf_size` on `BallTreeBuilder` rather than on `BallTree` itself, because `BallTree<T, D>`'s generics aren't inferable from a return type that's `BallTreeBuilder`, so `BallTree::with_leaf_size(6).build(points)` would have required (ugly imo) turbofish. Putting the constructor on the builder gets `BallTreeBuilder::with_leaf_size(6).build(points)` and inference works.

geo/src/algorithm/ball_tree.rs
urschrei: same `indices` cause as above.

geo/src/algorithm/ball_tree.rs
urschrei: Did some more testing. 16 seems better for 2d cases…

geo/src/algorithm/ball_tree.rs
michaelkirk: Sorry, I forgot a `&`

```
for point in &points[start..end] {
...
}
```

geo/src/algorithm/ball_tree.rs
michaelkirk: Ah, good point! Better how you have it.

geo/src/algorithm/ball_tree.rs
michaelkirk: I think I misunderstood what was happening here. 

I thought `tree_r` (returned from knn) had the same type as `brute_force_nn`. It was totally fine how you had it before. I'm also fine with the new version, but I regret wasting your time with the original request. 😓

michaelkirk (approved):
I think the new API that consolidated the "bare Point" with the "Point with Data" methods turned out great. Thanks for doing that.

michaelkirk:
I pushed up a little doc fix to be expedient. TBH, I don't understand why it was failing.

PR #1508, opened 2026-02-27
switch to earcut crate for performance improvement

michaelkirk:
---

Followup to https://github.com/georust/geo/pull/1507#issuecomment-3971913129

TODO: Currently a test in `stitch_triangulation` is failing. I haven't dug in very deeply. I'm not sure if this represents a bug in our stitch method or if we are breaking some calling contract with the new triangulation output.

/cc @JasmineLowen  - any interest in taking a look at the stitch failure?

## Benches

Benches have improved, even with the most basic integration.
```
TriangulateEarcut - small polys
                        time:   [5.3343 ms 5.4070 ms 5.4954 ms]
                        change: [-16.874% -15.773% -14.370%] (p = 0.00 < 0.05)
                        Performance has improved.

TriangulateEarcut - large_poly
                        time:   [1.5546 ms 1.5575 ms 1.5608 ms]
                        change: [-23.623% -23.377% -23.112%] (p = 0.00 < 0.05)
                        Performance has improved.
```

## Performance

There are further opportunities for speedups:

1. Expose a way to maintain a persistent EarCutter instance.
2. Expose a way to pass in a(re-usable) output buffer.
3. Avoid twice allocating the input vertices
   - Lazily compute vertex iterator
   - Return dynamic "getter"

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -119,7 +119,7 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - run: cargo check --all-targets --no-default-features
       # we don't want to test `proj-network` because it only enables the `proj` feature
-      - run: cargo test --features "proj serde earcutr multithreading"
+      - run: cargo test --features "proj serde earcut multithreading"
 
   geo_traits:
     name: geo-traits
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -23,6 +23,10 @@
   It wasn't referenced by the `triangle_indices` cut by earcutr, but you may notice a different triangulation for a given input.
 - Deprecate `StitchTriangulation`. Instead convert your triangles to Polygon and use unary_union.
   - <https://github.com/georust/geo/pull/1514>
+- Replace `earcutr` crate with the faster `earcut` crate.
+  - <https://github.com/georust/geo/pull/1508>
+  - The high level `TriangulateEarcut::earcut_triangles` API hasn't changed, but may return a different triangulation.
+  - BREAKING: The low level `TriangulateEarcut::earcut_triangles_raw` API now groups coordinates `[[x0, y0],[x1 y1]]]`, previously it was flattened: `[x0, y1, x1, y1]`
 
 ## 0.32.0 - 2025-12-05
 
--- a/geo/Cargo.toml
+++ b/geo/Cargo.toml
@@ -13,7 +13,9 @@ rust-version = "1.88"
 categories = ["science::geo"]
 
 [features]
-default = ["earcutr", "spade", "multithreading"]
+default = ["earcut", "spade", "multithreading"]
+# deprecated spelling earcutr->earcut
+earcutr = ["earcut"]
 proj-network = ["proj", "proj/network"]
 serde = ["dep:serde", "geo-types/serde"]
 multithreading = ["i_overlay/allow_multithreading", "geo-types/multithreading"]
@@ -26,7 +28,7 @@ use-serde = ["serde"]
 __allow_deprecated_features = []
 
 [dependencies]
-earcutr = { version = "0.5.0", optional = true }
+earcut = { version = "0.4.5", optional = true }
 spade = { version = "2.10.0", optional = true }
 float_next_after = "2.0.0"
 geo-types = { version = "0.7.18", features = ["approx", "rstar_0_12"] }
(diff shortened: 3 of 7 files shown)


frewsxcv (approved):

frewsxcv:
Approved in the sense that I'm +1 to switch away from my crate

michaelkirk:
This is a nice speedup. I'd like to incorporate it, but unfortunately it causes our stitch triangulation code to produce invalid output in some cases. 

From what I can tell, the earcut crate is returning a reasonable, valid triangulation, but there seems to be some kind of assumption in our stitch code that is violated by the earcut crate, but not the earcutr crate. I haven't quite nailed down what this is though.

If no one is using the triangulation code, we could deprecate it with a warning about correctness. It is notably faster than unioning via boolean operations if you already have a triangulation. 

As I recall it was originally introduced as an incremental step towards boolean operations back when our existing boolean-ops was panicky. We've since resolved the panic, so the value proposition has changed slightly.

I'm not personally that interested in digging into the implementation unless I know someone is using it.

michaelkirk:
With our speedy unary union operation, stitch triangles is actually not very fast. I think we should deprecate it. See #1514

JasmineLowen:
Hey, thanks for pinging me. Unfortunately I don't have as much time to work on geo related things anymore after a job switch sometime last year, so I would just politely decline to work on this any further. It pains me a bit but I just don't have the personal time resources necessary.

michaelkirk:
This is now based on #1514, so please merge that first.

michaelkirk:
I realized this is a breaking change for our `earcut_triangles_raw` method (and updated the changelog accordingly). 

Still OK to merge?

urschrei:
> I realized this is a breaking change for our `earcut_triangles_raw` method (and updated the changelog accordingly).
> 
> Still OK to merge?

I think this is a seldom-used corner of our public API. As ever, I appreciate your diligence in pointing out breakage; this is a reasonable trade-off IMO.

urschrei (approved):

frewsxcv:
The raw thing is basically an `rgis` exclusive method hah. Feel free to get rid of it

michaelkirk:
> The raw thing is basically an rgis exclusive method hah. Feel free to get rid of it

The MaplibreTile format is also using it, so I think we should keep it.

nyurik:
Would it make sense to on-board the `earcut` crate by Taku Fukada (@ciscorn) as a part of geo crate?

### PROs
* consistent usage of the same version of geo-types as all other algorithms, avoiding any kind of data conversion
* ability to use `Coord<i32>` instead of the built-in `Coord<f64>`
* wider community to maintain the code

### CONs
* slower iteration

michaelkirk:
Sorry, I don't understand what you're asking. Can you clarify what you'd like to see happen?

nyurik:
@michaelkirk I think it might be better to copy the earcut algorithm into the `geo` crate itself - as one of its algorithms.  This way that earcut algorithm would use exactly the same geo-types as all other algorithms, and it will be possible to avoid data conversion, more rapid iterations within the context of the geo crate, and most importantly - avoid `impl<T: CoordFloat>` requirement - so that `T` could be `i32` (important for MLT)

ciscorn:
@nyurik @michaelkirk Hi, I'm the author of the `earcut` crate.

I strongly agree that community maintenance would be better. I'd be happy to donate the crate to GeoRust. Please let me know if you'd like me to start the transfer.

About `Coord<i32>`: the earcut algorithm needs floating-point arithmetic internally. Since the API accepts an iterator of input coordinates, you can convert `Coord<i32>` to `Coord<f64>` there. The copy and conversion cost should be very small.

I also think it would be good to keep the `earcut` interface independent of `geo`, so it can be used in other areas such as computer graphics.

As @michaelkirk mentioned, the bigger performance issue is avoiding frequent repeated allocations. The crate already supports buffer reuse — it just depends on how `geo` exposes it.

michaelkirk:
> About Coord<i32>: the earcut algorithm needs floating-point arithmetic internally. Since the API accepts an iterator of input coordinates, you can convert Coord<i32> to Coord<f64> there. The copy and conversion cost should be very small.

I'm not very familiar with the earcutting algorithm, but a quick perusal shows some divisions that could be problematic for integers. However, at least some of them seem avoidable. I might have a go at this and see how far I can get.

> I also think it would be good to keep the earcut interface independent of geo, so it can be used in other areas such as computer graphics.

I totally agree. If anything, it might be useful to expose a data access trait or something that consumers could implement to avoid the cost of building `self.data` iff that data is already in the proper format.

ciscorn:
@michaelkirk You're right that an integer-optimized version like `EarcutInt` could be possible. I think it's worth exploring.

A data access trait seems a good idea!

Other than keeping the interface independent of `geo`, I don't have strong opinions on the API design. I'd be happy to see the crate improved by the community.

PR #1521, opened 2026-04-11
Perf: GeometryGraph constructor builds RTree

michaelkirk:
---

My original intent was to be able to use GeometryGraph outside of the context of PreparedGeometry without having to copy the setup code in `prepare_geometry` (specifically while playing around with #1472).

The nice side side effect of this is an incidental speedup. I think it's because we are rebuilding the geometry RTree in a couple places. If we build it once up front, we don't have to worry about this cost.

```
relate overlapping 50-point polygons
                        time:   [27.251 µs 27.693 µs 28.099 µs]
                        change: [-21.806% -19.731% -17.678%] (p = 0.00 < 0.05)
                        Performance has improved.

entire jts test suite   time:   [344.95 ms 346.44 ms 348.46 ms]
                        change: [-0.4298% +0.0268% +0.6061%] (p = 0.93 > 0.05)

jts test suite matching *Relate*
                        time:   [33.497 ms 33.642 ms 33.796 ms]
                        change: [-6.1824% -5.5581% -4.9284%] (p = 0.00 < 0.05)
                        Performance has improved.

disjoint polygons       time:   [19.555 µs 19.599 µs 19.656 µs]
                        change: [+0.1938% +0.4692% +0.9024%] (p = 0.00 < 0.05)

large rotated polygons  time:   [4.5333 ms 4.5564 ms 4.5896 ms]
                        change: [-28.900% -28.393% -27.871%] (p = 0.00 < 0.05)
                        Performance has improved.

offset polygons         time:   [4.2994 ms 4.3199 ms 4.3457 ms]
                        change: [-30.030% -29.516% -28.997%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/geo/src/algorithm/indexed/prepared_geometry.rs
+++ b/geo/src/algorithm/indexed/prepared_geometry.rs
@@ -1,10 +1,9 @@
 use crate::geometry::*;
-use crate::relate::geomgraph::{GeometryGraph, RobustLineIntersector};
+use crate::relate::geomgraph::GeometryGraph;
 use crate::{BoundingRect, GeometryCow, HasDimensions};
 use crate::{GeoFloat, Relate};
 
 use std::fmt::{Debug, Formatter};
-use std::rc::Rc;
 
 use crate::dimensions::Dimensions;
 use rstar::RTreeNum;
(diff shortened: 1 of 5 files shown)


geo/src/algorithm/indexed/prepared_geometry.rs
```diff
@@ -66,14 +65,8 @@ where
...
 {
-    let mut geometry_graph = GeometryGraph::new(0, geometry.clone().into());
-    let r_tree = geometry_graph.build_tree();
-    geometry_graph.set_tree(Rc::new(r_tree));
+    let geometry_graph = GeometryGraph::new(0, geometry.clone().into());
     let bounding_rect = geometry_graph.geometry().bounding_rect();
-
-    // TODO: don't pass in line intersector here - in theory we'll want pluggable line intersectors
-    // and the type (Robust) shouldn't be hard coded here.
-    geometry_graph.compute_self_nodes(Box::new(RobustLineIntersector::new()));
```
michaelkirk, line 76: This line (and the comment) got moved into the `GeometryGraph::new`.

urschrei (approved):
Well well. The robot reviewed too! I would not have caught that suggestion on line 24.

michaelkirk:
> Well well. The robot reviewed too! I would not have caught that suggestion on line 24.

Your job as unpaid code reviewer is safe for now @urschrei. I will not be taking the robots suggestion, because I prefer calling a cheap getter twice rather than introducing a new variable.

PR #1522, opened 2026-04-11
fix simply connected interior validation

michaelkirk:
---

~~Draft as this depends on #1521~~ Merged!

Built off of #1472, with some simplifications and clarity changes. I hope you don't mind that I've taken your code and run with it @bretttully. I would appreciate if you have any opinions on my changes/comments since you've studied the algorithm.


geo/src/algorithm/validation/polygon.rs
```diff
@@ -153,10 +185,195 @@ impl<F: GeoFloat> Validation for Polygon<F> {
...
+    // Which rings are connected, even if vicariously, through other rings.
+    let mut global = UnionFind::new(rings.len());
+    for (_intersection_coord, (intersecting_rings, intersection_is_on_a_vertex)) in
+        all_intersections
+    {
+        if !intersection_is_on_a_vertex {
+            continue;
+        }
+        // Which rings are connected at *this* coordinate
+        // let mut local = UnionFind::new(rings.len());
```
michaelkirk: I don't have an intuitive grasp on what the `local` UnionFind accomplishes. I've read the description in #1472, and it sort of of makes sense.

Can you come up with a test case which would cover this @bretttully? Currently all the tests pass even with this commented out like this.

bretttully:
No worries! I won't be able to take a look for a week or two, but will asap.

geo/src/algorithm/validation/polygon.rs
michaelkirk: I removed the notion of the "local" union find, because I believe it was no longer necessary.

I'm now grouping the intersections by coordinate - there's no longer a mechanism for us to "revisit" a coordinate later, and so I think the problem goes away.

michaelkirk:
Ok, this is ready for review. Despite it being slightly longer than #1472, I think it's a bit simpler. 

primarily (vs. #1472):

- We no longer try to prevent reporting "redundant" validation errors. Previously, there was some code to suppress "disconnected interior" validation errors when the faulty ring has already triggered another validation error. This simplified the code a little bit.

- Because we're grouping intersections by coordinate, the union-find algorithm gets simpler - we no longer need a "global" vs "local" connectivity check, we'll only encounter each intersection coordinate once.

- more comments and tests

michaelkirk:
Pinging @bretttully for review

geo/src/algorithm/validation/polygon.rs
```diff
@@ -165,10 +185,277 @@ impl<F: GeoFloat> Validation for Polygon<F> {
...
+        }
+
         Ok(())
     }
 }
 
+/// Convert a GeometryGraph edge index to a [`RingRole`].
+///
+/// Edge 0 is the exterior ring; edge N (N >= 1) is interior ring N-1.
+fn edge_index_to_ring_role(edge_idx: usize) -> RingRole {
```
urschrei: I think this will return incorrect indices for `RingRole` if a Polygon has empty interior rings: `GeometryGraph::add_polygon_ring` returns early for empty rings, so graph edge indices don't line up with interior-ring indices, but `edge_index_to_ring_role` assumes edge N = interior N−1. Our [test](https://github.com/georust/geo/pull/1522/changes#diff-79a6a081936ec82404df9c713d15e3d950e916b3389d7eb1806066b18a3bbe55R28) uses `wkt!`, which maybe doesn't add empty rings? But building those test geometries using `Polygon::new` should surface the problem.

Building the edge-index to `RingRole` map by enumerating only non-empty rings, reusing the skip logic already at line 133 above should fix it.

urschrei:
I think we need a test for a hole touching the exterior at two distinct points, which is probably the most common real-world disconnection. This should correctly error  with `InteriorNotSimplyConnected(Exterior, Interior(0))`: `POLYGON((0 0,10 0,10 10,0 10,0 0),(0 5,5 2,10 5,5 8,0 5))`

geo/src/algorithm/validation/polygon.rs
```diff
@@ -27,6 +31,16 @@ pub enum InvalidPolygon {
...
+    /// The polygon interior is not simply connected because rings touch in a way that
+    /// disconnects the interior into multiple regions.
+    ///
+    /// Per OGC Simple Feature Specification (ISO 19125-1), section 6.1.11.1:
+    /// "The interior of every Surface is a connected point set."
+    ///
+    /// This can occur when:
+    /// - Two rings share two or more vertices (creating a "corridor" that cuts through the interior)
+    /// - Rings form a cycle of single-vertex touches that encloses part of the interior
+    InteriorNotSimplyConnected(RingRole, RingRole),
```
urschrei, line 49: This will break downstream matchers because it's public and not `non-exhaustive`. Highly unlikely to happen in practice, but we should prob tag it as BREAKING to avoid any wrath.

urschrei (changes requested):

geo/src/algorithm/validation/polygon.rs
urschrei: NB: the Polygon will be correctly flagged as invalid anyway, so this isn't a huge deal.

michaelkirk:
> I think we need a test for a hole touching the exterior at two distinct points

There was one in the JTS test suite, but I think you're right - it's worth highlighting the happy (invalid? 🙃)  path explicitly in the module. Done!

(ready for re-review)

urschrei (approved):

PR #1524, opened 2026-04-13
prepare for geo 0.33.0 release

michaelkirk:
---

I think we might be waiting for #1502, which seems very close? Anything else on the horizon?

#1522 isn't quite ready, and would be a reasonable non-breaking followup release, so I don't think we need to wait for it.

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -1,6 +1,6 @@
 # Changes
 
-## Unreleased
+## 0.33.0 - 2026-4-15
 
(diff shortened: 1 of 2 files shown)


urschrei:
I'll see if I can get to it today.

michaelkirk:
Rebased now that #1502 was merged. Anything further to wait for?

urschrei (approved):
Let's push this one out the door.

michaelkirk:
published: https://crates.io/crates/geo/0.33.0
tagged: https://github.com/georust/geo/releases/tag/geo-0.33.0

PR #1525, opened 2026-04-20
fix: Add missing AsPrimitive<u32>

leomeinel:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/main/CODE_OF_CONDUCT.md).
- [ ] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
---

This fixes a build failure on wasm-32-unknown-unknown with error: the trait bound `T: AsPrimitive<u32>` is not satisfied.

earcut_triangles and earcut_triangles_iter methods have been auto-generated by clippy's auto fix feature.

### Success

Builds successfully in the same project after changes for platforms (*My [ci/cd](https://github.com/leomeinel/slimy_mist/actions/runs/24661823056) can be referenced for this*):

- [x] `wasm32-unknown-unknown`
- [x] `aarch64-linux-android` + `armv7-linux-androideabi` + `x86_64-linux-android`
- [x] `x86_64-unknown-linux-gnu`
- [x] `aarch64-unknown-linux-gnu`
- [x] `x86_64-pc-windows-msvc`
- [x] `x86_64-apple-darwin` + `aarch64-apple-darwin`
- [x] `aarch64-apple-ios` + `x86_64-apple-ios` + `aarch64-apple-ios-sim`

### Failure

Seems like your ci tests did pass, so this shouldn't be a regression.

Just to document this, 3 tests failed for me:

<details>
<summary>Test log</summary>

```
❯ cargo test --all --all-features --all-targets -q

running 1110 tests
....................................................................................... 87/1110
....................................................................................... 174/1110
....................................................................................... 261/1110
....................................................................................... 348/1110
....................................................................................... 435/1110
....................................................................................... 522/1110
....................................................................................... 609/1110
....................................................................................... 696/1110
...................................... 734/1110
algorithm::line_measures::metric_spaces::geodesic::tests::destination::interpolate_point::points_along_line_without_endpoints --- FAILED
algorithm::line_measures::metric_spaces::geodesic::tests::destination::interpolate_point::points_along_line_with_endpoints --- FAILED
........ 744/1110
algorithm::line_measures::metric_spaces::geodesic::tests::test_non_standard_geoid --- FAILED
..................................................................i..i................. 832/1110
....................................................................................... 919/1110
....................................................................................... 1006/1110
....................................................................................... 1093/1110
.................
failures:

---- algorithm::line_measures::metric_spaces::geodesic::tests::destination::interpolate_point::points_along_line_without_endpoints stdout ----

[...]

PR #1527, opened 2026-04-20
release 0.33.1: fixup earcut semver

michaelkirk:
---
Fixes upstream issue: https://github.com/ciscorn/earcut-rs/issues/27

This is a PR for the fix, and for the 0.33.1 hotfix release. 

Alternatively we could just wait a bit and see how quickly `earcut` addresses the issue. Given that we've already had 2 reports, it seems like it's bothering people, so I wouldn't want to wait too long.

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -1,5 +1,10 @@
 # Changes
 
+## 0.33.1 - 2026-4-20
+
+- Fix build breakage introduced by semver incompatible earcut dependency release
+  - <https://github.com/georust/geo/pull/1527>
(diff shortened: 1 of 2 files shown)


urschrei (approved):

leomeinel (approved):
This looks like a better solution than #1525 for now to me. Thanks for the quick fix!

michaelkirk:
published:  https://crates.io/crates/geo/0.33.1
tagged: https://github.com/georust/geo/releases/tag/geo-0.33.1

PR #1529, opened 2026-04-26
Add linear referencing functionality

urschrei:
---

Currently, there is no way to slice a `LineString` between two distance-along-line values, and `LineLocatePoint` (the existing point-location primitive) is Euclidean-only.

This PR proposes to add two new trait pairs in algorithm/line_measures/, parallel to the existing `InterpolatableLine` / `InterpolateLine`:

- `Substring` / `SubstringableLine`: extract a sub-linestring of a `Line` or `LineString` between two bounds. Both _by_distance and _by_ratio variants. Generic over any metric space implementing `InterpolatePoint + Length` (Euclidean, Haversine, Geodesic, Rhumb, ...). Interior parent vertices are **preserved** between the two endpoints, which is the property that distinguishes substring from interpolate(start) + interpolate(end)
  - Terminology note: both PostGIS and Shapely use "Substring"

We also add a new error enum (`SubstringError`), plus a new "Linear referencing" section in lib.rs that picks up `LineLocatePoint`, `LineStringSegmentize` and the interpolation traits.

How our functionality compares to GEOS `LengthIndexedLine` / `LocationIndexedLine`:

| GEOS                              | This PR                                       |
| --------------------------------- | --------------------------------------------- |
| `project(point) → length` | Already covered by `LineLocatePoint` (existing, Euclidean only, returns ratio*) |
| `extractPoint(length) → point`    | `InterpolateLine::point_at_distance_*` (any metric, already in geo) |
| `extractLine(start, end) → line`  | `Substring::substring_by_distance` (any metric) |
| `getLength`                       | `Length::length` (already in geo)             |
| Length ↔ ratio variants           | `_by_ratio` and `point_at_ratio_*` mirrors    |
| `indexOfAfter(point, minIndex)`   | **out of scope** (closest-projection only; but open to adding it)    |
| Negative indices, GEOS reversal   | **out of scope**: too confusing (this PR clamps; rejects `start > end`) |
| `MultiLineString` support         | **out of scope** (semantics surprising in GEOS itself, we only offer LineString support, like PostGIS does) |

* Note that GEOS returns length; it's easy for the caller to calculate, but we don't do it explicitly.

### DIFF
--- a/geo/src/algorithm/line_measures/substring.rs
+++ b/geo/src/algorithm/line_measures/substring.rs
@@ -1,0 +1,552 @@
+//! Extract a sub-linestring of a [`Line`] or [`LineString`] between two
+//! distance-along-line (or ratio-along-line) values.
+//!
+//! See [`Substring`] for the metric-generic extension trait, and
+//! [`SubstringableLine`] for the line-side trait. Both are wired up so
+//! you can call from either receiver:
+//!
+//! ```ignore
+//! line_string.substring_by_distance(&Euclidean, 5.0, 10.0)
+//! Euclidean.substring_by_distance(&line_string, 5.0, 10.0)
+//! ```
+//!
+//! This is the analogue of PostGIS `ST_Line_Substring` and GEOS
+//! `LengthIndexedLine::extractLine`. Unlike composing `line_locate_point`
+//! and `line_interpolate_point`, this preserves the original interior
+//! vertices of the parent linestring between the two endpoints.
+
+use super::{InterpolatePoint, Length};
+use crate::algorithm::vector_ops::Vector2DOps;
+use geo_types::{Coord, CoordFloat, Line, LineString};
+
+/// Errors returned by substring extraction. See [`SubstringableLine`].
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SubstringError {
+    /// Input line string contains no coordinates.
+    EmptyLineString,
+    /// Input line string contains a single vertex (no segments).
+    SingleVertex,
+    /// Input contains a non-finite coordinate (NaN or infinity).
+    NonFiniteCoordinate,
+    /// Range invalid: `start > end`, or `start`/`end` is not finite.
+    InvalidRange,
+}
+
+impl std::fmt::Display for SubstringError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            SubstringError::EmptyLineString => {
+                write!(f, "LineString contains no coordinates")
+            }
+            SubstringError::SingleVertex => {
+                write!(f, "LineString contains only a single vertex")
+            }
+            SubstringError::NonFiniteCoordinate => {
+                write!(f, "input contains a non-finite coordinate")
+            }
+            SubstringError::InvalidRange => {
+                write!(
+                    f,
+                    "invalid range: start must be <= end and both must be finite"
+                )
+            }
+        }
+    }
+}
+
+impl std::error::Error for SubstringError {}
+
+/// A linear geometry from which a sub-linestring can be extracted between
+/// two distance-along-line (or ratio-along-line) bounds.
+///
+/// Interior vertices of the parent geometry that fall strictly between
+/// the two bounds are preserved in the output.
+pub trait SubstringableLine<F: CoordFloat> {
+    /// The output type of substring extraction. Both `Line` and
+    /// `LineString` impls return `Result<LineString<F>, SubstringError>`
+    /// for a uniform call site.
+    type Output;
+
+    /// Extract a sub-linestring between `start` and `end`, both expressed
+    /// as a distance from the start of the line in the units of
+    /// `metric_space`.
+    ///
+    /// Out-of-range values are clamped to `[0, total_length]`. NaN, or
+    /// `start > end`, returns `Err(SubstringError::InvalidRange)`.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use geo::{Euclidean, SubstringableLine, wkt};
+    ///
+    /// let line_string = wkt!(LINESTRING(0.0 0.0, 10.0 0.0, 20.0 0.0, 30.0 0.0));
+    /// let result = line_string
+    ///     .substring_by_distance(&Euclidean, 5.0, 25.0)
+    ///     .unwrap();
+    /// // Interior parent vertices at x=10 and x=20 are preserved.
+    /// assert_eq!(result, wkt!(LINESTRING(5.0 0.0, 10.0 0.0, 20.0 0.0, 25.0 0.0)));
+    /// ```
+    fn substring_by_distance<M: InterpolatePoint<F> + Length<F>>(
+        &self,
+        metric_space: &M,
+        start: F,
+        end: F,
+    ) -> Self::Output;
+
+    /// Extract a sub-linestring between `start` and `end`, both expressed
+    /// as a fraction of the line's total length in `[0, 1]`.
+    ///
+    /// Out-of-range values are clamped to `[0, 1]`. NaN, or
+    /// `start > end`, returns `Err(SubstringError::InvalidRange)`.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use geo::{Euclidean, SubstringableLine, wkt};
+    ///
+    /// let line_string = wkt!(LINESTRING(0.0 0.0, 10.0 0.0, 20.0 0.0, 30.0 0.0));
+    /// let first_half = line_string
+    ///     .substring_by_ratio(&Euclidean, 0.0, 0.5)
+    ///     .unwrap();
+    /// assert_eq!(first_half, wkt!(LINESTRING(0.0 0.0, 10.0 0.0, 15.0 0.0)));
+    /// ```
+    fn substring_by_ratio<M: InterpolatePoint<F> + Length<F>>(
+        &self,
+        metric_space: &M,
+        start: F,
+        end: F,
+    ) -> Self::Output;
+}
+
+/// Extension trait providing substring extraction on any metric space
+/// implementing [`InterpolatePoint`] and [`Length`].
+///
+/// This is the metric-side counterpart to [`SubstringableLine`]; both
+/// expose the same operations and you can use whichever is more
+/// convenient at the call site.
+pub trait Substring<F: CoordFloat>: InterpolatePoint<F> + Length<F> + Sized {
+    /// Distance-bounded substring; see
+    /// [`SubstringableLine::substring_by_distance`].
+    ///
+    /// # NOTES
+    /// On [LineString], when start == end we deliberately produce a two-coord
+    /// [LineString] with both coords equal, rather than an `Err`. The result
+    /// is a degenerate linestring at a user-specified
+    /// location
+    fn substring_by_distance<L: SubstringableLine<F>>(
+        &self,
+        line: &L,
+        start: F,
+        end: F,
+    ) -> L::Output {
+        line.substring_by_distance(self, start, end)
+    }
+
+    /// Ratio-bounded substring; see
+    /// [`SubstringableLine::substring_by_ratio`].
+    fn substring_by_ratio<L: SubstringableLine<F>>(&self, line: &L, start: F, end: F) -> L::Output {
+        line.substring_by_ratio(self, start, end)
+    }
+}
+
+impl<F: CoordFloat, M: InterpolatePoint<F> + Length<F>> Substring<F> for M {}
+
+impl<F: CoordFloat> SubstringableLine<F> for Line<F> {
+    type Output = Result<LineString<F>, SubstringError>;
+
+    fn substring_by_distance<M: InterpolatePoint<F> + Length<F>>(
+        &self,
+        metric_space: &M,
+        start: F,
+        end: F,
+    ) -> Self::Output {
+        if !self.start.is_finite() || !self.end.is_finite() {
+            return Err(SubstringError::NonFiniteCoordinate);
+        }
+        if start.is_nan() || end.is_nan() || start > end {
+            return Err(SubstringError::InvalidRange);
+        }
+
+        let total = metric_space.length(self);
+        let start = start.max(F::zero()).min(total);
+        let end = end.max(F::zero()).min(total);
+
+        let start_pt =
+            metric_space.point_at_distance_between(self.start_point(), self.end_point(), start);
+        let end_pt =
+            metric_space.point_at_distance_between(self.start_point(), self.end_point(), end);
+
+        Ok(LineString::from(vec![start_pt.0, end_pt.0]))
+    }
+
+    fn substring_by_ratio<M: InterpolatePoint<F> + Length<F>>(
+        &self,
+        metric_space: &M,
+        start: F,
+        end: F,
+    ) -> Self::Output {
+        let total = metric_space.length(self);
+        self.substring_by_distance(metric_space, start * total, end * total)
+    }
+}
+
+impl<F: CoordFloat> SubstringableLine<F> for LineString<F> {
+    type Output = Result<LineString<F>, SubstringError>;
+
+    fn substring_by_distance<M: InterpolatePoint<F> + Length<F>>(
+        &self,
+        metric_space: &M,
+        start: F,
+        end: F,
+    ) -> Self::Output {
+        if self.0.is_empty() {
+            return Err(SubstringError::EmptyLineString);
+        }
+        if self.0.len() == 1 {
+            return Err(SubstringError::SingleVertex);
+        }
+        for c in &self.0 {
+            if !c.x.is_finite() || !c.y.is_finite() {
+                return Err(SubstringError::NonFiniteCoordinate);
+            }
+        }
+        if start.is_nan() || end.is_nan() || start > end {
+            return Err(SubstringError::InvalidRange);
+        }
+
+        let total = metric_space.length(self);
+        let start = start.max(F::zero()).min(total);
+        let end = end.max(F::zero()).min(total);
+
+        // NOTE: when start == end we deliberately produce a two-coord
+        // LineString with both coords equal, rather than an Err. The result
+        // is a degenerate linestring at a user-specified
+        // location, not a default zero. The rationale is that e.g. callers iterating duplicate
+        // mileposts shouldn't have to special-case this. If something like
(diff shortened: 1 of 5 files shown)


dabreegster:
Related to #986. Really really excited to try this out soon! I've been using a copy of the code from #1050 in my own projects.

urschrei:
> Related to #986. Really really excited to try this out soon! I've been using a copy of the code from #1050 in my own projects.

Please do! There isn't much new code; it's _mostly_ wiring together existing functionality and then tests. I had completely forgotten about #1050, so if there's missing API surface you want to see here lmk

geo/src/algorithm/line_measures/locate.rs
```diff
@@ -0,0 +1,425 @@
...
+//!
+//! Currently only `Locate for Euclidean` ships. The trait shape is
+//! forward-compatible for non-Euclidean impls (Haversine via
+//! cross-track / along-track machinery; Geodesic harder) — adding one
+//! is purely additive.
+//!
+//! This is the analogue of PostGIS `ST_Line_Locate_Point` and GEOS
+//! `LengthIndexedLine::project`. On a self-intersecting linestring,
+//! the **closest** projection is returned — there is no
+//! `indexOfAfter`-style minimum-index variant in this PR.
```
michaelkirk:
```suggestion
/// returned. There is no `indexOfAfter`-style minimum-index variant.
```

What is `indexOfAfter`? It feels like since I don't know what it is, I don't need to worry about it because it doesn't exist, but it's a bit tantalizing.

Also, we shouldn't mention "this PR" in a module comment - it's weird point in time reference.

geo/src/lib.rs
```diff
@@ -98,16 +98,25 @@
...
+//! ## Linear referencing
+//!
+//! Locate points on a line by their distance along it, and slice a line
+//! between two such positions. Equivalent to PostGIS `ST_Line_Substring`
+//! / `ST_Line_Locate_Point` and GEOS `LengthIndexedLine`.
+//!
+//! - **[`Substring`]**: Extract a sub-linestring between two distance- or ratio-along-line bounds (any metric space).
+//! - **[`Locate`]**: Locate a point along a line, returning distance or ratio at the closest projection (Euclidean only).
+//! - **[`InterpolateLine`]**: Interpolate a point along a line by distance or ratio (any metric space).
+//! - **[`LineLocatePoint`]**: Closest point on a line as a fraction of length (Euclidean only; superseded by [`Locate`]).
```
michaelkirk: > `LineLocatePoint` [...] Euclidean only; superseded by [`Locate`]

Locate is also Euclidean only - can you explain what we're gaining with the new trait introduction? The potential to add other implementations in the future, is that the goal?

Is there any reason `LineLocatePoint` should not be deprecated? It seems confusing to have two same but different ways of achieving the same thing.

This is a large PR, and Locate / LocatableLine seem completely separate from `Substring / SubstringableLine`. Can you separate that functionality to make this PR more focused or is there a reason they should be together?

geo/src/lib.rs
```diff
@@ -182,8 +191,6 @@
 //! - **[`Centroid`]**: Calculate the centroid of a geometry
 //! - **[`ChaikinSmoothing`]**: Smoothen `LineString`, `Polygon`, `MultiLineString` and `MultiPolygon` using Chaikin's algorithm
 //! - **[`proj`]**: Project geometries with the `proj` crate (requires the `proj` feature)
-//! - **[`LineStringSegmentize`]**: Segment a LineString into `n` segments
-//! - **[`LineStringSegmentizeHaversine`]**: Segment a LineString using Haversine distance
```
michaelkirk, line 188: Two more traits that seem ripe for unifying (some day)!

michaelkirk (commented):
I have a couple preliminary questions before I really dig into the review.

geo/src/algorithm/line_measures/substring.rs
```diff
@@ -0,0 +1,545 @@
...
+    ///     .unwrap();
+    /// assert_eq!(first_half, wkt!(LINESTRING(0.0 0.0, 10.0 0.0, 15.0 0.0)));
+    /// ```
+    fn substring_by_ratio<M: InterpolatePoint<F> + Length<F>>(
+        &self,
+        metric_space: &M,
+        start: F,
+        end: F,
+    ) -> Self::Output;
+}
```
dabreegster, line 122: #1050 included `line_split_many`, which I've found very convenient. Sometimes you might build up multiple fractions or distances where a split should occur, sort them, and try to split by all these at once. Some of the results may be empty (ie, if you ask for splits at `[0.2, 0.3, 0.3, 0.6]`).

A caller could build this on top of what you've defined here, but with presumably worse performance of redoing work for each call. Is it worth trying to figure out this case now, or save it for later? Either way, the APIs as you've defined them here look fine to me

geo/src/algorithm/line_measures/substring.rs
```diff
@@ -0,0 +1,545 @@
...
+        }
+        if start.is_nan() || end.is_nan() || start > end {
+            return Err(SubstringError::InvalidRange);
+        }
+
+        let total = metric_space.length(self);
+        let start = start.max(F::zero()).min(total);
+        let end = end.max(F::zero()).min(total);
+
+        // NOTE: when start == end we deliberately produce a two-coord
```
dabreegster, line 224: Worth calling this out too in the API docstring?

And this decision also matters for the `multi` API -- it'd return either `Vec<LineString>` with some degnerate ones, or `Vec<Option<LineString>>`

geo/src/algorithm/line_measures/locate.rs
urschrei: I'll drop the PR bit. `indexOfAfter` is GEOS jargon for "first projection _past_ a given index". I chose _before_ because it matches the existing `LineLocatePoint` semantics, and it's what PostGIS does.

geo/src/lib.rs
urschrei: > Locate is also Euclidean only - can you explain what we're gaining with the new trait introduction? The potential to add other implementations in the future, is that the goal?

Yep, certainly for the Haversine metric space, since we can use cross-track distance for the calculation. I don't think Geodesic is easily doable, and Rhumb is kinda niche.

> Is there any reason LineLocatePoint should not be deprecated?

Not really; we held off previously because there was no replacement. `Locate` is the replacement, so I'm happy to deprecate it in this PR.

> Can you separate that functionality to make this PR more focused?

They're separable in principle, but they form a single feature set: Locate maps `point → position`, `Substring` maps `two positions → sub-line`. Together that's the linear-referencing primitive set. PostGIS bundles them under the `ST_Line_*` family; GEOS bundles them as `LengthIndexedLine` / `LocationIndexedLine`. `Locate`, then `Substring` can be separate PRs for review purposes if you'd prefer.

geo/src/algorithm/line_measures/substring.rs
urschrei: Yep, done. A note one the `Multi` case: I went with PostGIS here, which does not allow linear reference ops on `MultiLineString` because the semantics are unclear, and those in GEOS are (imo) super-confusing (component order leaks into result, extracted lines might not exist in the original, projecting a point to equidistant coords is ambiguous)

geo/src/algorithm/line_measures/substring.rs
urschrei: Oh yeah that seems useful. Let's see how we fare with this PR, with the intent to add this as follow-up work?

geo/src/lib.rs
michaelkirk: Since we already have Euclidean `point → position` with `LineLocatePoint`, I'd prefer you make this PR smaller by excluding the new redundant `Locate` trait. 

Arguably it makes more sense to introduce a new `point → position` trait when we have more than just the Euclidean implementation ready to go. In the meanwhile, unless I'm misunderstanding, removing it makes this PR easier to review with no loss of functionality.

michaelkirk: >>  Locate is also Euclidean only - can you explain what we're gaining with the new trait introduction? The potential to add other implementations in the future, is that the goal?
>
> Yep, certainly for the Haversine metric space, since we can use cross-track distance for the calculation. I don't think Geodesic is easily doable, and Rhumb is kinda niche.

Bear in mind that cross track distance is the distance to a great circle - not a haversine line **segment**. Maybe that doesn't significantly complicate things in this case - I haven't thought through the algorithm, but I recall I thought the haversine linestring distance calculation would be trivial given the existing crosstrack impl, but it was more than trivial (for me anyway).

urschrei: OK done.

geo/src/algorithm/line_measures/substring.rs
dabreegster: Fine by me. Thanks again for this work! I'll cutover to it in a few projects after it merges

geo/src/algorithm/line_measures/substring.rs
```diff
@@ -0,0 +1,551 @@
...
+        end: F,
+    ) -> Self::Output {
+        if !self.start.x.is_finite()
+            || !self.start.y.is_finite()
+            || !self.end.x.is_finite()
+            || !self.end.y.is_finite()
+        {
+            return Err(SubstringError::NonFiniteCoordinate);
+        }
+        if start.is_nan() || end.is_nan() || start > end {
```
michaelkirk, line 169: minor nit: These `is_nan` checks are redundant with the [`is_finite`](https://docs.rs/num/latest/num/trait.Float.html#tymethod.is_finite) checks above.

geo/src/algorithm/line_measures/substring.rs
```diff
@@ -0,0 +1,551 @@
...
+impl<F: CoordFloat> SubstringableLine<F> for Line<F> {
+    type Output = Result<LineString<F>, SubstringError>;
+
+    fn substring_by_distance<M: InterpolatePoint<F> + Length<F>>(
+        &self,
+        metric_space: &M,
+        start: F,
+        end: F,
+    ) -> Self::Output {
+        if !self.start.x.is_finite()
```
michaelkirk: alternatively we have these in `VectorOps`: `!self.start.is_finite() || !self.end.is_finite()`

geo/src/algorithm/line_measures/substring.rs
```diff
@@ -0,0 +1,551 @@
...
+    /// [`SubstringableLine::substring_by_ratio`].
+    fn substring_by_ratio<L: SubstringableLine<F>>(&self, line: &L, start: F, end: F) -> L::Output {
+        line.substring_by_ratio(self, start, end)
+    }
+}
+
+impl<F: CoordFloat, M: InterpolatePoint<F> + Length<F>> Substring<F> for M {}
+
+impl<F: CoordFloat> SubstringableLine<F> for Line<F> {
+    type Output = Result<LineString<F>, SubstringError>;
```
michaelkirk: Seems like this should return a `Line`?

geo/src/algorithm/line_measures/substring.rs
```diff
@@ -0,0 +1,551 @@
...
+        end: F,
+    ) -> Self::Output {
+        if self.0.is_empty() {
+            return Err(SubstringError::EmptyLineString);
+        }
+        if self.0.len() == 1 {
+            return Err(SubstringError::SingleVertex);
+        }
+        for c in &self.0 {
+            if !c.x.is_finite() || !c.y.is_finite() {
```
michaelkirk, line 212:
```suggestion
            if !c.is_finite() {
```
(same as my comments in `Line`)

geo/src/algorithm/line_measures/substring.rs
```diff
@@ -0,0 +1,551 @@
...
+                return Err(SubstringError::NonFiniteCoordinate);
+            }
+        }
+        if start.is_nan() || end.is_nan() || start > end {
```
michaelkirk, line 216:
```suggestion
        if start > end {
```

geo/src/algorithm/line_measures/substring.rs
```diff
@@ -0,0 +1,551 @@
...
+        // change is localised to this function plus the
+        // `substring_linestring_zero_length_range` test.
+        let mut coords: Vec<Coord<F>> = Vec::with_capacity(self.0.len() + 1);
+        let mut cum = F::zero();
+        let mut emitted_start = false;
+        // When the bound lands exactly on a parent vertex, prefer
+        // the original coord over re-interpolating: it's exact, and
+        // it avoids floating-point drift in non-Euclidean metrics
+        // where `point_at_distance_between(a, b, len(a, b)) != b`.
+        let snap = |segment: &Line<F>, seg_len: F, local: F| -> Coord<F> {
```
michaelkirk:
```suggestion
        let clamp = |segment: &Line<F>, seg_len: F, local: F| -> Coord<F> {
```

Take it or leave it naming suggestion - like [`f64::clamp`](https://doc.rust-lang.org/std/primitive.f64.html#method.clamp).

It took me a dozen seconds to figure out what `snap` was. At first I assumed it was some kind of "grid snapping" thing like we do with our DelaunayTriangulation.

I think the name `local` could be more helpful too, maybe `distance_along` or something similar.

geo/src/algorithm/line_measures/substring.rs
```diff
@@ -0,0 +1,551 @@
...
+        // This test fails to compile if the re-exports in
+        // algorithm/mod.rs aren't wired correctly.
+        use crate::{Substring, SubstringError, SubstringableLine};
+        fn _t<S: Substring<f64>, L: SubstringableLine<f64>>(_s: &S, _l: &L) {}
+        let _e: SubstringError = SubstringError::EmptyLineString;
+    }
+
+    #[test]
+    fn substring_line_full_range() {
+        let line = Line::new(Point::new(0.0, 0.0), Point::new(10.0, 0.0));
```
michaelkirk:
```suggestion
        let line = wkt!(LINE(0.0 0.0,10.0 0.0));
```

You can spell it this way, but it's non-standard (won't work in JTS), so I won't insist! Just FYI.

geo/src/algorithm/line_measures/substring.rs
```diff
@@ -0,0 +1,551 @@
...
+            } else if local >= seg_len {
+                segment.end
+            } else {
+                metric_space
+                    .point_at_distance_between(segment.start_point(), segment.end_point(), local)
+                    .0
+            }
+        };
+
+        for segment in self.lines() {
```
michaelkirk, line 255: I was convinced this loop could be simplified, and tried to do so for a while, and eventually ended up slowly rebuilding a worse version than what you have here. Thanks for all the tests!

michaelkirk (approved):
The only feedback of substance is that I think slicing a Line should return a `Line`, not a `LineString`. If you'd like to push back we can discuss it more. 

Other than that, feel free to merge at will because everything else is just quibbling/code golf which I don't feel strongly about.

geo/src/algorithm/line_measures/substring.rs
urschrei: My thinking on uniformly returning LineString:

1. It's what all the other libraries do (even though they don't have `Line`).
2. I prefer to return OGC geometries from operations. This may not be a totally rational preference

I'm not wedded to my position here. WDYT?

geo/src/algorithm/line_measures/substring.rs
urschrei: `start > end` doesn't catch `NaN`. IEEE 754 says **any** comparison involving `NaN` is false, so `f64::NAN > 5.0` is `false`, causing validation to fall through. Then a couple of lines later:

`let start = start.max(F::zero()).min(total);`

`f64::max` and `f64::min` propagate the non-NaN operand when one side is NaN, so `NaN.max(0.0).min(10.0)` evaluates to `0.0`. The NaN silently disappears, and the `substring_line_invalid_range_nan_bound` test returns `Ok(LINESTRING(0 0, 5 0))` instead of the expected Err(InvalidRange).

geo/src/algorithm/line_measures/substring.rs
michaelkirk: I'd still prefer returning a Line pretty much whenever semantically possible, but I wouldn't block merging on this!

> I prefer to return OGC geometries from operations. This may not be a totally rational preference

Can you clarify what you mean by "operation"? Are all the methods we have currently which return a Line not "operations"? Or are they wrong for returning a `Line`?

---

> It's what all the other libraries do (even though they don't have Line)

I think the fact that they don't have `Line` is significant here. For me it seems like the same argument applies everywhere we return a `Line` — it *could* be a `LineString`, but what's gained is more than what's lost.

Another way to think about it: 

If we return a `Line`, but the person wants a `LineString`, converting a Line to a LineString is easy, requires no unwraps, and involves only the necessary amount of heap allocation.

However, if we return a `LineString` but the person actually wants a `Line`, they need to:
1. read the documentation to remind themselves there will *always* be 2 points, even with degenerate (0-D lines)
2. use `unwrap`
3. swallow the unnecessary heap allocation.

If you feel strongly that returning `LineString` uniformly is what people are likely to want, then feel free to leave it as is, but consider getting rid of the `Output` associated type, since it's always a LineString anyway. 

All this spilled ink is kind of funny though, because I'd wager that very few people will use Line's in this way anyway. Maybe because they aren't an OGC type, I don't think people reach for them much - at least I don't.

michaelkirk (approved):
I have one final plea, but happy to merge it as is.

geo/src/algorithm/line_measures/substring.rs
michaelkirk: My point was that `is_finite` already includes a nan check. Take a look https://docs.rs/num/latest/num/trait.Float.html#tymethod.is_finite

geo/src/algorithm/line_measures/substring.rs
michaelkirk: Oh whoops. I got this confused with the similar but different code in the `Line` impl.

In that case though, do you also want to check for infinity?

```rust
if !start.is_finite() || !end.is_finite() || start > end {
...
}
```

geo/src/algorithm/line_measures/substring.rs
michaelkirk: blargghhh.... I got `start` confused with `self.start`. I think you're right to leave this check, but maybe check for `!start.is_finite()` instead of only checking for `start.is_nan()` here?

michaelkirk:
If this is hung up on the Line vs LineString return type - I'm totally fine to merge as is.

urschrei:
I think I just wandered away from this one and then forgot about it. "Input" `Line`s now return `Line`.

PR #1533, opened 2026-04-28
chore: Unpin earcut version

ciscorn:
Unpins the earcut version that was pinned (#1527) due to my own semver violation (sorry).

The `AsPrimitive<u32>` trait bound was reverted upstream, so we can take advantage of the performance improvements without any breaking changes.

Benchmark improvements from earcut 0.4.5 -> 0.4.9:

| Bench                            |   0.4.5 |   0.4.9 |       Δ |
| -------------------------------- | ------: | ------: | ------: |
| TriangulateEarcut - small polys  | 5.31 ms | 3.88 ms | -26.9 % |
| TriangulateEarcut - large_poly   | 1.58 ms | 1.19 ms | -23.8 % |

---

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -1,5 +1,10 @@
 # Changes
 
+## Unreleased
+
+- Unpin the `earcut` dependency now that the upstream semver violation has been reverted. 
+  - <https://github.com/georust/geo/pull/1533>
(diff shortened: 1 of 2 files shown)


michaelkirk (approved):
Thanks!

PR #1534, opened 2026-04-28
Add reusable Earcutter instance

ciscorn:
As noted in https://github.com/georust/geo/pull/1508#issue-4002434807, exposing a reusable `Earcutter` instance (and the `earcut_triangulation_ref` trait method) avoids memory allocations across multiple triangulations.

This is especially effective when processing a bunch of simple polygons (e.g. OSM buildings), where per-call allocation cost would otherwise be significant.

## Benchmarks

| Bench       | per-call malloc | reuse scratch |       Δ |
| ----------- | ---------------------: | -----------------: | ------: |
| tiny polys  |                ~245 µs |            ~130 µs | -47 %   |
| small polys |               ~3.96 ms |           ~3.76 ms |  -5 %   |
| large_poly  |               ~1.22 ms |           ~1.15 ms |  -5 %   |

---

### DIFF
--- a/geo-benches/src/triangulate.rs
+++ b/geo-benches/src/triangulate.rs
@@ -2,8 +2,30 @@ use criterion::{criterion_group, criterion_main};
 use geo::algorithm::{TriangulateDelaunay, TriangulateDelaunayUnconstrained, TriangulateEarcut};
 use geo::geometry::Polygon;
 use geo::triangulate_delaunay::DelaunayTriangulationConfig;
+use geo::triangulate_earcut::Earcutter;
 
 fn criterion_benchmark(c: &mut criterion::Criterion) {
+    c.bench_function("TriangulateEarcut - tiny polys", |bencher| {
+        let multi_poly = geo_test_fixtures::nl_plots_wgs84::<f64>();
+        bencher.iter(|| {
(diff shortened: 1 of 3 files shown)


michaelkirk:
I had a few suggestions @ciscorn. I hope you won't be offended, but I implemented them here since I thought it would be faster than explaining:
 https://github.com/georust/geo/compare/main...mkirk%2Fearcuter

There's no change in the functionality, but it improves the existing naming and makes the new API's more discoverable.

How would you feel about pulling this into your branch?

ciscorn:
@michaelkirk Thanks! No worries, I often explain things in code too.

I was unsure about adding it to the trait in the same style, but I agree this is better for discoverability.

I also considered `u32` indices since they're sufficient and GPU-friendly, but I agree `usize` is better for consistency with existing types.

I pulled it into this branch.

michaelkirk (approved):
LGTM - I'll leave it open for another day or two in case any one else wants to weigh in.

(Sometimes I'll merge things right away if I think they're very unlikely to be controversial like #1533)

michaelkirk:
Sorry for the delay! I got swept up in some other stuff last week.

PR #1535, opened 2026-05-22
clamp in haversine distance calculation to prevent NaN

qkniep:
---

Clamp the intermediate value `a` in the haversine distance calculation to prevent a NaN value for inputs that are nearly antipodal. The added test shows an example of an input that returns NaN before this change and no longer does after.

This (very deliberately) changes the output of `Haversine.distance()` for some inputs.

### DIFF
--- a/geo/src/algorithm/line_measures/metric_spaces/haversine.rs
+++ b/geo/src/algorithm/line_measures/metric_spaces/haversine.rs
@@ -271,7 +271,8 @@ impl<F: CoordFloat + FromPrimitive> Distance<F, Point<F>, Point<F>> for Haversin
         let delta_lambda = (destination.x() - origin.x()).to_radians();
         let a = (delta_theta / two).sin().powi(2)
             + theta1.cos() * theta2.cos() * (delta_lambda / two).sin().powi(2);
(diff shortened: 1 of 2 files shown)


geo/src/algorithm/line_measures/metric_spaces/haversine.rs
```diff
@@ -572,7 +573,17 @@ mod tests {
...
             );
         }
+
+        #[test]
+        fn near_antipodal_points() {
+            let d: f64 = Haversine.distance(
+                Point::new(-121.96981239217565, 65.09034271683089),
+                Point::new(58.03018760782425, -65.0903427168311),
+            );
+            assert_relative_eq!(d, 20015114.442035925);
```
michaelkirk, line 583: I pushed up a more specific test rather than checking for nan.

michaelkirk (approved):
Thanks! Sorry for the slow review. 

I'll wait at least another day before merging incase someone else wants to weigh in.

geo/src/algorithm/line_measures/metric_spaces/haversine.rs
qkniep: yes, that's better

Issue #1537, opened 2026-05-28
Add _idx variants for ConvexHull and SimplifyVwPreserve to enable dimension passthrough downstream


willcohen:
I'm looking at incorporating geoarrow-rs geospatial algorithms as a performance optimization compared to vanilla JTS / a webassembly build of JTS when they're available, which, as I understand it, wraps the core operations from georust/geo.

I'm noticing that JTS can generally operate on coordinate sequence/buffers with extra dimensions (XYZ, XYM, XYZM), and can always pass them through. 

`Simplify` and `SimplifyVw` have their `SimplifyIdx` and `SimplifyVwIdx` variants, so geoarrow-rs can just nab the relevant indices from those and pull the extra dimensions through in their output. 

`SimplifyVwPreserve` and `ConvexHull`, however, don't have `idx` variants, and so geoarrow-rs has no way to route the Z/M values through.

Would a PR to add this be welcomed?

kylebarron:
Not sure but I thought maybe SimplifyVwPreserve would occasionally have to add new points to ensure the output is valid? Like perhaps it wouldn't always be able to define just indices into the existing coordinates?

willcohen:
I'm happy to be wrong but I'm not seeing any addition of new vertices, only the dropping of them (via https://github.com/georust/geo/blob/8126f0a5cc41aee51bd08a953af4b0b29db7e611/geo/src/algorithm/simplify_vw.rs#L366).

urschrei:
Just to confirm the preserve variant never adds coords: it skips removal to ensure validity.

michaelkirk:
Done in https://github.com/georust/geo/pull/1538

willcohen:
Thank you!

PR #1538, opened 2026-05-28
Add index variants to SimplifyVwPreserve and ConvexHull

willcohen:
---
This is a followup to #1537 @kylebarron.

I want to flag that Rust is NOT my primary programming language -- I've dabbled in C and can kind of read what's happening here. I'm much more comfortable in JTS and Clojure/JVM stuff -- geospatial algorithms are not new to me, but georust and rust are. Thus, this is heavily LLM assisted, but I have tried my best to use what I'd consider what appear to be reasonable programming best practices in thinking about tests and trying to follow my basic understanding of project conventions. I've also been using this branch downstream in other uses and it solves the problems flagged by the initial issue.

I fully expect that I have messed up many non-idiomatic Rust-like conventions, and apologize in advance about them! I hope to use this project much more intensively on JS and JVM soon, and it would be my gateway into this language and ecosystem.

@kylebarron the first commit is a test based on your question -- it simply surfaces the lack of adding new points, and if it's superfluous I'm happy to remove.

### DIFF
--- a/geo/src/algorithm/simplify.rs
+++ b/geo/src/algorithm/simplify.rs
@@ -187,7 +187,7 @@ where
 /// An `epsilon` less than or equal to zero will return an unaltered version of the geometry.
 pub trait Simplify<T, Epsilon = T> {
     /// The index-output type of [`Simplify::simplify_idx`], which varies by geometry.
-    type Output;
+    type IndexOutput;
 
     /// Returns the simplified representation of a geometry, using the [Ramer–Douglas–Peucker](https://en.wikipedia.org/wiki/Ramer–Douglas–Peucker_algorithm) algorithm
     ///
@@ -221,7 +221,7 @@ pub trait Simplify<T, Epsilon = T> {
         T: GeoFloat;
 
     /// Returns the indices of the points retained by [`Simplify::simplify`],
-    /// relative to the input geometry. [`Self::Output`] is `Vec<usize>` for
+    /// relative to the input geometry. [`Self::IndexOutput`] is `Vec<usize>` for
     /// `LineString`, `Vec<Vec<usize>>` for `MultiLineString`, [`PolygonIndices`]
     /// for `Polygon`, and `Vec<PolygonIndices>` for `MultiPolygon`.
     ///
@@ -243,7 +243,7 @@ pub trait Simplify<T, Epsilon = T> {
     ///
     /// assert_eq!(indices, vec![0_usize, 1, 2, 4]);
     /// ```
-    fn simplify_idx(&self, epsilon: T) -> Self::Output
+    fn simplify_idx(&self, epsilon: T) -> Self::IndexOutput
     where
         T: GeoFloat;
 }
@@ -302,7 +302,7 @@ impl<T> Simplify<T> for LineString<T>
 where
     T: GeoFloat,
 {
-    type Output = Vec<usize>;
+    type IndexOutput = Vec<usize>;
 
     fn simplify(&self, epsilon: T) -> Self {
         LineString::from(rdp::<_, _, LINE_STRING_INITIAL_MIN>(
@@ -330,7 +330,7 @@ impl<T> Simplify<T> for MultiLineString<T>
 where
     T: GeoFloat,
 {
-    type Output = Vec<Vec<usize>>;
+    type IndexOutput = Vec<Vec<usize>>;
 
     fn simplify(&self, epsilon: T) -> Self {
         MultiLineString::new(self.iter().map(|l| l.simplify(epsilon)).collect())
@@ -347,7 +347,7 @@ impl<T> Simplify<T> for Polygon<T>
 where
     T: GeoFloat,
 {
-    type Output = PolygonIndices;
+    type IndexOutput = PolygonIndices;
 
     fn simplify(&self, epsilon: T) -> Self {
         Polygon::new(
@@ -380,7 +380,7 @@ impl<T> Simplify<T> for MultiPolygon<T>
 where
     T: GeoFloat,
 {
-    type Output = Vec<PolygonIndices>;
+    type IndexOutput = Vec<PolygonIndices>;
 
     fn simplify(&self, epsilon: T) -> Self {
         MultiPolygon::new(self.iter().map(|p| p.simplify(epsilon)).collect())
--- a/geo/src/algorithm/simplify.rs
+++ b/geo/src/algorithm/simplify.rs
@@ -142,6 +142,33 @@ where
     vec![first, last]
 }
 
+/// Per-ring retained-index output for the index-returning simplification
+/// methods (`simplify_idx`, `simplify_vw_idx`, `simplify_vw_preserve_idx`) when
+/// applied to a `Polygon`. The exterior and interior ring indices are kept
+/// distinct rather than flattened into a single `Vec<usize>`; each index is
+/// relative to its own ring's coordinate sequence.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PolygonIndices {
+    pub exterior: Vec<usize>,
+    pub interiors: Vec<Vec<usize>>,
+}
+
+// Indices are relative to this ring's own coord sequence, not polygon-global.
+fn rdp_indices<T, const INITIAL_MIN: usize>(ring: &LineString<T>, epsilon: T) -> Vec<usize>
+where
+    T: GeoFloat,
+{
+    calculate_rdp_indices::<_, INITIAL_MIN>(
+        &ring
+            .0
+            .iter()
+            .enumerate()
+            .map(|(index, &coord)| RdpIndex { index, coord })
+            .collect::<Vec<RdpIndex<T>>>(),
+        epsilon,
+    )
+}
+
 /// Simplifies a geometry.
 ///
 /// The [Ramer–Douglas–Peucker
@@ -159,6 +186,9 @@ where
 ///
 /// An `epsilon` less than or equal to zero will return an unaltered version of the geometry.
 pub trait Simplify<T, Epsilon = T> {
+    /// The index-output type of [`Simplify::simplify_idx`], which varies by geometry.
+    type Output;
+
     /// Returns the simplified representation of a geometry, using the [Ramer–Douglas–Peucker](https://en.wikipedia.org/wiki/Ramer–Douglas–Peucker_algorithm) algorithm
     ///
     /// # Examples
@@ -189,6 +219,33 @@ pub trait Simplify<T, Epsilon = T> {
     fn simplify(&self, epsilon: T) -> Self
     where
         T: GeoFloat;
+
+    /// Returns the indices of the points retained by [`Simplify::simplify`],
+    /// relative to the input geometry. [`Self::Output`] is `Vec<usize>` for
+    /// `LineString`, `Vec<Vec<usize>>` for `MultiLineString`, [`PolygonIndices`]
+    /// for `Polygon`, and `Vec<PolygonIndices>` for `MultiPolygon`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use geo::Simplify;
+    /// use geo::line_string;
+    ///
+    /// let line_string = line_string![
+    ///     (x: 0.0, y: 0.0),
+    ///     (x: 5.0, y: 4.0),
+    ///     (x: 11.0, y: 5.5),
+    ///     (x: 17.3, y: 3.2),
+    ///     (x: 27.8, y: 0.1),
+    /// ];
+    ///
+    /// let indices = line_string.simplify_idx(1.0);
+    ///
+    /// assert_eq!(indices, vec![0_usize, 1, 2, 4]);
+    /// ```
+    fn simplify_idx(&self, epsilon: T) -> Self::Output
+    where
+        T: GeoFloat;
 }
 
 /// Simplifies a geometry, returning the retained _indices_ of the input.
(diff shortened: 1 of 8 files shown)


willcohen:
Sorry @urschrei -- force pushed after `cargo fmt`.

willcohen:
No substantive changes, just rebasing to main.

kylebarron (commented):
From a cursory look this looks good, but I don't know this part of the codebase well

willcohen:
I can also prep a WIP PR for geoarrow-rs if it helps to exercise how this would/could get used downstream!

kylebarron:
In terms of merging here, I don't _think_ it's relevant to other geo maintainers how it would be used downstream in geoarrow-rs. It does seem like it would be useful

geo/src/algorithm/simplify_vw.rs
```diff
@@ -259,20 +259,38 @@ where
...
+fn vwp_wrapper<T, const INITIAL_MIN: usize, const MIN_POINTS: usize>(
+    exterior: &LineString<T>,
+    interiors: Option<&[LineString<T>]>,
+    epsilon: T,
+) -> Vec<Vec<Coord<T>>>
+where
+    T: GeoFloat + RTreeNum,
+{
+    let indices_per_ring =
+        vwp_wrapper_indices::<T, INITIAL_MIN, MIN_POINTS>(exterior, interiors, epsilon);
```
michaelkirk, line 285: Previously:

`-> Vec<Vec<Coord>>`

Now:

`-> Vec<Vec<usize>> -> Vec<Vec<Coord>>`

There's another round of allocation now. 

Are there benches for the original method? Can you verify if there is any measurable perf difference with the new indirection? (I'm ok with a tiny regression vs. complicating the code, I just wanted you to verify if it was significant)

michaelkirk (changes requested):
Does it complicate things to add these new methods to the existing traits rather than introducing new traits? My experience has been that it's easier to find methods when they're group together rather than 1 method per trait.

willcohen:
Just to make sure I've got this right -- you're asking about hanging the index methods off the existing `SimplifyVwPreserve` and `ConvexHull` traits instead of separate `*Idx` traits? I'd copied the separate-trait setup from the existing `Simplify/SimplifyIdx` and `SimplifyVw/SimplifyVwIdx` pairs, but I don't have a strong reason of my own to keep them split.

I don't know georust well enough to have a clear reasoning across the library, but for consistency it seems like you'd then want all the `*Idx` traits consolidated, not just these ones from the PR, which is a bigger change. Is that something I should look into?

I benched simplify_vw_preserve (and convex_hull, since the qhull refactor routes the public path through the new generic kernel) on main vs this branch, and it looks clean / within noise threshold:

```
simplify vwp f32        time:   [1.3645 ms 1.4029 ms 1.4685 ms]
  3 (3.00%) high mild
  3 (3.00%) high severe

simplify vwp f64        time:   [1.4975 ms 1.5352 ms 1.6101 ms]
Found 22 outliers among 100 measurements (22.00%)
  6 (6.00%) high mild
  16 (16.00%) high severe

convex hull f32         time:   [160.55 µs 166.38 µs 174.13 µs]
  3 (3.00%) high mild
  4 (4.00%) high severe

convex hull f64         time:   [159.39 µs 162.28 µs 166.77 µs]
  3 (3.00%) high mild
  2 (2.00%) high severe
```

```
simplify vwp f32        time:   [1.3863 ms 1.4329 ms 1.4941 ms]
                        change: [-0.3394% +2.6102% +5.5463%] (p = 0.09 > 0.05)
  5 (5.00%) high mild
  5 (5.00%) high severe

simplify vwp f64        time:   [1.5303 ms 1.5510 ms 1.5747 ms]
                        change: [-3.1479% +0.0120% +2.7981%] (p = 0.99 > 0.05)
  12 (12.00%) high severe

convex hull f32         time:   [160.85 µs 165.21 µs 170.95 µs]
                        change: [-6.4660% -1.8523% +1.9145%] (p = 0.45 > 0.05)
Found 23 outliers among 100 measurements (23.00%)
  19 (19.00%) high mild
  4 (4.00%) high severe

convex hull f64         time:   [159.78 µs 163.64 µs 171.28 µs]
                        change: [-9.9256% -5.0044% -0.2686%] (p = 0.04 < 0.05)
  3 (3.00%) high mild
  8 (8.00%) high severe
```

michaelkirk:
> Just to make sure I've got this right -- you're asking about hanging the index methods off the existing SimplifyVwPreserve and ConvexHull traits instead of separate *Idx traits?

Yes, correct.


> I'd copied the separate-trait setup from the existing Simplify/SimplifyIdx and SimplifyVw/SimplifyVwIdx pairs, but I don't have a strong reason of my own to keep them split.
>
> I don't know georust well enough to have a clear reasoning across the library, but for consistency it seems like you'd then want all the *Idx traits consolidated, not just these ones from the PR, which is a bigger change. Is that something I should look into?

Great point - thanks for pushing back. Rather than continuing to follow this one function per trait just because "it's the way we've always done things", I think it would be better to have these other two traits consolidated too. We should `#deprecate` the separate Idx traits, and point the user at their "new" home to ease migration. That could be done in a followup PR or you can do it here if you'd like.

willcohen:
Additional commit added. 

Again -- I am a Rust newbie, and though I understand interfaces/protocols/etc in other languages, traits/impl are a different flavor of that concept, so I am very much leaning on an LLM. Through this muddy lens, as best as I understand it, this does streamline out the `*Idx` traits.

Instead, we now have
- `Simplify::simplify_idx`, `SimplifyVw::simplify_vw_idx`,
  `SimplifyVwPreserve::simplify_vw_preserve_idx`: each gains a `type Output` that resolves to 
  `Vec<usize>` / `Vec<Vec<usize>>` / `PolygonIndices` / `Vec<PolygonIndices>` for
  `LineString` / `MultiLineString` / `Polygon` / `MultiPolygon`.
- `ConvexHull::convex_hull_idx` is the odd one out, since it always provides one ring, so it just provides a single flat `Vec<usize>`.

 I've done the best I can to review this line by line, but since this is touching more pre-existing code, consider this PR as needing more than a cursory review.

I put the deprecated version as `0.34.0`, but that's me assuming this deserves a minor semver bump.

On this hot New England day my poor laptop is getting throttled, so my benchmark run this time is noisy, but doesn't seem to show a regression. I suspect that the 'improvements' are just throttling noise from my environment.
```
cargo bench --bench simplify --bench simplifyvw --bench convex_hull -- --baseline main --measurement-time 10

convex hull f32         time:   [156.58 µs 158.46 µs 161.02 µs]
                        change: [-3.0415% -0.8138% +1.2871%] (p = 0.48 > 0.05)
  2 (2.00%) high mild
  3 (3.00%) high severe

convex hull f64         time:   [158.24 µs 160.91 µs 165.07 µs]
                        change: [-16.609% -8.3167% -1.7093%] (p = 0.03 < 0.05)
                        Performance has improved.
  1 (1.00%) high mild
  3 (3.00%) high severe

convex hull with collinear random i64
                        time:   [21.987 ms 22.282 ms 22.637 ms]
                        change: [-15.480% -13.981% -12.228%] (p = 0.00 < 0.05)
                        Performance has improved.
  5 (5.00%) high mild
  9 (9.00%) high severe

simplify simple f32     time:   [72.446 µs 73.468 µs 75.020 µs]
                        change: [-1.7000% +0.5499% +3.0906%] (p = 0.68 > 0.05)
  7 (7.00%) high mild
  3 (3.00%) high severe

simplify simple f64     time:   [78.573 µs 80.417 µs 82.817 µs]
                        change: [-49.321% -25.363% +1.6455%] (p = 0.25 > 0.05)
  1 (1.00%) high mild
  4 (4.00%) high severe

simplify vw simple f32  time:   [115.27 µs 117.80 µs 120.90 µs]
                        change: [-10.861% -8.3766% -6.1652%] (p = 0.00 < 0.05)
                        Performance has improved.
  4 (4.00%) high mild
  6 (6.00%) high severe

simplify vw simple f64  time:   [146.73 µs 149.80 µs 153.59 µs]
                        change: [-7.8570% -6.1825% -4.4260%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
  6 (6.00%) high severe

simplify vwp f32        time:   [1.3102 ms 1.3293 ms 1.3554 ms]
                        change: [-0.8172% +0.6363% +2.4719%] (p = 0.49 > 0.05)
Found 17 outliers among 100 measurements (17.00%)
  1 (1.00%) low mild
  6 (6.00%) high mild
  10 (10.00%) high severe

simplify vwp f64        time:   [1.4928 ms 1.5229 ms 1.5639 ms]
                        change: [-0.9339% +0.9816% +3.1579%] (p = 0.34 > 0.05)
Found 17 outliers among 100 measurements (17.00%)
  2 (2.00%) high mild
  15 (15.00%) high severe

```

willcohen:
I also want to flag that `PolygonIndices` may be the wrong choice here too -- my gut non-typed inclination is to go with a simpler `Vec<Vec<usize>>` and just have people know that the shell goes first, but it also seems like the exact kind of error that a type would be ideal for. Let me know which way you'd prefer!

geo/src/algorithm/simplify_vw.rs
willcohen: Benches with the latest changes further down!

willcohen:
Rebased on main -- just to avoid the merge commit I accidentally auto-clicked through here on GH. Code is still the same as noted above.

urschrei:
I think this is pretty much done. I'm in favour of keeping `PolygonIndices`, for the reason you mention. Some (minor, imo) issues to be addressed:

- This is breaking bc of new associated types and required method, as well as removal of `SimplifyIdx` from the prelude and `geo::algorithm` so CHANGES needs to reflect that: both literally, and pointing to where the breakage is
- `convex_hull_idx_polygon` only asserts index bounds and `len() >= 4`. Our tests usually assert the exact expected index sequence
- The new fixtures don't use `wkt!` for construction; they should!

willcohen:
@urschrei thank you for the review! All points are addressed in the latest final commit.

urschrei (approved):
I'll leave this open for a day or so in case anyone else has follow-ups.

willcohen:
Thank you!

geo/src/algorithm/simplify.rs
```diff
@@ -142,6 +142,33 @@ where
...
 }
 
+/// Per-ring retained-index output for the index-returning simplification
+/// methods (`simplify_idx`, `simplify_vw_idx`, `simplify_vw_preserve_idx`) when
+/// applied to a `Polygon`. The exterior and interior ring indices are kept
+/// distinct rather than flattened into a single `Vec<usize>`; each index is
+/// relative to its own ring's coordinate sequence.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PolygonIndices {
+    pub exterior: Vec<usize>,
```
michaelkirk: Please make these fields private with public getters.

geo/src/algorithm/simplify_vw.rs
```diff
@@ -528,6 +581,9 @@ pub trait SimplifyVwIdx<T, Epsilon = T> {
 ///
 /// An `epsilon` less than or equal to zero will return an unaltered version of the geometry.
 pub trait SimplifyVwPreserve<T, Epsilon = T> {
+    /// The index-output type of [`SimplifyVwPreserve::simplify_vw_preserve_idx`], which varies by geometry.
+    type Output;
```
michaelkirk: Since there are multiple methods with outputs in this trait, calling this associated type `Output` is a little confusing. How about `IdxOutput` or similar. Please apply this change to the other traits in this PR as well.

geo/src/algorithm/simplify_vw.rs
```diff
@@ -943,4 +1081,104 @@ mod test {
...
+
+    // Pins the invariant that `SimplifyVwPreserve` is strictly subtractive:
+    // every coord in the output must also exist in the input. The algorithm
+    // tracks deletions via a fixed-length `adjacent` linked list and emits the
+    // surviving slots from `orig.0`, so synthesizing a new vertex would be a
+    // regression. Fixture exercises the shared-rtree retention branch (where
+    // removing `outer[2]` would self-intersect against the inner ring).
+    #[test]
+    fn simplify_vw_preserve_output_is_subset_of_input() {
+        let outer = line_string![
```
michaelkirk: This test seems overly complicated.

The fixture looks copy/pasted from tests that are testing the "shared-rtree retention branch" case, but we're not actually doing anything interesting with that behavior in this test.

My read is that we're only testing the "output is a subset of input". That would be a fine, but we should then use a simpler test fixture, and not mention the bit about self-intersection in the function doc.

geo/src/algorithm/simplify_vw.rs
```diff
@@ -943,4 +1081,104 @@ mod test {
...
+            (x: -40.0, y: 45.),
+            (x: -29.513671875, y: 47.32027765985069),
+            (x: -22.869140625, y: 43.80817468459856),
+            (x: -24.451171875, y: 35.266685523707665),
+        ];
+        let poly = Polygon::new(outer, vec![inner]);
+
+        let simplified = poly.simplify_vw_preserve(95.4);
+
+        for ring in std::iter::once(simplified.exterior()).chain(simplified.interiors().iter()) {
```
michaelkirk: Not a big deal, so take it or leave it, but you could make this a bit more readable with coords_iter.

```rust
        let input_coords: Vec<_> = poly.coords_iter().collect();
        for c in simplified.coords_iter() {
            assert!(
                input_coords.contains(c),
                "simplify_vw_preserve emitted coord {:?} not in input",
                c
            )
        }
```

michaelkirk (changes requested):
I have some requests, but I _think_ they should be easy to address.

michaelkirk (approved):
Rather than hold up merging on nitpicks, I've pushed up some changes.

willcohen:
Lovely. Thank you for being faster than me. I appreciate everyone's help!

PR #1546, opened 2026-06-15
Add snappable cardinal, ordinal, eightwise, sixteenwise compass direction types

mikwielgus:
---

This PR adds cardinal, ordinal, eight-wise, sixteen-wise compass direction enums, convertible from `Coord` by calculating bearing and dividing it, and then convertible back into `Coord` as unit vectors and sign vectors.

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -12,6 +12,8 @@
   - <https://github.com/georust/geo/pull/1535>
 - Add `RectOps` trait with RectOps::rect_union()` and `RectOps::rect_intersection()` operations.
   - <https://github.com/georust/geo/pull/1542>
+- Add snappable cardinal, ordinal, eightwise, sixteenwise compass directions.
+  - <https://github.com/georust/geo/pull/1546>
 
 ## 0.33.1 - 2026-4-20
 
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -30,6 +30,8 @@
 - BREAKING: Tightens the bound on `Contains<MultiPoint<T>> for MultiPoint<T>` from `T: CoordNum` to `T: GeoNum`
   - <https://github.com/georust/geo/pull/1555>
 - Derive all error types via `thiserror` instead of hand-written `Display`/`Error` impls.
+- Add snappable cardinal, ordinal, eightwise, sixteenwise compass directions.
+  - <https://github.com/georust/geo/pull/1546>
 
 ## 0.33.1 - 2026-4-20
 
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -30,7 +30,7 @@
 - BREAKING: Tightens the bound on `Contains<MultiPoint<T>> for MultiPoint<T>` from `T: CoordNum` to `T: GeoNum`
   - <https://github.com/georust/geo/pull/1555>
 - Derive all error types via `thiserror` instead of hand-written `Display`/`Error` impls.
-- Add snappable cardinal, ordinal, eightwise, sixteenwise compass directions.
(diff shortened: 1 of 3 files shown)


mikwielgus:
I have added functions to turn the direction enums clockwise and counterclockwise.

I should disclose: for this whole PR, I ran Cursor Auto a few times to speed up writing code and tests (with many manual changes as well). Names of all public types and functions have been chosen by me, doc comments have been almost entirely written by me, I have reviewed the code and ran the tests by hand. This is not a low-effort patch.

The module may be slightly overtested, depending on your taste, but having more tests doesn't hurt if your interface is well-defined and stable, which is likely to be the case here.

mikwielgus:
I would like to ask, are there are any preliminary remarks about this from the maintainers? Are compass directions enums of any interest, or perhaps this patch is a bad fit for this library?

michaelkirk:
I can see how this could be useful.

Is this a common thing supported in other geometry or geographic libraries?

Is the bearing implementation redundant with existing code?

mikwielgus:
> Is this a common thing supported in other geometry or geographic libraries?

Haskell geometry library [hgeometry](https://hackage.haskell.org/package/hgeometry-0.14/docs/Data-Geometry-Directions.html) has enums for cardinal and ordinal (intercardinal) directions.

Javascript geospatial library [`geolib`](https://github.com/manuelbieh/geolib) (4.3k stars on GitHub) has a [function](https://github.com/manuelbieh/geolib#getcompassdirectionorigin-destination-bearingfunction--getrhumblinebearing) to snap a vector to a compass direction.

Java GIS library [`geotools`](https://geotools.org/) has compass direction [enums](https://docs.geotools.org/stable/javadocs/net/opengis/gml311/CompassPointEnumeration.html) too.

I could also find a few small Rust crates that implement these enums: [direction](https://docs.rs/direction), [ae-direction](https://docs.rs/ae-direction), [endgame_direction](https://docs.rs/endgame_direction/latest/endgame_direction/), [roguemetry](https://github.com/Tairesh/roguemetry) (I've also seen some small Python packages for that). However, these libraries have their own types for vectors, which would require additional conversions to interoperate with `geo`, and some of them are not as mature and well-maintained as `geo`. For me, it would be convenient to have these enums available in `geo`, which I already use, instead of having to rely on these crates.

mikwielgus:
> Is the bearing implementation redundant with existing code?

There are some differences relevant to this code.

`Euclidean::bearing()` returns `0.0` for zero vector as input, whereas `compass_bearing()` here returns `None` in that case, which is then short-circuited in functions that call it.

`compass_bearing()` also works on integer coordinates, because it uses `CoordNum` as its scalar bound, whereas `Euclidean::bearing()` works only on floats, as its scalar bound is `CoordFloat`. (it may be worth considering to make `Euclidean::bearing()` also work for integer scalars).

mikwielgus:
Finally, one more thing that may be worth your attention.

I named the eight-wise and sixteen-wise direction enums `EightwiseDirection` and `SixteenwiseDirection`. The use of the words "eight-wise" and "sixteen-wise" for this purpose is not common (though I like it). A more linguistically conservative option could be `Direction8` and `Direction16` or `CompassDirection8` and `CompassDirection16` instead.

mikwielgus:
I have fixed CW/CCW convention (it was inverted instead of assuming X goes east and Y goes north), added `.opposite()` methods, added `ALL` constants to iterate over all enum variants.

Moreover, in the next commit, I have added enums for compass axes (compass directions but without senses). This is also useful, but makes this PR bigger, so if you want, I can split this part into another PR.

mikwielgus:
Ping. Some time has passed. Do you have any opinion if and when this can be merged?

The functionality is rather straightforward, but reduces glue code considerably when working with grids. It would be convenient for me to have this in geo.

Issue #621, opened 2021-02-06
Add example to Contains documentation demonstrating self-containing behavior


frewsxcv:
Following up from https://github.com/georust/geo/issues/615, it could be good to demonstrate this case in the `# Examples` section of the documentation: https://docs.rs/geo/0.17.0/geo/algorithm/contains/trait.Contains.html#examples

Adding something like:

```
//! Because this algorithm conforms to [DE-9IM], a one-dimensional type like `LineString` does not
//! contain itself, compared to a two-dimensional type like `Polygon`:
//!
//! ```
//! let line_string = geo::line_string[...];
//! assert!(!line_string.contains(&line_string));
//!
//! let polygon = geo::Polygon::new(line_string, vec![]);
//! assert!(polygon.contains(&polygon));
//! ```
```

PR #1548, opened 2026-06-24
Document that a LineString's endpoints are not contained

caraka:
The `Contains` trait's `# Examples` currently shows only positive "X in Y" cases. This adds a short example documenting a subtler, easy-to-miss behavior: a `LineString`'s endpoints are part of its *boundary*, not its interior, so they are not "contained", while interior points are — a direct consequence of the DE-9IM matrix already described in the trait docs. The example is a doc-test, so it's compile- and assertion-checked.

(Spotted while looking at #621, whose requested example — a 1-D `LineString` not containing *itself* — no longer holds: `contains` is reflexive across dimensions today. That issue looks obsolete and could probably be closed.)

### DIFF
--- a/geo/src/algorithm/contains/mod.rs
+++ b/geo/src/algorithm/contains/mod.rs
@@ -31,6 +31,14 @@
 ///
 /// // Point in Polygon
 /// assert!(polygon.contains(&point!(x: 1., y: 1.)));
+///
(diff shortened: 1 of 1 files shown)


michaelkirk (approved):

michaelkirk:
Thanks!

PR #1550, opened 2026-07-02
Include license texts in all published crates

decathorpe:
We noticed that the license texts are currently missing when packaging the geo / geo-types crates for Fedora Linux as a dependency of the latest GNOME Papers (PDF / document viewer).

Both the Apache-2.0 and MIT licenses require that redistributed sources contain a copy of the original license text.

Including symbolic links should be enough since `cargo package` / `cargo publish` resolve all symbolic links prior to packaging / publishing - unless you run those commands on a system that doesn't support symlinks properly (i.e. Windows without symbolic link support enabled).

We don't need new releases of the crates - we can temporarily include the `LICENSE-{APACHE,MIT}` files from the GitHub repo until such time as there are new releases published to crates.io.

Thanks!

- [x] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users (not a user-visible change).

### DIFF
--- a/geo-traits/LICENSE-APACHE
+++ b/geo-traits/LICENSE-APACHE
@@ -1,0 +1,1 @@
+../LICENSE-APACHE
--- a/geo-traits/LICENSE-MIT
+++ b/geo-traits/LICENSE-MIT
@@ -1,0 +1,1 @@
+../LICENSE-MIT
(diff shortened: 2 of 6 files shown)


urschrei (approved):
Thanks!

lnicola (approved):

PR #1551, opened 2026-07-03
fix: compute reverse bearing exactly in HaversineClosestPoint

### LINKED ISSUE #1325: Haversine closest point is in angles-as-euclidian-coordinates?
Please correct me if my assumptions or test is incorrect. 

```rust
use geo::prelude::*;
use geo::point;
use approx::assert_abs_diff_eq;
fn main() {
    // Make sure midpoint returned when at high latitudes
    use geo::Geodesic;
    let pt_a = point!(x: 0.0, y: 80.0);
    let pt_b = point!(x: 179.99, y: 80.0);
    let ln = geo::Line::new(pt_a, pt_b);
    let d = Geodesic::distance(pt_a, pt_b);
    let midpoint = Geodesic::point_at_ratio_between(pt_a, pt_b, 0.5);
    eprintln!("the midpoint is {:?}", midpoint);
    let closest_point = {
        let closest_res = ln.haversine_closest_point(&midpoint);
        use geo::Closest::*;
        match closest_res {
            Intersection(p) => p,
            SinglePoint(p) => p,
            Indeterminate => {
                panic!("Indeterminate");
            }
        }
    };
    assert_abs_diff_eq!(closest_point.y(), midpoint.y(), epsilon = 1e-6);
    assert_abs_diff_eq!(closest_point.x(), midpoint.x(), epsilon = 1e-6);
}
```

Image

The returned closest point is neither the input point or on the geodesic line. 

It _appears_ that many of the "spherical" methods like this are actually just treating angles as Cartesian coordinates with wraparound?

again, please correct me if I've done something incorrect here.

mvanhorn:
## Summary

`HaversineClosestPoint` now returns correct results for segments at high latitudes: interior points resolve to themselves instead of snapping to a segment endpoint.

## Why this matters

The reporter in #1325 hit this with a segment at 80 degrees latitude: the closest point to the segment's own midpoint came back as an endpoint thousands of kilometers away. The cross-track check needs the bearing from the segment end back toward the start, and the code approximated it as `bearing(p1, p2) +/- PI` (haversine_closest_point.rs:104-110 before this change). On a great circle that reverse bearing is only correct near the equator; at high latitudes the approximation is off by tens of degrees, so interior points were classified as outside the arc. The fix computes `Haversine.bearing(p2, p1)` directly, which is exact at any latitude.

## Testing

Two regression tests cover the reported case: the midpoint of a high-latitude segment resolves to itself (`point_to_line_high_latitude_midpoint`), and a point genuinely outside the arc still snaps to the nearest endpoint (`point_to_line_high_latitude_outside_arc`). `cargo test -p geo` passes (1113 unit + 224 doc tests), `cargo fmt --check` is clean, and the two `cargo clippy` warnings in the package predate this change (relate_operation.rs, old_sweep/proc.rs).

Fixes #1325

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- FIX: `Line::haversine_closest_point` no longer returns an endpoint for valid high-latitude projections.
+  - <https://github.com/georust/geo/issues/1325>
 - Unpin the `earcut` dependency now that the upstream semver violation has been reverted.
   - <https://github.com/georust/geo/pull/1533>
 - Added `Earcutter` and `TriangulateEarcut::earcut_triangulation_ref` to avoid per-call memory allocations across multiple triangulations.
--- a/geo/src/algorithm/haversine_closest_point.rs
+++ b/geo/src/algorithm/haversine_closest_point.rs
@@ -101,14 +101,9 @@ where
             return Closest::SinglePoint(p1);
         }
 
-        let pi = T::from(std::f64::consts::PI).unwrap();
         let crs_ad = Haversine.bearing(p1, *from).to_radians();
         let crs_ab = Haversine.bearing(p1, p2).to_radians();
-        let crs_ba = if crs_ab > T::zero() {
-            crs_ab - pi
-        } else {
-            crs_ab + pi
-        };
+        let crs_ba = Haversine.bearing(p2, p1).to_radians();
         let crs_bd = Haversine.bearing(p2, *from).to_radians();
         let d_crs1 = crs_ad - crs_ab;
         let d_crs2 = crs_bd - crs_ba;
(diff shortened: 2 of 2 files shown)


urschrei:
Hi @mvanhorn, thanks for the PR. As [previously requested](https://github.com/georust/geo/pull/1545#issuecomment-4702160050), please write (or at the very least edit) your own PR descriptions. We don't have any particular objection to LLM-assisted code contributions, but we do want to know that we're talking to humans when we discuss those contributions.

mvanhorn:
Sorry, you're right, you asked on #1545 too. Short version of this one: the old code approximated the reverse bearing as `bearing(p1,p2) ± PI`, which is only right near the equator, so at high latitudes the cross-track check misfired and interior points snapped to an endpoint. Using the real `Haversine.bearing(p2,p1)` fixes the #1325 case. I'll write these myself going forward.

urschrei (approved):
Thanks, @mvanhorn. I'm happy with this – I'll leave it open for a day in case anyone else wants to weigh in.

PR #1552, opened 2026-07-04
Return None for Centroid on non-finite coordinates

LucaCappelletti94:
`Centroid` for `Polygon` panicked in debug builds via `debug_assert!(ring.is_closed())` when a ring contained `NaN` coordinates, because `LineString::is_closed` uses `==` comparison and `NaN != NaN`. In release builds the assert was skipped but the computation produced `Some(POINT(NaN NaN))`.

The fix adds non-finite coordinate checks at each public entry point in `CentroidOperation`: `add_polygon` validates all rings upfront, `add_line_string` guards the `LineString::centroid` path, and `add_multi_point` guards `MultiPoint::centroid`. Each returns early without contributing to the centroid, so the result is naturally `None`. The redundant `debug_assert!(ring.is_closed())` in `add_ring` is removed since `add_polygon` is its only caller.

### DIFF
--- a/geo/src/algorithm/centroid.rs
+++ b/geo/src/algorithm/centroid.rs
@@ -602,7 +602,10 @@ impl<T: GeoFloat> CentroidOperation<T> {
     }
 
     fn add_ring(&mut self, ring: &LineString<T>) {
-        debug_assert!(ring.is_closed());
+        // NaN/Inf rings can't be closed (NaN != NaN), so let them through and the arithmetic propagates the non-finite value into the centroid.
+        debug_assert!(
+            ring.is_closed() || ring.0.iter().any(|c| !c.x.is_finite() || !c.y.is_finite())
+        );
 
         let area = get_linestring_area(ring);
         if area == T::zero() {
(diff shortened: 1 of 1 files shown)


geo/src/algorithm/centroid.rs
```diff
@@ -602,8 +635,6 @@ impl<T: GeoFloat> CentroidOperation<T> {
     }
 
     fn add_ring(&mut self, ring: &LineString<T>) {
-        debug_assert!(ring.is_closed());
```
michaelkirk, line 605: please re-review all your PRs and restore the assertions so I don't have to keep making this comment 😂

michaelkirk (changes requested):

geo/src/algorithm/centroid.rs
LucaCappelletti94: Restored, and did the same sweep across the other PRs.

Currently, the PRs code filters non-finite coordinates at the public entry points (`add_polygon`, `add_line_string`, `add_multi_point`, and the equivalents in the relate/geodesic/sweep PRs) so the observable behavior on `NaN`/`Inf` input is a defined early return, and keep the internal asserts as invariant markers behind that filter, which should now never lead to a fuzzer panic.

I will provide a PR later on with a fuzzer, so that you can catch these errors before they explode in downhill crates, and we can eventually catch more solid errors. The first ones that the fuzzer catches are always trivialities like this one, but they prevent the fuzzer from finding more menacing vulnerabilities like the ones I reported in geozero.

pr-review-replies.md
```diff
@@ -0,0 +1,41 @@
+# PR Review Replies
```
michaelkirk: What is this file?

geo/src/algorithm/centroid.rs
```diff
@@ -1111,4 +1143,51 @@ mod test {
             .push(Rect::new(c(10., 10.), c(11., 11.)).into());
         assert_eq!(collection.centroid().unwrap(), point!(x: 10.5, y: 10.5));
     }
+
+    #[test]
+    fn polygon_with_nan_centroid_returns_none() {
+        let nan_ring = Polygon::new(LineString::from(vec![(f64::NAN, f64::NAN)]), vec![]);
+        assert!(nan_ring.centroid().is_none());
```
michaelkirk: I think it would be preferable to output a centroid with a NAN/inf coordinate rather than `None` - is that reasonably doable?

michaelkirk (approved):
Thanks!

Issue #1059, opened 2023-08-25
GeodesicBearing produces NAN when y value is outside of [-90, 90]


JosiahParry:
When using `geodesic_bearing()` method with a point where the latitude is outside of the coordinate range [-90, 90], the result will be NaN. However, there is no corresponding result when a longitude value falls outside of the range [-180, 180]. I understand that values that fall outside of these ranges are likely to be planar coordinates and are not intended to work. If that is the case, I suspect the result of this method should return `Option<f64>` instead of `f64` to better handle these cases. 

On the contrary, the `haversine_bearing()` method has no such `NaN` behavior. It will return an f64 for any value provided despite it only being intended for values that fall within the [-180, 180] x range and [-90, 90] y range. 

Below are reproducible examples:

```rust
use geo::{point, GeodesicBearing};

#[test]
fn y_out_of_positive_range() {

    // y is out of range of Lat vals will return NaN
    let x = point!( x: 0_f64 , y: 91_f64);
    let y = point!( x: 0_f64 , y: 0_f64);

    assert!(x.geodesic_bearing(y).is_nan())
}

#[test]
fn y_out_of_negative_range() {

    // y is out of range of Lat vals will return NaN
    let x = point!( x: 0_f64 , y: -91_f64);
    let y = point!( x: 0_f64 , y: 0_f64);

    assert!(x.geodesic_bearing(y).is_nan())
}

#[test]
fn x_out_of_range() {
    // x is out of long rannge but will compute fine
    let x = point!( x: -361_f64 , y: 0_f64);
    let y = point!( x: 0_f64 , y: 0_f64);

    assert!(!x.geodesic_bearing(y).is_nan())
}
// running 3 tests
// test y_out_of_negative_range ... ok
// test y_out_of_positive_range ... ok
// test x_out_of_range ... ok
```

JosiahParry:
Note that this is using geo v 0.26.0

michaelkirk:
This third one actually seems reasonable to me:

```
#[test]
fn x_out_of_range() {
    // x is out of long rannge but will compute fine
    let x = point!( x: -361_f64 , y: 0_f64);
    let y = point!( x: 0_f64 , y: 0_f64);

    assert!(!x.geodesic_bearing(y).is_nan())
}
```

Longitude is just being wrapped.

e.g.

```diff
#[test]
fn x_out_of_range() {
    // x is out of long rannge but will compute fine
    let x = point!( x: -361_f64 , y: 0_f64);
[...]

PR #1553, opened 2026-07-04
Clamp latitude to [-90, 90] in Karney geodesic methods

LucaCappelletti94:
`Geodesic.distance`, `Geodesic.bearing`, and `Geodesic.destination` return `NaN` when a point's latitude is outside `[-90, 90]`, because the Karney inverse solver assumes latitude is within the pole boundaries. The input is finite and well-formed, so the `NaN` result is silent.

The fix clamps latitude to `[-90, 90]` before passing coordinates to the underlying `geographiclib_rs` solver. A point just past a pole is treated as the pole itself, which is the most intuitive behavior for small coordinate errors.

Related: [georust/geo#1059](https://github.com/georust/geo/issues/1059).

### DIFF
--- a/geo/src/algorithm/line_measures/metric_spaces/geodesic.rs
+++ b/geo/src/algorithm/line_measures/metric_spaces/geodesic.rs
@@ -3,6 +3,11 @@ use crate::Point;
 use geographiclib_rs::{DirectGeodesic, InverseGeodesic};
 use std::sync::LazyLock;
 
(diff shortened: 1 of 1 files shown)


michaelkirk:
What's wrong with returning NaN given invalid input?

LucaCappelletti94:
I re-read how postgis handles this, and it only clamps on the around of the borders plus returns an error, it does not simply clamps. I am closing this PR as it is not correct.

PR #1554, opened 2026-07-04
Use total_cmp in sweep line interval ordering

LucaCappelletti94:
`Geometry::interior_point()` panicked via `debug_assert!(intervals_overlap(...))` in the sweep line algorithm when processing polygons with near-zero coordinates containing negative zero (`-0.0`). The interval ordering used `<` (which treats `-0.0 == 0.0`) while sorting used `total_cmp` (which treats `-0.0 < 0.0`), violating the sweep's overlap invariant.

The fix uses `total_cmp` consistently for both interval ordering and sorting, so `-0.0` and `0.0` are handled uniformly.

Added also a targeted red test for the PR-reported `<` bug (not the change of this comment). The test builds a `SweepLineIndex` from `-0.0`/`0.0` endpoints and asserts the ordering invariant. Previously the only signal was the downstream `intervals_overlap` debug assert.


lnicola (approved):

geo/src/algorithm/sweep/mod.rs
```diff
@@ -248,7 +248,7 @@ impl<C: Cross> SweepLineIndex<C> {
 
         for segment in segments.into_iter() {
             let line = segment.line();
-            let (inserted_x, deleted_x) = if line.start.x < line.end.x {
+            let (inserted_x, deleted_x) = if line.start.x.total_cmp(&line.end.x).is_le() {
```
michaelkirk: You changed semantics here beyond NaN.

michaelkirk (changes requested):

geo/src/algorithm/sweep/mod.rs
LucaCappelletti94: Switched back to `.is_lt()`, but it is not a semantic change as you claim.

You can run the following snippet here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=3894721a107bcf66c16e1716ba0a4d2d

```rust
fn interval(s: f64, e: f64, le: bool) -> (u64, u64) {                                   
   let ord = s.total_cmp(&e);                                                          
   let take = if le { ord.is_le() } else { ord.is_lt() };                              
   let (a, b) = if take { (s, e) } else { (e, s) };                                    
   (a.to_bits(), b.to_bits())                                                          
}                                                                                       
                                                                                       
fn main() {                                                                             
   for (s, e) in [(-0.0, 0.0), (0.0, -0.0), (3.0, 3.0), (-1.0, 1.0), (1.0, -1.0)] {    
       assert_eq!(interval(s, e, true), interval(s, e, false));                        
   }                                                                                   
}
```

michaelkirk: Please do not make code reviews harder for us.

michaelkirk (approved):
Thanks!

PR #1555, opened 2026-07-04
Replace partial_cmp unwrap with total_cmp in lex_cmp

LucaCappelletti94:
`lex_cmp` ordered coordinates by unwrapping `partial_cmp`, which panics on `NaN` input (`partial_cmp` returns `None` for `NaN`). Any algorithm that sorts coordinates through `lex_cmp` (convex hull, Voronoi, point containment, and others) aborted the process on a `NaN` coordinate.

The fix replaces `partial_cmp().unwrap()` with `total_cmp`, which provides a deterministic total ordering that handles `NaN` gracefully. The `GeoNum` trait already provides `total_cmp` for both integers (delegating to `cmp`) and floats (using `f64::total_cmp`), so the change is a strict improvement with no behavioral change on finite input. Redundant NaN debug asserts in `NodeKey::cmp` are removed.

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -26,6 +26,8 @@
 - FIX: `InterpolatePoint::points_along_line` no longer appends a duplicate final vertex (and zero-length segment) for `Haversine`, `Geodesic`, and `Rhumb`.
   - <https://github.com/georust/geo/pull/1559>
 - Update the `rstar` dependency to 0.13.
+- FIX: `lex_cmp` uses `total_cmp` and no longer panics on `NaN` coordinates. Tightens the bound on `Contains<MultiPoint<T>> for MultiPoint<T>` from `T: CoordNum` to `T: GeoNum` (BREAKING).
+  - <https://github.com/georust/geo/pull/1555>
 
 ## 0.33.1 - 2026-4-20
 
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -26,7 +26,8 @@
 - FIX: `InterpolatePoint::points_along_line` no longer appends a duplicate final vertex (and zero-length segment) for `Haversine`, `Geodesic`, and `Rhumb`.
   - <https://github.com/georust/geo/pull/1559>
 - Update the `rstar` dependency to 0.13.
-- FIX: `lex_cmp` uses `total_cmp` and no longer panics on `NaN` coordinates. Tightens the bound on `Contains<MultiPoint<T>> for MultiPoint<T>` from `T: CoordNum` to `T: GeoNum` (BREAKING).
+- FIX: `lex_cmp` uses `total_cmp` and no longer panics on `NaN` coordinates.
+- BREAKING: Tightens the bound on `Contains<MultiPoint<T>> for MultiPoint<T>` from `T: CoordNum` to `T: GeoNum`
   - <https://github.com/georust/geo/pull/1555>
 
 ## 0.33.1 - 2026-4-20
--- a/geo/src/algorithm/contains/point.rs
+++ b/geo/src/algorithm/contains/point.rs
@@ -2,7 +2,7 @@ use super::{Contains, impl_contains_from_relate, impl_contains_geometry_for};
 use crate::algorithm::{CoordsIter, HasDimensions};
 use crate::geometry::*;
 use crate::utils::lex_cmp;
-use crate::{CoordNum, GeoFloat};
+use crate::{CoordNum, GeoFloat, GeoNum};
 
 // ┌────────────────────────────────┐
 // │ Implementations for Point      │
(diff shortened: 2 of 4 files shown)


lnicola:
Seems fine, but are NaN coordinates desirable?

LucaCappelletti94:
> Seems fine, but are NaN coordinates desirable?

The issue is not the NaN, which can be generated in all sorts of ways (corrupted input data, adversarial data) but the fact that a developer using this code does not know that it will cause a panic and crash the application. Additionally, panics like this make it very hard to fuzz software downhill.

lnicola:
Yeah, sounds fair.

lnicola (approved):

geo/src/algorithm/contains/point.rs
```diff
@@ -169,7 +169,7 @@ where
 
 impl<T> Contains<MultiPoint<T>> for MultiPoint<T>
 where
-    T: CoordNum,
+    T: GeoNum,
```
michaelkirk, line 172: This is a breaking change - please add it to the change log.

geo/src/algorithm/relate/geomgraph/node_map.rs
```diff
@@ -39,10 +39,6 @@ struct NodeKey<F: GeoFloat>(Coord<F>);
 
 impl<F: GeoFloat> std::cmp::Ord for NodeKey<F> {
     fn cmp(&self, other: &NodeKey<F>) -> std::cmp::Ordering {
-        debug_assert!(!self.0.x.is_nan());
```
michaelkirk, line 42: Keep these asserts.

geo/src/utils.rs
```diff
@@ -86,36 +87,25 @@ pub fn partial_min<T: PartialOrd>(a: T, b: T) -> T {
...
-    p.x.partial_cmp(&q.x)
-        .unwrap()
-        .then(p.y.partial_cmp(&q.y).unwrap())
+pub fn lex_cmp<T: GeoNum>(p: &Coord<T>, q: &Coord<T>) -> Ordering {
+    p.x.total_cmp(&q.x).then(p.y.total_cmp(&q.y))
 }
 
 /// Compute index of the least point in slice. Comparison is
 /// done using [`lex_cmp`].
-///
```
michaelkirk, line 102: restore this comment

geo/src/utils.rs
```diff
@@ -86,36 +87,25 @@ pub fn partial_min<T: PartialOrd>(a: T, b: T) -> T {
...
         .unwrap()
         .0
 }
-
 /// Compute index of the lexicographically least _and_ the
 /// greatest coordinate in one pass.
-///
-/// Should only be called on a non-empty slice with no `nan`
-/// coordinates.
-pub fn least_and_greatest_index<T: CoordNum>(pts: &[Coord<T>]) -> (usize, usize) {
```
michaelkirk: restore this comment

michaelkirk (changes requested):

geo/src/algorithm/contains/point.rs
LucaCappelletti94: Added to `Unreleased`

geo/src/algorithm/relate/geomgraph/node_map.rs
LucaCappelletti94: Restored.

geo/src/utils.rs
LucaCappelletti94: Restored

geo/src/utils.rs
LucaCappelletti94: Accidental drop during rebase, restored.

michaelkirk (approved):
Thanks!

geo/CHANGES.md
```diff
@@ -26,6 +26,8 @@
 - FIX: `InterpolatePoint::points_along_line` no longer appends a duplicate final vertex (and zero-length segment) for `Haversine`, `Geodesic`, and `Rhumb`.
   - <https://github.com/georust/geo/pull/1559>
 - Update the `rstar` dependency to 0.13.
+- FIX: `lex_cmp` uses `total_cmp` and no longer panics on `NaN` coordinates. Tightens the bound on `Contains<MultiPoint<T>> for MultiPoint<T>` from `T: CoordNum` to `T: GeoNum` (BREAKING).
```
michaelkirk:
```suggestion
- FIX: `lex_cmp` uses `total_cmp` and no longer panics on `NaN` coordinates.
- BREAKING: Tightens the bound on `Contains<MultiPoint<T>> for MultiPoint<T>` from `T: CoordNum` to `T: GeoNum`
```

Sorry to be pedantic, but makes the change log easier to scan at release time.

PR #1556, opened 2026-07-04
Guard Relate geometry graph construction against NaN coordinates

LucaCappelletti94:
`Relate` panicked via `debug_assert!(linear_ring.is_closed())` in geometry graph construction when a polygon ring contained `NaN` coordinates, because `is_closed` uses `==` comparison and `NaN != NaN`. This affected every relate-based predicate (`is_intersects`, `is_contains`, `is_within`, and others).

The fix adds non-finite coordinate checks at each geometry entry point in `GeometryGraph`: `add_polygon_ring`, `add_line_string`, `add_line`, and `add_point`. Each returns early without contributing to the graph when NaN coordinates are encountered, so the relate operation completes with a degenerate but defined result instead of panicking. A now-unreachable NaN assert in `NodeMap` is removed.

### DIFF
--- a/geo/src/algorithm/relate/geomgraph/geometry_graph.rs
+++ b/geo/src/algorithm/relate/geomgraph/geometry_graph.rs
@@ -206,6 +206,13 @@ where
         cw_left: CoordPos,
         cw_right: CoordPos,
     ) {
+        if linear_ring
+            .0
+            .iter()
+            .any(|c| !c.x.is_finite() || !c.y.is_finite())
+        {
+            return;
+        }
         debug_assert!(linear_ring.is_closed());
         if linear_ring.is_empty() {
             return;
@@ -264,6 +271,13 @@ where
         if line_string.is_empty() {
             return;
         }
+        if line_string
+            .0
+            .iter()
+            .any(|c| !c.x.is_finite() || !c.y.is_finite())
+        {
+            return;
+        }
 
         let mut coords: Vec<Coord<F>> = Vec::with_capacity(line_string.0.len());
         for coord in &line_string.0 {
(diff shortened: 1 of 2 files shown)


geo/src/algorithm/relate/geomgraph/geometry_graph.rs
```diff
@@ -292,8 +306,14 @@ where
...
     fn add_line(&mut self, line: &Line<F>) {
+        if !line.start.x.is_finite()
+            || !line.start.y.is_finite()
+            || !line.end.x.is_finite()
+            || !line.end.y.is_finite()
+        {
+            return;
+        }
         self.insert_boundary_point(line.start);
-        self.insert_boundary_point(line.end);
```
michaelkirk, line 296: Why did this go away?

geo/src/algorithm/relate/geomgraph/node_map.rs
```diff
@@ -77,13 +77,7 @@ where
         }
     }
     /// Adds a `NF::Node` with the given `Coord`.
-    ///
-    /// Note: Coords must be non-NaN.
     pub fn insert_node_with_coordinate(&mut self, coord: Coord<F>) -> &mut NF::Node {
-        debug_assert!(
```
michaelkirk, line 83: Leave the assert please. The point is to enforce things that "should never happen".

michaelkirk (changes requested):

geo/src/algorithm/relate/geomgraph/geometry_graph.rs
LucaCappelletti94: That is me doing merge collisions poorly, but it surprises me that the test suite did not go red. I am doing now some mutation testing to make the test suite more resilient.

geo/src/algorithm/relate/geomgraph/node_map.rs
LucaCappelletti94: Ok

PR #1557, opened 2026-07-08
Add BallTreeNode accessor API and cache-friendly point layout

urschrei:
`BallTree` currently supports only single-point queries (nearest_neighbour, within_distance, etc.). Dual-tree algorithms – clustering, k-NN graph construction, batch neighbour queries – instead need to walk the tree structurally: descend from the root, prune subtree pairs using ball-to-ball distance bounds, and scan the points under a leaf. Without a public surface for that, any consumer has to use private fields. This PR adds a small, read-only accessor API:

- `BallTree::node_count()` and `BallTree::node(idx)` return a `BallTreeNode`, a cheap `Copy` wrapper over `(tree, node index)`. The root of a non-empty tree is node 0.
- `BallTreeNode` exposes `center()`, `radius()`, `kind()`, `indices()` (original-input indices of the points under the node), `position_range()`, and `points()`.
- `NodeKind` distinguishes leaves from branches; branches hold child node indices for `BallTree::node`.
- `BallTreeNode::min_distance_to()` provides the ball-to-ball Euclidean lower bound used for dual-tree pruning, so each consumer doesn't re-calculate it.

It also adds an internal `points_ordered` array – the input points reordered to match the tree's leaf layout – so `points()` returns a contiguous slice rather than gathering through the indices permutation. This costs one extra copy of the input points (roughly `n * size_of::<Point<T>>()` bytes) in exchange for sequential leaf scans.

Notes:

- `NodeKind::Branch { left, right }` exposes raw `usize` indices rather than a pair of `BallTreeNodes`: returning nodes directly would tie `NodeKind` to a lifetime and lose `Copy`. Since `BallTree::node(idx)` is the only way to materialise an accessor, invalid indices can't be fabricated accidentally
- The API is `pub` rather than `pub(crate)`: external dual-tree implementations need the same API
- `node_count()` mirrors `len()` (which counts points, not nodes); `num_nodes()` is fine too if preferred.

Existing query APIs and behaviour are unchanged.

---

### DIFF
--- a/geo/src/algorithm/ball_tree.rs
+++ b/geo/src/algorithm/ball_tree.rs
@@ -176,10 +176,23 @@ impl<T: GeoFloat, D> BallTreePoint<T> for PointWithData<T, D> {
 ///
 /// Construction uses the KD algorithm: recursive median split along the axis
 /// of maximum dispersion, producing a balanced binary tree in O(n log n) time.
+///
+/// # Memory
+///
+/// The tree keeps two copies of the input points: one in original-input order
+/// (accessible via the standard index lookups) and one reordered to match the
+/// tree's leaf layout (exposed through [`BallTreeNode::points`] for
+/// cache-friendly leaf scans). This costs roughly `2 * n * size_of::<Point<T>>()`
+/// in bytes beyond the node metadata.
 #[derive(Debug, Clone)]
 pub struct BallTree<T: GeoFloat, D = ()> {
     nodes: Vec<Node<T>>,
     points: Vec<Point<T>>,
+    /// Points reordered so that `points_ordered[i] == points[indices[i]]`.
+    /// Lets leaf scans read points sequentially through a node's `start..end`
+    /// range, turning the otherwise scattered `points[indices[j]]` gather into
+    /// a contiguous walk.
+    points_ordered: Vec<Point<T>>,
     data: Vec<D>,
     indices: Vec<usize>,
 }
@@ -193,10 +206,25 @@ struct Node<T: GeoFloat> {
     kind: NodeKind,
 }
 
+/// Shape of a [`BallTree`] node: either a leaf, or an internal branch whose
+/// children are referenced by node index.
+///
+/// Use [`BallTree::node`] to turn those indices into [`BallTreeNode`]
+/// accessors.
 #[derive(Debug, Clone, Copy)]
-enum NodeKind {
+pub enum NodeKind {
+    /// A leaf node: all points in its ball are stored contiguously in the
+    /// tree's internal index array.
     Leaf,
-    Branch { left: usize, right: usize },
+    /// An internal node: its ball bounds the union of its two children's
+    /// balls. The fields are node indices that can be passed to
+    /// [`BallTree::node`].
+    Branch {
+        /// Node index of the left child.
+        left: usize,
+        /// Node index of the right child.
+        right: usize,
+    },
 }
 
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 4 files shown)


geo/src/algorithm/ball_tree.rs
```diff
@@ -254,6 +282,121 @@ impl<T: GeoFloat, D> BallTree<T, D> {
...
+    /// Node indices in the range `0..node_count()` can be passed to
+    /// [`BallTree::node`] to obtain a [`BallTreeNode`] accessor.
+    pub fn node_count(&self) -> usize {
+        self.nodes.len()
+    }
+
+    /// Returns an accessor for the node at the given index.
+    ///
+    /// For a non-empty tree the root is node `0`. Child indices reached via
+    /// [`NodeKind::Branch`] are also valid. Panics if `idx >= node_count()`
```
michaelkirk: We aren't super consistent about this, so take it or leave it, but I like the panics section for its uniformity.

```
  /// # Panics
```

geo/src/algorithm/ball_tree.rs
```diff
@@ -254,6 +282,121 @@ impl<T: GeoFloat, D> BallTree<T, D> {
...
+    /// if this bound exceeds the current best candidate, no pair of points
+    /// drawn from the two subtrees can improve the candidate.
+    ///
+    /// The computation uses the squared centre-to-centre distance and then
+    /// a square root, so with extreme coordinate magnitudes (roughly beyond
+    /// `1e100` for `f64`) the squared term can overflow to infinity and
+    /// propagate through the subtraction. Callers on well-scaled real-world
+    /// inputs don't need to worry about this; algorithms that may see
+    /// pathological magnitudes should clamp or pre-validate coordinates.
+    #[inline]
```
michaelkirk, line 398: The compiler will inline when it thinks it's useful. I don't think we should be trying to prematurely outsmart the compiler with all these inline comments unless something extraordinary is happening.

It's just a personal pet peeve - a paper-cut at worst - so I wouldn't block merging on it.

geo/src/algorithm/ball_tree.rs
```diff
@@ -176,10 +176,23 @@ impl<T: GeoFloat, D> BallTreePoint<T> for PointWithData<T, D> {
...
+/// cache-friendly leaf scans). This costs roughly `2 * n * size_of::<Point<T>>()`
+/// in bytes beyond the node metadata.
 #[derive(Debug, Clone)]
 pub struct BallTree<T: GeoFloat, D = ()> {
     nodes: Vec<Node<T>>,
     points: Vec<Point<T>>,
+    /// Points reordered so that `points_ordered[i] == points[indices[i]]`.
+    /// Lets leaf scans read points sequentially through a node's `start..end`
+    /// range, turning the otherwise scattered `points[indices[j]]` gather into
+    /// a contiguous walk.
```
michaelkirk, line 194: Is the idea that it's worth spending this extra memory because it's significantly faster than having some kind of `points_ordered() -> Iter<>` that uses the underlying `points` and `indices`?

geo/src/algorithm/ball_tree.rs
```diff
@@ -254,6 +282,121 @@ impl<T: GeoFloat, D> BallTree<T, D> {
...
+    ///
+    /// The computation uses the squared centre-to-centre distance and then
+    /// a square root, so with extreme coordinate magnitudes (roughly beyond
+    /// `1e100` for `f64`) the squared term can overflow to infinity and
+    /// propagate through the subtraction. Callers on well-scaled real-world
+    /// inputs don't need to worry about this; algorithms that may see
+    /// pathological magnitudes should clamp or pre-validate coordinates.
+    #[inline]
+    pub fn min_distance_to(&self, other: &BallTreeNode<'_, T, D>) -> T {
+        let center_dist = coord_distance_sq(self.center(), other.center()).sqrt();
```
michaelkirk, line 400:
```suggestion
        let center_dist = Euclidean.distance(self.center(), other.center());
```

Then you could remove the comment about overflow. Or is there some reason to prefer what you have?

geo/src/algorithm/ball_tree.rs
```diff
@@ -1259,4 +1405,136 @@ mod tests {
         let (bf_idx, _) = brute_force_nn(&points, &point!(x: 0.1, y: 0.1));
         assert_eq!(nn.index, bf_idx);
     }
+
+    // -- BallTreeNode accessor API ---------------------------------------
+
+    /// Collect the indices of every node in the tree by walking the
+    /// structure through the public accessor API.
+    fn walk_nodes(tree: &BallTree<f64>) -> Vec<usize> {
```
michaelkirk, line 1417: Presumably you have a use case for the new pub methods in this PR. I'm wondering if this method too would be a good candidate to add to the public API. Did you end up needing to implement something like this in your use case code?

If we were to make a public method like this, I have a couple concerns:

1. it seems like it would panic on an empty ball tree.
2. It might be better to have a lazy iter rather than allocating another vec (but I'd want to measure it)

On the other hand, if this kind of traversal is likely only useful for testing, I'm happy to leave it as is.

geo/src/algorithm/ball_tree.rs
```diff
@@ -1259,4 +1405,136 @@ mod tests {
...
+        }
+    }
+
+    #[test]
+    fn test_min_distance_to_overlapping_balls_is_zero() {
+        let tree = BallTreeBuilder::with_leaf_size(2).build(make_points());
+        let root = tree.node(0);
+        // A ball trivially overlaps itself and each of its children.
+        assert_eq!(root.min_distance_to(&root), 0.0);
+        if let NodeKind::Branch { left, right } = root.kind() {
```
michaelkirk: nit: `let else panic` seems more appropriate for a test to make sure we're actually running the test code in the following branch.

michaelkirk (approved):
Makes sense! I have no blocking concerns, but some optional ones if the spirit moves you.

geo/src/algorithm/ball_tree.rs
michaelkirk: I also remembered that `#[inline]` is necessary for cross-crate inlining. Michael today feels much less grumpy than Michael yesterday felt about this. Sorry.

michaelkirk:
Oh, oops. I only now realized that your HDBSCAN work was targeting this branch, so I merged the first of the stack into this branch. I feel like that's probably not what you intended to have happen. Sorry about that.

urschrei:
Hahahahaha oh god. Let me see if I can detangle it.

urschrei:
OK done. This is now the new bottom of the stack, nothing else changed.

geo/src/algorithm/ball_tree.rs
urschrei: I don't make anything similar public elsewhere in the stack! We could document that there's an ordering guarantee on `node_count()`/`NodeKind::Branch`: children always have higher indices than their parent. That would give external users the same capability with no allocation and no panic-on-empty (the range is simply empty). WDYT?

geo/src/algorithm/ball_tree.rs
urschrei: Yep: a lazy `points_ordered()` -> `impl Iterator` computing `points[indices[i]]` would remove the memory cost but keep the per-element random gather which is what the reordering removes.

The consumer is the leaf–leaf kernel in the dual-tree Boruvka traversal, an O(leaf_size^2) innermost loop over both nodes' points; with `points_ordered` we walk contiguous memory and return `&[Point<T>]` (rather than an iterator) (and we keep the slice API).

NB: there's no specific slice-vs-gather benchmark in geo-benches, so I'm relying on "this is a more vectorisation-friendly structure" here. We _could_ measure it.

geo/src/algorithm/ball_tree.rs
urschrei: `Euclidean.distance` for coords is implemented with `hypot`, which is overflow-safe but a much slower libm call than `sqrt(dx^2 + dy^2)`. `min_distance_to` is on the hot path, so we're trading the overflow caveat for speed.

michaelkirk: Ah, this gets used later in the PR stack. There's currently no way to measure it, so I'll take your word that this is the right tradeoff. 

I know we've all been taught "big PR ==big bad", but a bunch of small PR's that are less coherent are also a challenging review experience, because there's intermediate state which does not make sense. 

Personally I'm less afraid of big PR's than incomplete ones - especially when we wouldn't want to merge the first unless we're committed to merging the rest of the stack, which can be hard to know without reviewing the rest of the stack.

geo/src/algorithm/ball_tree.rs
michaelkirk: Without seeing how this code was ultimately consumed, I had a hunch this would be a useful public method, but sounds like I was wrong. Just skip it until/unless there's a reason for it.

michaelkirk (approved):

michaelkirk:
I wanted to enqueue this for merging, but I'm not sure how. Maybe you'll have to do it manually since I merged in the make_moons generator PR.

Screenshot 2026-08-19 at 10 59 23

geo/src/algorithm/ball_tree.rs
urschrei: > Personally I'm less afraid of big PR's than incomplete ones - especially when we wouldn't want to merge the first unless we're committed to merging the rest of the stack, which can be hard to know without reviewing the rest of the stack.

This is something we probably need to hash out in a bit more detail if we decide to use stacks going forward: my assumed way of working with them is to review individual PRs (in any order, but usually bottom-to-top), with the context of "this stack is landing feature x" so we have to begin with the question of whether feature x is something we want at all. Assuming we _do_ (based on e.g. PR description, cursory glance over the "shape" of the stack), we don't merge any of the stack until each PR is reviewed and approved, at which point the whole stack merges all at once (kind of like the rollups that the Rust repo uses).

That having been said, I've been ensuring that each PR passes CI individually and is logically self-contained to the greatest extent possible, just so it's reviewable on its merits, as opposed to "this a big feature, here it is in fragmented form".

Totally possible that this approach hasn't worked out in this case, _or_ is not going to work out for geo, _or_ (secret third confounder) that LLM-assisted features are a problem due to verbosity / quality / any other misgiving.

michaelkirk: "Do we want the feature" is definitely a good first line discriminator (I think HDBSCAn is something we want!)

A harder question to answer, more reflective of the reality is:

Given that we want the feature, should reviewing the amount code this feature requires take priority over the other things I want to work on... that's  a  much squishier line! And it's definitely more complicated now that we all have nuclear powered code generating devices in our pocket.

> cursory glance over the "shape" of the stack

I confess, I did not do this. I should have. I will try to get better at reviewing stacks. Thank you for bearing with me.

urschrei: > Given that we want the feature, should reviewing the amount code this feature requires take priority over the other things I want to work on... that's a much squishier line! And it's definitely more complicated now that we all have nuclear powered code generating devices in our pocket.

Yeah, I know, and I'm aware it touches both the "it's way easier to blam out a load of code now" and the "not many people review geo PRs, but we want features" concerns. I am also aware that this feature is both massive and extremely complex (probably the most complex code in the project, if it merges).

urschrei:
> I wanted to enqueue this for merging, but I'm not sure how. Maybe you'll have to do it manually since I merged in the make_moons generator PR.
> 
> Screenshot 2026-08-19 at 10 59 23

If you want, `gh stack merge 1557` should work to merge just this PR. You'll need the `ghost stack` tool to do it, though: `gh extension install github/gh-stack`

michaelkirk:
Raging against the machine (yelling at the clouds), I have opted to `git merge` and `git push` your branch.

urschrei:
Excited to find out whether GitHub is going to treat that as an edge case.

PR #1559, opened 2026-07-18
Fix duplicate final vertex in points_along_line

binggao1230:
`points_along_line` (the `InterpolatePoint` trait method) builds the interior points with a floating-point accumulator compared against `1.0`:

```rust
let interval = 1.0 / number_of_points;
let mut current_step = interval;
while current_step < 1.0 {
    points.push(/* point at current_step */);
    current_step += interval;
}
```

Summing `1/N` repeatedly drifts just below `1.0` for about 40% of `N`, so the guard runs one extra iteration and pushes an interior point at ratio ≈ 1.0 — coincident with `end`, which is then appended immediately after. The result is a duplicate final vertex and a zero-length final segment.

Three of the four metric spaces share this loop verbatim:

- `Haversine` (`metric_spaces/haversine.rs`)
- `Geodesic` (`metric_spaces/geodesic.rs`)
- `Rhumb` (`RhumbCalculations::intermediate_fill`)

`Euclidean` is unaffected because it interpolates through `densify_between`, which iterates over an integer count. Its comment even notes it "could be a good candidate to be *the single* basis for a unified generic of `points_along_line` for all metric spaces".

### Repro

New York → London at 300 km steps (`include_ends = true`):

| space | expected points | actual | final segment |
|---|---|---|---|
| Haversine | 20 | 21 | 2.5e-9 m |
| Geodesic | 20 | 21 | 1.2e-9 m |

`route[19] == route[20] == end`. Over a random battery, ~41% of Haversine/Rhumb lines and ~41% of Geodesic lines gain a spurious extra vertex. The gap never *exceeds* `max_distance`, so the interpolated positions themselves are fine — the defect is the extra vertex / wrong point count against the documented contract and the zero-length final segment, which breaks topology checks, `Simplify`, per-segment length weighting, and some renderers.

### Fix

Iterate over an integer step count in each of the three impls, mirroring `densify_between` (including its fallible `.to_u64().expect(...)`). Interior ratios become `frac * step` (a single multiply) rather than a running sum, so they equal the intended math and non-triggering outputs are byte-for-byte unchanged — the existing `points_along_line` unit tests pass unmodified.

Added a table-driven regression test in `interpolate_point.rs` covering all four metric spaces (Euclidean included as the already-correct reference): a triggering New York → London case plus a random battery, asserting no interior point coincides with `end` and that every sub-segment is a full step.

I did not fold the three impls onto `densify_between` in this PR: it interpolates via `point_at_ratio_between`, which for `Geodesic` re-solves the inverse per point and would change interior-point values. Keeping this surgical; that unification would make a reasonable follow-up.

### DIFF
--- a/geo/src/algorithm/line_measures/interpolate_point.rs
+++ b/geo/src/algorithm/line_measures/interpolate_point.rs
@@ -233,4 +233,104 @@ mod tests {
             assert_eq!(points, vec![]);
         }
     }
+
+    mod regular_spacing {
+        use super::*;
+        use crate::Distance;
+
+        // A line longer than `max_distance` must be split into equal sub-segments: the two ends
+        // plus the interior points, with no interior point coincident with `end`. Every gap is a
+        // full step - greater than half of `max_distance` (so there is no zero-length final
+        // segment) and never more than `max_distance` (the documented contract).
+        fn assert_regular<MS>(space: &MS, start: Point, end: Point, max_distance: f64)
+        where
+            MS: Distance<f64, Point, Point> + InterpolatePoint<f64>,
+        {
+            assert!(space.distance(start, end) > max_distance);
+
+            let with_ends: Vec<Point> = space
+                .points_along_line(start, end, max_distance, true)
+                .collect();
+            let without_ends: Vec<Point> = space
+                .points_along_line(start, end, max_distance, false)
+                .collect();
+
+            assert_eq!(with_ends[0], start);
+            assert_eq!(*with_ends.last().unwrap(), end);
+            assert_eq!(with_ends.len(), without_ends.len() + 2);
+            assert!(with_ends.len() >= 3);
+
+            for gap in with_ends.windows(2) {
+                let seg = space.distance(gap[0], gap[1]);
+                assert!(seg > max_distance / 2.0, "degenerate segment of {seg}");
+                assert!(
+                    seg <= max_distance * (1.0 + 1e-6),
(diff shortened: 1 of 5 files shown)


geo/src/algorithm/line_measures/interpolate_point.rs
```diff
@@ -233,4 +233,104 @@ mod tests {
...
+            // invariant (units here are coordinate degrees).
+            let (a, b) = (Point::new(0.0, 0.0), Point::new(0.0, 80.0));
+            assert_regular(&Euclidean, a, b, 7.0);
+            assert_eq!(Euclidean.points_along_line(a, b, 7.0, true).count(), 13);
+        }
+
+        #[test]
+        fn regular_spacing_battery() {
+            // Before the fix, ~40% of random lines produced a spurious duplicate end vertex in
+            // each of the three affected metric spaces. Deterministic LCG for reproducibility.
```
urschrei, line 312: What is `LCG`?

urschrei (approved):
This seems like a reasonable fix.

Please let us know that you used an LLM to produce this code.

gaoflow:
Yes — this was produced with AI assistance under my direction, and validated against the existing test suite plus new regression tests before submitting. Thanks for the review!

michaelkirk (approved):

michaelkirk:
Thank you for the fix! 

If you make another contribution (please do!) take care to edit down the LLM a bit more.

PR #1560, opened 2026-07-21
Optimise the separable geometry distance fast path

urschrei:
Several improvements to `separable_geometry_distance_fast` (with my robot friend `Fable`)

- Hoist the invariant scale factor `sqrt(1 + slope²)` out of the pruning loop and unify the two threshold formulas. The previous update formula `sqrt(min² · (1 + slope²))` overflowed to infinity for minimum distances beyond `~1e154`, silently disabling pruning
- Compute the closed-ring flags once per search instead of on every segment-distance evaluation
- Skip the robust Line–Line intersection predicate when computing segment distances: the fast path is only entered when the bounding boxes are strictly separated along an axis, so segments from opposite geometries can never intersect
- Use a single adjacent segment at open linestring endpoints instead of a degenerate zero-length duplicate
- Replace the `ControlFlow` `fold`s with plain loops, and use binary search to skip right-hand vertices whose projections trail the current left-hand vertex by more than the pruning threshold. The distance bound is symmetric in the projection delta, but the previous code only pruned in the forward direction, so overlapping projections caused a full scan of the trailing prefix.

Existing polygon distance benchmarks improve by ~12%. For geometries whose projections onto the centroid-connector axis overlap (elongated shapes offset diagonally, neither super-common nor unheard of), the prefix pruning replaces quadratic growth in vertex count with linear: 2.7× faster at 1,000 vertices, 9.1× at 4,000, and 35× at 16,000, with identical results.

Adds a benchmark exercising the overlapping-projection case. No new correctness tests are added by this PR as it's strictly a perf improvement.

---

### DIFF
--- a/geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
+++ b/geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
@@ -9,7 +9,6 @@ use crate::{CoordFloat, GeoFloat, GeoNum};
 use num_traits::{Bounded, Float};
 use rstar::RTree;
 use rstar::primitives::CachedEnvelope;
-use std::ops::ControlFlow;
 
 // Distance is a symmetric operation, so we can implement it once for both
 macro_rules! symmetric_distance_impl {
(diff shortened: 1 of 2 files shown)


geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -626,15 +636,27 @@ fn separable_geometry_distance_fast<F: GeoFloat>(
...
 
     // Step 4b: calculate the upper bound for a projection delta that could yield a smaller distance
     // This threshold allows us to skip vertex pairs that are too far apart by breaking early
     // this is the key piece of the algorithm!
-    let max_projection_delta = min_distance * (F::one() + (slope * slope)).sqrt();
-
-    // Step 5: minimum distance calculation. This is a bit hairy as an iterator, but the logic is straightforward:
+    // The scale factor is constant for the whole search, so compute it once. Note that
+    // min_distance * scale is preferred over sqrt(min_distance² * (1 + slope²)): the squared
+    // form overflows to infinity for min_distance beyond ~1e154, which would silently disable
```
michaelkirk, line 653: I suppose this constant depends on the numeric type. I know llm's like to insert "helpful" factoids like this, but I think it's unlikely to be useful and sometimes wrong.

michaelkirk (approved):
Great!

PR #1561, opened 2026-07-21
Improve MultiPolygon distance perf

urschrei:
Based on #1560, so merge that first.

Prune multi-geometry Euclidean distance using bounding-rectangle lower bounds (with my robot friend `Fable`)

The distance between two bounding rectangles is a _lower bound_ on the
distance between the geometries they enclose. The multi-geometry distance
implementations previously computed the full member-to-target distance for
every member. Members are now sorted by this lower bound, ascending, and
exact distances are **only** computed while the bound could still improve on
the running minimum: the first member whose bound reaches the current
minimum ends the search, as every subsequent bound is at least as large.
The search also ends early if the minimum reaches zero.

Members without a bounding rectangle (empty geometries) are given a zero
bound so they are never pruned, preserving existing empty-input behaviour.
Non-finite bounds prove nothing and never terminate the search.

Benchmark (10x10 grid of 64-gon polygons against an offset target
polygon): 163.5 µs before, 10.4 µs after, a roughly 16x improvement.

---

### DIFF
--- a/geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
+++ b/geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
@@ -351,25 +351,33 @@ impl_euclidean_distance_for_polygonlike_geometry!(&Rect<F>,      [&Point<F>, &Li
 // └───────────────────────────────────────────┘
 
 /// Euclidean distance implementation for multi geometry types.
+///
+/// Rather than folding the minimum over full member-to-target distance
+/// computations, members are first sorted by the distance between their
+/// bounding rectangle and the target's bounding rectangle. The exact distance
+/// is then only computed while a member's bounding-rectangle distance could
+/// still improve on the running minimum; the remaining members are skipped.
 macro_rules! impl_euclidean_distance_for_iter_geometry {
     ($iter_geometry:ty,  [$($to_geometry:ty),*]) => {
         impl<F: GeoFloat> Distance<F, $iter_geometry, $iter_geometry> for Euclidean {
             fn distance(&self, origin: $iter_geometry, destination: $iter_geometry) -> F {
-                origin
-                    .iter()
-                    .fold(Bounded::max_value(), |accum: F, member| {
-                        accum.min(self.distance(member, destination))
-                    })
+                bbox_pruned_min_distance(
+                    origin.iter(),
+                    destination.bounding_rect().into(),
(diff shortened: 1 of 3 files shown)


geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -439,6 +446,71 @@ impl<F: GeoFloat> Distance<F, &Geometry<F>, &Geometry<F>> for Euclidean {
 // │ Implementations utilities │
 // └───────────────────────────┘
 
+/// Minimum distance between two axis-aligned rectangles
+///
+/// Returns zero if the rectangles overlap or touch
+#[inline]
+fn rect_rect_distance<F: GeoFloat>(a: Rect<F>, b: Rect<F>) -> F {
+    let dx = (a.min().x - b.max().x)
```
michaelkirk: Can this be `Euclidean.distance(a, b)` rather than a new private method?

urschrei: If we swap that in, we call `Euclidean.distance(&Rect, &Rect)`, and we get the slow version: `origin.to_polygon()` -> `Distance<&Polygon, &Rect>` -> `to_polygon()` again -> full polygon–polygon distance: that's two heap allocations, an `intersects` sweep over 4x4 segments, and then _either_ `separable_geometry_distance_fast` or `nearest_neighbour_distance` (two R*-trees). The bench goes from ~10µs to ~35µs.

But we can drop the private function and give `Rect–Rect` a specialised trait impl with the same closed form, then call `Euclidean.distance` at the call site. That requires splitting the `polygonlike` macro so the self-to-self arm is generated for Triangle _only_, with Rect–Rect hand-written. See 7dbf114d for the diff.

michaelkirk: Perfect (btw can you shorten up the AI responses, it takes too much time to read and understand)!

My only qualm are the Rect/Rect distance calculation doc comments. Feel free to merge after patching that up.

michaelkirk:
much better, thanks! (and thanks for the speedup!)

urschrei:
> much better, thanks! (and thanks for the speedup!)

<3 (hit the approve button if you're happy?)

michaelkirk (approved):

PR #1562, opened 2026-07-22
feat: add `Geometry::static_name()` for type identification

nyurik:
Many of my projects end up having a simplified Display impl for Geometry to print just the type of the geometry. Might as well move this to the core.

Notes:
* there is another any-based impl that includes geometry type - this is different, as it is more suitable for dev than for end user
* I have also considered using [strum::IntoStaticStr](https://docs.rs/strum_macros/latest/strum_macros/derive.IntoStaticStr.html) as a simpler and more popular path, but using it for a single case seems a bit overkill

---

### DIFF
--- a/geo-types/CHANGES.md
+++ b/geo-types/CHANGES.md
@@ -8,6 +8,7 @@
 - Add `rstar` support for `Rect`.
   - <https://github.com/georust/geo/pull/1549>
 - Derive all error types via `thiserror` instead of hand-written `Display`/`Error` impls.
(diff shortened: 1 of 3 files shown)


michaelkirk (approved):
Thanks! I'll leave it open for a bit in case someone else cares to weigh in.

urschrei (approved):
Sorry about the merge conflict, but it's an easy fix.

nyurik:
no worries, fixed

PR #1563, opened 2026-07-22
chore: simplify all errors with `thiserror`

nyurik:
Tons of manually-maintained code can be done much cleaner and bug-free with `thiserror` crate. Result - shorter by 300 lines.

Note that `TriangulationError` error is duplicated for some reason in `triangulate_delaunay.rs` and `triangulate_spade.rs`, and they are nearly identical. Also, for some weird reason their error message was just the default Debug impl - not helpful. But this is all for another PR.

---

### DIFF
--- a/geo-types/Cargo.toml
+++ b/geo-types/Cargo.toml
@@ -12,7 +12,7 @@ edition = "2021"
 
 [features]
 default = ["std"]
-std = ["approx?/std", "num-traits/std", "serde?/std"]
+std = ["approx?/std", "num-traits/std", "serde?/std", "thiserror/std"]
 multithreading = ["rayon"]
 serde = [ "dep:serde", "rstar_0_8?/serde", "rstar_0_9?/serde", "rstar_0_10?/serde", "rstar_0_11?/serde", "rstar_0_12?/serde", "rstar_0_13?/serde"]
 
@@ -47,6 +47,7 @@ rstar_0_12 = { package = "rstar", version = "0.12", optional = true }
 rstar_0_13 = { package = "rstar", version = "0.13", optional = true }
 serde = { version = "1", optional = true, default-features = false, features = ["alloc", "derive"] }
 spade_2 = { package = "spade", version = "2", optional = true }
+thiserror = { version = "2.0.19", default-features = false }
 
 [dev-dependencies]
 approx = ">= 0.4.0, < 0.6.0"
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Derive all error types via `thiserror` instead of hand-written `Display`/`Error` impls. The error variants for the underlying-triangulation cases (`VoronoiError::Triangulation`, `TriangulationError::SpadeError`) and the nested validation errors now expose the wrapped error through `Error::source()`. `InvalidGeometry` variants are now `#[error(transparent)]` and gain `From` conversions from each inner validation error (e.g. `From<InvalidPolygon> for InvalidGeometry`).
 - FIX: `Line::haversine_closest_point` no longer returns an endpoint for valid high-latitude projections.
   - <https://github.com/georust/geo/issues/1325>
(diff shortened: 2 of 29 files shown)


geo-types/Cargo.toml
```diff
@@ -47,6 +47,7 @@ rstar_0_12 = { package = "rstar", version = "0.12", optional = true }
 rstar_0_13 = { package = "rstar", version = "0.13", optional = true }
 serde = { version = "1", optional = true, default-features = false, features = ["alloc", "derive"] }
 spade_2 = { package = "spade", version = "2", optional = true }
+thiserror = { version = "2.0", default-features = false }
```
michaelkirk:
```suggestion
thiserror = { version = "2.0.19", default-features = false }
```

geo/Cargo.toml
```diff
@@ -39,6 +39,7 @@ proj = { version = "0.31.0", optional = true }
 robust = "1.1.0"
 rstar = "0.13.0"
 serde = { version = "1.0", optional = true, features = ["derive"] }
+thiserror = "2.0"
```
michaelkirk:
```suggestion
thiserror = "2.0.19"
```

geo/CHANGES.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Derive all error types via `thiserror` instead of hand-written `Display`/`Error` impls. The error variants for the underlying-triangulation cases (`VoronoiError::Triangulation`, `TriangulationError::SpadeError`) and the nested validation errors now expose the wrapped error through `Error::source()`. `InvalidGeometry` variants are now `#[error(transparent)]` and gain `From` conversions from each inner validation error (e.g. `From<InvalidPolygon> for InvalidGeometry`).
```
michaelkirk: > `InvalidGeometry` variants are now `#[error(transparent)]`

Is there a behavioral change here, or is this only an implementation detail?

jts-test-runner/Cargo.toml
```diff
@@ -10,6 +10,7 @@ include_dir = { version = "0.7.2", features = ["glob"] }
 log = "0.4.14"
 serde = { version = "1.0.105", features = ["derive"] }
 serde-xml-rs = "0.6.0"
+thiserror = "2.0"
```
michaelkirk:
```suggestion
thiserror = "2.0.19"
```

michaelkirk (approved):
This looks great. We should have done it years ago.

As far as you know, are there any breaking changes here @nyurik?

michaelkirk:
It's a large change. I did review it, but it'd be good to get a +1 from another maintainer at least in terms of direction.

geo/CHANGES.md
nyurik: @michaelkirk its not breaking for sure, but there might be some cosmetic differences, e.g. geo/src/algorithm/triangulate_spade.rs has different error strings than before, but same code.

nyurik (commented):
cosmetics only, plus a few additional `from` impls to auto-convert some internal errors into wrappers - it might be possible now to simplify a few cases of error conversion.

geo/CHANGES.md
michaelkirk: Here I am asking specifically about the `#[error(transparent)]` in the changelog - it seems not actionable by the user, and feels like something an overly verbose LLM dumped in.

urschrei (approved):
Yeah, this is long overdue. Thanks for doing this @nyurik. I'm happy – I just created a small merge conflict when I merged #1549. If you could fix, then squash into a single commit we'll merge today.

geo/CHANGES.md
nyurik: i specifically asked llm to use transparent (it used source and manual string instead initially)

nyurik: P.S. i might still not be following though -- are you asking to reword the changelog msg?   I just cleaned it up a bit for simplicity

michaelkirk: Yes exactly. Using transparent is well and good, I was only asking about your changelog entry. It included verbose details of the sort an LLM would add, which I was 90% sure were irrelevant. Thanks for simplifying.

PR #1567, opened 2026-07-27
Fix clipping for unbounded Voronoi edges

### LINKED ISSUE #1558: voronoi_edges panics (debug) / drops edges (release) when a circumcentre is outside the bounding box
**`geo` version:** 0.33.1

**Item:** `Voronoi::voronoi_edges`

Identified with the help of a robot friend.

## Summary

For a valid but very obtuse triangle whose circumcentre falls well outside the
vertices' bounding box, `voronoi_edges`:

- **debug builds:** panics with
  `No infinite edges intersect the bounding box. Degenerate or invalid geometry?`
- **release builds:** returns `Ok` but with the affected edges silently dropped
  (here, an empty edge set for a valid triangle).

## Root cause

In the `[Inner, Outer]` arm, each infinite ray is extended from the circumcentre
(`start`) by `width + height` (the *unpadded* extent of the input points) before
being intersected with the padded bounding box:

```rust
let extended_end = Point::new(
    start.x + dir.x * (width + height),
    start.y + dir.y * (width + height),
);
```

When the circumcentre is far outside the box, `width + height` is far too short
to reach back to it, so no intersection is found and

```rust
debug_assert!(!intersections.is_empty(), "No infinite edges intersect ...");
```

fires (debug), or the edge is dropped (release). The extension length is tied to
the input's size rather than to the distance from `start` to the box.

For the triangle below, the circumcentre is at ~`(1, -4.95)`, the box is
`x in [0, 2], y in [0, 0.1]`, and `width + height = 2.1` which is nowhere near the ~5-unit gap.

## Suggested fix

Size the extension by the distance from `start` to the bounding box (or clip the
full infinite line to the padded rectangle, e.g. [Liang–Barsky](https://en.wikipedia.org/wiki/Liang–Barsky_algorithm)), so rays are
extended far enough regardless of where the circumcentre lies. A ray whose
finite portion _actually_ lies outside the box should then yield no edge without
tripping an assertion.

## Repro

```rust
#[test]
fn voronoi_edges_circumcentre_outside_bounds() {
    // A very obtuse triangle: its circumcentre is at about (1, -4.95), far
    // outside the vertices' bounding box (x in [0, 2], y in [0, 0.1]). Each
    // outer Voronoi ray is extended from the circumcentre by only
    // `width + height` (= 2.1), which is too short to bridge the ~5-unit gap
    // back to the padded box. In debug this trips the "No infinite edges
    // intersect the bounding box" debug_assert!; in release the edges are
    // dropped, leaving a valid triangle with no Voronoi edges.
    let triangle = wkt!(POLYGON((0.0 0.0, 1.0 0.1, 2.0 0.0, 0.0 0.0)));

    let edges = triangle.voronoi_edges().unwrap();

    // Part of the diagram lies within the padded box, so clipping must return
    // at least one edge.
    assert!(
        !edges.is_empty(),
        "expected clipped Voronoi edges for a valid triangle, got none"
    );
}
```

The same input as a `MultiPoint` reproduces it identically:

```rust
let sites = MultiPoint::from(vec![
    Point::new(0.0, 0.0),
    Point::new(1.0, 0.1),
    Point::new(2.0, 0.0),
]);
let _ = sites.voronoi_edges().unwrap();
```

LunaMeerkats:
## Summary

- clip unbounded Voronoi rays against the padded rectangle with a slab intersection
- preserve the part of an entering ray inside the rectangle, even when its circumcenter starts far outside
- add focused ray-clipping cases and the obtuse-triangle regression from #1558
- record the fix in the `geo` changelog

## Root cause

The previous implementation converted an infinite Voronoi ray into a finite
segment whose length was only the input width plus height. For a sufficiently
obtuse triangle, the circumcenter can be farther from the padded bounds than
that segment length. Debug builds then asserted because no intersection was
found, while release builds silently dropped the valid edges.

Clipping the full ray by its parameter interval avoids choosing an arbitrary
extension length. Rays entering from outside produce the rectangle's
entry-to-exit segment; rays starting inside produce start-to-exit; outward or
tangent-only rays produce no zero-length edge.

Fixes #1558.

## Validation

- `cargo fmt --all -- --check`
- `cargo test -p geo --lib voronoi` — 15 passed
- `cargo test -p geo --lib --release voronoi_edges_circumcenter_outside_bounds` — 1 passed
- `cargo check -p geo --all-targets`
- `cargo check -p geo --all-targets --no-default-features`
- `cargo clippy -p geo --all-targets -- -D warnings -A clippy::collapsible-match -A clippy::useless-conversion`
- broad `cargo test -p geo --lib` — 1,156 passed, 6 failed, 2 ignored; all six failures reproduce unchanged on the pinned base commit

The two allowed Clippy lints are unrelated Rust 1.96 diagnostics that also
reproduce on the base commit. All-features Clippy was unavailable locally
because the native PROJ/pkg-config/CMake toolchain is absent; Linux and MSRV
coverage are left to CI.

_Disclosure: this change was prepared and validated by an autonomous
open-source contribution agent under human-owned GitHub credentials._

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- FIX: Clip unbounded Voronoi edges when their circumcenter lies outside the
+  padded input bounds.
+  - <https://github.com/georust/geo/issues/1558>
 - FIX: `Line::haversine_closest_point` no longer returns an endpoint for valid high-latitude projections.
   - <https://github.com/georust/geo/issues/1325>
 - Add index-returning methods `Simplify::simplify_idx`, `SimplifyVw::simplify_vw_idx`, `SimplifyVwPreserve::simplify_vw_preserve_idx`, and `ConvexHull::convex_hull_idx`, returning input-relative indices of the retained vertices. Adds the `PolygonIndices` type and the `quick_hull_indices` free function.
--- a/geo/src/algorithm/voronoi.rs
+++ b/geo/src/algorithm/voronoi.rs
@@ -90,18 +90,15 @@ use crate::algorithm::triangulate_delaunay::{
     DelaunayTriangulationConfig, SpadeTriangulationFloat,
 };
 use crate::algorithm::vector_ops::Vector2DOps;
-use crate::line_intersection::{LineIntersection, line_intersection};
 use crate::utils::lex_cmp;
-use geo_types::{Coord, Line, LineString, Point, Polygon, Rect};
+use geo_types::{Coord, Line, LineString, Polygon, Rect};
 use num_traits::Float;
 use spade::Triangulation;
 use spade::handles::{VoronoiVertex::Inner, VoronoiVertex::Outer};
 
 use crate::algorithm::bool_ops::BoolOpsNum;
 use crate::algorithm::triangulate_delaunay::TriangulationError;
-use crate::{
-    BooleanOps, BoundingRect, Contains, Distance, Euclidean, TriangulateDelaunayUnconstrained,
-};
+use crate::{BooleanOps, BoundingRect, Contains, TriangulateDelaunayUnconstrained};
 
 /// Error type for Voronoi diagram computation.
 ///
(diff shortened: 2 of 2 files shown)


urschrei:
> Disclosure: this change was prepared and validated by an autonomous
open-source contribution agent under human-owned GitHub credentials.

We do not accept autonomous code contributions. If you are a human, review the code the agent(s) wrote under your direction, then open a PR in which you explain the scope and intent of the change. Please be prepared to answer questions from us regarding the changes.

PR #1568, opened 2026-07-27
State policies around agents and LLM contributions

urschrei:
Also tidy up old bors (RIP) mentions.

---


michaelkirk:
Thanks for writing this up. Looks good to me, but since it's also a policy doc, I think it makes sense to let it marinate for a while in case other maintainers want to weigh in.

michaelkirk (approved):

frewsxcv (approved):
Thanks for writing this up @urschrei

If anyone is feeling like they have the capacity, we should consider uplifting this to georust/meta and adding it to all the projects.

PR #1570, opened 2026-07-28
minor earcut bump to keep tests consistent

michaelkirk:
---

### DIFF
--- a/geo/src/algorithm/triangulate_earcut.rs
+++ b/geo/src/algorithm/triangulate_earcut.rs
@@ -8,7 +8,7 @@ pub trait TriangulateEarcut<T: CoordFloat> {
     /// # Examples
     ///
     /// ```
-    /// use geo::{coord, polygon, Triangle, TriangulateEarcut};
+    /// use geo::{polygon, wkt, TriangulateEarcut};
     ///
     /// let square_polygon = polygon![
     ///     (x: 0., y: 0.), // SW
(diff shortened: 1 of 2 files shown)


geo/src/algorithm/triangulate_earcut.rs
```diff
@@ -22,16 +22,8 @@ pub trait TriangulateEarcut<T: CoordFloat> {
...
-    ///             coord! { x: 0., y: 0. }, // SW
-    ///             coord! { x: 10., y: 0. }, // SE
-    ///             coord! { x: 10., y: 10. }, // NE
-    ///         ),
-    ///         Triangle(
-    ///             coord! { x: 10., y: 10. }, // NE
-    ///             coord! { x: 0., y: 10. }, // NW
-    ///             coord! { x: 0., y: 0. }, // SW
-    ///         ),
+    ///         wkt!(TRIANGLE(10.0 10.0,0.0 0.0,10.0 0.0)), // NE-SW-SE
```
michaelkirk, line 25: The vertices changed order in a recent minor update. It's not a winding change, just starting at a different vertex, so I think that's totally reasonable change for upstream to have made.

It's annoying that our tests are brittle, but I opted to simply update the fixtures (leaving them brittle) rather than writing a more robust test, gauging that it's more valuable that our tests be readable than robust. 

Reasonable minds could disagree!

urschrei (approved):

PR #1571, opened 2026-07-29
PreparedGeometry (and GeometryGraph) are Send

michaelkirk:
---

Supersedes #1198 (/cc @gauteh)

Benches are unremarkable except these two. I don't think it's a big deal. Note the 7% regression is a pretty small diff in absolute terms. 

```
jts test suite matching *Relate*
                        time:   [57.116 ms 57.467 ms 57.857 ms]
                        change: [+1.6262% +2.3887% +3.1370%] (p = 0.00 < 0.05)
                        Performance has regressed.
disjoint polygons       time:   [17.371 µs 17.706 µs 18.102 µs]
                        change: [+6.2106% +7.7836% +9.5288%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

### DIFF
--- a/geo/src/algorithm/relate/relate_operation.rs
+++ b/geo/src/algorithm/relate/relate_operation.rs
@@ -1,8 +1,8 @@
 use super::{EdgeEndBuilder, IntersectionMatrix};
 use crate::dimensions::{Dimensions, HasDimensions};
 use crate::relate::geomgraph::{
-    CoordNode, CoordPos, Edge, EdgeEnd, EdgeEndBundleStar, GeometryGraph, LabeledEdgeEndBundleStar,
-    RobustLineIntersector,
+    CoordNode, CoordPos, Edge, EdgeEnd, EdgeEndBundleStar, GeometryGraph, Label,
+    LabeledEdgeEndBundleStar, RobustLineIntersector,
     index::SegmentIntersector,
     node_map::{NodeFactory, NodeMap},
 };
@@ -10,8 +10,6 @@ use crate::{Coord, GeoFloat, GeometryCow};
 use crate::{CoordinatePosition, Relate};
 
 use geo_types::Rect;
-use std::cell::RefCell;
-use std::rc::Rc;
 
 /// Computes an [`IntersectionMatrix`] describing the topological relationship between two
 /// Geometries.
@@ -32,7 +30,6 @@ where
     geometry_b: &'a dyn Relate<F, Output = BBOX2>,
     nodes: NodeMap<F, RelateNodeFactory>,
     line_intersector: RobustLineIntersector,
-    isolated_edges: Vec<Rc<RefCell<Edge<F>>>>,
 }
 
 #[derive(PartialEq)]
@@ -61,7 +58,6 @@ where
             geometry_a,
             geometry_b,
             nodes: NodeMap::new(),
-            isolated_edges: vec![],
             line_intersector: RobustLineIntersector::new(),
         }
     }
@@ -116,7 +112,7 @@ where
 
         let mut nodes = NodeMap::new();
         std::mem::swap(&mut self.nodes, &mut nodes);
-        let labeled_node_edges = nodes
+        let labeled_node_edges: Vec<_> = nodes
             .into_iter()
             .map(|(node, edges)| (node, edges.into_labeled(&graph_a, &graph_b)))
             .collect();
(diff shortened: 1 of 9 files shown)


geo/src/algorithm/relate/relate_operation.rs
```diff
@@ -32,7 +30,6 @@ where
     geometry_b: &'a dyn Relate<F, Output = BBOX2>,
     nodes: NodeMap<F, RelateNodeFactory>,
     line_intersector: RobustLineIntersector,
-    isolated_edges: Vec<Rc<RefCell<Edge<F>>>>,
```
michaelkirk, line 35: This was a translation from JTS, which mutated self to hold the isolated_edges, and then immediately used them in one place. I've replaced that with a local variable, which obviates this field. It's not a big deal, but just replaces some more "global" reasoning with something more localized.

gauteh:
Thanks. I have tested, and this seems to work. I added an unsafe Sync, since (according to Claude) that should be safe:

```
#[derive(Clone)]
#[pyclass]
pub struct Shapes {
    prepped: PreparedGeometry<'static, Geometry>,
}

// PreparedGeometry is Send (PR georust/geo#1571). It contains RefCell internally so is not
// automatically Sync, but relate() only takes shared references and does not mutate across
// threads through the prepared graph, so it is safe to share.
unsafe impl Send for Shapes {}
unsafe impl Sync for Shapes {}

```

Matching seems to be fairly fast. For roaring-landmask loading the geometry and preparing it is slow compared to geos.

https://github.com/gauteh/roaring-landmask/pull/39

urschrei:
> loading the geometry

That's interesting @gauteh. What file format are you using?

gauteh:
wkb, it seems that it is preparing the geometry and not reading (using geozero), that takes time.

gauteh:
```
   ┌────────────────────────┬───────┬───────┐                                                                                                              ┃
   │ Step                   │ gshhg │ osm   │                                                                                                              ┃
   ├────────────────────────┼───────┼───────┤                                                                                                              ┃
   │ decompress             │ 1.7s  │ 2.3s  │                                                                                                              ┃
   ├────────────────────────┼───────┼───────┤                                                                                                              ┃
   │ geozero WKB parse      │ 224ms │ 256ms │                                                                                                              ┃
   ├────────────────────────┼───────┼───────┤                                                                                                              ┃
   │ PreparedGeometry::from │ 4.4s  │ 4.3s  │                                                                                                              ┃
   └────────────────────────┴───────┴───────┘                                                                                                              ┃
``` 

using:

```
use geo::PreparedGeometry;
use geozero::ToGeo;
use std::io::Read;
use std::time::Instant;

fn time_load(path: &str) {
    println!("=== {path} ===");

    // Step 1: decompress
    let t = Instant::now();
    let fd = std::fs::File::open(path).unwrap();
    let fd = std::io::BufReader::new(fd);
    let mut fd = xz2::bufread::XzDecoder::new(fd);
    let mut buf = Vec::new();
    fd.read_to_end(&mut buf).unwrap();
    println!("  decompress:             {:>8.1?}  ({} bytes)", t.elapsed(), buf.len());

    // Step 2: geozero WKB parse
    let t = Instant::now();
    let geom = geozero::wkb::Wkb(buf).to_geo().unwrap();
    println!("  geozero WKB parse:      {:>8.1?}", t.elapsed());

    // Step 3: PreparedGeometry construction
    let t = Instant::now();
    let _prepped = PreparedGeometry::from(geom);
    println!("  PreparedGeometry::from: {:>8.1?}", t.elapsed());
}

fn main() {
    time_load("assets/gshhg.wkb.xz");
    time_load("assets/osm.wkb.xz");
}
```

(assets from roaring-landmask repository).

urschrei:
My Claude thinks your `Sync` is unsound as written, btw:

<details>
<summary>🤖</summary>
 clone_for_arg_index clones the internal Rc-held R-tree (tree: self.tree.clone(), geometry_graph.rs:91) from &self — so two threads calling relate() concurrently on the shared PreparedGeometry race on a non-atomic Rc refcount. It won't reliably crash, but it's undefined behaviour, and it's the reason #1571 makes PreparedGeometry Send but deliberately not Sync. The safe pattern for the rayon use case is one clone per thread (or converting the internals to Arc, which would be a geo-side change).
</details>

But the larger question (sorry to derail this thread!) for your PR is whether `PreparedGeometry` is the correct approach for point-in-polygon. We have an [IntervalTree](https://docs.rs/geo/latest/geo/algorithm/indexed/interval_tree_multipolygon/struct.IntervalTreeMultiPolygon.html) which should be quite fast. Construction time alone should be 4-5x faster, and individual containment queries are 3 - 4 µs on my machine:

https://gist.github.com/urschrei/2526556bde173ab06c93eb8e3b2f56c1

(your timing code will need to be updated to replace `PreparedGeometry` steps with `IntervalTreeMultiPolygon::new`, and `test_geos_par_prepped.rs` can lose the `SyncPrepped` wrapper entirely: share `&Shapes` or clone it per rayon task, since clones are now an `Arc` bump).

If you need other types of geometry containment checks, `IntervalTree` won't work, of course.

gauteh:
Interesting! I only need contains, I will try out that. Good to know, was a bit suspicious about that unsafe reasoning...

urschrei:
I'm interested to see how this will do against GEOS: the interval tree is fairly close to optimal for this kind of query shape, and its underlying SIF tree implementation is very fast. I'd be unpleasantly surprised if we weren't close.

gauteh:
I updated the PR to use IntervalTree, looks very promising and also solves my Send/Sync issues... will do some more benchmarking.

gauteh:
Original geos:

```
test shapes::tests::benches::test_contains_in_ocean ... bench:       1,609.00 ns/iter (+/- 43.54)
test shapes::tests::benches::test_contains_on_land  ... bench:       1,528.50 ns/iter (+/- 26.84)
```

new IntervalTree:

```
running 2 tests
test shapes::tests::benches::test_contains_in_ocean ... bench:       2,168.67 ns/iter (+/- 113.98)
test shapes::tests::benches::test_contains_on_land  ... bench:       2,043.44 ns/iter (+/- 61.54)
```

so that is pretty good! geos is slightly better, but I think this may be good enough for me to switch over. Amazing.

urschrei:
Great news! I'll have a look to see whether there's any other perf to be chased down, but I don't think it's in the interval tree…

urschrei (approved):

PR #1573, opened 2026-07-31
Add HDBSCAN module scaffolding: core distances and union-find

urschrei:
First of five commits building up the HDBSCAN implementation. Adds the module skeleton, the core distance computation (Step 1: k-NN via ball tree, parallelised under the multithreading feature) with the `CoreData` struct that carries the retained k-NN table, and the two union-find variants shared by the MST and dendrogram steps.

Internal items are unused until the public API lands in the final commit of this stack, so the module temporarily carries `#![allow(dead_code)]`.

---

Part of the HDBSCAN stack (2 of 6). Based on #1557.

### DIFF
--- a/geo/src/algorithm/hdbscan/mod.rs
+++ b/geo/src/algorithm/hdbscan/mod.rs
@@ -1,0 +1,218 @@
+//! Hierarchical DBSCAN (HDBSCAN) clustering algorithm.
+//!
+//! HDBSCAN improves on DBSCAN by eliminating the epsilon parameter and
+//! producing a cluster hierarchy, making it more robust to clusters of varying
+//! density.
+//!
+//! # References
+//!
+//! Campello, R. J. G. B., Moulavi, D., & Sander, J. (2013).
+//! *Density-Based Clustering Based on Hierarchical Density Estimates.*
+//! In Advances in Knowledge Discovery and Data Mining (PAKDD 2013).
+
+// Temporary allowance while the algorithm is built up over several
+// commits; removed in the commit that adds the public API.
+#![allow(dead_code)]
+
+use crate::GeoFloat;
+use crate::Point;
+use crate::algorithm::ball_tree::BallTree;
+
+// ---------------------------------------------------------------------------
+// Step 1: Core distances via ball tree k-NN
+// ---------------------------------------------------------------------------
+
+/// Compute the core distance and k-NN indices for each point.
+///
+/// The *core distance* of a point is the distance to its `min_samples`-th
+/// nearest neighbour, **excluding the point itself**. This matches the
+/// reference [`hdbscan`](https://github.com/scikit-learn-contrib/hdbscan)
+/// library: `min_samples` counts other points, not the query point. We query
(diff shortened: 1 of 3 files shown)


geo/src/algorithm/hdbscan/mod.rs
```diff
@@ -0,0 +1,218 @@
...
+    let compute = |p: &Point<T>, dist: &mut T, row: &mut [usize]| {
+        let knn = tree.nearest_neighbours(p, k);
+        debug_assert_eq!(knn.len(), k, "min_samples must be <= n - 1");
+        *dist = knn[min_samples].distance;
+        for (slot, nn) in row.iter_mut().zip(knn.iter()) {
+            *slot = nn.index;
+        }
+    };
+
+    #[cfg(feature = "multithreading")]
```
michaelkirk, line 67: We already using rayon (via geo-types) when multithreading is enabled. This *seems* like it would fit well with rayon's par_iter. Did you consider using that instead of manually batching here?

geo/src/algorithm/hdbscan/mod.rs
```diff
@@ -0,0 +1,218 @@
...
+        Some(big)
+    }
+
+    fn size(&self, x: usize) -> usize {
+        self.size[x]
+    }
+}
+
+/// [`UnionForest`] with a synthetic per-component label, used by the
+/// single-linkage dendrogram construction in [`label`].
```
michaelkirk, line 182: What is `label`?

geo/src/algorithm/hdbscan/mod.rs
```diff
@@ -0,0 +1,218 @@
...
+/// (which only needs size-aware find+union) and the single-linkage
+/// dendrogram construction (which additionally tags each root with a
+/// synthetic label – see [`TreeUnionFind`]) can share the same
+/// primitives.
+struct UnionForest {
+    parent: Vec<usize>,
+    size: Vec<usize>,
+}
+
+impl UnionForest {
```
michaelkirk, line 140: We use UnionFind in the new Polygon validation - can we share that impl instead?

michaelkirk (approved):
I have some non-blocking suggestions.

michaelkirk:
BTW, I feel pretty strongly that we shouldn't merge any incomplete / unusable code into main. The first PR in the stack we already merged, but that one stood on its own (the balltree stuff), so that seems fine. 

Maybe the move is to merge the rest of the PR's into each other, and only merge the base one (this one) into main once they've all been reviewed?

urschrei:
> BTW, I feel pretty strongly that we shouldn't merge any incomplete / unusable code into main. The first PR in the stack we already merged, but that one stood on its own (the balltree stuff), so that seems fine.

Yep that one was slightly different (actually didn't you merge it by accident, and it just so happened that it was standalone?)

> Maybe the move is to merge the rest of the PR's into each other, and only merge the base one (this one) into main once they've all been reviewed?

Yes this is the move – the whole stack will merge at once when all of its PRs are approved.

PR #1576, opened 2026-07-31
Add cluster selection, label extraction, and GLOSH scores

urschrei:
Steps 5 to 7 of the HDBSCAN pipeline. Cluster stability is computed from the condensed tree following the reference implementation, and clusters are selected bottom-up: a parent whose stability exceeds the sum of its children's wins, otherwise it inherits the children's total. Flat labels are then assigned by walking each point up to its nearest selected ancestor, and GLOSH outlier scores are derived from the condensed tree independently of the flat selection, so noise points are scored too.

These steps have no direct unit tests; they are covered by the known-answer and end-to-end tests that land with the public API in the next pull request in the stack.

---

Part of the HDBSCAN stack (5 of 6). Based on #1557.

### DIFF
--- a/geo/src/algorithm/hdbscan/mod.rs
+++ b/geo/src/algorithm/hdbscan/mod.rs
@@ -21,6 +21,7 @@ use crate::Point;
 use crate::algorithm::ball_tree::BallTree;
 use crate::algorithm::ball_tree::NodeKind;
 use std::collections::HashMap;
+use std::collections::HashSet;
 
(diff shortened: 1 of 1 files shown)


michaelkirk (commented):
There seems to be several functions introduced here which aren't called. I'm assuming that they are used in subsequent PRs, but it's kind of weird to review them in isolation. Next time I'd say shoot for a smaller number of more complete PRs, even if it means the PRs will be larger.

PR #1577, opened 2026-07-31
Add Hdbscan trait, public API, and benchmark

urschrei:
Completes the HDBSCAN stack: the `Hdbscan` trait (with an impl for `MultiPoint` and slices of `Point`), the `HdbscanResult` and `HdbscanError` types, input validation, and the `hdbscan_impl` orchestration that wires the pipeline steps together. Exported from the algorithm module alongside `Dbscan`, with the crate docs updated.

Removes the temporary `#![allow(dead_code)]` now that every internal item has a caller.

Adds the known-answer tests (labels and GLOSH scores validated against the reference python `hdbscan` library), end-to-end and edge case tests, and a criterion benchmark using the shared `make_moons` generator.

---

Part of the HDBSCAN stack (6 of 6). Based on #1557.

### DIFF
--- a/geo/src/algorithm/hdbscan/mod.rs
+++ b/geo/src/algorithm/hdbscan/mod.rs
@@ -10,18 +10,370 @@
 //! *Density-Based Clustering Based on Hierarchical Density Estimates.*
 //! In Advances in Knowledge Discovery and Data Mining (PAKDD 2013).
 
-// Temporary allowance while the algorithm is built up over several
-// commits; removed in the commit that adds the public API.
-#![allow(dead_code)]
-
 use crate::Distance;
 use crate::Euclidean;
 use crate::GeoFloat;
+use crate::GeoNum;
+use crate::MultiPoint;
 use crate::Point;
 use crate::algorithm::ball_tree::BallTree;
 use crate::algorithm::ball_tree::NodeKind;
 use std::collections::HashMap;
 use std::collections::HashSet;
+
+/// Result of an HDBSCAN clustering run.
+///
+/// Contains both cluster labels and GLOSH outlier scores for every input point.
+#[derive(Debug, Clone, PartialEq)]
+pub struct HdbscanResult<T> {
+    /// Cluster assignment for each input point, in input order.
+    /// `Some(id)` for clustered points, `None` for noise.
+    pub labels: Vec<Option<usize>>,
+
+    /// GLOSH (Global-Local Outlier Scores from Hierarchies) outlier score for
+    /// each input point, in the range [0, 1]. Higher values indicate stronger
+    /// outliers.
+    pub outlier_scores: Vec<T>,
+}
+
+/// Errors that can occur during [`hdbscan`](Hdbscan::hdbscan) clustering.
+#[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
+pub enum HdbscanError {
+    /// `min_cluster_size` is out of range.
+    ///
+    /// `min_cluster_size` must be at least 2 and no greater than the number
+    /// of input points.
+    #[error(
+        "invalid min_cluster_size {min_cluster_size}: must be in 2..={n} (number of input points)"
+    )]
+    InvalidMinClusterSize {
+        /// The supplied `min_cluster_size`.
+        min_cluster_size: usize,
+        /// The number of input points.
+        n: usize,
+    },
+    /// `min_samples` is out of range.
+    ///
+    /// `min_samples` must be at least 1 and no greater than `n - 1`, since
+    /// the core distance is the distance to the `min_samples`-th nearest
+    /// neighbour excluding the point itself, and only `n - 1` other points
+    /// exist.
+    #[error(
+        "invalid min_samples {min_samples}: must be in 1..={} (n - 1, where n = {n} is the number of input points)",
+        .n.saturating_sub(1)
+    )]
+    InvalidMinSamples {
+        /// The supplied `min_samples`.
+        min_samples: usize,
+        /// The number of input points.
+        n: usize,
+    },
+    /// The minimum spanning tree could not connect all input points.
+    ///
+    /// HDBSCAN builds a single spanning tree over the mutual-reachability
+    /// graph. If the input contains non-finite coordinates (NaN or infinity),
+    /// or coordinates so large in magnitude that squared distances overflow
+    /// to infinity, the tree is left disconnected and any clustering would be
+    /// meaningless. Clean or rescale the input.
+    #[error(
+        "could not connect all input points: found {edges} spanning-tree edges, expected {expected}. Input may contain non-finite or extremely large coordinates"
+    )]
+    DisconnectedSpanningTree {
+        /// The number of spanning-tree edges that were found.
+        edges: usize,
+        /// The number of edges required to span all points (`n - 1`).
+        expected: usize,
+    },
+}
+
+/// Perform [HDBSCAN](https://en.wikipedia.org/wiki/HDBSCAN) (Hierarchical Density-Based Spatial Clustering of Applications with Noise) clustering on a set of points.
+///
+/// HDBSCAN extends DBSCAN by converting it into a hierarchical clustering
+/// algorithm, then extracting a flat clustering based on cluster stability.
+/// Unlike DBSCAN, it does not require an epsilon parameter and can find
+/// clusters of varying densities.
+///
+/// The central concept is the *mutual reachability distance* between two
+/// points *a* and *b*: `max(core_dist(a), core_dist(b), dist(a, b))`, where
+/// `core_dist(p)` is the distance from *p* to its *k*-th nearest neighbour
+/// (with *k* = `min_samples`). This inflates distances in sparse regions
+/// while leaving dense regions unchanged: in effect, a point in a sparse
+/// region is treated as farther from everything than it geometrically is.
+/// The minimum spanning tree (MST) of the resulting complete graph – the
+/// mutual reachability graph – therefore captures the density-based
+/// cluster structure of the data.
+///
+/// The MST makes it cheap to answer the question: "which points are grouped
+/// together if we only connect points closer than some threshold?" Imagine
+/// raising that threshold from zero: at first only the tightest pairs are
+/// connected; as it grows, small groups merge into larger ones, until every
+/// point belongs to a single group. Recording which groups merge at which
+/// threshold produces a tree, the *dendrogram*:
+///
+/// ```text
+///                    distance at which groups merge ──►
+///
+///  a ───┐
+///  b ───┴─────┐               {a,b} and {c,d} each merge at a small
+///  c ───┐     ├───────┐       distance; the two pairs join at a larger
+///  d ───┴─────┘       ├───    one; the isolated point e is absorbed
+///  e ─────────────────┘       last, at a much larger distance
+/// ```
+///
+/// This is a *single-linkage* hierarchy: the distance between two groups is
+/// simply the distance between their two closest members. Clusters that
+/// persist over a wide range of thresholds (like `{a, b}`) are "stable";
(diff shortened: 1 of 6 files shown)


geo/src/algorithm/hdbscan/mod.rs
```diff
@@ -10,19 +10,397 @@
...
+
+    /// GLOSH (Global-Local Outlier Scores from Hierarchies) outlier score for
+    /// each input point, in the range [0, 1]. Higher values indicate stronger
+    /// outliers.
+    pub outlier_scores: Vec<T>,
+}
+
+/// Errors that can occur during [`hdbscan`](Hdbscan::hdbscan) clustering.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum HdbscanError {
```
michaelkirk, line 40: Please use `thiserror`

geo/src/algorithm/hdbscan/mod.rs
```diff
@@ -10,19 +10,397 @@
...
+/// // Points in cluster 1 share the same label
+/// assert_eq!(result.labels[0], result.labels[1]);
+/// // Points in cluster 2 share a different label
+/// assert_eq!(result.labels[6], result.labels[7]);
+/// // The two clusters have different labels
+/// assert_ne!(result.labels[0], result.labels[6]);
+/// ```
+pub trait Hdbscan<T>
+where
+    T: GeoFloat + Send + Sync,
```
michaelkirk, line 256: Should we just make GeoNum Send/Sync?

geo/src/algorithm/hdbscan/mod.rs
```diff
@@ -10,19 +10,397 @@
...
+/// [`HdbscanError::InvalidMinSamples`] when `min_samples` is less than 1 or
+/// greater than `n - 1` (the core distance is the distance to the
+/// `min_samples`-th nearest *other* point); and
+/// [`HdbscanError::DisconnectedSpanningTree`] when the input contains
+/// non-finite or extremely large coordinates that leave the
+/// mutual-reachability graph disconnected.
+///
+/// # Trait bounds
+///
+/// This trait requires `T: GeoFloat + Send + Sync`, whereas the sibling
```
michaelkirk, line 174: This documentation seems a bit overly exhaustive at the expense of clarity for something which affects approximately nobody.

geo/src/algorithm/hdbscan/mod.rs
```diff
@@ -10,19 +10,397 @@
...
+/// the feature) keeps the feature additive. In practice `GeoFloat` is only
+/// implemented for `f32` and `f64`, both of which are `Send + Sync`, so this
+/// costs concrete callers nothing; only code generic over `T: GeoFloat` must
+/// add `+ Send + Sync` to call [`hdbscan`](Hdbscan::hdbscan).
+///
+/// # Algorithm
+///
+/// 1. **Compute core distances** – each point's distance to its
+///    `min_samples`-th nearest neighbour – using a
+///    [ball tree](https://en.wikipedia.org/wiki/Ball_tree) rather than
```
michaelkirk, line 188: link to our impl?

geo/src/algorithm/hdbscan/mod.rs
```diff
@@ -10,19 +10,397 @@
...
 use std::collections::HashSet;
 
+/// Result of an HDBSCAN clustering run.
+///
+/// Contains both cluster labels and GLOSH outlier scores for every input point.
+#[derive(Debug, Clone, PartialEq)]
+pub struct HdbscanResult<T> {
+    /// Cluster assignment for each input point, in input order.
+    /// `Some(id)` for clustered points, `None` for noise.
+    pub labels: Vec<Option<usize>>,
```
michaelkirk: I wonder if there'd be demand for an inverted form of this method - a way to iterate over the clusters.

Not necessary for this first pass, but I guess think about it as you (presumably?) use this code.

geo/src/algorithm/hdbscan/mod.rs
```diff
@@ -10,19 +10,397 @@
...
+///  d ───┴─────┘       ├───    one; the isolated point e is absorbed
+///  e ─────────────────┘       last, at a much larger distance
+/// ```
+///
+/// This is a *single-linkage* hierarchy: the distance between two groups is
+/// simply the distance between their two closest members. Clusters that
+/// persist over a wide range of thresholds (like `{a, b}`) are "stable";
+/// short-lived groupings are artefacts of a particular threshold. HDBSCAN
+/// condenses the dendrogram and keeps only the most stable clusters, and
+/// points such as `e` that never joined a stable cluster become noise.
```
michaelkirk, line 142: Great docs!

geo/src/algorithm/hdbscan/tests.rs
```diff
@@ -1,7 +1,296 @@
 use super::*;
+use crate::point;
 use crate::wkt;
 use approx::assert_relative_eq;
 
+#[test]
+fn test_empty_input() {
+    let points: Vec<Point<f64>> = vec![];
+    let result = points.hdbscan(5).unwrap();
```
michaelkirk, line 9: It's admittedly subjective, but I'm initially surprised this does not error while the `test_single_point` test returns an error.

My first instinct is that either `test_empty_input` should return:
```
Err(HdbscanError::InvalidMinClusterSize {
            min_cluster_size: 5,
            n: 0,
        }),
```

Or that `test_single_point` should simply return a valid result with no points being assigned to a cluster since none meet the criteria.

I realize you have this all spelled out in test_empty_vs_tiny_input_asymmetry - it's intentionally not symmetrical, but it didn't align with my guess as to what the behavior would be. If this is consistent with some existing reference implementation or how you think people will actually want to use it, feel free to leave it as is.

geo/src/algorithm/hdbscan/tests.rs
```diff
@@ -1,7 +1,296 @@
...
+            n: 3,
+        }),
+    );
+}
+
+/// Verify that the MultiPoint trait impl delegates correctly and
+/// produces the same result as calling on the underlying slice.
+#[test]
+fn test_multipoint_trait_impl() {
+    let mp = MultiPoint::new(vec![
```
michaelkirk, line 79: Can you please ask your llm to use `wkt!(MULTIPOINT ...)`  for test fixtures so I can easily visualize these ( and for the `[]` slice variants)

geo/src/algorithm/hdbscan/tests.rs
```diff
@@ -1,7 +1,296 @@
...
+            "point {i} should be in cluster C"
+        );
+    }
+}
+
+/// Uniform data with no density structure is labelled entirely noise.
+///
+/// A regular 4x4 grid has no denser sub-region that could form a cluster, so
+/// with the condensed-tree root excluded from selection (reference HDBSCAN
+/// with `allow_single_cluster = false`) every point is noise. Verified
```
michaelkirk, line 276: I don't see `allow_single_cluster` in the codebase. Is this a reference to the paper?

geo/src/algorithm/hdbscan/tests.rs
```diff
@@ -32,6 +321,280 @@ fn test_union_find_chain() {
     assert_eq!(uf.component_label(root), 6);
 }
 
+// -- Known-answer tests (ported from petal-clustering) --------------------
```
michaelkirk, line 324: I don't know what `petal-clustering` is - can you link to it?

geo/src/algorithm/hdbscan/tests.rs
```diff
@@ -32,6 +321,280 @@ fn test_union_find_chain() {
...
+
+/// Dataset from petal-clustering's `partial_labels` test (unsupervised).
+///
+/// Three spatial groups plus one isolated noise point:
+///   Group 1 (5 pts): indices 0..5, near (1-3, 7-9)
+///   Group 2 (4 pts): indices 5..9, near (5-6, 3-4)
+///   Group 3 (6 pts): indices 9..15, near (8-9, 1-3)
+///   Outlier:         index 15, at (7, 8)
+///
+/// With the Phase 1 correctness fixes, geo's labels and GLOSH scores match
```
michaelkirk, line 334: What are these correctness fixes about? Is this some development artifact or is this still useful for readers today?

michaelkirk (commented):
I think it's actually going to be easier for me to review top down rather than bottom up.

If that's going to break everything, I guess we could leave this unmerged until I make my way down.

geo/src/algorithm/hdbscan/mod.rs
urschrei: Seems like a good idea: https://github.com/georust/geo/pull/1594

geo/src/algorithm/hdbscan/mod.rs
urschrei: Fixed, but it'll be fully redundant anyway if #1594 merges

geo/src/algorithm/hdbscan/tests.rs
urschrei: Oops done. Two tests still don't use it: `test_large_n_correctness` (it would be a massive chunk of text) and `test_disconnected_spanning_tree_errors` (hopefully obvious: the wkt version would have 32 pairs with tiny magnitudes)

geo/src/algorithm/hdbscan/tests.rs
urschrei: It's from the reference scikit-learn hdbscan library. Reworded.

geo/src/algorithm/hdbscan/tests.rs
urschrei: Done.

geo/src/algorithm/hdbscan/tests.rs
urschrei: Development artifact from earlier iteration. Replaced with something more sensible.

geo/src/algorithm/hdbscan/tests.rs
urschrei: We already have the same asymmetry in `kmeans` and `dbscan`. An empty collection seems like a natural no-op to me: nothing in, nothing out. If it's non-empty but not clusterable, it's an error. The Python impl _does_ error btw. I've added a small justification to the test. WDYT?

michaelkirk: If you think it's right, we should leave it.

geo/src/algorithm/hdbscan/mod.rs
michaelkirk: Merged! So these qualifiers and the redundant docs can go away.

michaelkirk (approved):

michaelkirk:
Lol, ok I clearly don't understand how PR stacks are supposed to work. I was hoping I could merge 06 down into 05, but I guess not. Is it that we need to keep all these PR's open until the full stack is approved and then they all get merged in at once?

geo/src/algorithm/hdbscan/tests.rs
```diff
@@ -32,6 +306,267 @@ fn test_union_find_chain() {
...
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        0.0,
+        0.6,
+    ];
+    assert_eq!(result.outlier_scores.len(), expected_scores.len());
+    for (&got, &want) in result.outlier_scores.iter().zip(expected_scores.iter()) {
```
michaelkirk, line 389: minor code golf nit: approx supports slices, so you don't need to do this zipping:

```
assert_relative_eq!(result.outlier_scores.as_slice(), expected_scores.as_slice(), epsilon = 1e-9);
```

urschrei:
> Lol, ok I clearly don't understand how PR stacks are supposed to work. I was hoping I could merge 06 down into 05, but I guess not. Is it that we need to keep all these PR's open until the full stack is approved and then they all get merged in at once?

Yep, just leave them open.

PR #1579, opened 2026-08-07
Inline GeoNum::total_cmp impls for cross-crate perf

urschrei:

---

Found by Claude while searching for extra perf for https://github.com/georust/geo/pull/1571 (https://github.com/gauteh/roaring-landmask/pull/39).

<s>NB: this won't be visible to us for `geo` benchmarks (our in-crate calls are always inlined AFAIK); it only affects downstream users who don't use LTO.</s>

Update: as per https://github.com/georust/geo/pull/1579#issuecomment-5221990914 we use a separate crate for benchmarks, so this should be visible to us.

### DIFF
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -33,6 +33,7 @@
 - BREAKING: Tightens the bound on `Contains<MultiPoint<T>> for MultiPoint<T>` from `T: CoordNum` to `T: GeoNum`
   - <https://github.com/georust/geo/pull/1555>
 - Derive all error types via `thiserror` instead of hand-written `Display`/`Error` impls.
+- Inline total_cmp for downstream users who don't use LTO
 
 ## 0.33.1 - 2026-4-20
 
--- a/geo/src/lib.rs
+++ b/geo/src/lib.rs
@@ -410,6 +410,7 @@ macro_rules! impl_geo_num_for_float {
     ($t: ident) => {
         impl GeoNum for $t {
             type Ker = RobustKernel;
+            #[inline]
             fn total_cmp(&self, other: &Self) -> Ordering {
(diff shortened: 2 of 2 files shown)


gauteh:
Thanks. I don't see any immediate changes in the contains benchmark. With everything compiled and set up for python bindings, the python benchmarks are as fast or slightly faster than the geos version.

urschrei:
You mean you enabled thinLTO and it made no difference to your benchmarks, or you're testing against this branch and seeing no change?

gauteh:
I meant to comment on the roaring-landmask PR, sorry. I tested lto="thin", I haven't tested this PR.

michaelkirk (approved):

michaelkirk:
> NB: this won't be visible to us for geo benchmarks (our in-crate calls are always inlined AFAIK);

I hadnt thought about this before and can't investigate right now, but our geo-bench is a separate crate, so shouldn't that exercise the "external crate" aspect?

If I recall correctly, using LTO usually obviates the benefit of the inline attribute. Inline is still worthwhile for those who don't use LTO, so happy to merge it!

Mainly I mention because if we are doing LTO for our local benchmarks, it would similarly prevent exercising the "uninlined" case.

urschrei:
> our geo-bench is a separate crate, so shouldn't that exercise the "external crate" aspect?

That is a good point 🤔

urschrei:
> Mainly I mention because if we are doing LTO for our local benchmarks, it would similarly prevent exercising the "uninlined" case.

You're correct, so we aren't doing inlining or LTO, and this change should now show up in our benches that exercise lots of `total_cmp` (sweep line? relate?). But we should also add a bench for IntervalTree…

PR #1580, opened 2026-08-07
Add IntervalTreeMultiPolygon bench

urschrei:

---

See https://github.com/georust/geo/pull/1579 for context. This benchmark heavily depends on the change made in 1579.

### DIFF
--- a/geo-benches/src/interval_tree_multipolygon.rs
+++ b/geo-benches/src/interval_tree_multipolygon.rs
@@ -1,0 +1,79 @@
+//! Benchmarks for point-in-polygon queries via `IntervalTreeMultiPolygon`.
+
+use criterion::{Criterion, black_box, criterion_group, criterion_main};
+use geo::indexed::IntervalTreeMultiPolygon;
+use geo::{BoundingRect, Contains, MultiPolygon, Point, Polygon};
+
+/// A deterministic n x n grid of query points across the geometry's bounding
+/// rectangle, giving a stable mix of hits, misses, and near-boundary queries.
+fn query_grid(mp: &MultiPolygon, n: usize) -> Vec<Point> {
+    let rect = mp.bounding_rect().unwrap();
+    let (w, h) = (rect.width(), rect.height());
+    let mut points = Vec::with_capacity(n * n);
+    for i in 0..n {
+        for j in 0..n {
+            points.push(Point::new(
+                rect.min().x + w * (i as f64 + 0.5) / n as f64,
+                rect.min().y + h * (j as f64 + 0.5) / n as f64,
+            ));
+        }
+    }
+    points
+}
+
+fn criterion_benchmark(c: &mut Criterion) {
+    // Many small polygons: exercises candidate filtering across parcels.
+    let plots: MultiPolygon = geo_test_fixtures::nl_plots_wgs84();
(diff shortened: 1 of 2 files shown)


geo-benches/src/interval_tree_multipolygon.rs
```diff
@@ -0,0 +1,77 @@
...
+/// A deterministic n x n grid of query points across the geometry's bounding
+/// rectangle, giving a stable mix of hits, misses, and near-boundary queries.
+fn query_grid(mp: &MultiPolygon, n: usize) -> Vec<Point> {
+    let rect = mp.bounding_rect().unwrap();
+    let (w, h) = (rect.width(), rect.height());
+    let mut points = Vec::with_capacity(n * n);
+    for i in 0..n {
+        for j in 0..n {
+            points.push(Point::new(
+                rect.min().x + w * (i as f64 + 0.5) / n as f64,
```
michaelkirk, line 16: Why the `+ 0.5` ? It seems like such a minor adjustment that I feel like I'm missing the significance.

geo-benches/src/interval_tree_multipolygon.rs
```diff
@@ -0,0 +1,77 @@
...
+            &format!("IntervalTreeMultiPolygon contains, 1024 points ({label})"),
+            |bencher| {
+                bencher.iter(|| {
+                    let mut hits = 0;
+                    for point in &points {
+                        if index.contains(black_box(point)) {
+                            hits += 1;
+                        }
+                    }
+                    black_box(hits);
```
michaelkirk, line 55: assert_eq!(hits, 123) would be nice to sanity check.

michaelkirk (approved):
LGTM! Only minor quibbles - nothing blocking.

geo-benches/src/interval_tree_multipolygon.rs
urschrei: Done!

geo-benches/src/interval_tree_multipolygon.rs
urschrei: Without it, the grid starts at `i = 0, j = 0`, which places query points exactly on the bounding rectangle's minimum edges, which by definition pass through the geometry's extreme vertices: some queries would land exactly on polygon boundaries.

Those are the degenerate case for `contains`. The `+ 0.5` shifts every sample to the centre of its grid cell, so they're all guaranteed to be inside the bounding rect.

Fable says it also makes the sampling _symmetric_: without it the grid touches the min edges but stops short of the max edges at (n-1)/n (I had no idea).

michaelkirk: > Without it, the grid starts at i = 0, j = 0, which places query points exactly on the bounding rectangle's minimum edges, which by definition pass through the geometry's extreme vertices: some queries would land exactly on polygon boundaries.

Aren't those interesting cases worth testing? As much as any other? Or is it too pathological for the bench?

urschrei: I think they're interesting for _correctness_ (and we do in fact have tests covering that: e.g. `test_point_on_boundary_indexed_polygon` and a couple of others) but yeah, they represent a degenerate case, which (bc slower) affects the bench, which is measuring the common case (imo).

PR #1594, opened 2026-08-20
Require Send + Sync on GeoNum

urschrei:
Done by LLM; lmk if the CHANGES are too wordy.

---

### DIFF
--- a/geo/src/lib.rs
+++ b/geo/src/lib.rs
@@ -392,7 +392,7 @@ impl<T> GeoFloat for T where
 }
 
 /// A trait for methods which work for both integers **and** floating point
-pub trait GeoNum: CoordNum {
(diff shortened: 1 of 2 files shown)


geo/src/lib.rs
```diff
@@ -392,7 +392,11 @@ impl<T> GeoFloat for T where
 }
 
 /// A trait for methods which work for both integers **and** floating point
-pub trait GeoNum: CoordNum {
+///
+/// This trait requires `Send + Sync` so that algorithms behind the optional
```
michaelkirk: Overly verbose - we don't need to explain that we use Send and Sync to do the things that Send and Sync do.

michaelkirk (approved):
I (as usual) have some doc nitpicks, but LGTM otherwise.

PR #1600, opened 2026-08-26
Add explicit tolerances to geodesic tests

urschrei:
Three assertions in the geodesic tests used the default tolerance of `assert_relative_eq!`. The expected values were computed on glibc 2.31, and on glibc 2.36 (the default in bookworm) the libm results differ by a few ULP.

This surfaced in georust/docker-images#56, which moves the CI images from Debian bullseye (glibc 2.31) to bookworm (glibc 2.36).

Done in conjunction with Fable.

---


michaelkirk (approved):
Thanks!

PR #1601, opened 2026-08-27
Fix sweep line assertion failure across the signed-zero boundary

### LINKED ISSUE #1578: interior_point() panics on a polygon that check_validation() accepts, when coordinates contain -0.0
geo 0.33.1. A polygon containing a negative zero coordinate is reported as valid by check_validation(), but interior_point() then trips its own internal debug_assert! in the sweep line.

Minimal reproducer — only geo = "=0.33.1", with debug-assertions = true:

use geo::algorithm::{InteriorPoint, Validation};
use geo::{Coord, LineString, Polygon};

fn main() {
    let ring = LineString(vec![
        Coord { x: 5.0, y: 22.0 },
        Coord { x: 2.0, y: 7.0 },
        Coord { x: -0.0, y: 423.0 },   // negative zero
        Coord { x: 0.0, y: 3.0 },
        Coord { x: 9.0, y: 0.0 },
        Coord { x: 5.0, y: 22.0 },
    ]);
    let poly = Polygon::new(ring, vec![]);

    assert!(poly.check_validation().is_ok()); // passes
    let _ = poly.interior_point();            // panics
}

assertion failed: intervals_overlap(current_interval, overlapping_interval)
geo-0.33.1/src/algorithm/sweep/mod.rs:169

Replacing -0.0 with 0.0, leaving every other coordinate untouched, makes it return POINT(0.5084134615384616 211.5). The two polygons are numerically identical, since -0.0 == 0.0.

Root cause. src/algorithm/sweep/mod.rs mixes two incompatible orderings:

- line 266 sorts with total_cmp, and intervals_overlap (lines 277-280) compares with total_cmp — a total order, where -0.0 < +0.0;
- line 163 guards with the IEEE operator: if overlapping_interval.inserted_x > current_interval.deleted_x.

With deleted_x == -0.0 and inserted_x == 0.0 the IEEE guard is false, so the pair is not skipped, while intervals_overlap returns false because (-0.0).total_cmp(&0.0) is Less. The assertion then fails. Signed zero is precisely where the two orderings disagree on equal values.

Suggested fix. Make line 163 use the same ordering as the rest of the module:

if overlapping_interval.inserted_x.total_cmp(&current_interval.deleted_x).is_gt() {

Why it matters beyond the panic. It is a debug_assert!, so in release builds the check is compiled out and the sweep proceeds with a violated invariant — interior_point() can then return a wrong point silently. We only noticed because cargo-fuzz enables debug assertions.

Found by fuzzing a WKT parsing and geometry operations pipeline; the input was POLYGON((5 22, 2 7, -0 423, 0 3, 9 0, 5 22)).

jaideeppyne:
Closes #1578.

`Intersections` guards its inner loop with the IEEE comparison `overlapping_interval.inserted_x > current_interval.deleted_x`, but `intervals_overlap` — asserted immediately afterwards — compared the same values with `total_cmp`. Signed zero is the one place where finite values make those two orderings disagree: `(-0.0).total_cmp(&0.0)` is `Less`, while IEEE treats them as equal.

A segment whose maximum x is `-0.0`, followed by a segment inserted at `+0.0`, therefore passes the guard and then fails the assertion, even though the two coordinates denote the same position. `interior_point()` panics in debug builds on a polygon that `Validation::is_valid()` accepts:

    POLYGON((-5 0, -0 5, 0 10, 5 0, -5 0))

This is the case @urschrei identified in #1578 as still reachable after #1554; the original reporter's polygon now passes.

`intervals_overlap` now uses the IEEE operators, matching the guard. `total_cmp` is kept where a total order is actually required — sorting the intervals, and choosing each segment's endpoints in `SweepLineIndex::new`. Tightening the guard instead would be wrong: it would make genuine intersections be missed in release builds, where the assertion is compiled out. The doc comment on `intervals_overlap` is rewritten, since its claim about handling "very close values" did not describe what the function did.

Both new tests go through the existing `verify_intersections` brute-force comparison, so they check that the sweep finds the right intersections, not merely that it does not panic.

Two things worth stating plainly: this is a debug-mode panic rather than a silent wrong answer — release mode falls through to the full `line_intersection` check and stays correct. And the fixed assertion is now close to tautological given the guard; its remaining value is catching a future regression in the sort order.

An LLM agent was used for this change: it ran a property-test battery over the predicate and `relate` invariants to locate the defect, reproduced it, wrote the fix and the two regression tests, and drafted this description. The change has been reviewed and I am accountable for it.

### DIFF
--- a/geo/src/algorithm/interior_point.rs
+++ b/geo/src/algorithm/interior_point.rs
@@ -929,4 +929,38 @@ mod test {
         );
         let _ = poly.interior_point();
     }
+
+    #[test]
+    fn polygon_with_negative_zero_coordinate_does_not_panic() {
+        // Regression for https://github.com/georust/geo/issues/1578: an edge whose
+        // maximum `x` is `-0.0` tripped the sweep line's `intervals_overlap` assertion,
+        // even though `-0.0` and `+0.0` are the same position.
+        let poly = Polygon::new(
+            LineString::from(vec![
+                (-5.0, 0.0),
+                (-0.0, 5.0),
+                (0.0, 10.0),
+                (5.0, 0.0),
+                (-5.0, 0.0),
+            ]),
+            vec![],
+        );
+
+        let interior_point = poly.interior_point().unwrap();
+        assert!(poly.contains(&interior_point));
+
+        // Replacing `-0.0` with `+0.0` describes the very same polygon, so it must
+        // produce the very same interior point.
+        let positive_zero = Polygon::new(
+            LineString::from(vec![
+                (-5.0, 0.0),
+                (0.0, 5.0),
+                (0.0, 10.0),
+                (5.0, 0.0),
+                (-5.0, 0.0),
+            ]),
+            vec![],
+        );
+        assert_eq!(interior_point, positive_zero.interior_point().unwrap());
+    }
 }
--- a/geo/src/algorithm/sweep/mod.rs
+++ b/geo/src/algorithm/sweep/mod.rs
@@ -272,9 +272,15 @@ impl<C: Cross> SweepLineIndex<C> {
 /// Helper function to check if two intervals overlap on the `x`-axis
 ///
 /// This determines which segments might geometrically intersect during the sweep.
-/// We use `total_cmp` for robust floating-point comparisons to handle edge cases
-/// involving very close values.
+///
+/// The comparisons here are the IEEE operators rather than [`total_cmp`], because they
+/// are asking about *position*: `-0.0` and `+0.0` denote the same point on the `x`-axis,
+/// and a segment ending at `-0.0` does overlap one beginning at `+0.0`. `total_cmp` is
+/// used where a *total order* is genuinely required — sorting the intervals, and picking
+/// each segment's endpoints in [`SweepLineIndex::new`] — and signed zero is the only case
+/// where finite values make the two disagree.
+///
+/// [`total_cmp`]: f64::total_cmp
 fn intervals_overlap<C: Cross>(s0: &SweepLineInterval<C>, s1: &SweepLineInterval<C>) -> bool {
-    s0.inserted_x.total_cmp(&s1.inserted_x).is_le()
-        && s0.deleted_x.total_cmp(&s1.inserted_x).is_ge()
+    s0.inserted_x <= s1.inserted_x && s0.deleted_x >= s1.inserted_x
 }
--- a/geo/CHANGES.md
+++ b/geo/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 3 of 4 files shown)


geo/src/algorithm/sweep/mod.rs
```diff
@@ -272,9 +272,15 @@ impl<C: Cross> SweepLineIndex<C> {
...
-/// involving very close values.
+///
+/// The comparisons here are the IEEE operators rather than [`total_cmp`], because they
+/// are asking about *position*: `-0.0` and `+0.0` denote the same point on the `x`-axis,
+/// and a segment ending at `-0.0` does overlap one beginning at `+0.0`. `total_cmp` is
+/// used where a *total order* is genuinely required — sorting the intervals, and picking
+/// each segment's endpoints in [`SweepLineIndex::new`] — and signed zero is the only case
+/// where finite values make the two disagree.
+///
+/// [`total_cmp`]: f64::total_cmp
```
michaelkirk: Please skip the overly verbose AI comments and add a regression test to make sure we don't break it.

geo/src/algorithm/interior_point.rs
```diff
@@ -929,4 +929,38 @@ mod test {
         );
         let _ = poly.interior_point();
     }
+
+    #[test]
+    fn polygon_with_negative_zero_coordinate_does_not_panic() {
+        // Regression for https://github.com/georust/geo/issues/1578: an edge whose
+        // maximum `x` is `-0.0` tripped the sweep line's `intervals_overlap` assertion,
+        // even though `-0.0` and `+0.0` are the same position.
+        let poly = Polygon::new(
```
michaelkirk: please use the wkt! macro for test fixtures.

michaelkirk (changes requested):

geo/src/algorithm/sweep/mod.rs
jaideeppyne: Addressed in 9c2f49cc6e28a395f8377cb31dd808d73a6a8dc9. Dropped the long `intervals_overlap` commentary down to a one-line note that IEEE comparisons treat `-0.0` and `+0.0` as the same position. `sweep_spans_signed_zero_boundary` now asserts `intervals_overlap` directly on a segment deleted at `-0.0` and one inserted at `+0.0`, then still runs `verify_intersections` (including the crossing case) so a regression fails before the debug assertion would.

geo/src/algorithm/interior_point.rs
jaideeppyne: Addressed in 9c2f49cc6e28a395f8377cb31dd808d73a6a8dc9. Both interior_point fixtures now use `wkt!`:

```rust
let poly: Polygon = wkt!(POLYGON((-5.0 0.0, -0.0 5.0, 0.0 10.0, 5.0 0.0, -5.0 0.0)));
let positive_zero: Polygon = wkt!(POLYGON((-5.0 0.0, 0.0 5.0, 0.0 10.0, 5.0 0.0, -5.0 0.0)));
```

The `-0.0` literal is preserved through the macro, so this still exercises the signed-zero sweep path from #1578.

michaelkirk (approved):
LGTM - I'll leave it open another day in case someone else wants to review.

PR #1603, opened 2026-08-31
Fix unsound prefix pruning in the separable distance fast path

urschrei:
The prefix binary search added in #1560 skipped right-list vertices whose intercepts lie more than the projection threshold below the current left vertex. That bound is sound for vertex-vertex distances, but the search evaluates the segments adjacent to each vertex pair, and an edge adjacent to a skipped vertex can reach back into range: the fast path could overestimate the distance between bbox-separated geometries. (sorry, sorry)

Adds the counterexample as a regression test and a 20,000-case randomised comparison against the brute-force segment fold. The new test machinery is somewhat complex, but so is the algorithm.

Discovered by a property test during unrelated work with Fable. This should merge before we release.

---

### DIFF
--- a/geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
+++ b/geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
@@ -612,9 +612,19 @@ struct ProjectedVertex<F: GeoFloat> {
 /// - Any closer pair must have perpendicular separation `< d_min`
 /// - Therefore their intercept difference must be `< d_min * sqrt(1 + slope²)`
 ///
-/// If two `PQ` pair vertices have intercept difference `> max_projection_delta`, they CANNOT be closer
-/// than `d_min` regardless of where they lie along their respective parallel lines. This is
-/// conservative but guarantees we never skip a potentially optimal pair.
+/// If two `PQ` pair vertices have intercept difference `> max_projection_delta`, the two vertices
+/// CANNOT be closer than `d_min` regardless of where they lie along their respective parallel lines.
+///
+/// ## Why the Pruning is One-Sided
+///
+/// Pairs are scored on the segments *adjacent to* the two vertices, so the vertex bound
+/// transfers only in the iteration direction: for the realising edge pair, the
+/// higher-intercept left endpoint `u` and lower-intercept right endpoint `w` satisfy
+/// `intercept(w) - intercept(u) <= d * scale`, so `(u, w)` survives the forward breaks and
+/// evaluates the realising edges. The reverse difference is bounded only after adding the
+/// two edges' own intercept spans – an edge can run a long way back along the projection
+/// axis from the vertex that indexes it – so the prefix skip widens its threshold by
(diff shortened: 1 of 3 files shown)


geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -1608,6 +1642,85 @@ mod test {
...
+
+    #[test]
+    fn fast_path_matches_brute_force() {
+        // Cross-check the separable fast path against a brute-force segment sweep.
+        // Separation along x selects the fast path; the independently placed y windows
+        // make the centroid connector diagonal, which is what exercises the pruning.
+        // Seeded, so failures reproduce.
+        let mut rng = Pcg32::seed_from_u64(0x5EED);
+
+        for _ in 0..20_000 {
```
michaelkirk: I'm somewhat against including a fuzz tests like this in our normal unit test suite, primarily because it's slow.

I know it's just one test, but it adds about 10% runtime to `cargo test -p geo --lib` - and I wouldn't want to start a pattern of adding more tests like this.

I'd be in favor of:
- dropping it
- adding it as a separate fuzz target, but I'd expect it to _not_ use a reproducible seed so as to continue fuzzing.

It's not a deal breaker for me though if you want to keep it in.

geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -1608,6 +1642,85 @@ mod test {
...
+            .flat_map(|seg_a| {
+                b.lines()
+                    .map(move |seg_b| Euclidean.distance(&seg_a, &seg_b))
+            })
+            .fold(f64::INFINITY, f64::min)
+    }
+
+    /// A line string of 2 to 6 random vertices with `x` in `x_lo..x_hi` and `y` within
+    /// 6 of `y_centre`.
+    fn random_line_string(rng: &mut Pcg32, x_lo: f64, x_hi: f64, y_centre: f64) -> LineString<f64> {
```
michaelkirk: I think @frewsxcv did some work with Arbitrary for generating random line strings - but maybe you need a very particular shape here which precludes it.

michaelkirk (approved):
I would be lying if I said I really understand what's going on here, but I trust that you do.

urschrei:
I mostly-do. The fact that our test suite didn't catch it but a proptest immediately did kinda gives me pause. That's the nature of bugs I suppose.

geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
urschrei: Hmm OK it seemed pretty fast when I ran it under nextest, but I confess I didn't pay a huge amount of attention to that. We can move it to a separate target I think.

urschrei: Done!

Issue #1596, opened 2026-08-21
Euclidean distance panics on a single-coordinate LineString


hey-jj:
**Observed**

`Euclidean.distance(&a, &b)` between two `LineString<f64>` values panics when one of the line strings holds a single coordinate (one point, zero segments) and the two do not intersect. The panic is an `Option::unwrap()` on a `None` value:

```
thread 'main' panicked at geo-0.33.1/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs:451:59:
called `Option::unwrap()` on a `None` value
```

The single-coordinate line string reaches `nearest_neighbour_distance`, which builds an R-tree from `geom1.lines()`. A one-coordinate line string yields zero segments, so the tree is empty and `tree_a.nearest_neighbor(&point).unwrap()` unwraps `None`.

Tested against `geo` 0.33.1 from crates.io (current release).

**Expected**

Return a finite distance from the single coordinate to the nearest segment of the other line string, matching the behavior for line strings with two or more coordinates. A distance query should not panic on caller data.

**Reproducer**

`Cargo.toml`:

```toml
[package]
name = "repro"
version = "0.1.0"
edition = "2021"

[dependencies]
geo = "0.33.1"
```

`src/main.rs`:

```rust
use geo::{Coord, LineString};
use geo::algorithm::line_measures::{Distance, Euclidean};

fn main() {
    let a: LineString<f64> = LineString(vec![Coord { x: 0.0, y: 0.0 }]);
    let b: LineString<f64> = LineString(vec![
        Coord { x: 10.0, y: 10.0 },
        Coord { x: 11.0, y: 11.0 },
    ]);
    let d = Euclidean.distance(&a, &b);
    println!("distance = {d}");
}
```

`cargo run` panics at `distance.rs:451` instead of printing a distance. The same panic occurs when both line strings hold a single coordinate.

**Effect on callers**

A single-coordinate line string can arise from simplification or clipping that collapses a line to one vertex. When such a value reaches `Euclidean.distance`, the R-tree built from its zero segments is empty, `nearest_neighbor` returns `None`, and the `unwrap` aborts the calling thread. A caller that feeds geometry from an outside source into a distance query gets a thread abort on this input, with no `Result` to catch.

urschrei:
The panic is correctly observed but a single-coordinate LineString is invalid by construction, so the "expected" value is incorrect because this is unverified machine-generated nonsense.

michaelkirk:
[...]

PR #1609, opened 2026-09-02
Add Hegel tests for geo

DRMacIver:
Here are the promised Hegel tests for geo. I apologise for the fact that they're really quite a lot.

These were written by Claude Fable, under relatively light guidance from me on the specifics. I've got quite broad guidance on how to do this well, and I've spot checked individual tests to make sure they seem sane to me, but I do not understand geo well enough to be completely confident in them. The best I can say is that I've fixed every problem I've seen, the rest of what I've looked at looks reasonable-to-good, and they found a lot of bugs (which they currently have ignored tests for, associated with the specific issues).

As well as the ones I've already filed, there are a bunch of ignored cases in mod.rs that I don't fully understand the design considerations of. Here's what Claude had to say about them:

> Seven of the ignored tests in contains/mod.rs deliberately have no issue link. Each is a deterministic reproducer for a disagreement between Relate and the specialized Contains/Covers/Intersects implementations on degenerate or self-intersecting input: a zero-height Rect, a zero-length Line, a self-intersecting LineString, a zero-area MultiPolygon member, a polygon with an empty exterior but non-empty interior rings, and a ring holding -0.0, which panics in release builds. I haven't filed them because I don't know where you draw the line: InvalidRect only rejects non-finite coordinates and InvalidLineString permits self-intersection, so these inputs sit inside the crate's own validity model, but #1596 shows the question is live and the answers there weren't unanimous. The generators the property tests use stay inside the domain that model admits, so the suite is green as it stands and these seven are parked questions rather than failures. Tell me which you consider bugs and I'll file those and drop the rest. Two of them look like internal inconsistencies regardless of how the validity question lands: Geometry::covers disagreeing with the concrete impl it dispatches to, and Intersects consulting the interior rings of an empty polygon.

Anyway, I will be 0% offended if your response is "man this is way too much code, we can't possibly merge this". I was slightly startled when I realised how much there was here, but I guess geo has quite a broad test surface well suited to property-based testing.

If you would like me to attempt to triage it down into something smaller I can do my best, but I think most of these tests probably really are pulling their weight.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 **/Cargo.lock
 .idea
 .dir-locals.el
+.hegel/
--- a/geo/Cargo.toml
+++ b/geo/Cargo.toml
@@ -48,7 +48,7 @@ rand_pcg = "0.10.1"
 [dev-dependencies]
 approx = ">= 0.4.0, < 0.6.0"
 geo-test-fixtures = { path = "../geo-test-fixtures" }
-hegeltest = "0.35.0"
+hegel = { package = "hegeltest", version = "0.35.0" }
 jts-test-runner = { path = "../jts-test-runner" }
 pretty_env_logger = "0.4"
 wkt = "0.14.0"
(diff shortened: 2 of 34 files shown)


DRMacIver:
Apologies, on further review I've found a bunch of issues with this PR. I'm closing it for now and will open a fresh one when I'm more satisfied with its quality.

PR #1612, opened 2026-09-05
Fix closest-point NaNs from squared-length underflow

### LINKED ISSUE #1604: Very short segments give a NaN closest point, and panic in `Euclidean.distance`
```rust
use geo::{ClosestPoint, Line, Point, coord};

fn main() {
    let line = Line::new(coord! { x: 0.0, y: 0.0 }, coord! { x: 0.0, y: 1e-162 });
    println!("{:?}", line.closest_point(&Point::new(0.0, 0.0)));
}
```

Output:

```
Intersection(POINT(NaN NaN))
```

At `1e-161` and above the same call returns `Intersection(POINT(0.0 0.0))`. `Closest::Intersection` is documented as a point that "actually intersects with the object", and the line passes `check_validation()`.

The same scale of coordinates panics in `Euclidean.distance` on a release build:

```rust
use geo::{Distance, Euclidean, Rect, coord};

fn main() {
    let a = Rect::new(
        coord! { x: 0.0, y: 0.0 },
        coord! { x: 9.828413039546407e-237, y: 1.4830465425330546e-162 },
    );
    let b = Rect::new(
        coord! { x: 1.1113793747425387e-162, y: 1.1113793747425387e-162 },
        coord! { x: 4.914206519773204e-237, y: 2.513455854232436e-88 },
    );
    println!("{:?}", Euclidean.distance(&a.to_polygon(), &b.to_polygon()));
}
```

```
thread 'main' panicked at rstar-0.12.2/src/algorithm/nearest_neighbor.rs:46:52:
```

Both polygons pass `check_validation()`, and every coordinate is finite. On geo 0.33.1 `Euclidean.distance(&a, &b)` panics for the two `Rect`s directly as well; on current main that call returns `0.0` and only the polygon path panics.

Tested on geo 0.33.1 and current main.

(Found this while finally getting the hegel tests I wrote for geo in shape to give you)

Udonburo:
Refs #1604.

This fixes the `Line::closest_point` case where a non-zero segment has a squared length that rounds to zero. The existing calculation stays in place unless it would produce `0/0`.

The fallback scales the direction and query difference separately by powers of two and uses [compensated dot products](https://perso.ens-lyon.fr/jean-michel.muller/AnalysisCornea_final.pdf) to avoid the cancellation from unit normalization. The tests cover exact and near cancellation, subnormal interior points, and endpoint clamping in both directions. Polygon/RTree distance, coordinate subtraction errors and existing intersection-classification issues are outside this change.

The 25 focused tests pass in debug and release. Formatting passes, as do check and clippy with and without default features. The full suite still has two library-test failures and one doctest failure; all three reproduced with matching values on the unchanged base in a separate build directory. All-features validation remains incomplete because of the local PROJ/SQLite3 dependencies.

Ordinary-path timings were similar. The fallback cost about 2–3x for nearby tiny queries and reached about 1.5 µs in the extreme f64 cases measured.

Codex helped with reproduction, implementation, testing and numerical review.

- [x] I added an entry to `CHANGES.md`.

### DIFF
--- a/geo/src/algorithm/closest_point.rs
+++ b/geo/src/algorithm/closest_point.rs
@@ -67,7 +67,18 @@ impl<F: GeoFloat> ClosestPoint<F> for Line<F> {
         let direction_vector = Point::from(self.end - self.start);
         let to_p = Point::from(p.0 - self.start);
 
-        let t = to_p.dot(direction_vector) / direction_vector.dot(direction_vector);
+        let numerator = to_p.dot(direction_vector);
+        let squared_length = direction_vector.dot(direction_vector);
+        let t = if squared_length == F::zero() && numerator == F::zero() {
+            // Avoid 0/0 from squared-length underflow, but preserve non-zero
+            // numerators so that infinite projections still clamp to an endpoint.
+            // Project both vectors onto a unit direction to avoid squaring the
+            // length and keep t == 1 when p is the end point.
+            let unit_direction = direction_vector / line_length;
+            to_p.dot(unit_direction) / direction_vector.dot(unit_direction)
+        } else {
+            numerator / squared_length
+        };
 
         // check the cases where the closest point is "outside" the line
         if t < F::zero() {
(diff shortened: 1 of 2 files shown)


geo/src/algorithm/closest_point.rs
```diff
@@ -67,7 +67,15 @@ impl<F: GeoFloat> ClosestPoint<F> for Line<F> {
         let direction_vector = Point::from(self.end - self.start);
         let to_p = Point::from(p.0 - self.start);
 
-        let t = to_p.dot(direction_vector) / direction_vector.dot(direction_vector);
+        let numerator = to_p.dot(direction_vector);
+        let squared_length = direction_vector.dot(direction_vector);
+        let t = if squared_length == F::zero() && numerator == F::zero() {
+            // Avoid 0/0 from squared-length underflow, but preserve non-zero
+            // numerators so that infinite projections still clamp to an endpoint.
```
urschrei: This guard catches underflow but not overflow. Past about 1e154 the squared length is `+inf`, usually with the numerator, so the ordinary path is `inf / inf` and `t` is NaN. `scaled_projection` already pulls an overflowing pair into range, so we only need to add:

```suggestion
        let t = if (squared_length == F::zero() && numerator == F::zero())
            || !squared_length.is_finite()
            || !numerator.is_finite()
        {
            // Avoid 0/0 from underflow and inf/inf from overflow, but preserve a
            // non-zero numerator over a zero squared length so that infinite
            // projections still clamp to an endpoint.
```

urschrei (dismissed):

urschrei:
Oops, I accidentally committed my own suggested change. Sorry! That having been said, I think this is sound. I think we should hold off to see whether #1609 is reopened and then try this on top.

urschrei (approved):

urschrei:
Two additional tests from Opus that cover the `inf/inf` path:

```rust
    // Squared length overflows to +inf, so the ordinary path is inf / inf.
    #[test]
    fn long_line_has_finite_closest_points() {
        let line = Line::<f64>::new((0.0, 0.0), (1e200, 1e200));
        let closest = line.closest_point(&Point::new(1e200, 0.0));
        assert_finite(closest);
        let Closest::SinglePoint(closest) = closest else {
            panic!("expected a single point, got {closest:?}");
        };
        assert_relative_eq!(closest.x(), 5e199, max_relative = 1e-15);
        assert_relative_eq!(closest.y(), 5e199, max_relative = 1e-15);
    }

    // 1e160 is finite; its square is not.
    #[test]
    fn line_with_overflowing_squared_length_projects_in_range() {
        let line = Line::<f64>::new((0.0, 0.0), (1e160, 0.0));
        let closest = line.closest_point(&Point::new(5e159, 1.0));
        assert_finite(closest);
        let Closest::SinglePoint(closest) = closest else {
            panic!("expected a single point, got {closest:?}");
        };
        assert_relative_eq!(closest.x(), 5e159, max_relative = 1e-15);
        assert_relative_eq!(closest.y(), 0.0);
    }
```

michaelkirk:
> Two additional tests from Opus that cover the inf/inf path:

Do you want those two tests added to this branch @urschrei?

PR #1613, opened 2026-09-07
Make the point-to-segment distance scale-invariant

### LINKED ISSUE #1606: `simplify` drops a vertex farther than `epsilon` at large coordinates, and panics on a debug build
```rust
use geo::{coord, Coord, LineString, Simplify};

// Perpendicular distance from `p` to the line through `a` and `b`, scaled so the
// cross product cannot overflow.
fn distance(p: Coord, a: Coord, b: Coord) -> f64 {
    let s = (b.x - a.x).abs().max((b.y - a.y).abs());
    let (dx, dy) = ((b.x - a.x) / s, (b.y - a.y) / s);
    let (px, py) = ((p.x - a.x) / s, (p.y - a.y) / s);
    (dx * py - dy * px).abs() / (dx * dx + dy * dy).sqrt() * s
}

fn main() {
    for y in [-1e100f64, -1e200, -8.988465674311469e307] {
        let line_string = LineString::from(vec![
            coord! { x: 0.0, y: y },
            coord! { x: 0.0, y: 0.0 },
            coord! { x: 3.0, y: -0.0 },
        ]);
        let simplified = line_string.simplify(2.0);
        let middle = line_string.0[1];
        println!(
            "y = {y:e}: kept the middle vertex = {}, which is {} from the output (epsilon = 2.0)",
            simplified.0.contains(&middle),
            distance(middle, simplified.0[0], simplified.0[simplified.0.len() - 1]),
        );
    }
}
```

On a release build:

```
y = -1e100: kept the middle vertex = true, which is 3 from the output (epsilon = 2.0)
y = -1e200: kept the middle vertex = false, which is 3 from the output (epsilon = 2.0)
y = -8.988465674311469e307: kept the middle vertex = false, which is 3 from the output (epsilon = 2.0)
```

On a debug build the first two lines print and the third input panics:

```
thread 'main' panicked at src/algorithm/simplify.rs:108:5:
assertion `left != right` failed
  left: 0
 right: 0
```

`Simplify` documents that "points closer than `epsilon` distance from the simplified output may be removed". From `y = -1e200` the middle vertex is removed although it is 3.0 from the output and `epsilon` is 2.0; at `y = -1e100` it is kept. Every coordinate is finite.

### LINKED ISSUE #1610: `Euclidean.distance` from a point to a segment returns 0 or NaN at large coordinates
```rust
use geo::{coord, Distance, Euclidean, Line, Point};

fn main() {
    for y in [-1e100f64, -1e200, -8.988465674311469e307] {
        let line = Line::new(coord! { x: 0.0, y: y }, coord! { x: 3.0, y: -0.0 });
        let p = Point::new(0.0, 0.0);
        println!(
            "y = {y:e}: distance to the segment = {}, to its end point = {}",
            Euclidean.distance(&p, &line),
            Euclidean.distance(&p, &Point::from(line.end)),
        );
    }
}
```

Output:

```
y = -1e100: distance to the segment = 3, to its end point = 3
y = -1e200: distance to the segment = 0, to its end point = 3
y = -8.988465674311469e307: distance to the segment = NaN, to its end point = 3
```

The nearest point of the segment to `(0, 0)` is its end point `(3, -0.0)`, 3 away, so I expected 3 in all three rows rather than only at `y = -1e100`. Both endpoints are finite and the line passes `check_validation()`.

(Found while looking into #1606, which came out of the hegel tests.)

urschrei:
The squares in line_segment_distance overflow above 1e154 and underflow below 1e-162, which makes the projection parameter NaN. That can't be clamped, so the result was a distance to a point off the segment: 0 for a segment of length 1e200, NaN for one of 1e-162. A difference of two finite coordinates can also overflow, which gave NaN.

The fix scales each vector by its largest component where necessary, in a cold path, so the ordinary path is unchanged. The two vectors need independent scales: one shared scale underflows when a distant point meets a short segment. A difference that overflows is taken between halved coordinates.

This also fixes `Line`'s rstar `PointDistance` impl and vertex ranking in our Ramer-Douglas-Peucker implementation.

This fix costs 0.0% to 8.5% on the euclidean_distance and simplify benchmarks on my machine (best of four interleaved runs).

Not covered: the products of the point offset and the segment are unchecked where the squared segment length is in range, so they can still overflow or underflow where the offset times the segment length is above 1e308 or below 1e-308. Checking them costs about another 10%.

Closes #1610
closes #1606

Also unblocks the polygon-distance half of #1604.

Done in conjunction with Opus.


geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -1002,6 +1004,37 @@ mod test {
...
+        ] {
+            // The end of the segment is the nearest point to the origin.
+            assert_relative_eq!(Euclidean.distance(&origin, &line), 3.0);
+        }
+    }
+
+    // https://github.com/georust/geo/issues/1604
+    #[test]
+    fn polygon_distance_at_small_coordinates() {
+        // Rect has no WKT form, so the polygons keep the shape of the report.
```
michaelkirk, line 1024: >  Rect has no WKT form, so the polygons keep the shape of the report.

Our (nonstandard) wkt macro understands Rect: `wkt!(RECT(...))`

(but I don't feel so strongly about using it since Rect has decent construction ergonomics already)

geo-types/src/private_utils.rs
```diff
@@ -72,50 +72,51 @@ where
...
         return line_euclidean_length(Line::new(point, start));
     }
-    let dx = end.x - start.x;
-    let dy = end.y - start.y;
-    let d_squared = dx * dx + dy * dy;
-    let r = ((point.x - start.x) * dx + (point.y - start.y) * dy) / d_squared;
+    let (mut dx, mut dy) = (end.x - start.x, end.y - start.y);
+    let (mut px, mut py) = (point.x - start.x, point.y - start.y);
+
+    // Squares of raw coordinates overflow above 1e154 and underflow below 1e-162. The
```
michaelkirk, line 78: Could you review these comments? 

I find these kind of llm trivia exact numbers kind of distracting, and I assume it depends on f64 vs f32 (granted most folks are using f64). 

"passes the tests against 0 and 1" is also not very illuminating. I think you could less distractedly say that they overflow/underflow to NaN.

michaelkirk (approved):
Good fix!

Comments are only nitpicking, and should be considered optional.

PR #1614, opened 2026-09-07
Do not unwrap the nearest neighbour in nearest_neighbour_distance

### LINKED ISSUE #1596: Euclidean distance panics on a single-coordinate LineString

urschrei:
A `LineString` with a single coord yields no segments, so its R-tree is empty and
`nearest_neighbor` returns `None`.

Where one tree is empty, the fold over that side contributes nothing and the opposite
direction still measures the coord against the other geometry's segments. Where both are
empty, neither side has a segment and the two coords give the distance directly.

Callers already return early for empty `LineString`s, so the arm with no coord to measure
is unreachable from the public API.

Closes #1596.

Done in conjunction with Opus.


urschrei:
(holding off on changelog entries until we're ready to merge, since we've got a few in-flight PRs and needless conflict resolution is just annoying for everyone)

urschrei:
As with my comment on #1612 I'm fine to hold off on merging (or even reviewing) this and #1613 while we wait for #1609 to be updated, so that fixes can be tested against it (whether we merge it or not)

geo/src/algorithm/line_measures/metric_spaces/euclidean/distance.rs
```diff
@@ -522,16 +522,29 @@ fn bbox_pruned_min_distance<F: GeoFloat, M>(
 fn nearest_neighbour_distance<F: GeoFloat>(geom1: &LineString<F>, geom2: &LineString<F>) -> F {
     let tree_a = RTree::bulk_load(geom1.lines().map(CachedEnvelope::new).collect());
     let tree_b = RTree::bulk_load(geom2.lines().map(CachedEnvelope::new).collect());
```
michaelkirk, line 524: As I understand it, `nearest_neighbor_distance` is an optimization added to take advantage of the RTree - so I'm not sure it makes sense to contort it to work with inputs unsuitable for RTree's. 

An alternative would be to add something that encodes that contract:
```
debug_assert!(geom1.lines().len() >= 1)
debug_assert!(geom2.lines().len() >= 1)
```

...and then handle degenerate input which would fail those asserts at a higher level. e.g. Typically we treat LineString's of length 1 as Points.

At the end of the day the output is the same, so don't consider this as blocking feedback.

michaelkirk (approved):

PR #1615, opened 2026-09-10
Ensure `Triangle::interior_point` is always within triangle bounds (hegel)

### LINKED ISSUE #1607: `Triangle::interior_point` returns a point that does not intersect a zero-area triangle
```rust
use geo::{coord, InteriorPoint, Intersects, Line, Triangle};

fn main() {
    let triangle = Triangle::new(
        coord! { x: 0.0, y: 0.0 },
        coord! { x: 0.0, y: 0.0 },
        coord! { x: 3.0, y: 87.0 },
    );
    let point = triangle.interior_point();
    println!("triangle: {:?} intersects = {}", point.0, triangle.intersects(&point));

    let line = Line::new(coord! { x: 0.0, y: 0.0 }, coord! { x: 3.0, y: 87.0 });
    let point = line.interior_point();
    println!("same segment as a Line: {:?} intersects = {}", point.0, line.intersects(&point));
}
```

Output:

```
triangle: COORD(1.4999999999999998 43.5) intersects = false
same segment as a Line: COORD(0.0 0.0) intersects = true
```

`InteriorPoint` documents the result as "a point that's guaranteed to intersect a given geometry, and will be strictly on the interior of the geometry if possible, or on the edge if the geometry has zero area". This triangle has zero area and the returned point is off it; the same segment as a `Line` returns a point that intersects.

max-amb:
Fixes #1607. This PR adds a Hegel test for ensuring `triangle.intersects(&triangle.interior_point())`.
We fix the sliver issue by adding `center` for `Triangle` (mirroring `Rect` which already had it).
Furthermore, we explicitly handle degenerate (straight line) triangles by setting the midpoint to the long edge.

This PR was prepared with the help of Claude Fable 5.1.

### DIFF
--- a/geo-types/CHANGES.md
+++ b/geo-types/CHANGES.md
@@ -1,5 +1,9 @@
 # Changes
 
+## Unreleased
+
+- Add `Triangle::center`, the mean of the triangle's vertices, mirroring `Rect::center`.
+
 ## 0.7.20 - 2026-07-31
 
 - Add support for `rstar` 0.13 via the `rstar_0_13` feature, alongside the existing `rstar` versions.
--- a/geo-types/src/geometry/triangle.rs
+++ b/geo-types/src/geometry/triangle.rs
@@ -1,4 +1,4 @@
-use crate::{polygon, Coord, CoordNum, Line, Point, Polygon};
+use crate::{coord, polygon, Coord, CoordFloat, CoordNum, Line, Point, Polygon};
 use core::cmp::Ordering;
 
 /// A bounded 2D area whose three vertices are defined by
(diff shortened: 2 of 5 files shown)


urschrei:
I don't think we're ready to begin incorporating Hegel tests like this, see #1609 for what will probably be the conversation around that.

DRMacIver:
Hi @urschrei, sorry this one is my bad, I should have checked in with you on this first.

I figured that the big problem with #1609 was that it was trying to do too much at once and that a better way to add Hegel would be one step at a time. @max-amb is an intern working with us here at Antithesis so I asked him to extract a single issue of the ones I reported and try to get a single Hegel test and a fix for it.

Happy to have more of a conversation about this - I assumed, maybe incorrectly, from your initial response you were happy for us to add some Hegel tests. What do you need from us / what would you like to know?

(Although I see from CI that we're incompatible with some of your build requirements. We're about to fix the WASM one. I'm not sure what's involved with making hegel work on 1.88 but it's probably doable)

urschrei:
@DRMacIver I think we (georust) need to have a conversation about if / how we want to integrate Hegel – I'm personally very keen to do it, having been a long-time Hypothesis fan and wanting more and better proptests in geo. I suspect that if we do decide to incorporate it, we'll "just" do so in a dedicated PR that adds it as a dev dependency, wires up CI (I don't know if we necessarily need WASM support, and we can certainly bump our MSRV a bit; we like to keep it as low as feasible, but this represents a valid reason to bump it), and runs a small smoke test (or something).

Once that's merged we could then start incorporating the actual tests you guys have written (such as this one). My only concern with this PR is conflating the two steps.

All that to say: please hold while we figure out how to move forward, and thanks for the work to date.

DRMacIver:
@urschrei Sure thing, no problem. We'll leave this PR open for now, and let us know if and when you're ready for it.

Would you like us to extract the code changes from this to fix the issue into a standalone PR without the hegel tests?

urschrei:
Yep, if you have time!

max-amb:
@urschrei will do. Also, wondering if you would like for us to have a look at how the CI migration might work?

michaelkirk:
Do you want to coordinate the discussion about adopting hegel @urschrei, since you have some interest in it?

urschrei:
> Do you want to coordinate the discussion about adopting hegel @urschrei, since you have some interest in it?

Yep, but it'll be next week.

PR #1616, opened 2026-09-10
Ensure Triangle::interior_point is always within triangle bounds (non-hegel)

### LINKED ISSUE #1607: `Triangle::interior_point` returns a point that does not intersect a zero-area triangle

max-amb:
This is #1615 without the addition of Hegel (instead adding to the `triangles` test). The only other difference is the addition of `Triangle::center` to `CHANGES.md`.

The description of #1615 is copied in here:

> Fixes #1607. This PR adds a Hegel test for ensuring `triangle.intersects(&triangle.interior_point())`.
> We fix the sliver issue by adding `center` for `Triangle` (mirroring `Rect` which already had it).
> Furthermore, we explicitly handle degenerate (straight line) triangles by setting the midpoint to the long edge.
> 
> This PR was prepared with the help of Claude Fable 5.1.
> 
> 
> - [X] I agree to follow the project's [code of conduct](https://github.com/georust/geo/blob/main/CODE_OF_CONDUCT.md).
> - [X] I added an entry to `CHANGES.md` if knowledge of this change could be valuable to users.
>

### DIFF
--- a/geo/src/algorithm/interior_point.rs
+++ b/geo/src/algorithm/interior_point.rs
@@ -5,6 +5,7 @@ use crate::algorithm::{
     centroid::Centroid,
     coords_iter::CoordsIter,
     dimensions::HasDimensions,
+    intersects::Intersects,
     line_intersection::LineIntersection,
     line_measures::{Distance, Euclidean},
     lines_iter::LinesIter,
@@ -372,7 +373,14 @@ where
     type Output = Point<T>;
 
     fn interior_point(&self) -> Self::Output {
-        self.centroid()
+        let center = self.centroid();
+
+        if self.intersects(&center) {
+            center
+        } else {
+            // Just use the first vertex if we cannot do anything else.
+            self.v1().into()
+        }
     }
 }
 
@@ -788,6 +796,12 @@ mod test {
         assert_eq!(
             Triangle::new(c(0., 0.5), c(0., 0.5), c(0., 0.5)).interior_point(),
             point!(x: 0., y: 0.5)
+        );
+
+        // degenerate (straight line) triangle
+        assert_eq!(
+            Triangle::new(c(0., 0.), c(0., 0.), c(3., 87.)).interior_point(),
+            point!(x: 0., y: 0.)
         )
     }
 
--- a/geo-types/CHANGES.md
+++ b/geo-types/CHANGES.md
@@ -1,5 +1,9 @@
 # Changes
 
+## Unreleased
+- Add `center` function for `Triangle` type, mirroring `Rect`. Part of ensuring `triangle.interior_point()` is always within the triangle.
+  - <https://github.com/georust/geo/issues/1607>
+
 ## 0.7.20 - 2026-07-31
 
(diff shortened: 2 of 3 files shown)


geo/src/algorithm/interior_point.rs
```diff
@@ -372,7 +375,32 @@ where
     type Output = Point<T>;
 
     fn interior_point(&self) -> Self::Output {
-        self.centroid()
+        let candidate =
+            if T::Ker::orient2d(self.v1(), self.v2(), self.v3()) == Orientation::Collinear {
+                // A triangle like this is degenerate in some way, for example, it could be that
+                // all the points are on a straight line. We use the midpoint of the longest
+                // distance edge in this case.
```
michaelkirk: I question whether it's worthwhile to have such an elaborate work around for degenerate input... pretty much all of our algorithms assume validity, and no dimensional collapse. Occasionally we try to work around it, but it's always a tradeoff as to how far backwards we want to bend. 

We could:
1. leave it as is, where the returned point might not be inside the triangle. Garbage in garbage out.
2. just always return self.v1() without trying to give "the best" answer for degenerate input, keeping the code simple
3. or double down and let property based testing drive the codebase into trying to hande all these edge cases. 

If we do want to go the route of 3 and try to handle garbage input at all costs, presumably we also want Line.interior_point to behave correctly, so it might be better to delegate to `Line::new(start, end).interior_point()` when the triangle collapses, and put any floating point gymnastics there so others can take advantage.

max-amb: Probably `self.v1()`. I agree its probably extraneous to complicate the logic to only provide a more correct answer for the degenerate input.

max-amb: For now I have left the `center` function in. What do you think about this?

michaelkirk (approved):
LGTM! I'll leave it open a bit in case someone else wants to weigh in.

michaelkirk:
CI is currently failing. It might be a pre-existing issue but we're not going to merge this with red CI

Issue #1611, opened 2026-09-03
`contains` and `relate` disagree on a `Rect` with zero width or height


DRMacIver:
```rust
use geo::algorithm::{Contains, Covers, Relate, Validation};
use geo::{Coord, Point, Rect};

fn main() {
    let square = Rect::new(Coord { x: 0.0, y: 0.0 }, Coord { x: 1.0, y: 1.0 });
    let flat = Rect::new(Coord { x: 0.0, y: 0.25 }, Coord { x: 1.0, y: 0.25 });
    let tall = Rect::new(Coord { x: 0.25, y: 0.0 }, Coord { x: 0.75, y: 1.0 });

    println!("flat.check_validation()  = {:?}", flat.check_validation());

    println!("square.contains(&flat)   = {}", square.contains(&flat));
    println!("square.relate(&flat)     = {}", square.relate(&flat).is_contains());
    println!("square.covers(&flat)     = {}", square.covers(&flat));
    println!("square.relate(&flat)     = {} (is_covers)", square.relate(&flat).is_covers());

    println!("square.contains(&tall)   = {}", square.contains(&tall));
    println!("square.relate(&tall)     = {}", square.relate(&tall).is_contains());

    let origin = Point::new(0.0, 0.0);
    let degenerate = Rect::new(Coord { x: 0.0, y: 0.0 }, Coord { x: 0.0, y: 0.0 });
    println!("origin.contains(&degenerate) = {}", origin.contains(&degenerate));
    println!("origin.relate(&degenerate)   = {}", origin.relate(&degenerate).is_contains());

    println!(
        "square.to_polygon().contains(&flat.to_polygon()) = {}",
        square.to_polygon().contains(&flat.to_polygon())
    );
}
```

Output:

```
flat.check_validation()  = Ok(())
square.contains(&flat)   = true
square.relate(&flat)     = false
square.covers(&flat)     = true
square.relate(&flat)     = false (is_covers)
square.contains(&tall)   = true
square.relate(&tall)     = true
origin.contains(&degenerate) = true
origin.relate(&degenerate)   = false
square.to_polygon().contains(&flat.to_polygon()) = false
```

`Contains` and `Covers` both say the square contains a zero-height `Rect` lying inside it, and `Relate` says it does not. `tall`, whose width and height are both non-zero, agrees. A `Point` and a zero-size `Rect` at the same coordinate disagree the same way. `check_validation` accepts the zero-height rect, so this is valid input.

[...]

PR #1617, opened 2026-09-11
Make `Contains<Rect> for Rect` conform to DE-9IM standard

max-amb:
We change `Contains<Rect<T>> for Rect<T>` conform to DE-9IM standards by handling degenerate cases explicitly.

Part of #1611.

### DIFF
--- a/geo/src/algorithm/contains/rect.rs
+++ b/geo/src/algorithm/contains/rect.rs
@@ -1,6 +1,7 @@
 use geo_types::CoordFloat;
 
 use super::{Contains, impl_contains_from_relate, impl_contains_geometry_for};
+use crate::dimensions::Dimensions;
 use crate::{Area, CoordsIter, HasDimensions, Intersects, geometry::*};
 use crate::{CoordNum, GeoFloat};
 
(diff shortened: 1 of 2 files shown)


geo/src/algorithm/contains/rect.rs
```diff
@@ -31,12 +32,109 @@ where
...
-        // All points of LineString must be in the polygon ?
-        self.min().x <= other.min().x
-            && self.max().x >= other.max().x
-            && self.min().y <= other.min().y
-            && self.max().y >= other.max().y
+        use Dimensions::*;
+
+        // We need to handle dimensions seperately to properly conform to DE-9IM
+        match (self.dimensions(), other.dimensions()) {
+            (TwoDimensional, TwoDimensional) => {
```
michaelkirk, line 39: Similar to #1616, I appreciate that we're trying to handle invalid input, but we have hundreds of methods like this, and it won't make sense to re-implement inline the trait for every cartesian possibility of dimensional collapse inside every method.

Maybe a better pattern would be to cast the collapsed geometry to the equivalent type - e.g. a zero dimensional `Rect` becomes a `Point`, then we delegate `contains` on the `Point`. Would you mind seeing if a pattern like that is feasible?

max-amb: Not really... it can be slightly simplified when we are handling with points however  a lot of `Contains` for `Line` rely on `GeoNum` (contrasted to `Rect` that uses `CoordNum` which is weaker).

michaelkirk: I personally don't think this change is worthwhile as phrased.

